PR #4, opened 2024-11-25
ci: add initial builds for Go development

julieqiu:
Initiate github.com/googleapis/generator as a Go module, and introduce
the initial CI builds to support Go development.

### DIFF
--- a/all_test.go
+++ b/all_test.go
@@ -1,0 +1,86 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


all_test.go
```diff
@@ -0,0 +1,86 @@
...
+	"errors"
+	"io/fs"
+	"os"
+	"os/exec"
+	"regexp"
+	"strings"
+	"testing"
+)
+
+var googleHeader = regexp.MustCompile(`^// Copyright 20\d\d Google LLC
```
jskeet, line 28: Such short term thinking, building up tech debt that will bite us in 76 years ;)

all_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+func TestGoModTidy(t *testing.T) {
+	rungo(t, "mod", "tidy", "-diff")
+}
+
+func TestGovulncheck(t *testing.T) {
+	rungo(t, "run", "golang.org/x/vuln/cmd/govulncheck@v1.1.3", "./...")
+}
+
+func rungo(t *testing.T, args ...string) {
```
jskeet, line 76: I'm assuming that `testing` finds all functions starting with `Test` and just runs them magically.

jskeet (approved):

all_test.go
julieqiu: yup, https://pkg.go.dev/testing for details

PR #7, opened 2024-11-25
containers/dotnet: add initial container

julieqiu:
Copied container from
https://github.com/jskeet/google-cloud-dotnet/tree/docker-generate/picard


jskeet (approved):
Not actually reviewed - I'm assuming it's the same as on my repo. (Don't have opportunity to review anything really right now.)

PR #9, opened 2024-11-25
impl(internal/gitrepo): add functionality to clone .git repos

julieqiu:
Update flags. Add functionality to open and clone Git repositories, and update the generator logic to clone `googleapis` and `google-cloud-<language>`.


internal/gitrepo/gitrepo.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// CloneOrOpen provides access to a Git repository.
+//
+// If a repository already exists at the specified directory path (dirpath),
+// it opens and provides access to that repository.
+//
+// Otherwise, it clones the repository from the given URL (repoURL) and saves it
+// to the specified directory path (dirpath).
+func CloneOrOpen(ctx context.Context, dirpath, repoURL string) (*Repo, error) {
```
jskeet, line 41: At some point we're likely to want a branch (or maybe committish) but that can come later.

internal/generate/run.go
```diff
@@ -16,13 +16,37 @@ package generate
...
+const googleapisURL = "https://github.com/googleapis/googleapis"
+
+func cloneGoogleapis(ctx context.Context) (*gitrepo.Repo, error) {
+	repoPath := filepath.Join(os.TempDir(), "/generator-googleapis")
+	return gitrepo.CloneOrOpen(ctx, repoPath, googleapisURL)
+}
+
+func cloneLanguageRepo(ctx context.Context, language string) (*gitrepo.Repo, error) {
+	languageRepoURL := fmt.Sprintf("https://github.com/googleapis/google-cloud-%s", language)
+	repoPath := filepath.Join(os.TempDir(), fmt.Sprintf("/google-cloud-%s", language))
```
jskeet, line 50: Maybe `/generator-google-cloud-%s` to be consistent with `generator-googleapis`?

jskeet (approved):
One potential change and a note for the future, but can be merged as-is.

internal/generate/flags.go
```diff
@@ -62,11 +60,15 @@ Flags:
...
+	if cfg.api == "" {
+		return fmt.Errorf("api must be provided")
+	}
+
 	switch cfg.language {
 	case "cpp":
 		return errNotImplemented
-	case "csharp":
-		return errNotImplemented
+	case "dotnet":
```
jskeet, line 70: We need to modify the help text for this - and remove the "command" part, as that's not currently used as far as I can see (and prevented the flags from being parsed).

internal/generate/run.go
julieqiu: Done

internal/generate/flags.go
julieqiu: Done

PR #15, opened 2024-11-27
impl(internal/gitrepo): clone recursively (include submodules)

jskeet:
It's unclear whether this is actually needed or not - I can't see how the "default" is actually defaulting otherwise, but I may be missing something.


internal/gitrepo/gitrepo.go
```diff
@@ -55,12 +55,13 @@ func CloneOrOpen(ctx context.Context, dirpath, repoURL string) (*Repo, error) {
...
-		SingleBranch:  true,
-		Depth:         1,
-		Tags:          git.NoTags,
+		URL:               repoURL,
+		Progress:          os.Stdout,
+		ReferenceName:     plumbing.HEAD,
+		SingleBranch:      true,
+		Depth:             1,
+		Tags:              git.NoTags,
+		RecurseSubmodules: git.DefaultSubmoduleRecursionDepth,
```
julieqiu, line 64: add a note that conformance tests for dotnet lives in a submodule

julieqiu (approved):

PR #19, opened 2024-12-04
impl(internal/generate): add framework for flags

julieqiu:
Create a framework that supports multiple command options, each with different flags and functionality.

To see this in action, run:

```
go run ./cmd/generator -help
go run ./cmd/generator generate -help
go run ./cmd/generator create -help
```

### DIFF
--- a/internal/generate/generate.go
+++ b/internal/generate/generate.go
@@ -34,18 +34,18 @@ func generatorGenerateCommand() *command {
 		name:  "generate",
 		short: "Generate a new client library",
 		usage: "generator generate [arguments]",
-		fs:    flag.NewFlagSet("generate", flag.ContinueOnError),
+		flags: flag.NewFlagSet("generate", flag.ContinueOnError),
 		run:   generate,
 	}
-	c.fs.StringVar(&apiFlag, "api", "", "name of API inside googleapis")
-	c.fs.Func("language", "language to generate", func(language string) error {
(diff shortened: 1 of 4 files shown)


internal/generate/generate.go
```diff
@@ -0,0 +1,103 @@
...
+	languageRepo, err := cloneLanguageRepo(ctx, languageFlag)
+	if err != nil {
+		return err
+	}
+	return runDocker(googleapisRepo.Dir, languageRepo.Dir, apiFlag)
+}
+
+const googleapisURL = "https://github.com/googleapis/googleapis"
+
+func cloneGoogleapis(ctx context.Context) (*gitrepo.Repo, error) {
```
jskeet, line 78: I'm pretty sure eventually we'll want common functionality like this separated from the actual commands, but I don't know where yet :)

jskeet (commented):
My intuition may be way off here around the nesting and the representation of "generator" as a command in the same way as "generate" / "create". Let's chat later.

(Could certainly merge as-is to unblock though.)

julieqiu:
> My intuition may be way off here around the nesting and the representation of "generator" as a command in the same way as "generate" / "create". Let's chat later.
> 
> (Could certainly merge as-is to unblock though.)

Filed https://github.com/googleapis/generator/issues/20. 

Will merge so you can make progress will I'm out this week.

jskeet (approved):

PR #21, opened 2024-12-05
impl(internal/generate): refactor generatorCommand

### LINKED ISSUE #20: internal/generator: refactor generatorCommand
generatorCommand is the top-level command and not like the others. Refactor it to struct accordingly.

julieqiu:
The top-level command is not like the others. Refactor the construction of the top-level command and remove unnecessary fields from the command struct.

Fixes #20

### DIFF
--- a/internal/generate/create.go
+++ b/internal/generate/create.go
@@ -23,11 +23,10 @@ func generatorCreateCommand() *command {
 	c := &command{
 		name:  "create",
 		short: "Create a new client library",
-		usage: "generator create [arguments]",
 		flags: flag.NewFlagSet("create", flag.ContinueOnError),
 		run:   create,
 	}
-	c.flags.Usage = constructUsage(c.flags, c.short, c.usage, c.commands, false)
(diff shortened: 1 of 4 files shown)


internal/generate/create.go
```diff
@@ -23,11 +23,11 @@ func generatorCreateCommand() *command {
 	c := &command{
 		name:  "create",
 		short: "Create a new client library",
-		usage: "generator create [arguments]",
+		usage: "create [arguments]",
```
jskeet, line 26: I suspect this could be inferred from "name" - given that we always expect to basically parse flags in a consistent way. (i.e. I'd remove this entirely, and let anything that uses `usage` at the moment use `name` instead.)

It's only a tiny bit of redundancy though.

jskeet (approved):
One comment for a future possible change, but otherwise looks good :)

internal/generate/create.go
julieqiu: yup sounds good - removed

PR #25, opened 2024-12-11
feat: implement generator and container interfaces

julieqiu:
Create initial implementation of generator and container interfaces.


internal/command/command.go
```diff
@@ -0,0 +1,190 @@
...
+		if _, err := cloneGoogleapis(ctx); err != nil {
+			return err
+		}
+		return container.Generate(ctx, flagLanguage, flagAPIRoot, flagAPIPath, flagOutput, flagGeneratorInput)
+	},
+}
+
+var CmdUpdateRepo = &Command{
+	Name:  "update-repo",
+	Short: "Configure a new API in a given language",
```
jskeet, line 92:
```suggestion
	Short: "Update a language repository with new API definitions",
```

(Or similar...)

internal/command/command.go
```diff
@@ -0,0 +1,190 @@
...
+		if flagAPIRoot == "" {
+			return fmt.Errorf("-api-root is not provided")
+		}
+		if flagAPIPath == "" {
+			return fmt.Errorf("-api-path is not provided")
+		}
+		if !supportedLanguages[flagLanguage] {
+			return fmt.Errorf("invalid -language flag specified: %q", flagLanguage)
+		}
+		if _, err := cloneGoogleapis(ctx); err != nil {
```
jskeet, line 83: We only need to clone googleapis if api-root isn't specified (in which case, the location of the clone is the default)

internal/command/command.go
```diff
@@ -0,0 +1,190 @@
...
+		addFlagOutput,
+	} {
+		fn(fs)
+	}
+
+	fs = CmdUpdateRepo.flags
+	for _, fn := range []func(fs *flag.FlagSet){
+		addFlagAPIPath,
+		addFlagAPIRoot,
+		addFlagBranch,
```
jskeet, line 171: Not used yet?

internal/container/container.go
```diff
@@ -14,22 +14,53 @@
...
+func runGenerate(googleapisDir, languageDir, api string) error {
+	args := []string{
+		"run",
+		"-v", fmt.Sprintf("%s:/apis", googleapisDir),
+		"-v", fmt.Sprintf("%s:/output", languageDir),
+		dotnetImageTag,
+		"--command=update",
+		"--api-root=/apis",
+		fmt.Sprintf("--api=%s", api),
+		"--output-root=/output",
```
jskeet, line 53: Just --output now

internal/container/container.go
```diff
@@ -14,22 +14,53 @@
...
-	return errNotImplemented
+func Configure(ctx context.Context, language, apiRoot, apiPath, generatorInput string) error {
+	return runCommand("echo", "configure not implemented")
 }
 
-func Build() error {
-	return errNotImplemented
+const dotnetImageTag = "picard"
+
+func runGenerate(googleapisDir, languageDir, api string) error {
```
jskeet, line 44: Let's change "api" to "apiPath" to be clear

internal/container/container.go
```diff
@@ -14,22 +14,53 @@
...
+	args := []string{
+		"run",
+		"-v", fmt.Sprintf("%s:/apis", googleapisDir),
+		"-v", fmt.Sprintf("%s:/output", languageDir),
+		dotnetImageTag,
+		"--command=update",
+		"--api-root=/apis",
+		fmt.Sprintf("--api=%s", api),
```
jskeet, line 52: This should be api-path, but only if it's actually specified.
(Or we could make the container contract say that it's *always* specified, and to ignore it if it's empty? I don't much mind either way.)

internal/container/container.go
```diff
@@ -14,22 +14,53 @@
...
 
-func Clean() error {
-	return errNotImplemented
+func Configure(ctx context.Context, language, apiRoot, apiPath, generatorInput string) error {
+	return runCommand("echo", "configure not implemented")
 }
 
-func Build() error {
-	return errNotImplemented
+const dotnetImageTag = "picard"
```
jskeet, line 42: This is the bit I'll fix soon, with any luck.

jskeet (approved):
Various tweaks needed, but we could merge and then I can work on the fixes if you'd prefer? Approving for that reason.

internal/command/command.go
```diff
@@ -0,0 +1,190 @@
...
+		}
+		return container.Generate(ctx, flagLanguage, flagAPIRoot, flagAPIPath, flagOutput, flagGeneratorInput)
+	},
+}
+
+var CmdUpdateRepo = &Command{
+	Name:  "update-repo",
+	Short: "Configure a new API in a given language",
+	Run: func(ctx context.Context) error {
+		if flagAPIRoot == "" {
```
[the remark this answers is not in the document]
julieqiu: yup going to resolve this and add that in a future PR

internal/command/command.go
```diff
@@ -0,0 +1,190 @@
...
+		if !supportedLanguages[flagLanguage] {
+			return fmt.Errorf("invalid -language flag specified: %q", flagLanguage)
+		}
+		if _, err := cloneGoogleapis(ctx); err != nil {
+			return err
+		}
+		if _, err := cloneLanguageRepo(ctx, flagLanguage); err != nil {
+			return err
+		}
+		if err := container.Generate(ctx, flagLanguage, flagAPIRoot, flagAPIPath, flagOutput, flagGeneratorInput); err != nil {
```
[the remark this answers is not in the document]
julieqiu: Filed https://github.com/googleapis/generator/issues/26

internal/container/container.go
julieqiu: I went with the second for now. We can change it later.

internal/command/command.go
```diff
@@ -0,0 +1,190 @@
...
+			return err
+		}
+		if err := container.Clean(ctx, flagLanguage, flagOutput, flagAPIPath); err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
julieqiu: Filed https://github.com/googleapis/generator/issues/27

internal/command/command.go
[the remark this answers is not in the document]
julieqiu: I'm still not sure I understand what this means, but I'll merge and leave it to you to address.

PR #33, opened 2024-12-11
fix: resolve panic on startup due to duplicate language flag declaration

### LINKED ISSUE #30: Panic on startup
Running `go run ./cmd/generator`:

```text
configure flag redefined: language
panic: configure flag redefined: language
```

burkedavison:
Resolves #30

### DIFF
--- a/internal/command/flags.go
+++ b/internal/command/flags.go
@@ -43,8 +43,7 @@ func addFlagBranch(fs *flag.FlagSet) {
 }
 
 func addFlagLanguage(fs *flag.FlagSet) {
-	fs.StringVar(&flagLanguage, "language", "", "language to generate code for")
-	fs.Func("language", "", func(language string) error {
+	fs.Func("language", "language to generate code for", func(language string) error {
 		if !supportedLanguages[language] {
 			return fmt.Errorf("invalid -language flag specified: %q", language)
 		}
--- a/internal/command/flags.go
+++ b/internal/command/flags.go
@@ -30,7 +30,7 @@ var (
 )
 
 func addFlagAPIRoot(fs *flag.FlagSet) {
-	fs.StringVar(&flagAPIRoot, "api-root", "", "(Required) location of googleapis repository")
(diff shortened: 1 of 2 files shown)


internal/command/flags.go
```diff
@@ -43,8 +43,7 @@ func addFlagBranch(fs *flag.FlagSet) {
 }
 
 func addFlagLanguage(fs *flag.FlagSet) {
-	fs.StringVar(&flagLanguage, "language", "", "language to generate code for")
```
julieqiu, line 46: oops =P

I actually meant to keep this line and delete the other one

internal/command/flags.go
```diff
@@ -31,26 +30,19 @@ var (
 )
 
 func addFlagAPIRoot(fs *flag.FlagSet) {
-	fs.StringVar(&flagAPIRoot, "api-root", "", "location of googleapis repository")
+	fs.StringVar(&flagAPIRoot, "api-root", "", "(Required) location of googleapis repository")
```
julieqiu, line 33: I don't _think_ this is required (see @jskeet's comments on https://github.com/googleapis/generator/pull/25/files)

julieqiu, line 33: I don't _think_ this is required (if flagAPIRoot is empty, we clone googleapis)

https://github.com/googleapis/generator/pull/25/files#diff-69d3575d337e6861d23e877e04bbdb99f437a82661f706f055a50e82b1e8282aR81

I think this needs to be fixed:

https://github.com/googleapis/generator/pull/25/files#diff-69d3575d337e6861d23e877e04bbdb99f437a82661f706f055a50e82b1e8282aR54

julieqiu (approved):

internal/command/flags.go
burkedavison: It does behave this way for generate -- I was playing around and found it required for the 'configure' command, but then it's also not yet implemented.

I've updated to reflect the generate command's behavior which is the current focus.

PR #36, opened 2024-12-11
feat: provide default output directory if none is specified

### LINKED ISSUE #26: flagOutput should default to tmp/generator-{timestamp}
If flagOutput hasn't been provided, we should default it (e.g. to "tmp/generator-{timestamp}") and create the directory ourselves.

_Originally posted by @jskeet in https://github.com/googleapis/generator/pull/25#discussion_r1879644734_

burkedavison:
Fixes #26

### DIFF
--- a/internal/command/command.go
+++ b/internal/command/command.go
@@ -18,6 +18,10 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"log/slog"
+	"os"
+	"path/filepath"
+	"time"
 
 	"github.com/googleapis/generator/internal/container"
 )
@@ -85,6 +89,14 @@ var CmdGenerate = &Command{
 			}
 			flagAPIRoot = repo.Dir
 		}
+		if flagOutput == "" {
+			defaultOutput, err := defaultOutput()
+			if err != nil {
+				return err
+			}
+			flagOutput = defaultOutput
+			slog.Info(fmt.Sprintf("No output directory specified. Defaulting to %s", defaultOutput))
+		}
 		return container.Generate(ctx, flagLanguage, flagAPIRoot, flagAPIPath, flagOutput, flagGeneratorInput)
 	},
 }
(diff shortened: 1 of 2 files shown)


internal/container/container.go
```diff
@@ -20,11 +20,13 @@ import (
...
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 )
 
 func Generate(ctx context.Context, language, apiRoot, apiPath, output, generatorInput string) error {
-	return runGenerate(apiRoot, fmt.Sprintf("google-cloud-%s", language), apiPath)
+	languageDir := filepath.Join(output, fmt.Sprintf("google-cloud-%s", language))
+	return runGenerate(apiRoot, languageDir, apiPath)
```
burkedavison, line 29: This logic is really just perpetuating what was already there, but I'm guessing this is not the behavior that we'll choose to keep.

internal/command/command.go
```diff
@@ -125,6 +145,25 @@ var CmdUpdateRepo = &Command{
 	},
 }
 
+func defaultOutput() (string, error) {
+	const yyyyMMddHHmmss = "20060102150405" // Expected format by time library
+
+	path := filepath.Join(os.TempDir(), "generator-"+time.Now().Format(yyyyMMddHHmmss))
```
julieqiu, line 151:
```
path := filepath.Join(os.TempDir(), fmt.Sprintf("generator-%s", time.Now().Format(yyyyMMddHHmmss))
```

internal/command/command.go
```diff
@@ -125,6 +145,25 @@ var CmdUpdateRepo = &Command{
...
+
+	if _, err := os.Stat(path); err == nil {
```
julieqiu, line 153: nit: I would recommend writing this like:

```
	_, err := os.Stat(path)
	switch {
	case os.IsNotExist(err):
		if err := os.Mkdir(path, 0755); err != nil {
			return "", fmt.Errorf("unable to create default output path '%s': %w", path, err)
		}
	case err == nil:
		return "", fmt.Errorf("default output path already exists: %s", path)
	default:
		return "", fmt.Errorf("unable to check directory '%s': %w", path, err)
	}
```

julieqiu (approved):

internal/container/container.go
julieqiu: 1. We probably want to pass this in as an argument at some point down the line

burkedavison: Sounds good

internal/command/command.go
burkedavison: Done

internal/command/command.go
burkedavison: Done

PR #42, opened 2024-12-12
feat: Perform defensive copying

### LINKED ISSUE #40: Work out when to create defensive copies
There are times when it's convenient for a container to have read/write access to a directory, even if we don't want them to make changes. Examples:

- The googleapis clone: in .NET, some APIs require a pregeneration step to tweak the protos
- The generator-input directory: any postgeneration code that gets built during generation may end up being built *in* generator-input (at least, that's what's easiest in .NET)

I suspect we should:

- When cloning, always use a fresh clone rather than opening an existing one. (A user can specify an existing directory if they want. It's unclear under what circumstances that should be copied - maybe "yes" for googleapis, "no" for language-repo.)
- When generating, copy generator-input and pass that into the container

I wonder whether we should have a /tmp/generator-{timestamp} directory that's used as the root for *all* temporary files for that run - so under there, we can just clone to googleapis, google-cloud-dotnet etc, create a copy in generator-input.

All of this can probably be post-prototype.

jskeet:
This also refactors our temporary directories so they're under one directory per CLI run, e. g. /tmp/generator-20241212T153039.

Fixes #40.

### DIFF
--- a/internal/command/command.go
+++ b/internal/command/command.go
@@ -18,11 +18,12 @@ import (
 	"context"
 	"flag"
 	"fmt"
-	"github.com/googleapis/generator/internal/container"
 	"log/slog"
 	"os"
 	"path/filepath"
 	"time"
+
+	"github.com/googleapis/generator/internal/container"
 )
 
 type Command struct {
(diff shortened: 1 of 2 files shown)


internal/command/command.go
```diff
@@ -74,67 +75,107 @@ var CmdGenerate = &Command{
...
 	Short: "Configure a new API in a given language",
 	Run: func(ctx context.Context) error {
+
 		if !supportedLanguages[flagLanguage] {
 			return fmt.Errorf("invalid -language flag specified: %q", flagLanguage)
 		}
+
+		// tmpRoot is a newly-created working directory under /tmp
+		// We do any cloning or copying under there.
+		tmpRoot, err := createTmpWorkingRoot(time.Now())
```
burkedavison, line 128: Nice.

PR #43, opened 2024-12-13
feat: Implement the "generate" workflow

jskeet:
This workflow is for "unconfigured generation, based only on the API definition". (This is only unconfigured in terms of any language-specific configuration which would live in generator-input; we'd still expect a service config etc.)

The new "build" flag says whether or not to attempt to build the result.

Currently only a single language is supported; we could potentially iterate over multiple specified languages instead.

When building "unconfigured" code, we pass "--generator-output" to the container, instead of "--repo-root", to indicate the nature of the code to be built.

### DIFF
--- a/internal/command/command.go
+++ b/internal/command/command.go
@@ -75,29 +75,29 @@ var CmdGenerate = &Command{
 	Name:  "generate",
 	Short: "Generate a new client library",
 	Run: func(ctx context.Context) error {
-		// tmpRoot is a newly-created working directory under /tmp
-		// We do any cloning or copying under there.
-		tmpRoot, err := createTmpWorkingRoot(time.Now())
-		if err != nil {
-			return err
-		}
-
 		if flagAPIPath == "" {
 			return fmt.Errorf("-api-path is not provided")
 		}
 		if !supportedLanguages[flagLanguage] {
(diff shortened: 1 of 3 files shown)


internal/command/command.go
```diff
@@ -75,29 +75,29 @@ var CmdGenerate = &Command{
...
 		if flagAPIPath == "" {
 			return fmt.Errorf("-api-path is not provided")
 		}
 		if !supportedLanguages[flagLanguage] {
 			return fmt.Errorf("invalid -language flag specified: %q", flagLanguage)
 		}
-
-		var apiRoot string
 		if flagAPIRoot == "" {
-			repo, err := cloneGoogleapis(ctx, tmpRoot)
```
burkedavison, line 94: No more automatic cloning?

burkedavison (approved):

internal/command/command.go
jskeet: Not for now - the "spec" (the doc) says that it's required. But I do want to do that in a follow-up. (I figured I'd implement what was written first.)

jskeet: (I was confused - this is generate, not configure.) It's required in the proposed spec, and I think it makes sense to keep it that way. The use cases for generate are generally about testing a speculative API that you already have locally - it doesn't make much sense (IMO) to generate a pristine clone.

PR #57, opened 2024-12-13
feat: add initial kokoro setup

burkedavison:


.kokoro/build.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
jskeet, line 1: Possibly rename to update-dotnet.sh?

.kokoro/build.sh
```diff
@@ -0,0 +1,20 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+go run ./cmd/generator update-repo \
```
jskeet, line 18: If you set the environment variable in the Kokoro config (see previous PR, I can't remember the name offhand) you should be able to omit the image name.

jskeet (approved):
I'm sure it works as-is - just a couple of suggestions.

.kokoro/build.sh
burkedavison: Done

.kokoro/build.sh
burkedavison: Done

PR #64, opened 2024-12-16
feat: install go 1.23.4 in Kokoro

burkedavison:


jskeet (approved):
LGTM as a workaround, but I assume we'll want to either publish our own Docker image, or do something similar that at least avoids having this in each script.

PR #65, opened 2024-12-16
fix: Clean non-API-specific files when configuring a new API

### LINKED ISSUE #63: Copy after "configure" fails due to non-API-specific file
I could have sworn this was working before, but this now fails:

```text
$ go run ./cmd/generator configure -api-path=google/cloud/blockchainnodeengine/v1 -language=dotnet
```

The copy part fails with:

```text
2024/12/16 15:38:15 open /tmp/generator-20241216T153643/google-cloud-dotnet/.github/renovate.json: file exists
```

We've generated renovate.json (correctly) but not cleaned the old one. We might need a tweak to "clean" to say "non-API-specific files only".

jskeet:
Fixes #63

### DIFF
--- a/internal/command/command.go
+++ b/internal/command/command.go
@@ -133,7 +133,10 @@ var CmdConfigure = &Command{
 		if err := container.Generate(ctx, image, apiRoot, outputDir, generatorInput, flagAPIPath); err != nil {
 			return err
 		}
-		// No need to clean here, as configure should fail for an existing API.
+		// We don't need to clean the newly-configured API, but we *do* need to clean any non-API-specific files.
+		if err := container.Clean(ctx, image, languageRepo.Dir, "none"); err != nil {
+			return err
+		}
 		if err := os.CopyFS(languageRepo.Dir, os.DirFS(outputDir)); err != nil {
 			return err
 		}


jskeet:
(This won't help until https://github.com/googleapis/google-cloud-dotnet/pull/14012 is merged and a new container built, mind you.)

internal/command/command.go
```diff
@@ -133,7 +133,10 @@ var CmdConfigure = &Command{
 		if err := container.Generate(ctx, image, apiRoot, outputDir, generatorInput, flagAPIPath); err != nil {
 			return err
 		}
-		// No need to clean here, as configure should fail for an existing API.
+		// We don't need to clean the newly-configured API, but we *do* need to clean any non-API-specific files.
+		if err := container.Clean(ctx, image, languageRepo.Dir, "none"); err != nil {
```
burkedavison, line 137: What's the difference between `clean --api-path=none` and not having an `--api-path` argument?

burkedavison (approved):

internal/command/command.go
jskeet: No --api-path means "clean all configured APIs" at the moment. I'm tempted to make it required and have two special values: `all` and `none` for clarity. But that's probably a 2025 job.

PR #72, opened 2025-01-03
feat: implement commit

### LINKED ISSUE #69: In update-repo, we should commit before the build
Any changes to the repo within the build step shouldn't be persisted - the simplest way of doing this is just to commit changes before the build instead of after.

Currently that would fail because commit just exits, but once we've got commit implemented, we should move it.

burkedavison:
Resolves #69

Example build: https://fusion2.corp.google.com/invocations/ef4a832b-54f1-4dd7-9778-46f08539eee5/targets/cloud-devrel%2Fclient-libraries%2Fcloud-sdk-production-pipeline%2Fcontinuous;config=default/log

Also:
* Only throw an error in push() if push is requested.
* Minor formatting improvement ("thick line") to docker command output separators.
* Pretty-print the git status to the logs.

### DIFF
--- a/internal/command/command.go
+++ b/internal/command/command.go
@@ -140,7 +140,8 @@ var CmdConfigure = &Command{
 		if err := os.CopyFS(languageRepo.Dir, os.DirFS(outputDir)); err != nil {
 			return err
 		}
-		if err := commitAll(ctx, languageRepo); err != nil {
+		msg := fmt.Sprintf("Configured API %s", flagAPIPath) // TODO: Improve info using googleapis commits and version info
+		if err := commitAll(ctx, languageRepo, msg); err != nil {
 			return err
 		}
 		if err := container.Build(ctx, image, "repo-root", languageRepo.Dir, flagAPIPath); err != nil {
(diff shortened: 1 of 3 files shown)


internal/container/container.go
```diff
@@ -173,7 +173,7 @@ func runCommand(c string, args ...string) error {
 	cmd := exec.Command(c, args...)
 	cmd.Stderr = os.Stderr
 	cmd.Stdout = os.Stdout
-	slog.Info(strings.Repeat("-", 80))
+	slog.Info(strings.Repeat("=", 80))
```
jskeet, line 176: Not sure of the purpose of this line. I don't particularly object, mind you. (Have you deliberately left the one below using '-'?)

internal/command/command.go
```diff
@@ -326,12 +326,27 @@ func createTmpWorkingRoot(t time.Time) (string, error) {
...
+	if err != nil {
+		return err
+	}
+	if status.IsClean() {
+		slog.Info("No modifications to commit.")
+		return nil
+	}
+
+	gitrepo.PrintStatus(ctx, repo)
+	// TODO: Get commit message from googleapis
```
jskeet, line 341: For now, perhaps we could pass a message in from the caller? ("Configured API {x}", "Regenerated all APIs" or "Regenerated API {x}" would all be at least marginally better than this.

jskeet (approved):
Happy as is, but a couple of questions/suggestions.

internal/container/container.go
burkedavison: When looking at logs, it's just easier for my eye to differentiate the sections when there's a clear "start" line rather than separators that can mean both 'start' and 'middle'.

Not super important, though.

jskeet: Right, that's fine - no problem :)

jskeet (approved):
Yup, that's exactly what I was thinking of :)

PR #78, opened 2025-02-10
fix: Fix update-repo ready for further testing

jskeet:
We now always clone repos fully instead of with depth=1 - the latter breaks anything that uses the commit log. (We can potentially add a flag for that later.)

The change to filtering in GetApiCommits mirrors what we do in .NET; this is a much, much faster way of finding the commits that actually change directories. We don't even try to handle multi-parent commits, but I don't think that's required at the moment.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/googleapis/generator
 
-go 1.23
+go 1.23.6
 
 require github.com/go-git/go-git/v5 v5.13.2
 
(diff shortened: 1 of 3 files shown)


jskeet:
<sigh> - will look at the crypto vulnerability stuff tomorrow

jskeet:
@ldetmer The second commit mitigates the CVE, so this PR is now ready to go.

ldetmer (approved):

PR #89, opened 2025-03-11
feat: Run docker as the current user, rather than as root

### LINKED ISSUE #81: Specify uid:gid when starting containers
Currently we end up creating directories in the mounted volumes as root. We should use the current user instead, by specifying that when running docker run. See https://docs.docker.com/engine/containers/run/#user

jskeet:
This means the containers will need to support being run in this fashion (i.e. not depend on root access) but the big advantage is that any files written by the container will be owned by the user - so can easily be cleaned up.

(This also removes any need to have "git safe" directories configured due to user mismatches, although we're hoping that won't be relevant anyway.)

Fixes #81.

### DIFF
--- a/internal/container/container.go
+++ b/internal/container/container.go
@@ -20,6 +20,7 @@ import (
 	"log/slog"
 	"os"
 	"os/exec"
+	"os/user"
 	"strings"
 )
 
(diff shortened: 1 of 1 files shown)


jskeet:
Will address the vulnerability in a separate PR.

internal/container/container.go
```diff
@@ -144,9 +145,18 @@ func runDocker(image string, mounts []string, containerArgs []string) error {
 		"run",
 		"--rm", // Automatically delete the container after completion
 	}
+	// Run as the current user in the container - primarily so that any
+	// files we create end up being owned by the current user (and easily deletable).
+	currentUser, err := user.Current()
```
ldetmer, line 150: @zhumin8 just wanted to confirm that this logic works with java hermetic build image? I thought we needed it to run as root?

jskeet: https://github.com/googleapis/sdk-platform-java/tree/main/hermetic_build explicitly has `-u "$(id -u):$(id -g)"` in the Docker commands, so I'm assuming it's okay.

ldetmer (approved):

PR #94, opened 2025-03-19
feat: added create-release-pr command

ldetmer:
command:

go run ./cmd/librarian create-release-pr -github-token xxx -language dotnet -repo-root ~/test_workspace/google-cloud-dotnet/ -image us-central1-docker.pkg.dev/ldetmer-sanbox/fake-release-pr-image/dotnet:latest -push true -skipBuild 
true

example PR: https://github.com/ldetmer/google-cloud-dotnet/pull/78/


internal/command/command.go
```diff
@@ -355,10 +356,246 @@ var CmdUpdateApis = &Command{
...
 
+var CmdCreateReleasePR = &Command{
+	Name:  "create-release-pr",
+	Short: "Generate a PR for release",
+	Run: func(ctx context.Context) error {
+		err := checkFlags()
+		if err != nil {
+			return err
+		}
+		languageRepo, repoPath, inputDirectory, err := setupLanguageAndInputFolders(ctx)
```
jskeet, line 371: This is specific to releasing, right? We might want to rename the function - but that can wait.

This file is getting large enough that we should really be looking at refactoring, I think... (I'm not sufficiently expert in Go to know the best refactoring approach here, but we can at least look at some options - after this PR though.

internal/command/command.go
```diff
@@ -355,10 +356,246 @@ var CmdUpdateApis = &Command{
...
+		if err != nil {
+			return err
+		}
+
+		pipelineState, err := loadState(languageRepo)
+		if err != nil {
+			fmt.Println("Error loading pipeline state:", err)
```
jskeet, line 378: Let's use slog.Info like we do elsewhere

internal/command/command.go
```diff
@@ -355,10 +356,246 @@ var CmdUpdateApis = &Command{
...
+		err := checkFlags()
+		if err != nil {
+			return err
+		}
+		languageRepo, repoPath, inputDirectory, err := setupLanguageAndInputFolders(ctx)
+		if err != nil {
+			return err
+		}
+
+		pipelineState, err := loadState(languageRepo)
```
jskeet, line 376: I suspect we'll end up wanting to load the state within the loop, when we implement "revert commit if the test failed". This is probably okay for now.

internal/command/command.go
```diff
@@ -355,10 +356,246 @@ var CmdUpdateApis = &Command{
...
+			return nil, "", "", err
+		}
+	} else {
+		languageRepo, err = gitrepo.Open(ctx, flagRepoRoot)
+		if err != nil {
+			return nil, "", "", err
+		}
+	}
+
+	inputDir := filepath.Join(tmpRoot, "inputs")
```
jskeet, line 407: It's not yet clear to me what this is going to be used for. Hopefully it'll become obvious later :)

internal/command/command.go
```diff
@@ -355,10 +356,246 @@ var CmdUpdateApis = &Command{
...
+	if err := os.Mkdir(inputDir, 0755); err != nil {
+		fmt.Println("Failed to create input directory:", err)
+		return nil, "", "", err
+	}
+
+	repoPath := flagRepoRoot
```
jskeet, line 413: I suspect we don't need this - we already have languageRepo which should have the right path. (see `gitrepo.Repo.Dir`).

internal/command/command.go
```diff
@@ -355,10 +356,246 @@ var CmdUpdateApis = &Command{
...
+	}
+
+	return languageRepo, repoPath, inputDir, nil
+}
+
+func checkFlags() error {
+	if !supportedLanguages[flagLanguage] {
+		return fmt.Errorf("invalid -language flag specified: %q", flagLanguage)
+	}
+	if flagImage == "" {
```
jskeet, line 425: We should use the pipeline state for the image - allow it to be overridden by a flag, but default to the one in the pipeline state. Basically `deriveImage` should already be okay for this, I think.

internal/command/command.go
```diff
@@ -355,10 +356,246 @@ var CmdUpdateApis = &Command{
...
+		return fmt.Errorf("Create release pr requires image")
+	}
+	return nil
+}
+
+func generateReleasePr(ctx context.Context, librariesToRelease map[string]string, pipelineState *statepb.PipelineState, repoPath string, repo *gitrepo.Repo, prDescription string) error {
+	if len(librariesToRelease) > 0 {
+		updateLibraryMetadata(librariesToRelease, pipelineState, repoPath)
```
jskeet, line 433: I think we should update the pipeline state on disk as we create the other commits - so that each commit is self-contained. That way:

- If we revert a commit due to testing failing (in a later change) we stay consistent
- Each library release commit is consistent in its versioning, so it doesn't matter what the release build uses to read the version (it could use the pipeline state or the in-language version)
- If the PR is reviewed manually, a maintainer can drop any commit for a library they don't want to release, and the PR itself still remains valid.

internal/command/command.go
```diff
@@ -355,10 +356,246 @@ var CmdUpdateApis = &Command{
...
+
+/*
+this goes through each library in pipeline state and checks if any new commits have been added to that library since previous commit tag
+*/
+func generateReleaseCommitForEachLibrary(ctx context.Context, repoPath string, repo *gitrepo.Repo, inputDirectory string, pipelineState *statepb.PipelineState) (map[string]string, string, error) {
+	libraries := pipelineState.LibraryReleaseStates
+	var prDescription string
+	var librariesToRelease map[string]string
+	librariesToRelease = make(map[string]string)
+	for i := 0; i < len(libraries); i++ {
```
jskeet, line 464: Use range instead? `for _, library : range libraries`? (Ditto below for the source paths.)

internal/gitrepo/gitrepo.go
```diff
@@ -284,6 +287,22 @@ func GetApiCommits(ctx context.Context, repo *Repo, path string, commit string)
 	return commits, nil
 }
 
+// Returns all commits since tagName that contains files in path
+func GetApiCommitsSinceTagForSource(repo *Repo, path, tagName string) ([]object.Commit, error) {
```
jskeet, line 291: Let's rename this (and the above) to remove "Api" - in this particular case it's not API commits at all, but language repo commits.

internal/gitrepo/gitrepo.go
```diff
@@ -268,7 +271,7 @@ func GetApiCommits(ctx context.Context, repo *Repo, path string, commit string)
 		}
 		parentPathEntry, err := parentTree.FindEntry(path)
 		if err != nil {
-			return err
+			//return err
```
jskeet, line 274: Revert?

internal/gitrepo/gitrepo.go
```diff
@@ -224,10 +224,13 @@ func PrintStatus(ctx context.Context, repo *Repo) error {
 // Returns the commits in an API rooted at the given path,
 // stopping looking at the given commit (which is not included in the results).
 // The returned commits are ordered such that the most recent commit is first.
-func GetApiCommits(ctx context.Context, repo *Repo, path string, commit string) ([]object.Commit, error) {
+func GetApiCommits(repo *Repo, path string, commit string, retrieveAfterTimestamp *time.Time) ([]object.Commit, error) {
```
jskeet, line 227: I suspect we should refactor this - let's work out what querying we want to support, then design the functions we want.

(I think a function of "does commit X affect path Y" would be useful - and then a method to find all commits matching a given predicate, with another predicate as a stopping point.)

internal/command/command.go
```diff
@@ -355,10 +356,246 @@ var CmdUpdateApis = &Command{
...
+			return err
+		}
+
+		return generateReleasePr(ctx, librariesToRelease, pipelineState, repoPath, languageRepo, prDescription)
+	},
+}
+
+func setupLanguageAndInputFolders(ctx context.Context) (*gitrepo.Repo, string, string, error) {
+	startOfRun := time.Now()
+	tmpRoot, err := createTmpWorkingRoot(startOfRun)
```
jskeet, line 393: Return an error if err != nil.

internal/command/command.go
```diff
@@ -355,10 +356,246 @@ var CmdUpdateApis = &Command{
...
+*/
+func generateReleaseCommitForEachLibrary(ctx context.Context, repoPath string, repo *gitrepo.Repo, inputDirectory string, pipelineState *statepb.PipelineState) (map[string]string, string, error) {
+	libraries := pipelineState.LibraryReleaseStates
+	var prDescription string
+	var librariesToRelease map[string]string
+	librariesToRelease = make(map[string]string)
+	for i := 0; i < len(libraries); i++ {
+		library := libraries[i]
+		var commitMessages []string
+		//TODO: need to add common paths as well as refactor to see if can check all paths at 1 x
```
jskeet, line 467: I suggest we find "all commits since the last release" first, then filter those commits by paths.

internal/command/command.go
```diff
@@ -355,10 +356,246 @@ var CmdUpdateApis = &Command{
...
+		if len(commitMessages) > 0 && isReleaseWorthy(commitMessages) {
+			releaseNotes, err := createReleaseNotes(library, commitMessages, inputDirectory)
+			releaseVersion, err := calculateNextVersion(library)
+
+			//
+			if err := container.UpdateReleaseMetadata(flagImage, repoPath, inputDirectory, library.Id, releaseVersion); err != nil {
+				slog.Info(fmt.Sprintf("Received error running container: '%s'", err))
+
+			}
+			//TODO: make this configurable so we don't have to run per library
```
jskeet, line 490: Having fixed the .NET container so that it only builds the right library, I wouldn't expect we'd need this. Even when testing, I'd hope the build should be quick enough.

internal/command/command.go
```diff
@@ -355,10 +356,246 @@ var CmdUpdateApis = &Command{
...
+		for j := 0; j < len(library.SourcePaths); j++ {
+			//TODO: figure out generic logic
+			previousReleaseTag := library.Id + "-" + library.CurrentVersion
+			commits, err := gitrepo.GetApiCommitsSinceTagForSource(repo, library.SourcePaths[j], previousReleaseTag)
+			if err != nil {
+				fmt.Println("Error searching commits:", err)
+				//TODO update PR description with this data and mark as not humanly resolvable
+			}
+			for _, commit := range commits {
+				commitMessages = append(commitMessages, commit.Message)
```
jskeet, line 477: I think it might be better to handle this as:

- Find all distinct commits which affect the paths for this library
- Parse the commit messages for all commits into some useful structure (which may well include "don't consider this commit for library X" or similar)
- Use the parsed structure to determine whether or not it's worth a release (and ideally the new version)
- Create release notes from the parsed structure

internal/command/command.go
```diff
@@ -355,10 +356,246 @@ var CmdUpdateApis = &Command{
...
+			err = os.Remove(filepath.Join(inputDirectory, "release-notes.txt"))
+			if err != nil {
+				return nil, "", err
+			}
+		}
+	}
+	return librariesToRelease, prDescription, nil
+}
+
+func createCommit(ctx context.Context, err error, repo *gitrepo.Repo, releaseNotes string) {
```
jskeet, line 514: Not sure why you're passing err in here... but we should probably return an error.

internal/command/command.go
```diff
@@ -355,10 +356,246 @@ var CmdUpdateApis = &Command{
...
+	_, err = gitrepo.AddAll(ctx, repo)
+	if err != nil {
+		fmt.Println("Error adding files:", err)
+		//TODO update PR description with this data and mark as not humanly resolvable
+	}
+	//TODO: what should the description be here?
```
jskeet, line 520: I think the description should be "Release library X version Y" and then the release notes.

internal/command/command.go
```diff
@@ -355,10 +356,246 @@ var CmdUpdateApis = &Command{
...
+	if err := gitrepo.Commit(ctx, repo, releaseNotes); err != nil {
+		slog.Info(fmt.Sprintf("Received error trying to commit: '%s'", err))
+		//TODO update PR description with this data and mark as not humanly resolvable
+	}
+}
+
+// TODO: update with actual logic
+func createReleaseNotes(library *statepb.LibraryReleaseState, commitMessages []string, inputDirectory string) (string, error) {
+	releaseNotes := fmt.Sprintf("Library: %s\n", library.Id)
```
jskeet, line 529: I'd suggest the release notes file shouldn't include this - it should just be the changes.
(I realize all of this will change.)

internal/command/command.go
```diff
@@ -355,10 +356,246 @@ var CmdUpdateApis = &Command{
...
+
+	for _, commitMessage := range commitMessages {
+		releaseNotes += fmt.Sprintf("%s\n", commitMessage)
+	}
+
+	path := filepath.Join(inputDirectory, "release-notes.txt")
```
jskeet, line 535: I'd suggest we create a file per library, e.g. {library-id}-{version}-release-notes.txt. That way when running locally, it's easy to see the results.

internal/command/command.go
jskeet: Much later on: it looks like this is only used for release notes, so let's call it that :)

internal/command/command.go
```diff
@@ -355,10 +356,246 @@ var CmdUpdateApis = &Command{
...
+	}
+	return releaseNotes, nil
+}
+
+// TODO: need to add handling of suffix
+func calculateNextVersion(library *statepb.LibraryReleaseState) (string, error) {
+	if library.NextVersion != "" {
+		return library.NextVersion, nil
+	}
+	parts := strings.Split(library.CurrentVersion, ".")
```
jskeet, line 553: Let's see if there's a SemVer library we can use to avoid reinventing the wheel.
https://pkg.go.dev/golang.org/x/mod/semver is encouraging.

internal/command/command.go
```diff
@@ -355,10 +356,246 @@ var CmdUpdateApis = &Command{
...
+
+	//increment minor version
+	minor++
+
+	return fmt.Sprintf("%d.%d.%s", major, minor, patch), nil
+}
+
+func updateLibraryMetadata(releasedLibraries map[string]string, pipelineState *statepb.PipelineState, repoPath string) {
+	for i := 0; i < len(pipelineState.LibraryReleaseStates); i++ {
+		library := pipelineState.LibraryReleaseStates[i]
```
jskeet, line 582: As noted before, let's do this within generateReleaseCommitForEachLibrary. We should also copy the "last generated commit" for each API into the library's metadata.

jskeet (commented):
Lots of comments, but I'm okay with us getting something in and then we can alternate doing refactoring (and genuine functionality-change) PRs based on our time zones. Let's chat about next steps later today.

internal/command/command.go
ldetmer: originally i had that, but i wasn't sure if we wanted to make it more dynamic so we could add other stuff if necessary?  if so I can leave a comment, if not I can make it more specific.

internal/command/command.go
ldetmer: renamed, just so you could see changes i left as is, but i would like to pull all these functions out to a separate create_release_pr.go file

internal/command/command.go
ldetmer: i will clear this file out on every container call though because I want to only pass the container the release notes for that library - how about having the inputDirectory be labeled {library-id}-{version} ?

jskeet: Yup, we could do that. I don't think there's very much risk of containers taking advantage of there being more release notes available than just the one for the library it's being asked to handle.

ldetmer: ok fair enough, i'll update file name

internal/command/command.go
ldetmer: noted, but defering as i need to do some research, ideally we would get a library that provides the minor version

internal/command/command.go
jskeet: I'm happy for you to do all the refactoring you'd like, and I can review afresh tomorrow. (I won't be able to look through everything tonight anyway.)

Please let me know if you're happy for *me* to merge your PR tomorrow morning UK time if I don't have any other blocking comments - I'd then be able to work on other changes to build on top of it.

internal/command/command.go
ldetmer: noted, added todo

internal/command/command.go
ldetmer: makes sense, will work on fixing up

internal/gitrepo/gitrepo.go
ldetmer: will be part of refactoring session

internal/command/command.go
ldetmer: i'm fine with you merging, but there are a couple of major changes still

1) filter "all commits since the last release" first, then filter those commits by paths.
2) change the way i'm processing commits
3) move all the generate PR logic to a separate class
4) fix semvar logic

jskeet (approved):
Merging as-is, to unblock further work in the UK time zone.

jskeet (commented):
Not sure why it thinks there are unresolved comments...

PR #96, opened 2025-03-21
feat: Parse commit messages and format them as release notes

jskeet:
Lots more to do here, such as:
- Including a link to the commit for each line
- Deduping (only use each commit once, and if multiple commits have the same line, have a single line and link to all the commits)
- Highlight breaking changes
- Maybe report dependency updates

This commit also fails the whole process if we fail to commit a change. (That's really fatal.) We should only "undo and continue" if *build* fails - in a later change, we should move the build to after the commit.

### DIFF
--- a/internal/command/commitmessage.go
+++ b/internal/command/commitmessage.go
@@ -1,0 +1,143 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


internal/command/commitmessage.go
```diff
@@ -0,0 +1,143 @@
...
+		CommitHash:         commit.Hash.String(),
+	}
+}
+
+// Returns whether this commit should trigger a release for the given library.
+func IsReleaseWorthy(message *CommitMessage, libraryId string) bool {
+	if slices.Contains(message.NoTriggerLibraries, libraryId) {
+		return false
+	}
+	if slices.Contains(message.NoTriggerLibraries, libraryId) {
```
ldetmer, line 131: should this be message.triggerLibraries?

ldetmer (approved):
looks great! just left one comment about logic

internal/command/commitmessage.go
jskeet: Thank you! Yes, absolutely :)

PR #97, opened 2025-03-21
chore: add handling of errors when creating a release PR

ldetmer:
this adds text to release PR if an error occurs when trying to generate any library, it also adds the do-not-merge label

example 
https://github.com/ldetmer/google-cloud-dotnet/pull/84

### DIFF
--- a/internal/gitrepo/gitrepo.go
+++ b/internal/gitrepo/gitrepo.go
@@ -38,6 +38,10 @@ type Repo struct {
 	repo *git.Repository
 }
 
+type PullRequestMetadata struct {
+	Number int
+}
+
 // CloneOrOpen provides access to a Git repository.
 //
 // If a repository already exists at the specified directory path (dirpath),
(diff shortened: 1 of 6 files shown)


internal/gitrepo/gitrepo.go
```diff
@@ -361,9 +360,52 @@ func CreatePullRequest(ctx context.Context, repo *Repo, remoteBranch string, acc
...
+	labels := []string{label}
+
+	_, _, err = gitHubClient.Issues.AddLabelsToIssue(ctx, organization, repoName, prNumber, labels)
+	if err != nil {
+		return fmt.Errorf("failed to add label: %w", err)
+	}
 	return nil
 }
+
+func CleanWorkingTree(repo *Repo) error {
```
jskeet, line 392: We should work out the differences between this and the existing ResetHard call. It could be that this is a better option and we should remove ResetHard. (In particular, I suspect that doesn't remove files, so it's effectively a little broken at the moment...)

jskeet (approved):
I like it :)

internal/gitrepo/gitrepo.go
ldetmer: i think we still need both.  I added API docs, but basically for me, I just want to drop the local changes, not all the previous local commits that haven't been pushed yet in this release PR, but for resetHard you want to drop any local changes and match up with origin.  Let me know if I misunderstood your use case and I'll be happy to update

jskeet: I think in both cases we want to reset the working copy and index to a particular commit so that it's clean. In some cases that can be "HEAD" and in some cases it can be an older commit. My existing use of ResetHard is only trying to drop the local changes rather than earlier commits (as I'm not committing anything on the repo that I'm calling it on - it's on googleapis). I think we can probably drop my reset and keep clean.

PR #98, opened 2025-03-24
feat: Add update-image-tag command

jskeet:


internal/command/updateimagetag.go
```diff
@@ -0,0 +1,190 @@
...
+
+		return push(ctx, languageRepo, startOfRun, "chore: update generation image tag", "")
+	},
+}
+
+func regenerateApi(ctx context.Context, apiRepo *gitrepo.Repo, languageRepo *gitrepo.Repo, generatorInput string, image string, outputRoot string, apiState *statepb.ApiGenerationState) error {
+	if err := gitrepo.Checkout(ctx, apiRepo, apiState.LastGeneratedCommit); err != nil {
+		return err
+	}
+	slog.Info(fmt.Sprintf("Generating '%s'", apiState.Id))
```
zhumin8, line 169: Just want to confirm my understanding: I might missed this in the docs, is "AutomationLevel" only affect [updateapi](https://github.com/googleapis/librarian/blob/aa7b45bc089c835ca90ac8125e28aa19e31fe9b7/internal/command/updateapis.go#L175-L178), but not updateimagetag? iiuc, this means even if automation level is blocked, we still want to update the generated code for generator change, just not api proto changes?

internal/command/updateimagetag.go
```diff
@@ -0,0 +1,190 @@
...
+
+		// Take a defensive copy of the generator input directory from the language repo.
+		generatorInput := filepath.Join(tmpRoot, "generator-input")
+		if err := os.CopyFS(generatorInput, os.DirFS(filepath.Join(languageRepo.Dir, "generator-input"))); err != nil {
+			return err
+		}
+
+		// Perform "generate, clean" on each element in ApiGenerationStates.
+		for _, apiState := range state.ApiGenerationStates {
+			err = regenerateApi(ctx, apiRepo, languageRepo, generatorInput, image, outputDir, apiState)
```
zhumin8, line 138: I wonder why to regenerate for each api instead of using container.generate() for all? Is it because we do not record a global equivalent to apiState.LastGeneratedCommit in the pipeline-state file?

zhumin8 (approved):
LGTM. Got two questions, but don't think they are blocking

internal/command/updateimagetag.go
```diff
@@ -0,0 +1,190 @@
...
+				return err
+			}
+			clean, err := gitrepo.IsClean(ctx, apiRepo)
+			if err != nil {
+				return err
+			}
+			if !clean {
+				return errors.New("api repo must be clean before updating the language image tag")
+			}
+		}
```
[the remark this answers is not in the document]
jskeet: At some point, potentially. I think we may wait until At some point, potentially. I think we may wait until we've got all the CLI commands implemented - so the interface is stable - and then do it.

internal/command/updateimagetag.go
jskeet: It's because we want to regenerate at whatever commit it was previously generated. It could be that "latest" is broken for some APIs, for example.

internal/command/updateimagetag.go
jskeet: Yes, correct. Even if an API is blocked, the commit should refer to the last *successful* generation, so we'd expect it to be okay with the new image.

jskeet (commented):
Thanks for the review - will merge.

jskeet:
Humbug, it looks like you're not configured in the team at the moment @zhumin8 - will see how easy it is to fix that...

burkedavison (approved):

PR #99, opened 2025-03-25
feat: split generation into generate-raw and generate-library

jskeet:
Recommended to review one commit at a time.

This fixes some issues that I found along the way - but primarily it's about making commands which deal with a repo (as opposed to "raw" generation) more library-centric.

For .NET, the changes required are in https://github.com/googleapis/google-cloud-dotnet/pull/14521


zhumin8 (approved):
LGTM. 
I haven't gone through some changes in detail (createreleasepr.go and cofigure.go that I am less familiar), but also do not want to block your work. I will take another pass tomorrow, feel free to merge if you need.

jskeet:
Merging mostly to unblock Laura who will need to adapt #97

PR #111, opened 2025-04-02
feat: remove output flag to simplify things

jskeet:
This also refactors the configure code for consistency.


ldetmer:
@zhumin8  FYI we're removing the output dir flag, in case you're using that for java (it'll now be work-root + "/output"

jskeet:
> @zhumin8 FYI we're removing the output dir flag, in case you're using that for java (it'll now be work-root + "/output"

It's still going to be in the container, btw. It's just not in the CLI interface.

PR #116, opened 2025-04-07
feature: add storage of env vars to file in release-please command

ldetmer:
additionally add the integration tests to release-please command

### DIFF
--- a/internal/command/command.go
+++ b/internal/command/command.go
@@ -264,6 +264,7 @@ func init() {
 		addFlagGitUserName,
 		addFlagRepoRoot,
 		addFlagSkipBuild,
+		addFlagEnvFile,
 	} {
 		fn(fs)
 	}
--- a/internal/command/command.go
+++ b/internal/command/command.go
@@ -31,7 +31,7 @@ import (
 	"google.golang.org/protobuf/encoding/protojson"
 )
 
-const releaseIdEnvVarName = "RELEASE_ID"
+const releaseIDEnvVarName = "RELEASE_ID"
 
 type Command struct {
 	Name  string
--- a/internal/command/command.go
+++ b/internal/command/command.go
@@ -31,7 +31,7 @@ import (
 	"google.golang.org/protobuf/encoding/protojson"
 )
 
-const RELEASE_ID_ENV_VAR_NAME = "RELEASE_ID"
+const releaseIdEnvVarName = "RELEASE_ID"
 
 type Command struct {
 	Name  string
--- a/internal/command/command.go
+++ b/internal/command/command.go
@@ -31,6 +31,8 @@ import (
 	"google.golang.org/protobuf/encoding/protojson"
 )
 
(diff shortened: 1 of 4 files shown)


internal/command/command.go
```diff
@@ -31,6 +31,8 @@ import (
 	"google.golang.org/protobuf/encoding/protojson"
 )
 
+const releaseIdEnvVarName = "RELEASE_ID"
```
jskeet, line 34: Not sure if this should conventionally be releaseIDEnvVarName - I'm fine with it having a lower 'd' for now though.

internal/command/createreleasepr.go
```diff
@@ -66,6 +72,7 @@ var CmdCreateReleasePR = &Command{
 		}
 
 		releaseID := fmt.Sprintf("release-%s", formatTimestamp(startOfRun))
+		utils.AppendToFile(flagEnvFile, fmt.Sprintf("%s:%s\n", releaseIdEnvVarName, releaseID))
```
jskeet, line 75: This uses a colon - the other one uses an equals sign. Presumably they should be the same? Not sure which though.

internal/command/flags.go
```diff
@@ -59,6 +60,10 @@ func addFlagBuild(fs *flag.FlagSet) {
 	fs.BoolVar(&flagBuild, "build", false, "whether to build the generated code")
 }
 
+func addFlagEnvFile(fs *flag.FlagSet) {
```
jskeet, line 63: Perhaps this should be optional? (When running locally for testing purposes it's not really needed?) Happy either way though.

internal/utils/fileutils.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"os"
+	"path/filepath"
+)
+
+// AppendToFile writes the content to a file in the specified filePath.
+// It creates the file if it does not exist, otherwise it appends to existing file.
+func AppendToFile(filePath string, content string) error {
+	file, err := os.OpenFile(filePath, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0666)
```
jskeet, line 25: I suspect we need to do something to close the file handle afterwards (now and below). I can't remember the language construct off-hand...

internal/command/command.go
ldetmer: changed to D

internal/command/createreleasepr.go
ldetmer: thanks for catching, should be =

internal/command/flags.go
ldetmer: i changed it to be optional, but made default /tmp/env-vars.txt, let me know if that sounds okay to you

internal/utils/fileutils.go
ldetmer: good call, added

jskeet (approved):
Had meant that we needn't write the file at all if the flag wasn't specified - but using /tmp is fine too. (Or we can put it in the working root - all of this is fine.)

PR #118, opened 2025-04-08
feat: Provide environment variables to container commands

jskeet:
This integrates with Secret Manager so that environment variable values in the container can either come from secrets or from host environment variables.


internal/container/container.go
```diff
@@ -238,9 +230,17 @@ func runDocker(image string, mounts []string, containerArgs []string) error {
 	for _, mount := range mounts {
 		args = append(args, "-v", mount)
 	}
-
+	if containerEnv != nil {
+		if err := writeEnvironmentFile(containerEnv, commandName); err != nil {
+			return err
+		}
+		args = append(args, "--env-file")
```
ldetmer, line 237: maybe rename this arg to be "--container-env-file"?  since we're using --env-file to pass in the output env vars location to the librarian? so its obvious they are different files?

proto/pipeline.proto
```diff
@@ -97,3 +98,31 @@ enum AutomationLevel {
...
+// An environment variable to be provided to a container command.
+message CommandEnvironmentVariable {
+  // The name of the environment variable (e.g. TEST_PROJECT).
+  string name = 1;
+
+  // The name of the secret to be used to fetch the value of the environment
+  // variable when it's not present in the host system. If this is not specified,
+  // or if a Secret Manager project has not been provided to Librarian,
+  // any missing environment variables are skipped.
+  string secret_name = 2;
```
ldetmer, line 127: this is very specific to env vars that are secrets.  Do you think we'll want to reuse this for other env vars? maybe worth it to add a type, and if the type is secret we use this field to look up, otherwise there could be other logic.  If this is only meant for secrets then maybe rename to CommandSecretsEnvironmentVariable? if you think its easier to add these changes when we need them, that's fine.

internal/container/environment.go
```diff
@@ -0,0 +1,116 @@
...
+		return "# No environment variables", nil
+	}
+	var builder strings.Builder
+	for _, variable := range commandConfig.EnvironmentVariables {
+		value, present := os.LookupEnv(variable.Name)
+		if present {
+			builder.WriteString(fmt.Sprintf("%s=%s\n", variable.Name, value))
+			continue
+		}
+		if containerEnv.secretManagerClient == nil {
```
ldetmer, line 90: this is where having a type might be helpful, if it is type secret and this is null should we return an error.

ldetmer (approved):
cool implementation.  I just left some comments for thought, the only one I feel strongly about is the -env-file rename, but happy to discuss

internal/container/container.go
jskeet: This is the docker run command. env-file is the name of the parameter docker uses - we can't change that. https://docs.docker.com/reference/cli/docker/container/run/#env

proto/pipeline.proto
jskeet: I've been discussing with Amanda that we could use Parameter Manager as another source - I don't think we need to specify a type particularly, as we can just have a precedence (like we have "existing environment variable takes precedence over secret") or have a oneof with secret_name and parameter_name for example, to stop both being specified. (Or *maybe* it's okay to specify both sources, because one caller might want to put the value in Secret Manager and another in Parameter Manager...)

internal/container/environment.go
jskeet: The choice at the moment leaves it to the container to decide whether the value is actually required or not. We could add required/optional in the config file, but I think it's probably simpler to leave it up to the container (especially as it may only be required for some APIs).

jskeet:
Merging as-is - we can chat more about other changes tomorrow.

PR #121, opened 2025-04-09
refactor: Refactor command implementations to reduce duplication

### LINKED ISSUE #117: Refactor commands to extract common code
Everything creates a working root, lots of code loads the pipeline state, etc. We should make that common.

jskeet:
All commands create a work-root and do language validation; many also obtain a language repo by opening/cloning. This change puts all common information into a CommandContext struct, setting it up in one place and then passing it into an execute function.

Further refactoring may be desirable for container execution, to set up the image and environment aspects within a single struct that is then passed into all the container functions.

Fixes #117.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/googleapis/librarian
 
-go 1.23.6
+go 1.23.8
 
 require github.com/go-git/go-git/v5 v5.14.0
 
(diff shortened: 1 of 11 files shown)


jskeet:
The test error is due to a vulnerability, which I'll fix in a separate PR first.

PR #140, opened 2025-04-24
feat: Handle update-apis in the same way as configure/create-release-pr

jskeet:
Generation/build errors should be reported, but should not prevent other libraries from being generated.

Refactored the existing code to remove duplication.

### DIFF
--- a/internal/command/createreleasepr.go
+++ b/internal/command/createreleasepr.go
@@ -15,7 +15,6 @@
 package command
 
 import (
-	"errors"
 	"flag"
 	"fmt"
 	"log/slog"
@@ -36,11 +35,6 @@ import (
 
 const prNumberEnvVarName = "_PR_NUMBER"
 const baselineCommitEnvVarName = "_BASELINE_COMMIT"
-
-type ReleasePrDescription struct {
-	Releases []string
-	Errors   []string
-}
 
 var CmdCreateReleasePR = &Command{
 	Name:  "create-release-pr",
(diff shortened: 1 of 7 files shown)


internal/command/updateapis.go
```diff
@@ -200,16 +186,26 @@ func updateLibrary(ctx *CommandContext, apiRepo *gitrepo.Repo, generatorInput st
 	}
 
 	// Once we've committed, we can build - but then check that nothing has changed afterwards.
-	if err := container.BuildLibrary(containerConfig, languageRepo.Dir, library.Id); err != nil {
-		return err
-	}
+	// We consider a "something changed" error as fatal, whereas a build error just needs to
```
ldetmer, line 189: why is this considered fatal?  why wouldn't we also log error and revert commit?

internal/command/createreleasepr.go
```diff
@@ -86,81 +80,47 @@ var CmdCreateReleasePR = &Command{
...
-	if prMetadata != nil {
-		if err := appendResultEnvironmentVariable(ctx, prNumberEnvVarName, strconv.Itoa(prMetadata.Number)); err != nil {
+		prMetadata, err := createPullRequest(ctx, prContent, "chore: Library release", "release")
+		if err != nil {
 			return err
 		}
-	}
-	return nil
+		// If we've actually created a release PR, we have more to do.
+		// (We may have finished with no changes, or we may not be pushing.)
```
ldetmer, line 97: we can remove this comment right? you're taking care of this logic later on in the createPullRequestMethod?

ldetmer (commented):
This is a nice refactor!

ldetmer (approved):

internal/command/createreleasepr.go
jskeet: Editing comment for clarity.

internal/command/updateapis.go
jskeet: (As discussed offline) It's fatal because it's a bug in the container rather than probably a problem with an API.

PR #142, opened 2025-04-24
chore: update generate command to call language container generate library command if library exists for given API

ldetmer:


internal/command/generate.go
```diff
@@ -71,3 +85,27 @@ var CmdGenerate = &Command{
 		return nil
 	},
 }
+
+// Checks if the library with the given API path exists in the remote pipeline state
+func checkIfLibraryExistsInLanguageRepo(ctx *CommandContext) string {
+	if flagRepoUrl == "" {
```
jskeet, line 91: I think we should probably handle repo-root as well, but that could come later. Maybe file an issue and I'll look at it another time?

internal/command/generate.go
```diff
@@ -47,19 +49,31 @@ var CmdGenerate = &Command{
...
 
-		if err := container.GenerateRaw(ctx.containerConfig, apiRoot, outputDir, flagAPIPath); err != nil {
+		// check if the library exists in the remote pipeline state, if so use GenerateLibrary command
+		// otherwise use GenerateRaw command
+		// In case of error when looking up library, we will fallback to GenerateRaw command
+		// and log the error
+		libraryId := checkIfLibraryExistsInLanguageRepo(ctx)
+		if libraryId != "" {
+			ctx.languageRepo, err = cloneOrOpenLanguageRepo(ctx.workRoot)
+			slog.Warn("Unable to checkout language repo ", "error", err)
```
jskeet, line 69: I think you're missing a condition here :) (And returning an error if we can't check out the repo.)

internal/command/generate.go
```diff
@@ -47,19 +49,31 @@ var CmdGenerate = &Command{
...
+			generatorInput := filepath.Join(ctx.languageRepo.Dir, "generator-input")
+			err = container.GenerateLibrary(ctx.containerConfig, apiRoot, outputDir, generatorInput, libraryId)
```
jskeet, line 71: Let's just use "return container.GenerateLibrary" here and the equivalent below, and get rid of the final bits.

internal/command/generate.go
```diff
@@ -47,19 +49,31 @@ var CmdGenerate = &Command{
...
+		} else {
+			slog.Info("No matching library found for ", "flagAPIPath", flagAPIPath)
```
jskeet, line 73: Maybe say " - performing raw generation" at the end of this?

internal/command/generate.go
```diff
@@ -47,19 +49,31 @@ var CmdGenerate = &Command{
...
 			return err
 		}
 		slog.Info(fmt.Sprintf("Code will be generated in %s", outputDir))
 
-		if err := container.GenerateRaw(ctx.containerConfig, apiRoot, outputDir, flagAPIPath); err != nil {
+		// check if the library exists in the remote pipeline state, if so use GenerateLibrary command
+		// otherwise use GenerateRaw command
+		// In case of error when looking up library, we will fallback to GenerateRaw command
+		// and log the error
+		libraryId := checkIfLibraryExistsInLanguageRepo(ctx)
```
jskeet, line 66: I think Go conventions would suggest libraryID here

internal/command/generate.go
```diff
@@ -47,19 +49,31 @@ var CmdGenerate = &Command{
...
+		if libraryId != "" {
+			ctx.languageRepo, err = cloneOrOpenLanguageRepo(ctx.workRoot)
+			slog.Warn("Unable to checkout language repo ", "error", err)
+			generatorInput := filepath.Join(ctx.languageRepo.Dir, "generator-input")
```
jskeet, line 70: Add an equivalent slog.Info here to say "Performing refined generation for library {libraryId}"? (That way however the generation happens, it's clear what's going on before we get to the container part.)

jskeet (commented):
Various nits, but just to confirm, this is definitely what I was expecting :)

internal/command/generate.go
ldetmer: we still need to build after this, so can't return directly, but I did a refactor to try to make this more readable

PR #153, opened 2025-05-07
feat: Add -skip-integration-tests flag

jskeet:
This must be set to a bug number (although we only validate that it starts with b/) and can be specified in create-release-pr and create-release-artifacts.

This commit removes the -skip-build flag; I believe we *always* want to build when creating a release PR. (Whether we run unit tests or not is potentially separate.)


jskeet:
Fixes b/416172658 (from a Librarian perspective, anyway)

ldetmer (approved):

PR #161, opened 2025-05-12
fix: add additional packages necessary for code generation for librarian

ldetmer:
this allows us to use librarian image when running tap presubmits

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -65,7 +65,5 @@ RUN echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] https://packages.
 # Install the gcloud CLI
 RUN apt-get update && apt-get install -y google-cloud-sdk
 
-RUN gcloud --version
-
 COPY --from=build /src/librarian .
 ENTRYPOINT ["/app/librarian"]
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -28,13 +28,20 @@ RUN CGO_ENABLED=0 GOOS=linux go build ./cmd/librarian
...
 RUN apt update
-RUN apt install -y ca-certificates curl
+RUN apt-get install -y unzip \
+  gnupg \
+  apt-transport-https \
+  ca-certificates \
+  curl \
+  && rm -rf /var/lib/apt/lists/*
+
+  # Add Docker's official GPG key
```
jskeet, line 44: Nit: outdent the comment.

Dockerfile
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.23 AS build
+FROM golang:1.24 AS build
```
jskeet, line 15: If we're requiring Go 1.24 here, I wonder whether we should do so in go.mod as well.

jskeet (approved):

Dockerfile
ldetmer: will open separate PR for that

PR #166, opened 2025-05-15
feat: Make merge-release-pr a one-stop-merging-shop

jskeet:
This now polls (once per minute) for all the conditions we want to enforce until it's ready to merge, then merges. See the comment on waitForPullRequestReadinessSingleIteration for details of the conditions that are enforced.

### DIFF
--- a/internal/command/mergereleasepr.go
+++ b/internal/command/mergereleasepr.go
@@ -21,12 +21,22 @@ import (
 	"log/slog"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/google/go-github/v69/github"
 	"github.com/googleapis/librarian/internal/githubrepo"
 	"github.com/googleapis/librarian/internal/gitrepo"
 	"github.com/googleapis/librarian/internal/statepb"
 )
+
+// The label used to avoid users merging the PR themselves.
+const DoNotMergeLabel = "do-not-merge"
+const DoNotMergeAppId = 91138
+const ConventionalCommitsAppId = 37172
+
+// The label used to indicate "I've noticed a problem with this PR; I won't check it again
+// until you've done something".
+const MergeBlockedLabel = "merge-blocked-see-comments"
 
 var CmdMergeReleasePR = &Command{
 	Name:  "merge-release-pr",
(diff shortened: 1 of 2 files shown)


internal/command/mergereleasepr.go
```diff
@@ -67,40 +77,198 @@ func mergeReleasePRImpl(ctx *CommandContext) error {
...
+}
+
+// A single iteration of the loop in waitForPullRequestReadiness,
+// in a separate function to make it easy to indicate an "early out".
+// Returns true for "PR is ready to merge", false for "keep polling".
+// If this function returns false (with no error) the reason will have been logged.
+// Checks performed:
+// - The PR must not be merged
+// - The PR must not have the label with the name specified in MergeBlockedLabel
+// - The PR must have the label with the name specified in DoNotMergeLabel
```
ldetmer, line 114: do we need to have this check? if we don't have donotmerge label does it prevent the flow from continuing?

internal/command/mergereleasepr.go
```diff
@@ -67,40 +77,198 @@ func mergeReleasePRImpl(ctx *CommandContext) error {
...
+func waitForPullRequestReadinessSingleIteration(ctx *CommandContext, prMetadata githubrepo.PullRequestMetadata) (bool, error) {
+	slog.Info("Checking pull request for readiness")
+	pr, err := githubrepo.GetPullRequest(ctx.ctx, prMetadata.Repo, prMetadata.Number)
+	if err != nil {
+		return false, err
+	}
+
+	// If the PR has been merged by someone else, abort this command. (We can skip this step in the flow, if we still want to release.)
+	if pr.GetMerged() {
+		return false, errors.New("pull request already merged")
```
ldetmer, line 130: we're failing because we want someone to confirm its still okay to release?

internal/command/mergereleasepr.go
```diff
@@ -67,40 +77,198 @@ func mergeReleasePRImpl(ctx *CommandContext) error {
...
+	}
+	for _, checkRun := range checkRuns {
+		// Skip the do-not-merge check and conventional commits checks
+		// (Once b/416489721 has been fixed, we can remove the conventional commits check)
+		if checkRun.GetApp().GetID() == DoNotMergeAppId || checkRun.GetApp().GetID() == ConventionalCommitsAppId {
+			continue
+		}
+
+		// For now, we assume that every other check must be complete and successful.
+		// We can't get at the required status checks with the current access token;
```
ldetmer, line 169: this is definitely going to be a problem for java (we have many failing non required checks), but if it works for you for now, we can fix it for the next language

ldetmer (approved):

internal/command/mergereleasepr.go
jskeet: Currently it would make the operation of removing the do-not-merge label fail. (We do that just before merging.)
Basically we should get everyone in the habit of leaving that, I think. But we can tweak all of this in the future, of course.

internal/command/mergereleasepr.go
jskeet: It's more because we don't know what's going on at this stage. If whoever merged it manually wants to skip this step of the flow and go onto the next one, they can do so (although we'll need to fetch the merge commit separately) but I didn't want to just pass as if this is normal.

Definitely worth getting feedback on this once we've tried it.

internal/command/mergereleasepr.go
jskeet: Sounds good. It's not clear to me what the best way of identifying checks is - for these two, the app ID is enough... but for others it might not be.

PR #167, opened 2025-05-15
feat: Allow merge-release-pr to wait for the merge commit to be synchronized

jskeet:
This is intended to be somewhat general-purpose, so the only URL requirement is that we can put the merged commit hash as a suffix of the supplied prefix.


jskeet:
Marked as do-not-merge because we want to get the previous change definitely working before trying this.

ldetmer (approved):

PR #175, opened 2025-05-19
chore: create the work root if it's specified but doesn't exist

jskeet:
(The duplication here is slightly annoying, but it's different to the version below


jskeet:
Closing - we'll create the directory instead.

PR #178, opened 2025-05-21
feat: improve command help

julieqiu:


cmd/librarian/main.go
```diff
@@ -24,12 +24,6 @@ import (
 
 func main() {
 	ctx := context.Background()
-
-	log.Println("Arguments passed to librarian:")
```
jskeet, line 28: This has been really useful when diagnosing CI stuff - but I understand it's not useful for a lot of interactive work. Would adding a -v option or similar be feasible?

internal/command/createreleaseartifacts.go
```diff
@@ -38,7 +38,7 @@ type LibraryRelease struct {
 
 var CmdCreateReleaseArtifacts = &Command{
 	Name:  "create-release-artifacts",
-	Short: "Create release artifacts from a merged release PR",
+	Short: "Generate release artifacts from a merged PR.",
```
jskeet, line 41: Generate sounds like it's running a generator. This is about building/packaging...

internal/command/publishreleaseartifacts.go
```diff
@@ -31,7 +31,7 @@ import (
 
 var CmdPublishReleaseArtifacts = &Command{
 	Name:  "publish-release-artifacts",
-	Short: "Publishes previously-created release artifacts to package managers etc",
+	Short: "Publish release artifacts to package managers.",
```
jskeet, line 34: I'd like there to be something to indicate that this shouldn't be doing a build.

cmd/librarian/main.go
julieqiu: I added a log line later on if that works?

jskeet (approved):
Yup, this LGTM, thanks :)

PR #192, opened 2025-06-04
impl(internal/command): rename ctx to state

### LINKED ISSUE #189: internal/command: rename all references to `ctx` where it means `commandState` to `state`
See https://github.com/googleapis/librarian/commit/ed7795d942279f7e1f2716be3689db20a4854d00 for context

julieqiu:
The term context and the variable ctx has a specific connotation in Go. See https://go.dev/blog/context-and-structs. Rename references to ctx to state.

Fixes #189

### DIFF
--- a/internal/command/command.go
+++ b/internal/command/command.go
@@ -183,10 +183,10 @@ func RunCommand(c *Command, ctx context.Context) error {
 	return c.execute(cmdContext)
 }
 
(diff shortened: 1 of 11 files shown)


jskeet (approved):
Works for me. Obviously we also have the pipeline state, so there's a clash either way, but it's probably easier to differentiate between command state and pipeline state.

PR #193, opened 2025-06-04
impl(internal/command): move execute logic to separate functions

julieqiu:
Move execution steps into separate helper functions to improve readability for each command, particularly on pkg.go.dev.


jskeet (approved):
I'd been doing this for new commands (with an "impl" name instead, as you've probably seen by now) but hadn't found the time to refactor the existing ones.

Issue #194, opened 2025-06-04
internal/command: merge maybeGetLanguageRepo, maybeLoadStateAndConfig, and execute into a single Run function


julieqiu:
At the moment, the call stack for librarian is:

main --> librarian.Run --> command.RunCommand --> maybeGetLanguageRepo + maybeLoadStateAndConfig + execute

This creates a level of abstraction that requires all commands to follow the abstraction of maybeGetLanguageRepo + maybeLoadStateAndConfig + execute, which isn't always applicable (at the moment, it only applies to 5 out of the 7 existing commands). 

To eliminate the `command.RunCommand` abstraction, change the current Command struct:

```
type Command struct {
	// Name is the unique identifier for the command.
	Name string

	// Short is a concise description shown in the 'librarian -h' output.
	Short string

	// maybeGetLanguageRepo attempts to obtain a language-specific Git
	// repository, cloning if necessary.  Returns nil if not applicable.
	maybeGetLanguageRepo func(workRoot string) (*gitrepo.Repo, error)

	// maybeLoadStateAndConfig attempts to load pipeline state and config, even if no
	// language repo is present.
	maybeLoadStateAndConfig func(languageRepo *gitrepo.Repo) (*statepb.PipelineState, *statepb.PipelineConfig, error)

	// execute runs the command's with the provided context.
	execute func(*commandState) error

	// flagFunctions are functions to initialize the command's flag set.
	flagFunctions []func(fs *flag.FlagSet)

	// flags is the command's flag set for parsing arguments and generating
	// usage messages. This is populated for each command in init().
	flags *flag.FlagSet
}
```

To:

```
type Command struct {
	// Name is the unique identifier for the command.
	Name string

	// Short is a concise description shown in the 'librarian -h' output.
	Short string

	// Run executes the command.
	Run func(ctx context.Context) error

	// flagFunctions are functions to initialize the command's flag set.
	flagFunctions []func(fs *flag.FlagSet)

	// flags is the command's flag set for parsing arguments and generating
	// usage messages. This is populated for each command in init().
	flags *flag.FlagSet
}
```

julieqiu:
I think this is now fixed, so closing the issue to reduce clutter. Feel free to repoen if needed.

PR #200, opened 2025-06-04
impl(internal/command): implement Run for remaining commands

julieqiu:
For #194

### DIFF
--- a/internal/command/command.go
+++ b/internal/command/command.go
@@ -155,6 +155,9 @@ func cloneOrOpenLanguageRepo(workRoot string) (*gitrepo.Repo, error) {
 
 // RunCommand executes a given command, setting up its context including work
 // directory, language repository, pipeline state, and container configuration.
+//
+// TODO(https://github.com/googleapis/librarian/issues/194): remove this
+// function once all code has switched over to using Command.Run.
 func RunCommand(c *Command, ctx context.Context) error {
 	startTime := time.Now()
 	workRoot, err := createWorkRoot(startTime)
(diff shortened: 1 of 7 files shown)


internal/command/command.go
```diff
@@ -189,6 +192,40 @@ func RunCommand(c *Command, ctx context.Context) error {
 	return c.execute(cmdContext)
 }
 
+func createContainerForLanguage(ctx context.Context) (*commandState, error) {
```
jskeet, line 195: Maybe createCommandStateForLanguage? (But again, happy to change later potentially.)

And I'm assuming *this* is how we end up avoiding duplication - because most commands should end up being able to use this.

internal/command/generate.go
```diff
@@ -49,10 +51,48 @@ var CmdGenerate = &Command{
...
+	startTime := time.Now()
+	workRoot, err := createWorkRoot(startTime)
+	if err != nil {
+		return err
+	}
+	workRoot, err = resolveLibraryWorkRoot(workRoot)
+	if err != nil {
+		return err
+	}
+	repo, err := cloneOrOpenLanguageRepo(workRoot)
```
jskeet, line 67: This and the loadRepoStateAndConfig should be conditionalized on resolveLibraryWorkRoot not returning an empty string. (Otherwise, `repo`, `config` and `state` should all be null.)

(And yes, I regret the naming of `resolveLibraryWorkRoot`...)

jskeet (commented):
*Right now* this would break the generate command (so I'm not approving just yet) but it's possible that another chnage later in the stack would fix it anyway, in which case I'll approve at that point (with a suitable warning).

jskeet (approved):
Approving to unblock - we should make the runGenerate change as a very fast-follow.

jskeet:
Filed https://github.com/googleapis/librarian/issues/223 for the changes to runGenerate

internal/command/generate.go
julieqiu: filed an issue.

PR #203, opened 2025-06-04
impl(internal/command): refactor openOrCloneLanguageRepoIfLibraryExists

julieqiu:


internal/command/generate.go
```diff
@@ -120,13 +120,24 @@ func runGenerateCommand(state *commandState, outputDir string) (string, error) {
 // Checks if the library with the given API path exists in the repo specified either
 // by a URL or a local path, and opens or clones it if so.
 func openOrCloneLanguageRepoIfLibraryExists(workRoot string) (*gitrepo.Repo, error) {
+	workRoot, err := resolveLibraryWorkRoot(workRoot)
+	if err != nil {
+		return nil, err
+	}
+	// Otherwise (if the library *does* exist), clone or open it as normal.
```
jskeet, line 127: This is going to try to clone in two situations it shouldn't:

- If neither the repo-url nor the repo-root has been specified (at which point it will fail)
- If the repo-root/url does exist, but the library isn't configured in the repo.

In both cases, we'll end up passing an empty string into cloneOrOpenLanguageRepo as the workRoot. In the latter case, I suspect we could end up cloning into the *current* directory (and then ignoring that).

We could change this function to return `nil, nil` if workRoot is empty at this point, but personally I'm finding this more confusing than before (particularly because we're assigning back to `workRoot`).

This may be a matter of not having had enough coffee though.

jskeet: Testing discussed elsewhere. I'm happy for this to be fixed after everything else is merged, as a fast-follow. (And I can file an issue for that if it's useful.)

jskeet (approved):
Given that my requested change would need to be addressed in different places in at least two commits, I suggest we merge as-is and then fix it immediately afterwards.

internal/command/generate.go
julieqiu: yup, filed an issue

Issue #205, opened 2025-06-04
internal/command: split into two packages for command framework and librarian business logic


julieqiu:
At the moment, internal/command mixes the command framework with the librarian business logic. Ideally, there would be a clear separation between the two to improve modularity, simplify testing, and make it easier to evolve the business logic independently of the command infrastructure.

julieqiu:
We have internal/cli and internal/librarian, which seems good enough for now.

PR #212, opened 2025-06-04
impl(internal/command): split CLI framework from librarian logic

julieqiu:
For https://github.com/googleapis/librarian/issues/205


jskeet:
Ah - I hadn't seen that these PRs are building on top of each other. (I don't think I've ever worked with a PR which wasn't targetting a feature branch or main.) Will re-evaluate everything when I get a chance (but probably not today)

jskeet:
I'd typically create a single PR with multiple commits, and then ideally enable "rebase and merge" as an option in the repo so that we could maintain that set of commits when we merge it (so that it's easier to understand than a single huge commit). Then you can easily see the overall set of changes, or the changes for each individual commit within the PR.

jskeet (approved):
This and #213 each make a lot more sense after seeing that they're intended to be combined :)

PR #217, opened 2025-06-04
chore: Update renovate options

jskeet:
These are based on https://github.com/googleapis/gapic-generator-go/blob/main/renovate.json


julieqiu:
Would you be able to update the commit message to describe the effect of this change? As someone who is not familiar with renovate, I'm not sure what this change does.

julieqiu (approved):

PR #218, opened 2025-06-04
impl(internal/cli): add Command.SetFlags

julieqiu:
Instead of having Flags and FlagFunctions as exported fields, add Command.SetFlags to handle this fields. This makes it clear that there is only one place where they should set (at init), and simplifies the Commands struct.


jskeet (approved):
This PR was a lot easier to understand having learned about `init()` in go :)

PR #232, opened 2025-06-05
fix: Fix generate command to only clone/open repo for configured libraries

### LINKED ISSUE #223: (Post-194) Fix generate command flow for library that isn't configured
#200 and #203 change the behavior of the `generate` command when either the repo-root and repo-url aren't specified, or there's no configured library that generates them.

To avoid complicating a multi-PR merge, we may well want to merge a lot of those and then immediately fix that scenario.

The change should be: If neither repo-url and repo-root or set, *or* if one of them is set but there's no configured library, then we should end up with no languageRepo, pipelineState or pipelineConfig in the CommandState.

### LINKED ISSUE #228: This and the loadRepoStateAndConfig should be conditionalized on resolveLibraryWorkRoot not returning an empty string. (Otherwise, `repo`, `config` and `state` should all be null.)
This and the loadRepoStateAndConfig should be conditionalized on resolveLibraryWorkRoot not returning an empty string. (Otherwise, `repo`, `config` and `state` should all be null.)

(And yes, I regret the naming of `resolveLibraryWorkRoot`...)

_Originally posted by @jskeet in https://github.com/googleapis/librarian/pull/200#discussion_r2126670444_

### LINKED ISSUE #229: This is going to try to clone in two situations it shouldn't:
This is going to try to clone in two situations it shouldn't:

- If neither the repo-url nor the repo-root has been specified (at which point it will fail)
- If the repo-root/url does exist, but the library isn't configured in the repo.

In both cases, we'll end up passing an empty string into cloneOrOpenLanguageRepo as the workRoot. In the latter case, I suspect we could end up cloning into the *current* directory (and then ignoring that).

We could change this function to return `nil, nil` if workRoot is empty at this point, but personally I'm finding this more confusing than before (particularly because we're assigning back to `workRoot`).

This may be a matter of not having had enough coffee though.

_Originally posted by @jskeet in https://github.com/googleapis/librarian/pull/203#discussion_r2125852201_

jskeet:
Fixes #229.
Fixes #228.
Fixes #223.

### DIFF
--- a/internal/librarian/generate.go
+++ b/internal/librarian/generate.go
@@ -27,6 +27,7 @@ import (
 	"github.com/googleapis/librarian/internal/cli"
 	"github.com/googleapis/librarian/internal/container"
 	"github.com/googleapis/librarian/internal/githubrepo"
+	"github.com/googleapis/librarian/internal/gitrepo"
 	"github.com/googleapis/librarian/internal/statepb"
 )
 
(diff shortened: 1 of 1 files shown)


internal/librarian/generate.go
```diff
@@ -51,25 +52,43 @@ func init() {
 }
 
 func runGenerate(ctx context.Context) error {
-	startTime := time.Now()
-	workRoot, err := createWorkRoot(startTime)
-	if err != nil {
+	if err := validateRequiredFlag("api-path", flagAPIPath); err != nil {
```
jskeet, line 55: The diff is being unhelpful here. Really I've just moved the checks that were at line 92 to the start of `runGenerate`.

ldetmer (approved):

PR #240, opened 2025-06-08
impl(cmd/librarian): add support for version

julieqiu:
A librarian subcommand is added which prints the binary version.

Fixes https://github.com/googleapis/librarian/issues/238

### DIFF
--- a/internal/cli/version.go
+++ b/internal/cli/version.go
@@ -48,11 +48,13 @@ func version(info *debug.BuildInfo) string {
 		return "not available"
 	}
 
+	// Construct the pseudo-version string per
+	// https://go.dev/ref/mod#pseudo-versions.
 	buf := strings.Builder{}
 	buf.WriteString("0.0.0")
 	if revision != "" {
 		buf.WriteString("-")
-		buf.WriteString(revision[:12])
+		buf.WriteString(revision)
 	}
 	if at != "" {
 		// commit time is of the form 2023-01-25T19:57:54Z
--- a/internal/cli/version.go
+++ b/internal/cli/version.go
@@ -55,7 +55,9 @@ func version(info *debug.BuildInfo) string {
 	buf.WriteString("0.0.0")
 	if revision != "" {
 		buf.WriteString("-")
-		buf.WriteString(revision)
+		// Per https://go.dev/ref/mod#pseudo-versions, only use the first 12
+		// letters of the commit hash.
+		buf.WriteString(revision[:12])
 	}
 	if at != "" {
 		// commit time is of the form 2023-01-25T19:57:54Z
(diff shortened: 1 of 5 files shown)


internal/cli/version.go
```diff
@@ -0,0 +1,66 @@
...
+		if s.Key == "vcs.time" {
+			at = s.Value
+		}
+	}
+
+	if revision == "" && at == "" {
+		return "not available"
+	}
+
+	buf := strings.Builder{}
```
jba, line 51: `var buf strings.Builder` is more idiomatic.

Grepping the stdlib, the tree rooted at go/src:
`rg -tgo 'var [A-Za-z0-9_]+ [A-Za-z0-9_.]+'`: 804 lines
`rg -tgo '[A-Za-z0-9_]+ := [A-Za-z0-9_.]+\{\}'`: 2 lines

(`rg` is ripgrep)

internal/cli/version.go
```diff
@@ -0,0 +1,66 @@
...
+	buf.WriteString("0.0.0")
+	if revision != "" {
+		buf.WriteString("-")
+		buf.WriteString(revision[:12])
```
jba, line 55: where does 12 come from?

I recently saw a comment (by gri I think) like:
```
   ... 8 ... // len("moosejaw")
```
(not the exact string)

internal/librarian/version.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/googleapis/librarian/internal/cli"
+)
+
+var CmdVersion = &cli.Command{
+	Name:  "version",
+	Short: "Print version information.",
+	Run:   runVersion,
+}
+
+func runVersion(ctx context.Context) error {
```
jba, line 30: Inline the function.

internal/cli/version.go
```diff
@@ -0,0 +1,66 @@
...
+
+package cli
+
+import (
+	"runtime/debug"
+	"strings"
+	"time"
+)
+
+// Version prints the version information for the binary.
```
jskeet, line 23: Perhaps "returns" rather than "prints"? It might be worth mentioning that while it's derived from machine readable sources, the returned value is meant to be *human* readable rather than anything that's then parsed further. (Assuming I've understood that correctly.)

internal/librarian/version.go
jskeet: I'd be interested in whether there are common conventions around this for Go. I can see that it's nice and short - but if every other command is using a pattern of a function called `runXyz`, is that consistency worth continuing?

jskeet (commented):
This is great - I was thinking that we needed something like this just last night. (But it would have taken me ages to find BuildInfo I suspect.)

Not approving just because there are still nits, but very happy to see the overall PR :)

jskeet (approved):

PR #241, opened 2025-06-08
impl(internal/librarian): add basic test

julieqiu:


jba (approved):

jskeet (approved):
We need to figure out (and then document) how we're going to work on what are effectively multi-commit PRs. I *think* if I merge this, it'll go into #240 as another commit - which will then end up being squashed into a single commit. For this pair of changes, that's probably fine (we don't really need the test separate from the implementation) but for other PRs it would be less ideal.

Will leave it just approved for now though, to give us options.

jskeet (approved):
Reapproving.

PR #242, opened 2025-06-08
feat(internal/cli): add Command.Usage and Command.Long fields

julieqiu:
The librarian commands were previously undocumented. Added Usage and Long fields to the Command struct to serve as placeholders for usage text.  For now, these fields are populated with a TODO as a placeholders.

When constructing the usage information, the code will now panic if Short, Long, or Usage are empty, so that we ensure new commands are documented moving forward.

For https://github.com/googleapis/librarian/issues/237

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -25,8 +25,14 @@ type Command struct {
 	// Name is the unique identifier for the command.
 	Name string
 
-	// Short is a concise description shown in the 'librarian -h' output.
+	// Short is a concise one-line description of the command.
 	Short string
+
+	// Usage is the one line usage.
+	Usage string
+
+	// Long is the full description of the command.
+	Long string
 
 	// Run executes the command.
 	Run func(ctx context.Context) error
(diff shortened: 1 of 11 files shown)


internal/cli/cli.go
```diff
@@ -66,20 +72,24 @@ func Lookup(name string, commands []*Command) (*Command, error) {
...
+	output := fmt.Sprintf("Name:\n  %s\n\n", c.Name)
+	output += fmt.Sprintf("Usage:\n  %s\n\n", c.Usage)
+	output += fmt.Sprintf("Description:\n  %s\n", c.Long)
+	output += "\nFlags:\n"
 	return func() {
-		fmt.Fprint(fs.Output(), output)
-		fs.PrintDefaults()
-		fmt.Fprintf(fs.Output(), "\n\n")
+		fmt.Fprint(c.flags.Output(), output)
+		c.flags.PrintDefaults()
```
jba, line 92: Can c.flags be nil? (You checked for nilness somewhere else, in a previous PR.)

jba (dismissed):

internal/cli/cli.go
```diff
@@ -66,20 +72,24 @@ func Lookup(name string, commands []*Command) (*Command, error) {
 // SetFlags registers a list of functions that configure flags for the command.
 func (c *Command) SetFlags(flagFunctions []func(fs *flag.FlagSet)) {
 	c.flags = flag.NewFlagSet(c.Name, flag.ContinueOnError)
-	c.flags.Usage = constructUsage(c.flags, c.Name)
+	c.flags.Usage = c.usage()
```
jskeet, line 75: Just to check my understanding:

- `SetFlags` is called from `init()` in each command's source file
- `SetFlags` calls `usage()` which panics if information is missing
- `init()` will be called automatically (and unconditionally) for all source files within the package
- Therefore we get a panic if any command is missing information, even if that's not the one we're trying to run

I'm pretty confident that's the behavior, and that's what I'd want, but correct me if I'm wrong.

internal/cli/cli.go
jskeet: It's being set in line 74 directly before the call to this function - so I'm assuming it's okay unless `usage()` is called anywhere else.

internal/cli/cli.go
```diff
@@ -66,20 +72,24 @@ func Lookup(name string, commands []*Command) (*Command, error) {
...
-	output := fmt.Sprintf("Usage:\n\n  librarian %s [arguments]\n", name)
-	output += "\nFlags:\n\n"
+func (c *Command) usage() func() {
+	if c.Name == "" || c.Short == "" || c.Usage == "" || c.Long == "" {
+		panic(fmt.Sprintf("command %q is missing documentation", c.Name))
+	}
+
+	output := fmt.Sprintf("Name:\n  %s\n\n", c.Name)
+	output += fmt.Sprintf("Usage:\n  %s\n\n", c.Usage)
+	output += fmt.Sprintf("Description:\n  %s\n", c.Long)
```
jskeet, line 88: I would imagine that Long will be multiple lines in some cases. Is it worth splitting `Long` on `\n` and then joining with the two-space prefix? (Definitely a nice-to-have that we could do later, of course.)

jskeet (dismissed):

jskeet:
@julieqiu Are you happy to merge this now? Once this is in, I'll address #237.

julieqiu:
@jskeet this should be ready now

Issue #263, opened 2025-06-08
plan: create tasks for documents to be added to doc/


julieqiu:
See https://github.com/googleapis/librarian/issues/262 for context. Some that come to mind:

- High-level architecture
- Onboarding a new language
- CLI design (this is https://github.com/googleapis/librarian/issues/237)
- README cleanup
- How we do dependency management

@jskeet to create tasks, add them to https://github.com/googleapis/librarian/milestone/15, and add as sub-issues of this. I did 2 as an example.

jskeet:
@julieqiu: If we think this is the right set of issues, should we then move *this* issue to milestone 15? (Basically let's talk about how you see these umbrella issues, milestones and individual tasks interacting.)

julieqiu:
> [@julieqiu](https://github.com/julieqiu): If we think this is the right set of issues, should we then move _this_ issue to milestone 15? (Basically let's talk about how you see these umbrella issues, milestones and individual tasks interacting.)

Yup, that SGTM. Let's close this and just add to https://github.com/googleapis/librarian/milestone/15 moving forward.

Issue #247, opened 2025-06-08
doc: document how we work


julieqiu:
Right now, CONTRIBUTING.md doesn’t fully reflect how we handle contributions, things like code review expectations, commit message standards, copyright headers, test coverage, etc. It also still says “We'd love to accept your patches...” which may not match how we're actually operating.

We should update CONTRIBUTING.md, so that expectations clearer for everyone.

It would also be good to have standards that we can use for new repositories across googleapis (/cc @meltsufin).

Here are a few examples that highlight why it might be helpful to clarify and update our contribution guidelines for this repository:

- https://github.com/googleapis/librarian/issues/8 has been open since 2024, and https://github.com/googleapis/librarian/pull/17 was submitted by an external contributor, was ultimately closed.  If we’re not accepting external contributions, it might be good to state that explicitly so contributors aren’t investing time unnecessarily. We should also fix that issue, for the reasons in https://github.com/googleapis/librarian/issues/8#issuecomment-2954187896.

- https://github.com/googleapis/librarian/pull/88 has been open since February 25. There’s been code review feedback and several follow-up pings, but no updates. When a PR is unlikely to move forward, it may be better to close it to respect reviewer time. It’s also not clear what response SLOs we're following for issues or PRs, so it can be hard to know when or whether to follow up.

- The repository currently has 284 labels at https://github.com/googleapis/librarian/labels, and it’s not always clear which ones are active or meaningful. For example, I filed https://github.com/googleapis/librarian/issues/202  with p0 and 🚨 labels, and did not get a response after 5 days, which suggests p0 may mean something different than I expected (I started adding tests in https://github.com/googleapis/librarian/pull/235 and https://github.com/googleapis/librarian/pull/241)

- It’s unclear whether it’s okay to assign issues to individuals directly, or if there’s a rotation or process for triage.

In general, it might be helpful to update CONTRIBUTING.md (or create a brief governance doc somewhere in this repository) to clarify things like contribution model, SLOs, code review expectations, labeling, and issue triage. This would help set shared expectations and make it easier for both new and existing contributors to engage.

julieqiu:
Adding https://github.com/googleapis/librarian/pull/241#pullrequestreview-2909072295

julieqiu:
This work is still necessary, but based on today’s conversations, it sounds like @meltsufin likely won’t be able to get to it this month. I think it’s important to have this in place before the end of June, so I’m assigning it to myself and will coordinate with @jskeet @ldetmer @codyoss as needed.

PR #302, opened 2025-06-10
docs: initial contributing guidelines

jskeet:
This will want to be refined over time, but at least addresses the concern in #247 around a lack of clarity for external contributions.

Our team processes should be more thoroughly documented, which is tracked by #263 - but they need to be determined first.


julieqiu:
I think it is fine to merge this PR as a starting point, but I'm not sure it currently addresses the examples I listed in the issue. Approving this, but let's leave #247 open until that's addressed.

julieqiu (approved):

PR #314, opened 2025-06-10
tests(internal/librarian): unit tests for simple bits of pullrequest.go

jskeet:
This doesn't touch createPullRequest which is the hard bit (due to interacting with the Git repo and GitHub). We could refactor some of the code in that function (in particular the description formatting part) to unit test that though.

Towards #299.

### DIFF
--- a/internal/librarian/pullrequest_test.go
+++ b/internal/librarian/pullrequest_test.go
@@ -17,7 +17,6 @@ package librarian
 import (
 	"errors"
 	"slices"
-	"strings"
 	"testing"
 )
 
@@ -28,12 +27,12 @@ func TestAddErrorToPullRequest(t *testing.T) {
 	}
 	addErrorToPullRequest(&pr, "library-id", errors.New("bang"), "generating")
 	if !slices.Equal([]string{"s1", "s2"}, pr.Successes) {
-		t.Errorf("addErrorToPullRequest modified successes: %s", strings.Join(pr.Successes, ", "))
+		t.Errorf("addErrorToPullRequest modified successes: %q", pr.Successes)
 	}
 	// Note the error isn't included here; it's logged, but not added to the PullRequestContent
-	expected := "Error while generating library-id"
-	if !slices.Equal([]string{"e1", "e2", expected}, pr.Errors) {
-		t.Errorf("unexpected errors after addErrorToPullRequest: %s", strings.Join(pr.Errors, ", "))
+	want := "Error while generating library-id"
+	if !slices.Equal([]string{"e1", "e2", want}, pr.Errors) {
+		t.Errorf("unexpected errors after addErrorToPullRequest: %q", pr.Errors)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


internal/librarian/pullrequest_test.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	if !slices.Equal([]string{"e1", "e2"}, pr.Errors) {
+		t.Errorf("addSuccessToPullRequest modified errors: %s", strings.Join(pr.Errors, ", "))
+	}
+}
+
+func TestFormatListAsMarkdownEmptyList(t *testing.T) {
+	got := formatListAsMarkdown("Title", []string{})
+	if got != "" {
+		t.Errorf("formatting empty list produced non-empty output '%s'", got)
```
codyoss, line 57: nit: this is in a few places in this file but you can use %q for a quoted string.

internal/librarian/pullrequest_test.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+}
+
+func TestFormatListAsMarkdownNonEmptyList(t *testing.T) {
+	got := formatListAsMarkdown("Title", []string{"x1", "x2"})
+	want := "## Title\n\n- x1\n- x2\n\n\n"
+	if got != want {
+		t.Errorf("want '%s'; got '%s'", want, got)
```
codyoss, line 65: nit: in Go this is usually expressed as got, want: https://go.dev/wiki/CodeReviewComments#useful-test-failures

internal/librarian/pullrequest_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"testing"
+)
+
+func TestAddErrorToPullRequest(t *testing.T) {
+	pr := PullRequestContent{
+		Successes: []string{"s1", "s2"},
+		Errors:    []string{"e1", "e2"},
+	}
+	addErrorToPullRequest(&pr, "library-id", errors.New("bang"), "generating")
+	if !slices.Equal([]string{"s1", "s2"}, pr.Successes) {
```
codyoss, line 30: not required here by any means but if you want to compare the whole struct to another struct with nice diffs a lot of our Go libraries use: https://pkg.go.dev/github.com/google/go-cmp/cmp#Diff

codyoss (dismissed):

internal/librarian/pullrequest_test.go
jskeet: Will leave this one for now, but we can chat about idiomatic struct testing when we discuss testing in general.

internal/librarian/pullrequest_test.go
jskeet: Great to know, thanks :) (And fixing.)

jskeet (commented):
@codyoss I've added a new commit (which I'll squash) to adapt the formatting and order - having found that %q not only quotes individual strings, but produces usable slice-of-strings formatting, that simplifies things a lot :)

jskeet:
Requested re-review from Cody just to check that the changes in the second commit look okay. Then I'll squash and merge.

internal/librarian/pullrequest_test.go
codyoss: ack

codyoss (approved):

PR #331, opened 2025-06-12
docs(internal/librarian): initial usage help for update-apis

jskeet:
Towards #237

This PR only documents a single command, so that we can discuss formatting, detail level etc. Once we're comfortable with that, I expect to be able to create one PR to document all the other commands.


codyoss (approved):
Nice!

Issue #333, opened 2025-06-12
librarian: refactor code to make it more testable


codyoss:
- [x] Make "client" types for subpackages
- [x] Turn functions into methods
- [x] Make a app type to hold all the clients
- [x] Define some interfaces for easier mocking/facking of things like git and github

codyoss:
This is done, forgot to close.

PR #338, opened 2025-06-12
refactor: add client and methods for testability

codyoss:
There is likely further work to be done here to make some better abstractions in the librarian package, but I am trying to limit one big change per PR atm.

Updates: #333

### DIFF
--- a/internal/librarian/updateapis.go
+++ b/internal/librarian/updateapis.go
@@ -117,7 +117,7 @@ func updateAPIs(state *commandState) error {
 	cleanWorkingTreePostGeneration := true
 	if flagAPIRoot == "" {
 		var err error
-	apiRepo, err = cloneGoogleapis(state.workRoot)
+		apiRepo, err = cloneGoogleapis(state.workRoot)
 		if err != nil {
 			return err
 		}


internal/githubrepo/githubrepo.go
```diff
@@ -76,134 +94,112 @@ func CreateRelease(ctx context.Context, repo GitHubRepo, tag, commit, title, des
...
 }
 
-func GetRawContent(ctx context.Context, repo GitHubRepo, path, ref string) ([]byte, error) {
-	gitHubClient := github.NewClient(nil)
+func (c *Client) GetRawContent(ctx context.Context, repo *Repository, path, ref string) ([]byte, error) {
 	options := &github.RepositoryContentGetOptions{
 		Ref: ref,
 	}
-	closer, _, err := gitHubClient.Repositories.DownloadContents(ctx, repo.Owner, repo.Name, path, options)
+	closer, _, err := c.Repositories.DownloadContents(ctx, repo.Owner, repo.Name, path, options)
```
ldetmer, line 185: check me, but I believe this is using the new client that requires access token? Originally I don't think this function required an access token. Please confirm this won't force commands that didn't require github tokens, to now have auth.  You should be  able to run almost all the commands without a push flag without a github token (except for merge-PR and publish-release artifacts)

codyoss: good catch I restored the old logic for validating this for now.

internal/librarian/createreleasepr.go
```diff
@@ -133,9 +133,6 @@ func createReleasePR(state *commandState) error {
 	if err := validateSkipIntegrationTests(); err != nil {
 		return err
 	}
-	if err := validatePush(); err != nil {
```
ldetmer, line 136: I think we want to restore this too?

internal/librarian/createreleasepr.go
```diff
@@ -187,7 +184,11 @@ func createReleasePR(state *commandState) error {
 	// Final steps if we've actually created a release PR.
 	// - We always add the do-not-merge label so that Librarian can merge later.
 	// - Add a result environment variable with the PR number, for the next stage of the process.
-	err = githubrepo.AddLabelToPullRequest(state.ctx, *prMetadata, DoNotMergeLabel)
+	ghClient, err := githubrepo.NewClient()
```
ldetmer, line 187: just to confirm, we are now creating the client in the commands (instead of githubrepo), so we can unit test githubrepo.go?

can we mock things for githubrepo, so we can unit test the commands?

internal/librarian/createreleasepr.go
codyoss: oops, fixed.

internal/librarian/createreleasepr.go
codyoss: yes, at least that is the end goal of all this. I think I will have 2 or 3 more PRs till we get there. Need to make these things as "injectable" pieces controlled and initiated at these higher levels to we can mock away packages at a time so we can test the commands.

ldetmer (approved):

Issue #393, opened 2025-06-14
publish-release-artifacts


julieqiu:
See https://github.com/googleapis/librarian/issues/387 and https://github.com/googleapis/librarian/issues/386 for context

Issue #392, opened 2025-06-14
librarian: fix usage for create-release-artifacts


julieqiu:
See https://github.com/googleapis/librarian/issues/387 and https://github.com/googleapis/librarian/issues/386 for context

Issue #391, opened 2025-06-14
librarian: fix usage for merge-release-pr


julieqiu:
See https://github.com/googleapis/librarian/issues/387 and https://github.com/googleapis/librarian/issues/386 for context

Issue #390, opened 2025-06-14
internal/cli: fix usage for update-image-tag


julieqiu:
See https://github.com/googleapis/librarian/issues/387 and https://github.com/googleapis/librarian/issues/386 for context

PR #396, opened 2025-06-15
impl(internal/librarian): fix usage text

julieqiu:
For https://github.com/googleapis/librarian/issues/393
For https://github.com/googleapis/librarian/issues/392
For https://github.com/googleapis/librarian/issues/391
For https://github.com/googleapis/librarian/issues/390
For https://github.com/googleapis/librarian/issues/389
For https://github.com/googleapis/librarian/issues/388
For https://github.com/googleapis/librarian/issues/387
For https://github.com/googleapis/librarian/issues/386

### DIFF
--- a/internal/librarian/configure.go
+++ b/internal/librarian/configure.go
@@ -34,13 +34,15 @@ import (
 
 var CmdConfigure = &cli.Command{
 	Short: "configure configures libraries for new APIs in a language",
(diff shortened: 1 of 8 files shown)


jskeet (dismissed):
It feels like if the usage is just `librarian <short> <required flags> [flags]` then if we can indicate which flags are required (for this command) we should be able to generate the usage and simplify the validation code.

Filed https://github.com/googleapis/librarian/issues/408 but obviously that's not urgent.

Issue #198, opened 2025-06-04
librarian: split flag parsing logic from application logic


julieqiu:
Currently, the business logic for the librarian command is mixed with the interface for the CLI itself. This makes it confusing to understand where variables are being defined. 

For example, in this code snippet:

```
	maybeLoadStateAndConfig: func(languageRepo *gitrepo.Repo) (*statepb.PipelineState, *statepb.PipelineConfig, error) {
		// Load the state and config from the artifact directory. These will have been created by create-release-artifacts.
		state, err := loadPipelineStateFile(filepath.Join(flagArtifactRoot, pipelineStateFile))
		if err != nil {
			return nil, nil, err
		}
		config, err := loadPipelineConfigFile(filepath.Join(flagArtifactRoot, pipelineConfigFile))
		if err != nil {
			return nil, nil, err
		}
		return state, config, nil
	},
```

It is not intuitive that `flagArtifactRoot` is a flag and `pipelineStateFile` is a global variable, just by looking at the code. We should refactor this such that it is clearer.

julieqiu:
I don't want to move code around until we have tests. This is blocked on https://github.com/googleapis/librarian/issues/202.

/cc @jskeet

PR #460, opened 2025-06-18
fix(librarian): add tests for validate function in flags.go

### LINKED ISSUE #297: tests: tests for internal/librarian/flags.go

zhumin8:
closes #297

### DIFF
--- a/internal/librarian/flags.go
+++ b/internal/librarian/flags.go
@@ -169,7 +169,7 @@ func validateSkipIntegrationTests() error {
 // start executing the command. (At least for simple cases where a flag is required;
 // note that this isn't always going to be the same for all commands for one flag.)
 func validateRequiredFlag(name, value string) error {
-	if value == "" {
+	if strings.TrimSpace(value) == "" {
 		return fmt.Errorf("required flag -%s not specified", name)
 	}
 	return nil
--- a/internal/librarian/flags_test.go
+++ b/internal/librarian/flags_test.go
@@ -1,0 +1,208 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


internal/librarian/flags.go
```diff
@@ -169,7 +169,7 @@ func validateSkipIntegrationTests() error {
 // start executing the command. (At least for simple cases where a flag is required;
 // note that this isn't always going to be the same for all commands for one flag.)
 func validateRequiredFlag(name, value string) error {
-	if value == "" {
+	if strings.TrimSpace(value) == "" {
```
julieqiu, line 172: I don't think we want to trim whitespace on the value. We probably want to return an error, and tell the user to type the accurate input.

internal/librarian/flags_test.go
```diff
@@ -0,0 +1,208 @@
...
+			name:           "push is false, no token, no error",
+			setFlagPush:    false,
+			setGitHubToken: true,
+			expectedErr:    nil,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			resetAllFlags()
```
julieqiu, line 59: what's the reason that this function needs to be executed?

internal/librarian/flags_test.go
```diff
@@ -0,0 +1,208 @@
...
+
+			// Set up global flag and environment variable based on test case
```
julieqiu, line 61: we should clean this up as part of https://github.com/googleapis/librarian/issues/198

julieqiu:
@zhumin8 - I started reviewing this PR, but I'm not sure it actually makes sense to test these functions. Not sure if @jskeet agrees. 

Additionally, checkout https://github.com/googleapis/librarian/blob/main/CONTRIBUTING.md#sending-a-pull-request for details on how to structure commit message, etc.

jskeet:
I think #535 will make all of this a lot simpler. I'm struggling to imagine right now what the eventualy state of flags looks like - I'd previously thought that a test for flag parsing might also serve as an example for how to write tests that deal with flags, but using a config instead should make that obsolete, basically. Maybe close for now (as some of the flag implementation changes will render this PR obsolete) and then re-examine after the flag changes have gone through?

julieqiu:
Going to close this — thanks @zhumin8 for putting it up! I’ve filed a lot of good first issue tickets since you opened this, so it might be better to pick up one of those. This umbrella issue has several solid options: https://github.com/googleapis/librarian/issues/487

Issue #492, opened 2025-06-22
librarian: pass global flags as explicit arguments to runGenerate


julieqiu:
See https://github.com/googleapis/librarian/issues/487 for context

julieqiu:
As of [#493](https://github.com/googleapis/librarian/pull/493), this is done except for [#486](https://github.com/googleapis/librarian/issues/486), which needs to be fixed to allow `workRoot` to be passed explicitly.

zhumin8:
#486 seems to be done in https://github.com/googleapis/librarian/blob/6f0702c855369c36348ac05f4c3144544cef39b4/internal/librarian/generate.go#L214?

julieqiu:
@zhumin8 seems like it. Closing, and we can always reopen if we missed something,

PR #493, opened 2025-06-22
fix(internal/librarian): pass arguments directly to runGenerate

julieqiu:
Update runGenerate to accept various flags as explicit parameters instead of reading from global flags. This makes the function easier to test, reuse, and reason about in isolation.

For https://github.com/googleapis/librarian/issues/492

### DIFF
--- a/internal/librarian/generate.go
+++ b/internal/librarian/generate.go
@@ -69,7 +69,15 @@ There is no "clean" operation or copying of the generated code in raw generation
 other source code to be preserved/cleaned. Instead, the "build-raw" command is provided with the same
 output directory that was specified for the "generate-raw" command.
 `,
-	Run: runGenerate,
+	Run: func(ctx context.Context) error {
+		if err := validateRequiredFlag("api-path", flagAPIPath); err != nil {
+			return err
+		}
+		if err := validateRequiredFlag("api-root", flagAPIRoot); err != nil {
+			return err
+		}
+		return runGenerate(ctx, flagAPIPath, flagAPIRoot, flagRepoRoot, flagRepoUrl, flagSecretsProject, flagBuild)
+	},
 }
 
 func init() {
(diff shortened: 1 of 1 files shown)


internal/librarian/generate.go
```diff
@@ -86,30 +94,25 @@ func init() {
 	})
 }
 
-func runGenerate(ctx context.Context) error {
-	if err := validateRequiredFlag("api-path", flagAPIPath); err != nil {
-		return err
-	}
-	if err := validateRequiredFlag("api-root", flagAPIRoot); err != nil {
+func runGenerate(ctx context.Context, apiPath, apiRoot, repoRoot, repoURL, secretsProject string, build bool) error {
```
jba, line 97: This signature is error-prone because there are so many string args in a row. Consider taking a struct.

internal/librarian/generate.go
```diff
@@ -235,12 +239,12 @@ func detectIfLibraryConfigured() (bool, error) {
...
-	libraryID := findLibraryIDByApiPath(pipelineState, flagAPIPath)
+	libraryID := findLibraryIDByApiPath(pipelineState, apiPath)
 	if libraryID == "" {
-		slog.Info(fmt.Sprintf("API path %s not configured in repo", flagAPIPath))
+		slog.Info(fmt.Sprintf("API path %s not configured in repo", apiPath))
 		return false, nil
 	}
 
-	slog.Info(fmt.Sprintf("API path %s configured in repo library %s", flagAPIPath, libraryID))
+	slog.Info(fmt.Sprintf("API path %s configured in repo library %s", apiPath, libraryID))
```
jba, line 248: If you're using slog, might as well go all-in on structured logging:
```
slog.Info("API configured", "path", apiPath, "library", libraryID)
```

julieqiu: done

internal/librarian/generate.go
julieqiu: Actually I figured out a way. Did this on a follow up PR https://github.com/googleapis/librarian/pull/535. Let me know what you think.

jskeet:
@julieqiu: #535 is configured to merge into the branch for this PR, so I *think* if we merge #535 into this first, we can rereview this PR (which will then have two commits in, which we'll squash to one). Shout if I've misunderstood though.

julieqiu:
> @julieqiu: #535 is configured to merge into the branch for this PR, so I _think_ if we merge #535 into this first, we can rereview this PR (which will then have two commits in, which we'll squash to one). Shout if I've misunderstood though.

The way I’ve been handling multi-commit PRs is to merge the one based on main first, then rebase the next one on top—since they get squashed in the end anyway. Otherwise, I’d just stack it directly on top of this one instead of creating a separate PR.

internal/librarian/generate.go
jba: LGTM.

Issue #505, opened 2025-06-22
librarian: refactor deriveImage to eliminate global dependencies


julieqiu:
The current implementation of `deriveImage` relies on global state:

- `flagLanguage` and `flagImage` are global flags
- `defaultRepositoryEnvironmentVariable` is read via `os.Getenv` inside the function
- state is the only explicit input

https://github.com/googleapis/librarian/blob/a5e8509f7206833ca757d0c2bb0f5cbba22603bb/internal/librarian/command.go#L160

This creates hidden dependencies that make the function harder to test, reuse, and reason about in isolation.

Refactor `deriveImage` to take all required inputs as explicit parameters:

```
func deriveImage(language, imageOverride, defaultRepo string, state *statepb.PipelineState) string
```

The caller is then responsible for passing in these values:

```
	flagLanguage,
	flagImage,
	os.Getenv(defaultRepositoryEnvironmentVariable),
	state,
```

The resulting code would look something like this: https://go.dev/play/p/xv1N2V3GauL

PR #513, opened 2025-06-22
fix(issue): Address issue #505 - librarian: refactor deriveImage to eliminate global dependencies

Harshitshukla0208:
This PR is to address issue #505.

The `deriveImage` function currently relies on global variables and environment variables, which makes it difficult to test and reason about. By refactoring it to accept all necessary parameters explicitly, we can eliminate these hidden dependencies. This change will improve the function's reusability and testability, as it will no longer depend on external state. The function signature will be updated to take `language`, `imageOverride`, `defaultRepo`, and `state` as parameters, allowing the caller to provide these values directly.

### DIFF
--- a/internal/librarian/command.go
+++ b/internal/librarian/command.go
@@ -157,7 +157,7 @@ func appendResultEnvironmentVariable(workRoot, name, value string) error {
 	return appendToFile(envFile, fmt.Sprintf("%s=%s\n", name, value))
 }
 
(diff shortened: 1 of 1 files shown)


julieqiu:
Closing this out until https://github.com/googleapis/librarian/pull/513#issuecomment-2994117192 has been addressed, per guidelines at https://github.com/googleapis/librarian/blob/main/CONTRIBUTING.md.

Issue #504, opened 2025-06-22
librarian: pass createCommandStateForLanguage explicitly to createCommandStateForLanguage


julieqiu:
The function createCommandStateForLanguage currently accesses the global flags `flagSecretsProject` directly:

https://github.com/googleapis/librarian/blob/a5e8509f7206833ca757d0c2bb0f5cbba22603bb/internal/librarian/command.go#L117

This creates implicit dependencies that make the function harder to test, reuse, and reason about in isolation.

Instead, update the function signature to accept these values as explicit parameters:

```
func createCommandStateForLanguage(ctx context.Context, secretsProject string) (*commandState, error)
```

This makes the function’s inputs explicit and ensures its behavior is a pure function of its arguments.

PR #514, opened 2025-06-22
fix(issue): Address issue #504 - librarian: pass createCommandStateForLanguage explicitly to createCommandStateForLanguage

Harshitshukla0208:
This PR is to address issue #504.
The function `createCommandStateForLanguage` currently accesses the global variable `flagSecretsProject`, which creates implicit dependencies making the function harder to test and reason about. By modifying the function signature to accept `secretsProject` as a parameter, we make the function's dependencies explicit. This change enhances testability and reusability, as the function can now be called with different values without relying on global state.


julieqiu:
Closing this out until https://github.com/googleapis/librarian/pull/514#issuecomment-2994118262 has been addressed, per guidelines at https://github.com/googleapis/librarian/blob/main/CONTRIBUTING.md.

Issue #522, opened 2025-06-22
test: add linter for doc comments


julieqiu:
Add a linter to enforce basic adherence to Go's official comment guidelines: https://tip.golang.org/doc/comment.

julieqiu:
Going to close this since the t.Skip call has been removed, and will open a separate issue for https://github.com/googleapis/librarian/pull/579

Issue #521, opened 2025-06-22
all: document exported functions


julieqiu:
We should document all packages, constants, functions, types, and variables following the guidelines at https://go.dev/doc/comment. 

This includes:

- Adding a clear package-level comment that explains the purpose of the gitrepo package and what kinds of functionality it provides

<img width="1102" alt="Image" src="https://github.com/user-attachments/assets/944f6e30-36f9-4276-a746-67e8941eb5ff" />

- Ensuring that all exported types, functions, and variables are documented with complete, descriptive comments
- Keeping documentation concise, consistent, and useful for future maintainers and contributors.

jskeet:
Hmm. Running `pkgsite .` locally (on #586) I think I may have done something wrong.

The overview for a package includes documentation:

![Image](https://github.com/user-attachments/assets/424183c0-b14b-44f4-a10d-7afad3fc72c1)

... but the list of packages doesn't have anything:

![Image](https://github.com/user-attachments/assets/0bb40adc-df87-4f93-8ee1-257e92a5b2fb)

I'm hoping this is just something I'm doing wrong locally rather than in the doc comments themselves.

PR #523, opened 2025-06-22
feat(all): check that exported symbols have doc comments

julieqiu:
Add a test to ensure that exported funcs, types, consts, and vars must have comments.

This test is skipped until https://github.com/googleapis/librarian/issues/521 is addressed.

For https://github.com/googleapis/librarian/issues/522

### DIFF
--- a/all_test.go
+++ b/all_test.go
@@ -109,6 +109,18 @@ func TestGovulncheck(t *testing.T) {
 	rungo(t, "run", "golang.org/x/vuln/cmd/govulncheck@latest", "./...")
 }
 
+func rungo(t *testing.T, args ...string) {
+	t.Helper()
+
+	cmd := exec.Command("go", args...)
+	if output, err := cmd.CombinedOutput(); err != nil {
+		if ee := (*exec.ExitError)(nil); errors.As(err, &ee) && len(ee.Stderr) > 0 {
+			t.Fatalf("%v: %v\n%s", cmd, err, ee.Stderr)
+		}
+		t.Fatalf("%v: %v\n%s", cmd, err, output)
+	}
+}
+
 func TestExportedSymbolsHaveDocs(t *testing.T) {
 	// TODO(https://github.com/googleapis/librarian/issues/522): turn on once
 	// all existing symbols have docs.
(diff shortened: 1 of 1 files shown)


all_test.go
```diff
@@ -105,6 +109,80 @@ func TestGovulncheck(t *testing.T) {
 	rungo(t, "run", "golang.org/x/vuln/cmd/govulncheck@latest", "./...")
 }
 
+func TestExportedSymbolsHaveDocs(t *testing.T) {
+	// TODO(https://github.com/googleapis/librarian/issues/522): turn on once
+	// all existing symbols have docs.
+	t.Skip()
+
+	err := filepath.WalkDir(".", func(path string, d os.DirEntry, err error) error {
+		if err != nil || d.IsDir() || !strings.HasSuffix(path, ".go") || strings.HasSuffix(path, "_test.go") {
```
jba, line 118: Don't you want to return a non-nil error?

all_test.go
```diff
@@ -105,6 +109,80 @@ func TestGovulncheck(t *testing.T) {
...
+		}
+
+		fset := token.NewFileSet()
+		node, err := parser.ParseFile(fset, path, nil, parser.ParseComments)
+		if err != nil {
+			t.Errorf("failed to parse file %q: %v", path, err)
+			return nil
+		}
+
+		for _, decl := range node.Decls {
```
jba, line 129: // Visit every top-level declaration in the file.

all_test.go
```diff
@@ -105,6 +109,80 @@ func TestGovulncheck(t *testing.T) {
...
+			gen, ok := decl.(*ast.GenDecl)
+			if ok && (gen.Tok == token.TYPE || gen.Tok == token.CONST || gen.Tok == token.VAR) {
+				for _, spec := range gen.Specs {
+					switch s := spec.(type) {
+					case *ast.TypeSpec:
+						if s.Name.IsExported() {
+							if gen.Doc == nil {
+								t.Errorf("%s: exported type %q is missing doc comment",
+									path, s.Name.Name)
+							} else if !startsWithName(gen.Doc.Text(), s.Name.Name) {
```
jba, line 139: I'd say this is too strict. Sometimes we might write:
```
// A Foo is a bar.
type Foo ...
```

jba: Actually, startsWithName is the problem.

all_test.go
```diff
@@ -105,6 +109,80 @@ func TestGovulncheck(t *testing.T) {
...
+		}
+		return nil
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
+func startsWithName(comment, name string) bool {
+	words := strings.Fields(comment)
```
jba, line 179: for word := range strings.FieldsSeq(comment)

all_test.go
```diff
@@ -105,6 +109,80 @@ func TestGovulncheck(t *testing.T) {
...
+			return nil
+		}
+
+		for _, decl := range node.Decls {
+			gen, ok := decl.(*ast.GenDecl)
+			if ok && (gen.Tok == token.TYPE || gen.Tok == token.CONST || gen.Tok == token.VAR) {
+				for _, spec := range gen.Specs {
+					switch s := spec.(type) {
+					case *ast.TypeSpec:
+						if s.Name.IsExported() {
```
jba, line 135: Feels like there should be a common function
```
func checkDoc(t *testing.T, name *ast.Ident, doc *ast.CommentGroup)
```

jba (changes requested):

all_test.go
julieqiu: Done

all_test.go
```diff
@@ -116,3 +120,57 @@ func rungo(t *testing.T, args ...string) {
...
+	}
+}
+
+func checkDoc(t *testing.T, name *ast.Ident, doc *ast.CommentGroup, path string) {
+	t.Helper()
+	if !name.IsExported() {
+		return
+	}
+	if doc == nil {
+		t.Errorf("%s: %q is missing doc comment",
```
jskeet, line 173: Just for my curiosity: is there a reason this is split across lines, but the error at line 137 isn't? (I thought line breaks were really uncommon in Go, but I may be wrong or there may be a different convention at work here.)

all_test.go
```diff
@@ -116,3 +120,57 @@ func rungo(t *testing.T, args ...string) {
...
+		t.Fatal(err)
+	}
+}
+
+func checkDoc(t *testing.T, name *ast.Ident, doc *ast.CommentGroup, path string) {
+	t.Helper()
+	if !name.IsExported() {
+		return
+	}
+	if doc == nil {
```
jskeet, line 172: Is there something else that does the "starting with their name" part of the check (from the PR description), or is that still to be implemented?

all_test.go
jba: Line breaks are at the author's discretion. gofmt will not insert them: there is no maximum line length. We add them when we feel the line is getting long. My usual editor window is about 120 chars wide, maybe more, and I will break when the line wraps on that. But there's no rule, and Julie's break here is consistent with common practice.

julieqiu:
> One other thing: it's OK to document a const decl but not each spec. And sometimes not even that, if the type doc says it all:
> 
> ```
> // The Beatles.
> const (
>     John = iota
>     Paul
>     George
>     Ringo
> )
> ```
> 
> ```
> // Color is a valid primary color.
> type Color int
> 
> const (
>     Red Color = iota
>     Blue
>     Yellow
> )
> ```
> 
> A real-world example where both the type and the decl have doc, but not the specs: https://pkg.go.dev/log/slog#Level

Ack, deleted the check

all_test.go
julieqiu: I deleted it after this discussion: https://github.com/googleapis/librarian/pull/523#discussion_r2160513755. The goal is for this to be useful, so I'd rather have us miss something than have a check be noisy.

(thanks for the comment, I didn't catch that it was still in the commit message)

jba (approved):

PR #535, opened 2025-06-22
feat(internal/librarian): add Config struct and FromFlags constructor

julieqiu:
A Config struct and FromFlags function is introduced to centralize CLI flag values in a single place. This simplifies downstream usage and improves testability.


jskeet (approved):
I like this a lot better than pr7 on its own. Admittedly it means the config ends up being one giant config across all commands, but I think we can cope with that. (We could alternatively have a sort of CommonConfig for configuration that's common across multiple commands, and then a command-specific config struct, but that feels a bit messy.)

jba (approved):

PR #558, opened 2025-06-23
fix(internal/librarian): dedupe release note entries

### LINKED ISSUE #549: librarian: dedupe commit messages in release notes
If we see the same conventional commit line in multiple commits that contribute to a single library (e.g. separately for v1 and v1beta) we should only create a single entry for that commit line in the release notes.

It should be straighforward to just keep a set of "lines we've already included".

chingor13:
Fixes #549

### DIFF
--- a/internal/librarian/createreleasepr.go
+++ b/internal/librarian/createreleasepr.go
@@ -310,11 +310,9 @@ func generateReleaseCommitForEachLibrary(state *commandState, inputDirectory str
 // TODO(https://github.com/googleapis/librarian/issues/564): decide on release notes ordering
 func formatReleaseNotes(commitMessages []*CommitMessage) string {
 	// Group release notes by type, preserving ordering (FIFO)
-	features := []string{}
+	var features, docs, fixes []string
 	featuresSeen := make(map[string]bool)
-	docs := []string{}
 	docsSeen := make(map[string]bool)
-	fixes := []string{}
 	fixesSeen := make(map[string]bool)
 
 	// TODO(https://github.com/googleapis/librarian/issues/547): perhaps record breaking changes in a separate section


internal/librarian/createreleasepr.go
```diff
@@ -308,24 +310,29 @@ func generateReleaseCommitForEachLibrary(state *commandState, inputDirectory str
...
+			fixes[fix] = true
+		}
 	}
 
 	var builder strings.Builder
 
-	maybeAppendReleaseNotesSection(&builder, "New features", features)
-	maybeAppendReleaseNotesSection(&builder, "Bug fixes", fixes)
-	maybeAppendReleaseNotesSection(&builder, "Documentation improvements", docs)
+	maybeAppendReleaseNotesSection(&builder, "New features", slices.Collect(maps.Keys(features)))
```
jba, line 333: Sort for determinism.
Or if order matters, you need to use a slice, and use a map only to dedup.

julieqiu (approved):

internal/librarian/createreleasepr.go
chingor13: Rewrote to preserve the initial ordering (we use the map to only decide if we have seen the entry before)

Added a TODO linking to https://github.com/googleapis/librarian/issues/564

internal/librarian/createreleasepr.go
```diff
@@ -307,18 +307,37 @@ func generateReleaseCommitForEachLibrary(state *commandState, inputDirectory str
...
 
+// TODO(https://github.com/googleapis/librarian/issues/564): decide on release notes ordering
 func formatReleaseNotes(commitMessages []*CommitMessage) string {
+	// Group release notes by type, preserving ordering (FIFO)
 	features := []string{}
+	featuresSeen := make(map[string]bool)
 	docs := []string{}
+	docsSeen := make(map[string]bool)
 	fixes := []string{}
+	fixesSeen := make(map[string]bool)
```
julieqiu, line 318: nit:

```
var features, docs, fixes []string
```

julieqiu (approved):

internal/librarian/createreleasepr.go
chingor13: Done.

julieqiu (approved):

internal/librarian/createreleasepr.go
```diff
@@ -307,18 +307,35 @@ func generateReleaseCommitForEachLibrary(state *commandState, inputDirectory str
...
 
-	// TODO(https://github.com/googleapis/librarian/issues/549): deduping (same message across multiple commits)
 	// TODO(https://github.com/googleapis/librarian/issues/547): perhaps record breaking changes in a separate section
 	// TODO(https://github.com/googleapis/librarian/issues/550): include backlinks, googleapis commits etc
 	for _, commitMessage := range commitMessages {
-		features = append(features, commitMessage.Features...)
-		docs = append(docs, commitMessage.Docs...)
-		fixes = append(fixes, commitMessage.Fixes...)
+		for _, feature := range commitMessage.Features {
```
jba, line 321: These are all the same, so this could be written
```
dedup := func(s []string) []string {
    seen := make(map[string]bool)
    var r []string
    for _, el := range s {
      if !seen[el] {.....}
   }
   return r
}

features := dedup(commitMessage.Features)
...
...
```

chingor13: Filed https://github.com/googleapis/librarian/issues/600 to consider following up on this

Issue #526, opened 2025-06-22
librarian: create abstraction for go-git usage


julieqiu:
There are a couple of places where we use [github.com/go-git/go-git/v5/plumbing/object.Commit](https://pkg.go.dev/github.com/go-git/go-git/v5/plumbing/object#Commit) directly:

https://github.com/googleapis/librarian/blob/96dd98069587574167d82d0c4c641a1b225f0e85/internal/librarian/updateapis.go#L284

https://github.com/googleapis/librarian/blob/96dd98069587574167d82d0c4c641a1b225f0e85/internal/librarian/commitmessage.go#L47

Instead, we should provide an internal abstraction in [internal/gitrepo](https://github.com/googleapis/librarian/tree/main/internal/gitrepo) to isolate this dependency.

By centralizing our usage behind an internal abstraction, we make it easier to change dependencies in the future. For example, if we decide to switch to a different Git library, we would only need to update a single layer instead of refactoring the entire codebase. This also improves maintainability and testability by limiting third-party APIs to one place.

PR #559, opened 2025-06-23
fix(internal/gitrepo): abstract git.Commit in go-git usage

JoeWang1127:
Fixes: #526


internal/gitrepo/gitrepo.go
```diff
@@ -343,12 +358,13 @@ func (r *Repository) GetCommitsForPathsSinceTag(paths []string, tagName string)
...
-// with all commits having a single parent.
-func (r *Repository) GetCommitsForReleaseID(releaseID string) ([]object.Commit, error) {
+// GetCommitsForReleaseID returns all commits with the given release ID, i.e.,
+// where the commit message contains a line of Librarian-Release-Id: <release-id>.
+// These commits are expected to be contiguous, from head, with all commits
+// having a single parent.
+func (r *Repository) GetCommitsForReleaseID(releaseID string) ([]Commit, error) {
 	releaseIDLine := fmt.Sprintf("Librarian-Release-ID: %s", releaseID)
-	commits := []object.Commit{}
+	commits := []Commit{}
```
julieqiu, line 367: We generally use pointers for this

```
	commits := []*Commit{}
```

internal/gitrepo/gitrepo.go
```diff
@@ -373,7 +389,10 @@ func (r *Repository) GetCommitsForReleaseID(releaseID string) ([]object.Commit,
 		}
 
 		if gotReleaseID {
-			commits = append(commits, *candidateCommit)
+			commits = append(commits, Commit{
```
julieqiu, line 392: ```commits = append(commits, &Commit{...})```

re: pointers comment above

julieqiu (approved):

JoeWang1127:
> @JoeWang1127 - not sure if you intended for this to be a draft, but I reviewed it and it looks good :)

I plan to add an unit test for `GetCommitsForPathsSinceCommit` before approval.

internal/gitrepo/gitrepo.go
JoeWang1127: Done.

JoeWang1127:
The test succeeded in my local env but failed in CI:
```
--- FAIL: TestGetCommitsForPathsSinceCommit (0.02s)
    gitrepo_test.go:95: GetCommitsForPathsSinceCommit([local/first local/third]) expected 2 commit(s), got error reference not found
    gitrepo_test.go:101: GetCommitsForPathsSinceCommit([local/first local/third]) expected 2 commit(s), got 0
    gitrepo_test.go:95: GetCommitsForPathsSinceCommit([local/zero]) expected 0 commit(s), got error reference not found
```

JoeWang1127:
Looks like we need to add author info when using git-commit:
```
gitrepo_test.go:76: git-commit file has error: author field is required
```

.gitignore
```diff
@@ -23,3 +23,6 @@ go.work.sum
 
 # env file
 .env
+
+# Intellji file
+.idea/
```
julieqiu, line 28: nit: I would make a separate PR for this change, since it is outside the scope of this commit

internal/gitrepo/gitrepo.go
julieqiu: It looks like this is still `commits := []Commit{}` and needs to be changed

internal/gitrepo/gitrepo.go
julieqiu: It looks like this is still `commits = append(commits, Commit{` and needs to be changed.

julieqiu (changes requested):

internal/gitrepo/gitrepo.go
JoeWang1127: Sorry I missed this, I believe this time it changed.

PR #574, opened 2025-06-24
fix(internal): move LIBRARIAN_GITHUB_TOKEN to Config

### LINKED ISSUE #510: librarian: convert LIBRARIAN_GITHUB_TOKEN to flag

dangazineu:
Moved resolution of `LIBRARIAN_GITHUB_TOKEN` environment variable into the `config` package. 

Now `githubrepo.NewClient` takes a GH Access Token as a parameter instead of getting it from the environment, improving testability. Refactored callers of `githubrepo.NewClient` to retrieve that value from the `config.Config`. Also replaced any direct flag access in those functions to use values `Config` instance that is now available.

Fixes #510.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -121,15 +121,8 @@ func New() *Config {
 
 // IsValid ensures the values contained in a Config are valid.
 func (c *Config) IsValid() (bool, error) {
-	if err := c.validatePush(); err != nil {
-		return false, err
+	if c.Push && c.GitHubToken == "" {
+		return false, errors.New("no GitHub token supplied for push")
 	}
 	return true, nil
-}
-
-func (c *Config) validatePush() error {
-	if c.Push && c.GitHubToken == "" {
-		return errors.New("no GitHub token supplied for push")
-	}
-	return nil
 }


internal/config/config.go
```diff
@@ -111,5 +115,21 @@ func New() *Config {
...
+// IsValid ensures the values contained in a Config are valid.
+func (c *Config) IsValid() (bool, error) {
+	if err := c.validatePush(); err != nil {
+		return false, err
+	}
+	return true, nil
+}
+
+func (c *Config) validatePush() error {
+	if c.Push && c.GitHubToken == "" {
```
julieqiu, line 131: I would move this inside `IsValid` since it's just one line to avoid a level of indirection (unless there's a reason for separating that I might not be seeing)

internal/librarian/configure.go
```diff
@@ -99,17 +100,19 @@ func init() {
 }
 
 func runConfigure(ctx context.Context) error {
+	cfg := config.New()
```
julieqiu, line 103: Leaving a comment for me to remember to factor this out into internal/cli

julieqiu (changes requested):

internal/config/config.go
dangazineu: I imagine `IsValid()` will grow to include many other validations, so separating them by "theme" would make sense. For now though, I agree that the additional method only serves as extra indirection, removed.

julieqiu (approved):

Issue #578, opened 2025-06-24
librarian: rename references of container to docker


julieqiu:
Cleanup following https://github.com/googleapis/librarian/pull/571 and https://github.com/googleapis/librarian/issues/484

jskeet:
There are three separate concepts:

- Docker (the means to start a container)
- Image (effectively a snapshot)
- Container (something running, created by Docker from an image)

I think if we try to use Docker to mean all of these it'll get really confusing. I'd like to stick to the standard terminology as far as possible. We've certainly got terminology issues (caused by me) where we refer to a "language container" when we should probably say "language image" - or come up with another term entirely. (There will be a lot of design docs etc to change once we've got such a term.)

I'll remove "good first issue" as it's nuanced. Would welcome @codyoss's views too.

codyoss:
I think the only time we should use the word Docker is when we are talking about the Dockerfile. Other than that I do like tool agnostic terms like image and container.

julieqiu:
@codyoss - Let's cover this in your design doc, since we've already reversed direction on this once. We originally renamed container to docker to make the dependency explicit, so if we're considering renaming it back, I think we should clearly articulate the rationale and provide clear terminology for the team moving forward.

jskeet:
Something potentially relevant - in #652 @jba suggested we investigate using podman instead of docker. I've filed #714 to investigate that (and assigned it to myself for now, but left it as unplanned).

ldetmer:
I believe this is a stale issue, we are using the term docker when we are actually interacting with docker program [example](https://github.com/googleapis/librarian/blob/main/internal/docker/docker.go) and are using container when we refer to the image the language uses to interact with librarian.

PR #583, opened 2025-06-24
chore(internal/docker): rename some 'docker' back to 'container'

jskeet:
(In particular, anything talking about "a docker" should probably be "a container" or maybe "an image".)

Fixes the (self-resolved) comments in #577.

Happy to be persuaded on some of these.


julieqiu:
I originally considered removing the word "container" from the codebase entirely (see https://github.com/googleapis/librarian/issues/578). We might want to refer to these explicitly as "Docker containers" for clarity. That said, approving this for now since it’s a minor point.

PR #588, opened 2025-06-24
fix(all_test.go): make the linter ignore generated protobuf code

### LINKED ISSUE #587: tests: the linter checks generated code
We should exclude *.pb.go in the linter, as it's generated from the pipeline.proto file (which *does* have comments everywhere it can).

We can potentially remove the exclusion after implementing #326, although it wouldn't harm us to leave the filter in.

jskeet:
Fixes #587

### DIFF
--- a/all_test.go
+++ b/all_test.go
@@ -127,7 +127,7 @@ func TestExportedSymbolsHaveDocs(t *testing.T) {
 	t.Skip()
 
 	err := filepath.WalkDir(".", func(path string, d os.DirEntry, err error) error {
-		if err != nil || d.IsDir() || !strings.HasSuffix(path, ".go") || strings.HasSuffix(path, "_test.go") {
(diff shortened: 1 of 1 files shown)


all_test.go
```diff
@@ -127,7 +127,7 @@ func TestExportedSymbolsHaveDocs(t *testing.T) {
 	t.Skip()
 
 	err := filepath.WalkDir(".", func(path string, d os.DirEntry, err error) error {
-		if err != nil || d.IsDir() || !strings.HasSuffix(path, ".go") || strings.HasSuffix(path, "_test.go") {
+		if err != nil || d.IsDir() || !strings.HasSuffix(path, ".go") || strings.HasSuffix(path, "_test.go") || strings.HasSuffix(path, ".pb.go") {
```
jskeet, line 130: I'm happy to break the line here, but would appreciate suggestions on how/where to do so if so.

julieqiu (approved):

PR #594, opened 2025-06-24
docs(internal/config): expand configuration documentation

jskeet:
Each member that is now documented should be complete, but there are additional members still to be documented.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -22,43 +22,152 @@ import (
 // Config holds all configuration values parsed from flags or environment
 // variables.
 type Config struct {
-	// APIPath is the path to the API to be configured or generated.
+	// APIPath is the path to the API to be configured or generated,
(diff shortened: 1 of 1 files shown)


jskeet:
@julieqiu: I thought it would be worth sending this now, as I'm unavailable for a while, and feedback from the members that I *have* documented could be useful for later ones as well. I don't mind whether we address any comments in this PR and merge it with only the currently-documented members, or press on until it's complete. (I'd tend to favor the former, just to reduce the chance of conflicts, but I don't mind much.)

julieqiu:
SGTM, no preferences and also generally prefer smaller commits. I'll take a look later today. 

/cc @dangazineu @codyoss @zhumin8 @JoeWang1127 @chingor13 also for any input

julieqiu (approved):

Issue #592, opened 2025-06-24
docker: add Docker.run method


julieqiu:
Adds a Docker.run method to encapsulate Docker command execution. This isolates command construction, making it easier to test.

I have a WIP PR here: https://github.com/googleapis/librarian/pull/591

This is blocked on https://github.com/googleapis/librarian/issues/590

PR #595, opened 2025-06-24
feat(internal/docker): add Docker.run and test command construction

julieqiu:
A Docker.run method is added to encapsulate Docker command execution. This isolates command construction, making it easier to test.

A test is also added for Docker command construction across all supported operations, using a mock run method that compares the expected and actual docker command.

The test surfaced a bug in PublishLibrary, where the --version flag was incorrectly set to the library ID instead of the library version. This has been fixed.

Fixes https://github.com/googleapis/librarian/issues/592

### DIFF
--- a/internal/docker/docker_test.go
+++ b/internal/docker/docker_test.go
@@ -1,0 +1,223 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


internal/docker/docker_test.go
```diff
@@ -0,0 +1,223 @@
...
+		testAPIRoot         = "testAPIRoot"
+		testGeneratorInput  = "testGeneratorInput"
+		testGeneratorOutput = "testGeneratorOutput"
+		testImage           = "testImage"
+		testInputsDirectory = "testInputsDirectory"
+		testLanguageRepo    = "testLanguageRepo"
+		testLibraryID       = "testLibraryID"
+		testLibraryVersion  = "testLibraryVersion"
+		testOutput          = "testOutput"
+		testOutputDir       = "testOutputDir"
```
jskeet, line 36: Do we need both of these? (Just have `testOutput`?)

jskeet (approved):
One suggestion, but feel free to resolve and merge anyway. (If you do make the change, I'll attempt to review it quickly, but I'm about to stop work for the day.)

PR #610, opened 2025-06-24
fix: remove direct reference of global flag from updateImageTag

zhumin8:
For #498, remaining work to close it is in #560 and #609.

Get tag from config instead of global flag.


zhumin8:
close and include as part of #609

Issue #236, opened 2025-06-08
cli: refactor flagset to Command.FlagSet


julieqiu:
Why does this take a slice of functions instead of one function?

Or why not have Command.FlagSet that command authors can just access directly?

_Originally posted by @jba in https://github.com/googleapis/librarian/pull/235#discussion_r2134021371_

jba:
I was thinking that instead of `Command` having `flags *flag.FlagSet`, it would have
```
Flags flag.FlagSet
```
and then you wouldn't need a function at all. Anyone could call
```
var verbose = mycmd.Flags.Bool("v", ...)
```
or
```
addVerboseFlag(&mycmd.Flags)
```

PR #612, opened 2025-06-25
fix(internal/cli,librarian): remove cli.Command.SetFlags

julieqiu:
Command.Flags is now exported, which allows Command.SetFlags to be removed.

This simplifies flag setup by allowing direct access without needing SetFlags or other helper methods. Flags can now be added inline or via helper functions like:

  cmd.Flags.Bool("v", ...)
  addVerboseFlag(&cmd.Flags)

Fixes https://github.com/googleapis/librarian/issues/236

### DIFF
--- a/internal/librarian/configure.go
+++ b/internal/librarian/configure.go
@@ -17,7 +17,6 @@ package librarian
 import (
 	"context"
 	"errors"
-	"flag"
 	"fmt"
 	"io/fs"
 	"log/slog"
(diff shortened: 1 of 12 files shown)


internal/librarian/configure.go
```diff
@@ -84,19 +83,18 @@ commits will still be present in the language repo.
...
-		addFlagAPIRoot,
-		addFlagGitUserEmail,
-		addFlagGitUserName,
-		addFlagLanguage,
-		addFlagPush,
-		addFlagRepoRoot,
-		addFlagRepoUrl,
-		addFlagSecretsProject,
-	})
+	cmdConfigure.InitFlags()
```
jskeet, line 86: I have a *slight* concern that it's easy to copy/paste a line from a different command here, and forget that you need to change the argument for the function. I'm happy to wait until that actually becomes a problem.

Would it be feasible/sensible in a follow-up PR to remove the ".Flags" for each of these calls? And maybe even make these functions methods receiving a Command instead? So we'd have `cmdConfigure.addFlagImage()` etc. (That would potentially make the potential problem mentioned at the top of this comment less likely to happen, as the part to change would be at the start of each line.)

(That definitely doesn't need to be in this change - I'd be interested to know whether you were already considering it though, or whether it's infeasible/undesirable for some reason I haven't considered.)

jskeet (approved):

PR #625, opened 2025-06-25
fix(internal/github): rename githubrepo to github

julieqiu:
Rename the githubrepo package to github, since this package provides operations on the GitHub API. This also avoids confusion with the gitrepo package, which provides git operations on repositories.


julieqiu:
It looks like `github.com/google/go-github/v69/github` is used directly inside `internal/librarian/mergereleasepr.go`, so we need to refactor that first.

Issue #622, opened 2025-06-25
docker: move validation upstream


julieqiu:
The docker package should assume it receives valid, well-formed inputs. Currently, it includes checks for empty strings and other validation logic that would be better handled when cmd.Parse is called.

Example:
https://github.com/googleapis/librarian/blob/main/internal/docker/docker.go#L108-L116

To simplify the code, we should keep the docker package focused on command construction and execution, and assume that it is receiving valid arguments.

/cc @jskeet in case you have any input that would change this

PR #633, opened 2025-06-25
fix(internal/docker): clean method to unconditionally add the library-id flag

### LINKED ISSUE #630: docker: make the "clean" container command's --library-id flag required
The `Clean` method in docker.go expects that `libraryID` might be an empty string, and only conditionally adds the `--library-id` flag for the container command.

In practice, we always actually pass a library ID, so we should unconditionally add the flag. (We could validate that the parameter is non-empty, but #622 suggests we move away from that pattern anyway.)

zhumin8:
Remove check for empty string and add the `library-id` flag unconditionally.

Fixes #630

### DIFF
--- a/internal/docker/docker.go
+++ b/internal/docker/docker.go
@@ -205,8 +205,8 @@ func (c *Docker) BuildLibrary(repoRoot, libraryID string) error {
 	commandArgs := []string{
 		"--repo-root=/repo",
 		"--test=true",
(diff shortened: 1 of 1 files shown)


zhumin8:
Not adding unit test, per #622 we are moving away from validation within docker.go.

julieqiu (approved):

julieqiu (approved):

PR #640, opened 2025-06-25
feat(internal/githubrepo): implement pagination for GitHub API calls

### LINKED ISSUE #540: githubrepo: implement pagination or use go-github-paginate
In the githubrepo module, we have a few cases where we use the GitHub API to list items (e.g. commits). The API is paginated, but we're only retrieving the first page at the moment. We should either implement pagination manually, or use go-github-paginate which should do this transparently.

(Labeling as "good first issue" as if go-github-paginate is suitable, it should be simple - but I'd appreciate @codyoss's views on whether that's a suitable library.)

dangazineu:
Implemented pagination for GetPullRequestReviews and GetDiffCommits. Also removed unnecessary pagination logic for GetCommit.

Fixes #540.

### DIFF
--- a/internal/githubrepo/githubrepo.go
+++ b/internal/githubrepo/githubrepo.go
@@ -25,6 +25,9 @@ import (
 	"github.com/google/go-github/v69/github"
 )
 
+// pageSize is used in paginated requests. Extrated into a variable for better testability.
+var pageSize = 100
+
 // Client represents this package's abstraction of a GitHub client, including
 // an access token.
 type Client struct {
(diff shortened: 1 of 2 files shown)


julieqiu (changes requested):
Would you be able to add a test for these changes?

internal/githubrepo/githubrepo.go
```diff
@@ -175,10 +178,20 @@ func (c *Client) GetPullRequestCheckRuns(ctx context.Context, pullRequest *githu
 // GetPullRequestReviews fetches all reviews for the pull request identified
 // by prMetadata.
 func (c *Client) GetPullRequestReviews(ctx context.Context, prMetadata *PullRequestMetadata) ([]*github.PullRequestReview, error) {
-	// TODO(https://github.com/googleapis/librarian/issues/540): implement pagination or use go-github-paginate
-	listOptions := &github.ListOptions{PerPage: 100}
-	reviews, _, err := c.PullRequests.ListReviews(ctx, prMetadata.Repo.Owner, prMetadata.Repo.Name, prMetadata.Number, listOptions)
-	return reviews, err
+	var allReviews []*github.PullRequestReview
+	listOptions := &github.ListOptions{PerPage: pageSize}
```
julieqiu, line 182: I would pass in pageSize as an argument to GetPullRequestReviews, since this is the API for the client, and set a const in the package where it is called.

dangazineu:
Closing this so someone else can take it while I'm busy with other tasks.

Issue #561, opened 2025-06-23
librarian: pass context to detectIfLibraryConfigured


julieqiu:
Instead of calling `context.Background()` inside the function, it should be passed through as the first argument of 
detectIfLibraryConfigured

https://github.com/googleapis/librarian/blob/6f0702c855369c36348ac05f4c3144544cef39b4/internal/librarian/generate.go#L239

See https://go.dev/blog/context-and-structs

Issue #532, opened 2025-06-22
docker: pass context explicitly functions that use EnvironmentProvider


julieqiu:
EnvironmentProvider currently stores a context.Context on the struct:
https://github.com/googleapis/librarian/blob/96dd98069587574167d82d0c4c641a1b225f0e85/internal/container/environment.go#L37

We should remove the context field from the struct and instead pass it explicitly as the first argument to any method that requires it. It is preferred to pass context.Context as the first argument to functions, not as a field on a struct, since `context.Context` is meant to control the lifetime of one operation. 

Storing it on a long-lived struct (like `EnvironmentProvider`) can cause it to accidentally outlive the operation it was meant for, or be reused across unrelated operations

See https://go.dev/blog/context-and-structs for details.

PR #641, opened 2025-06-25
refactor(internal/librarian): pass context at runtime for secrets

### LINKED ISSUE #531: container: create package to abstract use of secretmanager
The container package interacts directly with cloud.google.com/go/secretmanager for retrieving secrets. This tight coupling makes the code harder to test.

We should create an `internal/secrets` package that provides an internal abstraction for Secret Manager access, and use that throughout the codebase instead.

chingor13:
Fixes #531

### DIFF
--- a/internal/docker/docker_test.go
+++ b/internal/docker/docker_test.go
@@ -15,6 +15,7 @@
 package docker
 
 import (
+	"context"
 	"fmt"
 	"testing"
 
@@ -50,7 +51,7 @@ func TestDockerRun(t *testing.T) {
 		{
 			name: CommandGenerateRaw,
 			runCommand: func() error {
-				return d.GenerateRaw(testAPIRoot, testOutput, testAPIPath)
+				return d.GenerateRaw(context.Background(), testAPIRoot, testOutput, testAPIPath)
 			},
 			want: []string{
 				"run", "--rm",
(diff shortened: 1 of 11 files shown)


internal/docker/docker_test.go
```diff
@@ -50,7 +51,7 @@ func TestDockerRun(t *testing.T) {
 		{
 			name: CommandGenerateRaw,
 			runCommand: func() error {
```
julieqiu, line 53: change this:

runCommand: func() error { ... }

to 

runCommand: func(ctx context.Context) error {

Then below call:

ctx := context.Background()
test.runCommand(ctx)

julieqiu (dismissed):

julieqiu:
> Fixes #531 Fixes #532 Fixes #561
> 
> There is likely more refactoring to be done if we want to compile all of the config before executing docker commands. Currently, config/environment files are fetched lazily right before we write the environment file.

Swap the order per https://github.com/googleapis/librarian/blob/main/CONTRIBUTING.md#commit-messages

internal/docker/docker_test.go
chingor13: Done

julieqiu (approved):

julieqiu (approved):

PR #665, opened 2025-06-27
refactor(internal): move user lookup to config

### LINKED ISSUE #590: docker: move user lookup to config
Moves the call to user.Current into the config package. 

https://github.com/googleapis/librarian/blob/f88310179258de5e6b6f33ab8eac0004b41dff59/internal/docker/docker.go#L261

This makes it easier to test and reuse, and keeps Docker-related logic focused on container setup
rather than host introspection.

zhumin8:
Moves the call to user.Current into the config package, keep Docker-related logic focused.

Fixes #590

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -166,13 +166,13 @@ type Config struct {
 	// GitUserName is specified with the -git-user-name flag.
 	GitUserName string
 
-	// GID is the group ID of the current user. It is used to run Docker
+	// UserGID is the group ID of the current user. It is used to run Docker
 	// containers with the same user, so that created files have the correct
 	// ownership.
 	//
 	// This is populated automatically after flag parsing. No user setup is
 	// expected.
-	GID string
+	UserGID string
 
 	// Image is the language-specific container image to use for language-specific
 	// operations. It is primarily used for testing Librarian and/or new images.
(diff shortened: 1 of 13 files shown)


internal/config/config.go
```diff
@@ -164,6 +166,14 @@ type Config struct {
 	// GitUserName is specified with the -git-user-name flag.
 	GitUserName string
 
+	// GID is the group ID of the current user. It is used to run Docker
+	// containers with the same user, so that created files have the correct
+	// ownership.
+	//
+	// This is populated automatically after flag parsing. No user setup is
+	// expected.
+	GID string
```
julieqiu, line 175: I would name this UserGid to match https://pkg.go.dev/os/user#User#Gid

internal/config/config.go
```diff
@@ -348,6 +358,14 @@ type Config struct {
 	// TagRepoURL is specified with the -tag-repo-url flag.
 	TagRepoURL string
 
+	// UID is the user ID of the current user. It is used to run Docker
+	// containers with the same user, so that created files have the correct
+	// ownership.
+	//
+	// This is populated automatically after flag parsing. No user setup is
+	// expected.
+	UID string
```
julieqiu, line 367: I would name this User.Uid to match https://pkg.go.dev/os/user#User#Uid

internal/config/config.go
```diff
@@ -368,6 +386,19 @@ func New() *Config {
 	}
 }
 
+// SetupUser performs late initialization of user-specific configuration,
+// determining the current user. This is in a separate method as it
+// can fail, and is called after flag parsing.
+func (c *Config) SetupUser() error {
```
julieqiu, line 392: since this gets called inside librarian. Run, I would remove this and move it to `New`:

```
func New() (*Config, error) {
	currentUser, err := user.Current()
	if err != nil {
		return fmt.Errorf("failed to get current user: %w", err)
	}
        ...
}
```

julieqiu (changes requested):

julieqiu (approved):

PR #668, opened 2025-06-27
fix(internal/librarian): don't display both usage twice

### LINKED ISSUE #666: librarian: requesting help for an individual flag shows overall usage too
Running:

```sh
go run ./cmd/librarian generate -h
```

... displays the help for the generate command, but then *also* displays the more general usage documentation starting with "Librarian manages client libraries for Google APIs."

I think if we should only show the general usage documentation if Librarian is run without a specific command.

jskeet:
If Command.Parse fails having found a specific command, we can expect that command to have displayed a suitable command-specific-usage message; we don't need to display the generic usage as well.

This affects both the case of "help explicitly requested with -h" as well as "a flag that's not registed for this command was supplied".

Fixes #666

### DIFF
--- a/internal/librarian/librarian.go
+++ b/internal/librarian/librarian.go
@@ -63,7 +63,9 @@ func Run(ctx context.Context, arg ...string) error {
 		return err
 	}
 	if err := cmd.Parse(arg[1:]); err != nil {
-		CmdLibrarian.Flags.Usage()
+		// We expect that if cmd.Parse fails, it will already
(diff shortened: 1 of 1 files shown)


jskeet:
This doesn't include a test as I couldn't figure out how to intercept output (from outside the cli package). I'm happy to close this and have another go next week if there's a straightforward approach that I've missed.

julieqiu (approved):

PR #704, opened 2025-06-30
chore(internal/config): rename environment variables

### LINKED ISSUE #682: librarian: update KOKORO* envvars to instead be prefixed as LIBRARIAN
Ref: go/librarian:cli-reimagined

We are renaming these envvars.

JoeWang1127:
Rename `KOKORO_*` environment variables to `LIBRARIAN_*`.

Fixes #682


zhumin8 (changes requested):
nit: these env var names are referred from the documentation comments here, can you also rename these?
https://github.com/googleapis/librarian/blob/fac59553a099f8650f183fed1e4509b6048ded9d/internal/config/config.go#L114

zhumin8 (approved):

Issue #326, opened 2025-06-11
all: migrate pipeline files to YAML


julieqiu:
See https://github.com/googleapis/librarian/issues/245

PR #707, opened 2025-06-30
refactor(internal): remove language flag and usages

### LINKED ISSUE #675: remove language flag
Ref: go/librarian:cli-reimagined

Remove the language flag and any detection log that was being used as a part of this flag. We will now require the CLI to be more explicitly with the language repo it is processing.

zhumin8:
Remove the language flag and logic that was being used as a part of this flag. 

- `deriveImage` now look for -image flag first,  If not specified, use an image configured in the `config.yaml`. If both are not provided, returns an error.  Note that pipeline_state.json to state.yaml migration is not implemented yet, thus taking the old `state.ImageTag` as placeholder for now. ref: https://github.com/googleapis/librarian/issues/326, go/librarian:cli-reimagined
- removed [`findApisToConfigure`](https://github.com/googleapis/librarian/commit/79d33c0d65307ec2c790b11a1f558526d8eb4322) from configure command.

Fixes #675

### DIFF
--- a/internal/librarian/command.go
+++ b/internal/librarian/command.go
@@ -156,8 +156,8 @@ func deriveImage(imageOverride string, state *statepb.PipelineState) (string, er
 	if state == nil {
 		return "", nil
 	}
-	// TODO: use image from state.yaml when switch to this config file. see go/librarian:cli-reimagined
-	// https://github.com/googleapis/librarian/issues/326
+	// TODO(https://github.com/googleapis/librarian/issues/326):
+	// use image from state.yaml when switch to this config file. see go/librarian:cli-reimagined
 	if state.ImageTag == "" {
 		return "", errors.New("pipeline state does not have image specified and no override was provided")
 	}


internal/librarian/command.go
```diff
@@ -146,24 +149,18 @@ func appendResultEnvironmentVariable(workRoot, name, value, envFileOverride stri
...
-		tag = "latest"
-	} else {
-		tag = state.ImageTag
+		return "", nil
 	}
-	if defaultRepository == "" {
-		return fmt.Sprintf("%s:%s", relativeImage, tag)
-	} else {
-		return fmt.Sprintf("%s/%s:%s", defaultRepository, relativeImage, tag)
+	// TODO: use image from state.yaml when switch to this config file. see go/librarian:cli-reimagined
```
julieqiu, line 159: Link to an issue corresponding to the TODO

internal/librarian/command_test.go
```diff
@@ -45,47 +45,50 @@ func TestCommandUsage(t *testing.T) {
...
+			imageOverride: "",
+			state:         &statepb.PipelineState{ImageTag: ""},
+			wantErr:       true,
 		},
 	} {
 		t.Run(test.name, func(t *testing.T) {
-			got := deriveImage(test.language, test.imageOverride, test.defaultRepository, test.state)
+			got, err := deriveImage(test.imageOverride, test.state)
 
+			if (err != nil) != test.wantErr {
```
julieqiu, line 88: I would structure like this for readability 

if test.wantErr {
...
}

if err != nil {
...
}

julieqiu (approved):

julieqiu (approved):

PR #721, opened 2025-07-01
chore(all): turn on codecov

### LINKED ISSUE #699: all: turn on codecov
https://github.com/codecov

JoeWang1127:
Fixes #699


zhumin8:
from earlier GVC session, @ldetmer to confirm we want to turn on codecov.

.github/workflows/librarian.yaml
```diff
@@ -27,7 +27,11 @@ jobs:
     - name: Display Go version
       run: go version
     - name: Run tests
-      run: go test ./...
+      run: go test -coverprofile=coverage.out ./...
```
julieqiu, line 30: how does this test work? when I run it locally it seems to just print the coverage percentage, but never fails

JoeWang1127: > how does this test work? when I run it locally it seems to just print the coverage percentage, but never fails

I've added the codecov config but it didn't fail the workflow when the coverage is below 90%.

According to this [step](https://docs.codecov.com/docs/github-2-getting-a-codecov-account-and-uploading-coverage), we may need to merge this pull request first.

ldetmer:
We will allow this, based off [google-cloud-rust's usage](https://github.com/googleapis/google-cloud-rust/blob/main/.codecov.yml) setting precendent.

ldetmer (approved):

PR #713, opened 2025-07-01
chore(golangci-lint): fix issues for lll and turn on

amghr:
Fixes #691

PR #732, opened 2025-07-02
chore(golangci-lint): add gofmt, goimports and golines and fix issues

### LINKED ISSUE #730: librarian: add formatters to .golangci
From https://github.com/googleapis/librarian/pull/713#issuecomment-3028433884 
Add formatters from https://golangci-lint.run/usage/formatters/ to .golangci to help developers fix linter issues.

When added, developer can run `golangci-lint run --fix` to fix common linter issues.

zhumin8:
Added formatters and fixed issues, also added documentation on usage.

Note golines formatter and lll linter together can leads to conflicts, even if aligning settings. Preferring golines as it can be used with `golangci-lint run --fix` for shorten long lines automatically.
Will close #713 to avoid conflict.

Fixes #730

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -39,6 +39,3 @@ formatters:
     - gofmt
     - goimports
     - golines
-  settings:
-    golines:
-      max-len: 120 # Align with lll's max-len
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -32,3 +32,13 @@ linters:
     - unused
     - usetesting
     
+formatters:
+  # See https://golangci-lint.run/usage/formatters for details.
+  default: none
+  enable:
+    - gofmt
+    - goimports
+    - golines
+  settings:
+    golines:
+      max-len: 120 # Align with lll's max-len
--- a/doc/howwewritego.md
+++ b/doc/howwewritego.md
@@ -127,6 +127,14 @@ handling:
 - [Generics Tutorial](https://go.dev/doc/tutorial/generics)
 - [Error Handling with Generics](https://go.dev/blog/error-syntax)
 
+## Linters and Formatter
+
+This repository has linters and formatters configured in .golangci.yaml.
+To install golangci-lint locally, see https://golangci-lint.run/welcome/install/#binaries
+
+To run linter: `golangci-lint run`
+
+To run formatter: `golangci-lint run --fix`
 
 ## Need Help? Just Ask!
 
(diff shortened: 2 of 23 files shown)


doc/howwewritego.md
```diff
@@ -127,6 +127,14 @@ handling:
 - [Generics Tutorial](https://go.dev/doc/tutorial/generics)
 - [Error Handling with Generics](https://go.dev/blog/error-syntax)
 
+## Linters and Formatter
```
julieqiu, line 130: I would write this as:

## Linters

All linters for this repository run as part of `go test ./...` and are included in [all_test.go](https://github.com/googleapis/librarian/blob/main/all_test.go).

Most linters run via [`TestGolangCILint`](https://github.com/googleapis/librarian/blob/4dbdf10432c152ea9e563f471b39aa5f564e284e/all_test.go#L99), and are configured using [.golangci.yaml](https://github.com/googleapis/librarian/blob/main/.golangci.yaml) following the [golangci-lint configuration guidelines](https://golangci-lint.run/usage/configuration/).

### Running Linters

To run `golangci-lint` as a test, run:

```sh
go test -run TestGolangCILint
```

Or to run it directly:

```
go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@latest run
```

To run the formatter and apply fixes:

```
go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@latest run --fix
```

internal/config/config_test.go
```diff
@@ -196,7 +196,8 @@ func TestIsValid(t *testing.T) {
 			if (err != nil) != test.wantErr {
 				t.Errorf("IsValid() got error = %v, want error = %t", err, test.wantErr)
 			}
-			if test.wantErr && err != nil && err.Error() != "no GitHub token supplied for push" && err.Error() != "unable to parse push config" {
+			if test.wantErr && err != nil && err.Error() != "no GitHub token supplied for push" &&
+				err.Error() != "unable to parse push config" {
```
julieqiu, line 200: We should never compare `err.Error()` as a test case, because it is brittle. Filing an issue for a follow up.

.golangci.yaml
```diff
@@ -32,3 +32,10 @@ linters:
     - unused
     - usetesting
     
+formatters:
+  # See https://golangci-lint.run/usage/formatters for details.
+  default: none
+  enable:
+    - gofmt
+    - goimports
+    - golines
```
julieqiu, line 41: I'm not sure that golines makes the code more readable based on the changes below, and gofmt and goimports should run as part of gopls. 

If you have the Go VS Code extension, it will automatically run when you save your files:
https://code.visualstudio.com/docs/languages/go

My recommendation would be to remove `golines`

Issue #774, opened 2025-07-07
librarian: remove all proto references and generate scripts


codyoss:
We will no longer be using protobufs at all in this repo. This issue should:
- delete the proto directory
- ensure we don't have a hard dependency on `google.golang.org/protobuf` in go.mod
- remove generate-protos.sh script

Blocked By: #772
Blocked By: #773

Issue #744, opened 2025-07-04
librarian: restore code coverage target to 80%


jskeet:
`codecov.yml` specifies the project coverage target. We're reducing it from 90% temporarily, to unblock work, but we want to restore it to 90%.

Aim to do this for 2025-07-11, but if we aren't there just yet, increase it then to whatever we're at, and then restore it to 90% by 2025-07-18.

codyoss:
we hit it!

jskeet:
@codyoss: Presumably we can now up the config? Let me know if you want me to do that tomorrow.

codyoss:
@jskeet Just opened the PR ;)

PR #800, opened 2025-07-08
refactor: remove dep on protos

codyoss:
Temporarily moving the definition of the state file into raw Go structs. The generated proto dir is a big hit to coverage so removing it more proactively will help us have more accurate metrics.

Updates: #744
Fixes: #774


codyoss:
This will likely affect changes other might have locally because of the prevalence of types in the repo, FYI.

zhumin8 (approved):

codyoss:
Using admin to bypass the patch failure.

PR #807, opened 2025-07-09
feat(internal/librarian): rework generate container contract

### LINKED ISSUE #770: librarian: rework generate container contract
Reference: go/librarian:cli-reimagined

context | type | description
-- | -- | --
/librarian/generate-request.json | mount | (read) | A JSON file that describes which library to generate. 
/input | mount (read/write) | The exact contents of the generator-input folder, e.g. google-cloud-go/.librarian/generator-input. Example contents are: - Templates ([example](https://github.com/googleapis/google-cloud-dotnet/blob/main/generator-input/README-template.md)) -Logic to run tweaks ([example](https://github.com/googleapis/google-cloud-dotnet/blob/main/generator-input/tweaks/Google.Cloud.AIPlatform.V1/pregeneration.sh)). This folder has read/write access to allow the container to add any new language specific configuration required.
/output | mount (write) | This folder is mounted into the container. It is meant to be the destination for any code generated by the container. Its output structure should match that of where the code should land in the resulting repository. For example if we are generating the [secretmanger v1](https://github.com/googleapis/google-cloud-go/tree/main/secretmanager/apiv1) client for Go, we would write files to `/output/secretmanager`.
/source | mount (read) | This folder is mounted into the container. It contains, for example, the whole contents of [googleapis](https://github.com/googleapis/googleapis). This will be needed in order to read the  service config files and likely also the BUILD.bazel files that hold a lot of configuration today.
command | Positional Argument | The value will always be `generate`` for this invocation.

JoeWang1127:
Fixes #770


internal/librarian/createreleaseartifacts.go
```diff
@@ -153,13 +153,13 @@ func copyMetadataFiles(outputRoot string, releases []LibraryRelease, languageRep
 	if err := languageRepo.Checkout(finalRelease.CommitHash); err != nil {
 		return err
 	}
-	sourceStateFile := filepath.Join(languageRepo.Dir, config.GeneratorInputDir, pipelineStateFile)
+	sourceStateFile := filepath.Join(languageRepo.Dir, config.LibrarianDir, pipelineStateFile)
```
JoeWang1127, line 156: I made this change because the pipeline state file should live in `.librarian` directory.

internal/librarian/generate.go
```diff
@@ -183,9 +183,10 @@ func runGenerateCommand(ctx context.Context, cfg *config.Config, outputDir strin
 		if libraryID == "" {
 			return "", errors.New("bug in Librarian: Library not found during generation, despite being found in earlier steps")
 		}
-		generatorInput := filepath.Join(languageRepo.Dir, config.GeneratorInputDir)
+		generateRequest := filepath.Join(languageRepo.Dir, config.LibrarianDir, config.GenerateRequest)
+		generatorInput := filepath.Join(languageRepo.Dir, config.LibrarianDir, config.GeneratorInputDir)
```
JoeWang1127, line 187: `generator-input` directory should be a subdirectory of `.librarian`.

zhumin8 (dismissed):

internal/config/config.go
```diff
@@ -24,10 +24,15 @@ import (
 )
 
 const (
+	DefaultPushConfig string = "noreply-cloudsdk@google.com,Google Cloud SDK"
 	// GeneratorInputDir is the default directory to store files that generator
 	// needs to regenerate libraries from an empty directory.
 	GeneratorInputDir string = "generator-input"
-	DefaultPushConfig string = "noreply-cloudsdk@google.com,Google Cloud SDK"
+	// GenerateRequest is a JSON file that describes which library to generate.
+	GenerateRequest string = "generate-request.json"
```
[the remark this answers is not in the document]
zhumin8: > Also, I think this const should live in the docker package as it will be the only place the name will be referenced.

They are also referenced in internal/librarian/generate.go

internal/librarian/generate.go
```diff
@@ -245,3 +250,31 @@ func (r *generateRunner) detectIfLibraryConfigured(ctx context.Context) (bool, e
 	slog.Info("API configured", "path", apiPath, "library", libraryID)
 	return true, nil
 }
+
+func toGenerateRequestJSON(state *config.PipelineState, filepath string) error {
```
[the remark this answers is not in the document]
JoeWang1127: Done.

internal/librarian/generate.go
```diff
@@ -245,3 +250,31 @@ func (r *generateRunner) detectIfLibraryConfigured(ctx context.Context) (bool, e
...
+	data, err := json.MarshalIndent(state, "", "  ")
+	if err != nil {
+		slog.Warn("failed to marshal state to JSON", "state", state, "error", err)
+		return err
```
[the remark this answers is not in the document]
JoeWang1127: Done.

internal/docker/docker.go
```diff
@@ -84,17 +84,19 @@ func New(workRoot, image, secretsProject, uid, gid string, pipelineConfig *confi
 // output specifies the empty output directory into which the command should
 // generate code, and libraryID specifies the ID of the library to generate,
 // as configured in the Librarian state file for the repository.
-func (c *Docker) Generate(ctx context.Context, cfg *config.Config, apiRoot, output, generatorInput, libraryID string) error {
+func (c *Docker) Generate(ctx context.Context, cfg *config.Config, apiRoot, output, generateRequest, generatorInput, libraryID string) error {
 	commandArgs := []string{
-		"--source=/apis",
+		"--request=/request",
```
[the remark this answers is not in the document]

internal/docker/docker.go
```diff
@@ -84,17 +84,19 @@ func New(workRoot, image, secretsProject, uid, gid string, pipelineConfig *confi
...
+		"--input=/input",
 		"--output=/output",
-		fmt.Sprintf("--%s=/%s", config.GeneratorInputDir, config.GeneratorInputDir),
+		"--source=/source",
 		fmt.Sprintf("--library-id=%s", libraryID),
 	}
 	mounts := []string{
-		fmt.Sprintf("%s:/apis", apiRoot),
+		fmt.Sprintf("%s:/request:ro", generateRequest), // readonly volume.
```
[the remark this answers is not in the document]

internal/librarian/generate_test.go
```diff
@@ -0,0 +1,111 @@
...
+			expectErr: false,
+		},
+		{
+			name:      "empty-pipelineState",
+			state:     &config.PipelineState{},
+			expectErr: false,
+		},
+	}
+
+	for _, tc := range testCases {
```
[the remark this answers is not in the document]
JoeWang1127: Done.

internal/librarian/generate.go
```diff
@@ -192,9 +192,10 @@ func (r *generateRunner) runGenerateCommand(ctx context.Context, outputDir strin
 		if libraryID == "" {
 			return "", errors.New("bug in Librarian: Library not found during generation, despite being found in earlier steps")
 		}
+		generateRequest := filepath.Join(r.repo.Dir, config.LibrarianDir, config.GenerateRequest)
```
codyoss, line 195: I think this can be figured out in the docker package too. I think the value should be in the outputDir, not in the repo.

testdata/empty-pipelineState.json
```diff
@@ -0,0 +1 @@
+{}
```
codyoss, line 1: nit: add newlines to both json files.

internal/docker/docker.go
```diff
@@ -84,17 +86,22 @@ func New(workRoot, image, secretsProject, uid, gid string, pipelineConfig *confi
 // output specifies the empty output directory into which the command should
 // generate code, and libraryID specifies the ID of the library to generate,
 // as configured in the Librarian state file for the repository.
-func (c *Docker) Generate(ctx context.Context, cfg *config.Config, apiRoot, output, generatorInput, libraryID string) error {
+func (c *Docker) Generate(ctx context.Context, cfg *config.Config, state *config.PipelineState, apiRoot, output, generateRequest, generatorInput, libraryID string) error {
```
codyoss, line 89: nit: This is a lot of inputs for a function. Could you refactor this into a `GenerateRequest struct` that holds all but the context value. That way we can better document what each field is as well.

internal/docker/docker.go
```diff
@@ -200,3 +207,23 @@ func (c *Docker) runCommand(cmdName string, args ...string) error {
 	slog.Info(fmt.Sprintf("=== Docker end %s", strings.Repeat("=", 65)))
 	return err
 }
+
+func toGenerateRequestJSON(state *config.PipelineState, filepath string) error {
+	data, err := json.MarshalIndent(state, "", "  ")
+	if err != nil {
+		return errors.New("failed to marshal state to JSON")
```
codyoss, line 214: It is almost always a good idea to preserve the original error context when returning errors in Go. The language has an error wrapping mechanism. For example this error would turn into:
```suggestion
		return fmt.Errorf("failed to marshal state to JSON: %w", err)
```

codyoss (changes requested):
Just a couple more small things. Thanks for the refactoring, looks good.

testdata/empty-pipelineState.json
JoeWang1127: Done.

internal/docker/docker.go

internal/docker/docker.go

internal/librarian/generate.go
JoeWang1127: I thought the `generate-request.json` should live in `.librarian` directory, which is at the root of the language repo.

internal/docker/docker.go
```diff
@@ -79,25 +125,27 @@ func New(workRoot, image, secretsProject, uid, gid string, pipelineConfig *confi
...
-		fmt.Sprintf("--%s=/%s", config.GeneratorInputDir, config.GeneratorInputDir),
-		fmt.Sprintf("--library-id=%s", libraryID),
+		"--source=/source",
+		fmt.Sprintf("--library-id=%s", request.libraryID),
 	}
 	mounts := []string{
-		fmt.Sprintf("%s:/apis", apiRoot),
-		fmt.Sprintf("%s:/output", output),
-		fmt.Sprintf("%s:/%s", generatorInput, config.GeneratorInputDir),
+		fmt.Sprintf("%s:/librarian:ro", request.requestJson), // readonly volume.
```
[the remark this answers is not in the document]
codyoss: librarian is correct on the side of the container. It does map to .librarian of the language repo.

internal/docker/docker.go
```diff
@@ -62,6 +65,51 @@ type Docker struct {
 	run func(args ...string) error
 }
 
+// GenerateRequest contains all the information required for a language
+// container to run the generate command.
+type GenerateRequest struct {
+	// repo is a pointer to the [gitrepo.Repository] struct, representing the
+	// language repository.
+	repo *gitrepo.Repository
```
codyoss, line 73: nit, the only information this package needs from this object is the dir. Maybe make this field a string and called RepoDir. That way you can decouple this whole package from needing to import gitrepo

codyoss (changes requested):

internal/docker/docker.go

codyoss (approved):
Thanks!

internal/librarian/generate.go
codyoss: this is a just temporary file that is never persisted. It is only used to communicate to-from the container contract.

internal/docker/docker.go
```diff
@@ -200,3 +227,26 @@ func (c *Docker) runCommand(cmdName string, args ...string) error {
 	slog.Info(fmt.Sprintf("=== Docker end %s", strings.Repeat("=", 65)))
 	return err
 }
+
+func toGenerateRequestJSON(state *config.PipelineState, jsonFilePath string) error {
+	data, err := json.MarshalIndent(state, "", "  ")
```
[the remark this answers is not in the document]
JoeWang1127: Thanks, now I understand the json file contains `LibraryState`.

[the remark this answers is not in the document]

codyoss (dismissed):

codyoss (approved):

zhumin8 (approved):
LGTM. 
Thanks for waiting!
nit: Can you add some description to this PR?(https://github.com/googleapis/librarian/blob/main/CONTRIBUTING.md#commit-messages)

PR #813, opened 2025-07-09
chore(all): exclude .idea from header tests

### LINKED ISSUE #572: `TestHeaders` conflicts with local files
`TestHeaders` fails for me because my IDE config files aren't excluded from the test. Filtering out for `.gitignore` patterns (plus global ignore rules, which are not stored in the repo's .gitignore) is overkill, and hardcoding every possible file to the test exclusions seems too brittle.

My suggestion is for the test to read an environment variable with values for exclusion from the test, allowing for different environments to exclude different files.

Files that are checked into the repo, and are not supposed to be tested for headers, should still be excluded in code. This issue is specifically for files that are environment-specific.

dangazineu:
For #572.

### DIFF
--- a/all_test.go
+++ b/all_test.go
@@ -54,7 +54,7 @@ func TestHeaders(t *testing.T) {
 			return err
 		}
 		if d.IsDir() {
-			if d.Name() == "testdata" || d.Name() == ".git" || d.Name() == ".vscode" {
+			if d.Name() == "testdata" || d.Name() == ".git" || d.Name() == ".vscode" || d.Name() == ".idea" {
 				return fs.SkipDir
 			}
 			return nil


dangazineu:
@codyoss I saw your suggestion on #572 to use an allowlist, and I think we could keep the issue open till that more comprehensive solution is implemented. For now I suggest we just add `.idea/` to the folder exclusion list.

codyoss:
Sounds good, I know I noticed this with .vscode which I am adding in another PR too.

codyoss (approved):

codyoss:
hmmm I am confused why codecov is unhappy here.

codyoss (approved):

Issue #753, opened 2025-07-04
test(generate): add unit tests for generate.go


dangazineu:
**Title:** test(generate): add unit tests for generate.go

**Body:**
The `internal/librarian/generate.go` file is responsible for generating client library code. This is a critical part of the release process, as it creates the code that is ultimately shipped to users.

Currently, this file has no unit tests, which makes it difficult to ensure that it correctly handles all generation scenarios.

This issue tracks the work to add comprehensive unit tests for the functions in `generate.go`.

#### `runGenerate`
The tests for `runGenerate` should cover:
- Both "refined" and "raw" generation modes.
- Handling of the `-build` flag.
- Mocking the `docker.Docker` and `gitrepo.Repository` dependencies to isolate the functions under test.
- Verifying that the correct container commands are called with the correct arguments.

#### `detectIfLibraryConfigured`
The tests for `detectIfLibraryConfigured` should cover:
- Handling of local and remote repositories.
- Correctly identifying configured libraries.
- Handling of missing or invalid pipeline state files.
- Mocking the `github.Client` to avoid making real API calls.

#### `executeGenerate`
The tests for `executeGenerate` should cover:
- Verifying that the correct `generate` command is called (`GenerateLibrary` or `GenerateRaw`).
- Verifying that the `Clean` and `BuildLibrary` commands are called when the `-build` flag is specified in refined generation mode.
- Verifying that the `BuildRaw` command is called when the `-build` flag is specified in raw generation mode.

### Acceptance Criteria
- All functions in `internal/librarian/generate.go` have unit tests.
- The new tests cover both success and error cases.
- The overall test coverage for `generate.go` is at least 90%.

Child issue of #202

PR #816, opened 2025-07-10
refactor(inernal/librarian): setup code for testing

codyoss:
Code is functionally equivalent to what was previously there, just moved the "loading" of data to a type constructor and refactored.

- Create a runner type to hold all preloaded state/config
- Turn functions into methods of runner
- reduce inputs to funcs

Updates: #753


zhumin8 (approved):
Thanks for the refactoring!
FYI - @JoeWang1127 possible merge conflicts for #807

PR #819, opened 2025-07-10
test(internal/librarian): add tests for generate.go

codyoss:
This is just the first pass at adding some testing to this layer. To get better coverage there is still some more refactoring that needs to be done to get abstract away some of the operations that involve cloning repositories and what not.

In this PR I made the following changes:
- Created interfaces for injected components so we can mock them away in tests
- Inject more information in the constructor for the generate runner
- Break up generate codepaths into more testable functions
- Add the tests
- Fixed a bug that had state/config files located in generator input

Updates: #753


zhumin8 (approved):
Thanks for the refactoring. This also makes testing some methods within stateandconfig.go easier.

Issue #756, opened 2025-07-04
test(clone): add unit tests for clone.go


dangazineu:
**Title:** test(clone): add unit tests for clone.go

**Body:**
The `internal/librarian/clone.go` file is responsible for cloning the googleapis repository. This is a critical part of the release process, as it provides the API definitions used to generate the libraries.

Currently, this file has no unit tests, which makes it difficult to ensure that it correctly handles all cloning scenarios.

This issue tracks the work to add comprehensive unit tests for the `cloneGoogleapis` function.

### Refactoring for Testability

The `cloneGoogleapis` function is a thin wrapper around `gitrepo.NewRepository`, which makes it difficult to test in isolation. To make it more testable, we should introduce an interface that `gitrepo.Repository` implements, and then use that interface in `cloneGoogleapis`. This will allow us to mock the `gitrepo.Repository` in our tests.

Alternatively, we can introduce a variable for the `gitrepo.NewRepository` function, which can be replaced with a mock implementation in our tests.

### Implementation Guidance
The tests should cover:
- Verifying that `gitrepo.NewRepository` is called with the correct arguments (`repoPath`, `MaybeClone`, `RemoteURL`, `CI`).
- Handling errors returned by `gitrepo.NewRepository`.

### Acceptance Criteria
- The `cloneGoogleapis` function in `internal/librarian/clone.go` has unit tests.
- The new tests cover both success and error cases.
- The overall test coverage for `clone.go` is at least 90%.

Child issue of #202

Issue #805, opened 2025-07-09
librarian: add push-config flag to generate


codyoss:
go/librarian:cli-reimagined

We noticed in #804 that this flag was not on generate yet. We will be removing the relevant code for this flag in the linked PR. We should add it back for generate and bolster test around this flag.

cuiy0006:
I noticed I can keep `PushConfig` in config and only delete the flag. So for this ticket, we only need to add back the flag.

catchyzheng:
Looks like the `push-config` flag is renamed to `push` in the newer design doc? go/librarian:generate-command

codyoss:
@catchyzheng This is true. For our July milestone lets assume the go/librarian:cli-reimagined is still the source of truth so we have a consistent target.

PR #856, opened 2025-07-11
test(internal/gitrepo): add tests, rm unused code

codyoss:
Some of this code will likely need to be restored in #805, but removing for now so we can evaluate if we need it in the future or not.

Fixes: #756


jskeet:
Drive-by non-review: this would get us up to 65% coverage, wahoo :)

zhumin8 (approved):

Issue #771, opened 2025-07-07
librarian: rework build container contract


codyoss:
Reference: go/librarian:cli-reimagined

context | type | description
-- | -- | --
/librarian | mount (read) | The exact contents of the `.librarian` folder in the language repository. Additionally this will contain a file name `build-request.json` describing the library being processed.
/repo | mount (read/write) | The whole language repo. The mount is read/write to make diff-testing easier. Any changes made to this directory will have no-effect on the generated code, it is a deep-copy.
command | Positional Argument | The value will always be `build` for this invocation.

PR #863, opened 2025-07-14
feat(internal/librarian): rework build container contract

JoeWang1127:
Rework build command container contract
- Create `build-request.json`
- Mount `.librarian` to `/librarian`
- Mount repo root to `/repo`

Fixes: #771


internal/config/config.go
```diff
@@ -24,6 +24,8 @@ import (
 )
 
 const (
+	// BuildRequest is a JSON file that describes which library to build/test.
+	BuildRequest      string = "build-request.json"
```
zhumin8, line 28: Question, what is different in content between "build-request.json" to "generate-request.json"?

internal/librarian/generate.go
```diff
@@ -212,23 +210,35 @@ func (r *generateRunner) runGenerateCommand(ctx context.Context, outputDir strin
...
+		RepoDir:   r.repo.Dir,
+	}
+
+	slog.Info("Build requested in the context of refined generation; cleaning and copying code to the local language repo before building.")
+	// TODO(https://github.com/googleapis/librarian/issues/775)
+	if err := os.CopyFS(r.repo.Dir, os.DirFS(outputDir)); err != nil {
+		return err
+	}
+
+	return r.containerClient.Build(ctx, buildRequest)
```
zhumin8, line 241: I missed this in the previous "generate" PR. But should we cleanup the json file after execution? Otherwise, we will end up with a lot of old generate/build request files in .librarian/ dir.

internal/docker/docker_test.go
```diff
@@ -188,9 +216,14 @@ func TestDockerRun(t *testing.T) {
 				return nil
 			}
 			ctx := t.Context()
-			if err := test.runCommand(ctx, test.docker); err != nil {
-				t.Fatal(err)
+			err := test.runCommand(ctx, test.docker)
+
+			if (err != nil) != test.wantErr {
```
zhumin8, line 221: nit: Can you break this condition down for readability?

zhumin8 (changes requested):

internal/docker/docker_test.go
JoeWang1127: I actually copy this from https://github.com/googleapis/librarian/blob/ac2ae0792c0c76ed11eda4b0f9c876dcc002b0b4/internal/librarian/generate_test.go#L132

internal/config/config.go
JoeWang1127: Looking at the two examples in go/librarian:language-onboarding, I don't see any difference.

zhumin8: fyi - I asked similar question [here](https://docs.google.com/document/d/1bQ5aXIfQEauI6cHlZKh8R6PBAG_sfHRaDxHcxRCQMYI/edit?resourcekey=0-TyGOVF1t5hGtF0Pydz572w&disco=AAABk_vU9oc). We should keep the current names for Java/Go onboarding.

internal/librarian/generate.go
JoeWang1127: How about creating an clean up issue so that I can remove this file in a follow up PR?

PR #866, opened 2025-07-14
chore: add GEMINI.md

codyoss:
I have been trying to lean into using gemini more in my personal devflow. I think adding this file could help others ground thier environment for better results.


codyoss:
Thanks for the feedback!

dangazineu (approved):

Issue #724, opened 2025-07-01
librarian: remove the default for push-config


cuiy0006:
If the flag `push-config` is not specified, do not create commit. Currently commits with a default name and email.

This is a followup to #678.

cuiy0006:
> I have resolved the comments, please proceed as documented in the design doc.

I found this reply in the comment,
```
just updated files in the output directory
```
Can you make it clearer,
1. generateReleaseCommit
2. createPullRequest
3. AddLabelToPullRequest

`push-config` is only used in `generateReleaseCommit` for `commitAll`, do we only skip `commitAll` and still create pull request? Or skip all commit/pr operations?

codyoss:
Lets not worry too much about this bug until later in the month. But for generate this comes into play after all containers have run if the flag has been specified.

catchyzheng:
@codyoss  IIUC, the requirement for this ticket has been satisfied from my last PR. 
"We should only create a commit if this flag is specified." is implemented [here](https://github.com/googleapis/librarian/blob/main/internal/librarian/command.go#L141)

"for generate this comes into play after all containers have run if the flag has been specified." see [here](https://github.com/googleapis/librarian/blob/main/internal/librarian/generate.go#L188)

PR #868, opened 2025-07-14
feat(internal/librarian): add push-config flag and corresponding functionality

catchyzheng:
feat(internal/librarian): add `push-config` flag and its functionality.

When push-config and github token both set and exist, we do the following:
1. add all changes to staging. 
2. commit them.
3. create pull request and push to the repo.

Fixes: https://github.com/googleapis/librarian/issues/805


internal/librarian/flags.go
```diff
@@ -34,6 +34,11 @@ func addFlagHostMount(fs *flag.FlagSet, cfg *config.Config) {
 	fs.StringVar(&cfg.HostMount, "host-mount", defaultValue, "a mount point from Docker host and within the Docker. The format is {host-dir}:{local-dir}.")
 }
 
+func addFlagPushConfig(fs *flag.FlagSet, cfg *config.Config) {
+	defaultValue := "noreply-cloudsdk@google.com,Google Cloud SDK"
```
zhumin8, line 38: FYI - This default value was added as a temporarily see [comment](https://docs.google.com/document/d/1bQ5aXIfQEauI6cHlZKh8R6PBAG_sfHRaDxHcxRCQMYI/edit?resourcekey=0-TyGOVF1t5hGtF0Pydz572w&disco=AAABk4ECl78) in go/librarian:cli-reimagined, #724

internal/librarian/generate.go
```diff
@@ -85,6 +85,7 @@ func init() {
 	addFlagAPI(fs, cfg)
 	addFlagBuild(fs, cfg)
 	addFlagHostMount(fs, cfg)
+	addFlagPushConfig(fs, cfg)
```
zhumin8, line 88: iiuc, #805 should include adding functionality of this flag to generate and related testing?

zhumin8 (changes requested):

internal/librarian/flags.go
zhumin8: Sure, it'be helpful if you could add a TODO linking to the above issue.

catchyzheng:
> Can you please fix the godoc issues reported via ci?

Done

internal/librarian/generate.go
```diff
@@ -209,6 +211,17 @@ func (r *generateRunner) runGenerateCommand(ctx context.Context, outputDir strin
 		return libraryID, r.containerClient.Generate(ctx, generateRequest)
 	}
 	slog.Info("No matching library found (or no repo specified)", "path", r.cfg.API)
+	// Handle push config flag
+	if r.cfg.PushConfig != "" {
```
zhumin8, line 215: iiuc, this logic needs to insert before return statement in L211?

internal/librarian/generate.go
```diff
@@ -209,6 +211,17 @@ func (r *generateRunner) runGenerateCommand(ctx context.Context, outputDir strin
...
+		if r.cfg.GitHubToken == "" {
+			return "", fmt.Errorf("GitHub token is required to push config")
```
zhumin8, line 217: I think it's better to do this check earlier in the flow and fail early. 
Maybe in newGenerateRunner()?

internal/librarian/generate.go
```diff
@@ -206,9 +208,24 @@ func (r *generateRunner) runGenerateCommand(ctx context.Context, outputDir strin
 			RepoDir:   r.repo.Dir,
 		}
 		slog.Info("Performing refined generation for library", "id", libraryID)
-		return libraryID, r.containerClient.Generate(ctx, generateRequest)
+		generateResponse := r.containerClient.Generate(ctx, generateRequest)
```
zhumin8, line 211: I believe `generateResponse` is just the error? we would want to early return if so.

internal/librarian/generate.go
```diff
@@ -206,9 +208,27 @@ func (r *generateRunner) runGenerateCommand(ctx context.Context, outputDir strin
...
+		err = r.containerClient.Generate(ctx, generateRequest)
+		if err != nil {
+			return libraryID, err
+		}
+
+		// If the push-config flag is set and the GitHub token is provided,
+		// we push the config changes to GitHub.
+		if r.cfg.PushConfig != "" {
+			if r.cfg.GitHubToken == "" {
+				return libraryID, fmt.Errorf("GitHub token is required to push config")
```
ldetmer, line 220: lets move this check to part of the validate flags command to first thing in generate runner, so we don't bother calling the language container generate if they are missing the githubtoken and this flag is setup

internal/librarian/generate.go
```diff
@@ -204,9 +208,17 @@ func (r *generateRunner) runGenerateCommand(ctx context.Context, outputDir strin
 			RepoDir:   r.repo.Dir,
 		}
 		slog.Info("Performing refined generation for library", "id", libraryID)
-		return libraryID, r.containerClient.Generate(ctx, generateRequest)
+		err = r.containerClient.Generate(ctx, generateRequest)
+		if err != nil {
+			return libraryID, err
+		}
+
```
zhumin8, line 215: This Go convention helps restrict scope for the err var (I just learned). Also return default values for other types if an error is returned.
```suggestion
if err := r.containerClient.Generate(ctx, generateRequest); err != nil {
			return "", err
		}

```

fyi - we are going to run into conflict with #871 depending which one merges in  first.

internal/librarian/flags.go
```diff
@@ -34,6 +34,12 @@ func addFlagHostMount(fs *flag.FlagSet, cfg *config.Config) {
 	fs.StringVar(&cfg.HostMount, "host-mount", defaultValue, "a mount point from Docker host and within the Docker. The format is {host-dir}:{local-dir}.")
 }
 
+func addFlagPushConfig(fs *flag.FlagSet, cfg *config.Config) {
+	// TODO(https://github.com/googleapis/librarian/issues/724):remove the default for push-config
+	defaultValue := "noreply-cloudsdk@google.com,Google Cloud SDK"
+	fs.StringVar(&cfg.PushConfig, "push-config", defaultValue, "The user and email for Git commits, in the format \"user:email\"")
```
zhumin8, line 40: This "user:email" format is not consistent with the parsing logic. Perhaps use language from http://go/librarian:cli-reimagined

If specified, will try to create a commit and pull request for the generated changes. The format should be `”{git-email-address},{author-name}”`. Also, when this field is specified it is expected a Github token will be provided with push access via the environment variable `LIBRARIAN_GITHUB_TOKEN`.

internal/librarian/generate.go
```diff
@@ -204,9 +208,17 @@ func (r *generateRunner) runGenerateCommand(ctx context.Context, outputDir strin
...
+		// Push the config changes to GitHub.
+		err := commitAndPush(ctx, r, r.cfg.PushConfig)
```
zhumin8, line 217: This error is not handled?

internal/github/github.go
```diff
@@ -108,3 +109,57 @@ func (c *Client) GetRawContent(ctx context.Context, path, ref string) ([]byte, e
...
+		Base:                github.Ptr("main"),
+		Body:                github.Ptr(body),
+		MaintainerCanModify: github.Ptr(true),
+	}
+	pr, _, err := c.PullRequests.Create(ctx, repo.Owner, repo.Name, newPR)
+	if err != nil {
+		return nil, err
+	}
+
+	fmt.Printf("PR created: %s\n", pr.GetHTMLURL())
```
jba, line 132: This seems out of place in a package function. Maybe log?

internal/github/github.go
```diff
@@ -108,3 +109,57 @@ func (c *Client) GetRawContent(ctx context.Context, path, ref string) ([]byte, e
...
+// GetGitHubRepoFromRemote parses the GitHub repo name from the remote for this repository.
+// There must only be a single remote with a GitHub URL (as the first URL), in order to provide an
+// unambiguous result.
+// Remotes without any URLs, or where the first URL does not start with https://github.com/ are ignored.
+func GetGitHubRepoFromRemote(repo *gitrepo.Repository) (*Repository, error) {
+	remotes, err := repo.Remotes()
+	if err != nil {
+		return nil, err
+	}
+	gitHubRemoteNames := []string{}
```
jba, line 146: more idiomatically:
`var gitHubRemoteNames []string`

internal/github/github.go
```diff
@@ -108,3 +109,57 @@ func (c *Client) GetRawContent(ctx context.Context, path, ref string) ([]byte, e
...
+	gitHubUrl := ""
```
jba, line 147: `gitHubURL`

internal/github/github.go
```diff
@@ -108,3 +109,57 @@ func (c *Client) GetRawContent(ctx context.Context, path, ref string) ([]byte, e
...
+		}
+	}
+
+	if len(gitHubRemoteNames) == 0 {
+		return nil, fmt.Errorf("no GitHub remotes found")
+	}
+
+	if len(gitHubRemoteNames) > 1 {
+		joinedRemoteNames := strings.Join(gitHubRemoteNames, ", ")
+		return nil, fmt.Errorf("can only determine the GitHub repo with a single matching remote; GitHub remotes in repo: %s", joinedRemoteNames)
```
jba, line 162: Inline the strings.Join, or at least give the variable a much shorter name (a single letter is fine).

internal/librarian/command.go
```diff
@@ -132,3 +134,48 @@ func createWorkRoot(t time.Time, workRootOverride string) (string, error) {
 	slog.Info("Temporary working directory", "dir", path)
 	return path, nil
 }
+
+// commitAndPush creates a commit and push request to Github for the generated changes.
+// It uses the GitHub client to create a PR with the specified branch, title, and description to the repository.
+func commitAndPush(ctx context.Context, r *generateRunner, pushConfig string) error {
+	// Ensure we have a GitHub repository
+	gitHubRepo, err := github.GetGitHubRepoFromRemote(r.repo)
```
jba, line 142: I think you can remove `Get` from the function name. It doesn't do anything. (Unless it's doing an HTTP GET and that's important.)

internal/librarian/command.go
```diff
@@ -132,3 +134,48 @@ func createWorkRoot(t time.Time, workRootOverride string) (string, error) {
...
+	if err != nil {
+		return err
+	}
+
+	userEmail, userName, err := parsePushConfig(pushConfig)
+	if err != nil {
+		return err
+	}
+	_, err = r.repo.AddAll()
```
jba, line 151: if _, err := ...; err != nil {

internal/librarian/command.go
```diff
@@ -132,3 +134,48 @@ func createWorkRoot(t time.Time, workRootOverride string) (string, error) {
...
+	if err != nil {
+		return err
+	}
+
+	// TODO: get correct language for description (https://github.com/googleapis/librarian/issues/885)
+	description := "Changes in this PR"
+	r.repo.Commit(description, userName, userEmail)
+
+	// Create a new branch, set title and description for the PR.
+	datetime_now := formatTimestamp(time.Now())
```
jba, line 161: `datetimeNow`, or just `now`

internal/librarian/command.go
```diff
@@ -132,3 +134,48 @@ func createWorkRoot(t time.Time, workRootOverride string) (string, error) {
...
+	titlePrefix := "Librarian pull request"
+	branch := fmt.Sprintf("librarian-%s", datetime_now)
+	title := fmt.Sprintf("%s: %s", titlePrefix, datetime_now)
+
+	_, err = r.ghClient.CreatePullRequest(ctx, gitHubRepo, branch, title, description)
```
jba, line 166: ditto

internal/librarian/command.go
```diff
@@ -132,3 +134,48 @@ func createWorkRoot(t time.Time, workRootOverride string) (string, error) {
...
+	if err != nil {
+		return fmt.Errorf("failed to create pull request: %w", err)
+	}
+	return nil
+}
+
+func parsePushConfig(pushConfig string) (string, string, error) {
```
jba, line 173: s/pushConfig/s/ 

or something else short.
Clearly it's a push config or you wouldn't be parsing it.

jba (commented):
This is a Go readability review.

jba (dismissed):

jba (approved):

ldetmer (approved):

internal/librarian/generate.go
```diff
@@ -204,9 +208,18 @@ func (r *generateRunner) runGenerateCommand(ctx context.Context, outputDir strin
...
 		slog.Info("Performing refined generation for library", "id", libraryID)
-		return libraryID, r.containerClient.Generate(ctx, generateRequest)
+		if err := r.containerClient.Generate(ctx, generateRequest); err != nil {
+			return "", err
+		}
+
+		// Push the config changes to GitHub.
+		if err := commitAndPush(ctx, r.repo, r.ghClient, r.cfg.PushConfig); err != nil {
+			return "", err
+		}
```
zhumin8, line 218: iiuc, we need to check on condition that PushConfig is provided to commit and push.

internal/librarian/generate.go
```diff
@@ -216,10 +220,18 @@ func (r *generateRunner) runGenerateCommand(ctx context.Context, outputDir strin
 		if err := r.cleanAndCopyLibrary(libraryID, outputDir); err != nil {
 			return "", err
 		}
+
+		// Push the config changes to GitHub.
+		if r.cfg.PushConfig != "" {
+			if err := commitAndPush(ctx, r.repo, r.ghClient, r.cfg.PushConfig); err != nil {
+				return "", err
+			}
+		}
```
zhumin8, line 229: Sorry I missed this in earlier reviews. But I believe this "commit and push" logic should be moved outside of `runGenerateCommand` and after build is done (whenever build flag specified).
https://github.com/googleapis/librarian/blob/3ee39a62fd90f2369301c56f350d1906fc23f879/internal/librarian/generate.go#L178-L181

Basically, Librarian should not create a PR if build is failing. 
I don't recall this explicitly called out in [go/librarian:cli-reimagined](http://goto.google.com/librarian:cli-reimagined), but this is no change from earlier implementations (see [doc](https://github.com/googleapis/librarian/blob/dcf81d04e47df28c9ccfbba885b923946713a834/internal/librarian/updateapis.go#L63-L65)), and you can also refer to decision diagram [here](https://docs.google.com/document/d/1hS1YHaYd_OFqoKGV8xIhub28sBN47tvMmdFzc-SyHcg/edit?resourcekey=0-oq3vLClMKxKxeiClt5W4DQ&tab=t.0#heading=h.kb6ic4eawdf)

cc. @codyoss

internal/librarian/generate.go
```diff
@@ -178,6 +182,13 @@ func (r *generateRunner) run(ctx context.Context) error {
 	if err := r.runBuildCommand(ctx, libraryID); err != nil {
 		return err
 	}
+
+	// Push the config changes to GitHub.
+	if r.cfg.PushConfig != "" {
+		if err := commitAndPush(ctx, r.repo, r.ghClient, r.cfg.PushConfig); err != nil {
+			return err
+		}
+	}
```
zhumin8, line 191: nit: Can you refactor this check inside of `commitAndPush` for consistency with r.runBuildCommand above? The two snippets handle conditional logic differently, makes it harder to read. Also please update the comment accordingly.

internal/librarian/command.go
```diff
@@ -138,6 +138,9 @@ func createWorkRoot(t time.Time, workRootOverride string) (string, error) {
 // commitAndPush creates a commit and push request to Github for the generated changes.
 // It uses the GitHub client to create a PR with the specified branch, title, and description to the repository.
 func commitAndPush(ctx context.Context, repo *gitrepo.Repository, ghClient GitHubClient, pushConfig string) error {
+	if pushConfig == "" {
+		return fmt.Errorf("push config flag is not set")
+	}
```
zhumin8, line 143: Skip if pushConfig not specified, use logic similar to 
https://github.com/googleapis/librarian/blob/30d0042f6a1bf2ffd9adc5fb713535d2ad071357/internal/librarian/generate.go#L260-L263

ldetmer (approved):

Issue #869, opened 2025-07-14
Clean up generate/build request files


JoeWang1127:
In #770 and #771 , the container contracts create `generate-request.json` and `build-request.json`. 

These two files should be removed after executing the docker command as they are ephemeral.

PR #870, opened 2025-07-15
chore(internal/docker): remove request files

JoeWang1127:
Remove `generate-request.json` and `build-request.json` after running docker container.

Fixes: #869

### DIFF
--- a/internal/docker/docker.go
+++ b/internal/docker/docker.go
@@ -126,7 +126,7 @@ func (c *Docker) Generate(ctx context.Context, request *GenerateRequest) error {
 	defer func(name string) {
 		err := os.Remove(name)
 		if err != nil {
-			slog.Warn("fail to remove", name, "err", err)
+			slog.Warn("fail to remove file", slog.String("name", name), slog.Any("err", err))
 		}
 	}(jsonFilePath)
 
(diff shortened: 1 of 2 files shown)


internal/docker/docker.go
```diff
@@ -159,7 +166,14 @@ func (c *Docker) Build(ctx context.Context, request *BuildRequest) error {
 		fmt.Sprintf("--library-id=%s", request.LibraryID),
 	}
 
-	return c.runDocker(ctx, request.Cfg, CommandBuild, mounts, commandArgs)
+	if err := c.runDocker(ctx, request.Cfg, CommandBuild, mounts, commandArgs); err != nil {
+		return err
+	}
+
+	// It's OK to not handle this error since this file is ephemeral.
+	os.Remove(jsonFilePath)
```
zhumin8, line 174: Maybe use Go's [`defer` statement](https://go.dev/tour/flowcontrol/12) to schedule the cleanup? It is idiomatic in Go and has the benefit of guaranteed clean up on all error path, including `c.runDocker()`  in L170.
Sample usage in this repo [here](https://github.com/googleapis/librarian/blob/ade9ce0bb270281213252159fb81576bd29ca96f/internal/github/github.go#L108)

zhumin8 (changes requested):

internal/docker/docker.go

Issue #777, opened 2025-07-07
librarian: detect the name of the service yaml


codyoss:
The state file now explicitly lists the service-yaml file for each API. Librarian needs to detect the correct file to put in the state file using a heuristic based off of the source path of the API in googleapis. This should happen before any container commands are invoked so this filename can be shared to the containers. 

Blocked by: #772

codyoss:
I did a quick check last week and this should find all the service yamls for libraries onboard in Go, other than grafeas which is a unique case.

jskeet:
> I did a quick check last week and this should find all the service yamls for libraries onboard in Go, other than grafeas which is a unique case.

I'm intrigued to know why it fails for Grafeas - doesn't https://github.com/googleapis/googleapis/blob/master/grafeas/v1/grafeas_v1.yaml uniquely fulfil those requirements?

Happy to chat offline.

PR #873, opened 2025-07-15
feat: parse service config from api source

JoeWang1127:
Fixes: #777

### DIFF
--- a/internal/librarian/stateandconfig.go
+++ b/internal/librarian/stateandconfig.go
@@ -17,8 +17,10 @@ package librarian
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
+	"strings"
 
 	"path/filepath"
 
@@ -33,11 +35,11 @@ const pipelineConfigFile = "pipeline-config.json"
 
 // Utility functions for saving and loading pipeline state and config from various places.
 
-func loadRepoStateAndConfig(languageRepo *gitrepo.Repository) (*config.LibrarianState, *config.PipelineConfig, error) {
+func loadRepoStateAndConfig(languageRepo *gitrepo.Repository, source string) (*config.LibrarianState, *config.PipelineConfig, error) {
 	if languageRepo == nil {
 		return nil, nil, nil
 	}
-	state, err := loadLibrarianState(languageRepo)
+	state, err := loadLibrarianState(languageRepo, source)
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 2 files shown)


internal/librarian/stateandconfig.go
```diff
@@ -87,17 +91,74 @@ func parsePipelineConfig(contentLoader func() ([]byte, error)) (*config.Pipeline
...
+			})
+		}
+		modifiedLibraries[len(modifiedLibraries)-1].APIs = modifiedAPIs
+	}
+	state.Libraries = modifiedLibraries
+
+	return nil
+}
+
+func findServiceConfigIn(contentLoader func(file string) ([]byte, error), path string) (string, error) {
```
zhumin8, line 140: Can you add comment to this method stating the heuristic it looks for? https://github.com/googleapis/librarian/issues/777#issuecomment-3044211865

internal/librarian/stateandconfig.go
```diff
@@ -87,17 +91,74 @@ func parsePipelineConfig(contentLoader func() ([]byte, error)) (*config.Pipeline
...
+func parseLibrarianState(contentLoader func(file string) ([]byte, error), path, source string) (*config.LibrarianState, error) {
+	bytes, err := contentLoader(path)
 	if err != nil {
 		return nil, err
 	}
 	var s config.LibrarianState
 	if err := yaml.Unmarshal(bytes, &s); err != nil {
 		return nil, fmt.Errorf("unmarshaling librarian state: %w", err)
 	}
+	if err := populateServiceConfig(&s, contentLoader, source); err != nil {
```
zhumin8, line 103: nit: can you rename to indicate it only populates when not already present in state.yaml?

internal/librarian/stateandconfig.go
```diff
@@ -87,17 +91,74 @@ func parsePipelineConfig(contentLoader func() ([]byte, error)) (*config.Pipeline
...
+		return nil, fmt.Errorf("populating service config: %w", err)
+	}
 	if err := s.Validate(); err != nil {
 		return nil, fmt.Errorf("validating librarian state: %w", err)
 	}
 	return &s, nil
 }
+
+func populateServiceConfig(state *config.LibrarianState, contentLoader func(file string) ([]byte, error), source string) error {
+	modifiedLibraries := []*config.LibraryState{}
```
zhumin8, line 113: The `modifiedLibraries` slice seems redundant. The code could directly modify the `state.Libraries` slice since  `library` is a pointer to a LibraryState struct. I think this should help simplify the logic here.

zhumin8 (changes requested):

internal/librarian/stateandconfig.go
JoeWang1127: Good suggestion, made change accordingly.

internal/librarian/stateandconfig.go
JoeWang1127: Done.

internal/librarian/stateandconfig.go
```diff
@@ -87,17 +91,73 @@ func parsePipelineConfig(contentLoader func() ([]byte, error)) (*config.Pipeline
...
+func parseLibrarianState(contentLoader func(file string) ([]byte, error), path, source string) (*config.LibrarianState, error) {
+	bytes, err := contentLoader(path)
 	if err != nil {
 		return nil, err
 	}
 	var s config.LibrarianState
 	if err := yaml.Unmarshal(bytes, &s); err != nil {
 		return nil, fmt.Errorf("unmarshaling librarian state: %w", err)
 	}
+	if err := populateServiceConfigIfEmpty(&s, contentLoader, source); err != nil {
```
zhumin8, line 103: nit: what do you think of adding a comment so this can be re-examined when implementing configure? For it is really only needed in configure? 
But I guess it's not much harm to keep it for all. Leaving it to you.

JoeWang1127: The function will be called when setting up a generate runner [(code),](https://github.com/googleapis/librarian/blob/7f8bf74c45e39596c525e2aa3f08e4e765c383f0/internal/librarian/generate.go#L106) so there's no difference if it's configure or not.

PR #915, opened 2025-07-18
fix(internal/librarian): load state.yaml from librarian dir

### LINKED ISSUE #916: librarian: state.yaml should be loaded from librarian dir
[go/librarian:cli-reimagined](http://goto.google.com/librarian:cli-reimagined)
pipeline state file is expected inside `.librarian` instead of `generate-input`

This was missed between #807 and #803

zhumin8:
State file is now expected in `.librarian/state.yaml`

Fixes #916


JoeWang1127 (approved):
Could you create an issue for this change?

Issue #861, opened 2025-07-14
Implement end-to-end test for `generate` command


JoeWang1127:
Tasks need to completed in this issue:

- [ ] Create a `./internal/librarian/generate_e2e_test.go`
- [ ] Create test state/config, golden files in ./testdata directory
- [ ] Write tests to execute `runner.run()`
- [ ] Modify the test container to accept flags and generate files based on the flag
- [ ] Mark the test with `e2e` tag
- [ ] Modify the workflow to execute `e2e` test suite

PR #921, opened 2025-07-21
chore(internal/librarian): create a e2e test suite for `generate` command

JoeWang1127:
Fixes: #861

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -27,8 +27,7 @@ const (
 	// BuildRequest is a JSON file that describes which library to build/test.
 	BuildRequest string = "build-request.json"
 	// ConfigureRequest is a JSON file that describes which library to configure.
-	ConfigureRequest  string = "configure-request.json"
-	DefaultPushConfig string = "noreply-cloudsdk@google.com,Google Cloud SDK"
+	ConfigureRequest string = "configure-request.json"
 	// GeneratorInputDir is the default directory to store files that generator
 	// needs to regenerate libraries from an empty directory.
 	GeneratorInputDir string = ".librarian/generator-input"
@@ -199,7 +198,7 @@ type Config struct {
 func New() *Config {
 	return &Config{
 		GitHubToken: os.Getenv("LIBRARIAN_GITHUB_TOKEN"),
-		PushConfig:  DefaultPushConfig,
+		PushConfig:  "",
 	}
 }
 
@@ -225,7 +224,7 @@ func (c *Config) IsValid() (bool, error) {
 		return false, errors.New("no GitHub token supplied for push")
 	}
 
-	if _, err := validatePushConfig(c.PushConfig, DefaultPushConfig); err != nil {
+	if _, err := validatePushConfig(c.PushConfig, ""); err != nil {
 		return false, err
 	}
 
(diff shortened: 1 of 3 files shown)


internal/librarian/generate_e2e_test.go
```diff
@@ -0,0 +1,239 @@
...
+	for _, test := range []struct {
+		name    string
+		cfg     *config.Config
+		wantErr bool
+	}{
+		{
+			name: "testRunSuccess",
+			cfg: &config.Config{
+				API:      "google/cloud/pubsub/v1",
+				Repo:     fmt.Sprintf("%s-%d", localRepoDir, rand.Intn(10000)),
```
chingor13, line 56: `Repo` and `WorkRoot` seem unrelated to the test case and can be calculated in the test setup code. Declaring them in the test input struct distracts from the contents of the test.

internal/librarian/generate_e2e_test.go
```diff
@@ -0,0 +1,239 @@
...
+		},
+		{
+			name: "testRunFailed",
+			cfg: &config.Config{
+				API:      "google/invalid/path",
+				Repo:     fmt.Sprintf("%s-%d", localRepoDir, rand.Intn(10000)),
+				Source:   "../../testdata/e2e/generate/api_root",
+				WorkRoot: filepath.Join(os.TempDir(), fmt.Sprintf("librarian-%d", rand.Intn(10000))),
+			},
+			wantErr: true,
```
chingor13, line 69: Can/should we have a more specific error check? (e.g. error message/type). If we just check to see if there is an error or not, we may not be confident in what is failing.

internal/librarian/generate_e2e_test.go
```diff
@@ -0,0 +1,239 @@
...
+					t.Fatalf("Unmarshal() error = %v", err)
+				}
+				if content.ErrorMessage == "" {
+					t.Fatalf("can not find error message in generate response")
+				}
+			}
+
+			// test cleanup
+			os.RemoveAll(test.cfg.Repo)
+			os.RemoveAll(test.cfg.WorkRoot)
```
chingor13, line 113: Not a golang expert, but is there a try/finally (defer?) way to ensure these are cleaned up?

testdata/e2e_func.go
chingor13, line 1: I don't understand how this file is used

chingor13 (changes requested):
Conceptually, I feel like an end-to-end test should invoke the librarian CLI as the entrypoint rather than invoking internal methods. Is the intent to be more of an integration/system test (rather than e2e)?

testdata/e2e_func.go
JoeWang1127: This file is executed within the docker container. 

It received the flags from librarian and write response, similarly to a language container.

internal/librarian/generate_e2e_test.go
JoeWang1127: Done.

JoeWang1127:
> Conceptually, I feel like an end-to-end test should invoke the librarian CLI as the entrypoint rather than invoking internal methods. Is the intent to be more of an integration/system test (rather than e2e)?

I'll try to refactor the test to do so.

internal/librarian/generate_e2e_test.go
JoeWang1127: In go/librarian:generate-command, there's no error types/return code defined for generate command.

If the generation failed, we can expect the `generate-response.json` contains a `error` field.

JoeWang1127:
The e2e test should pass once #724 is resolved

PR #922, opened 2025-07-21
chore: migrate multi-approvers workflow to Librarian repo

meltsufin:
Migrating from https://github.com/googleapis/infra-github/blob/main/.github/workflows/multi_approvers.yaml so that we can get rid of that repo entirely.


chingor13 (dismissed):

zhumin8:
@meltsufin can you add a header to this file?

Issue #933, opened 2025-07-22
librarian: rename flag source to api-source


zhumin8:
[go/librarian:cli-reimagined](http://goto.google.com/librarian:cli-reimagined)

https://github.com/googleapis/librarian/blob/10e32185114ee020512b1e9fb1448486b0e45766/internal/librarian/flags.go#L49-L51

PR #987, opened 2025-07-24
fix: rename flag source to api-source

catchyzheng:
rename flag source to api-source

fix: #933

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -53,6 +53,16 @@ type Config struct {
 	// API Path is specified with the -api flag.
 	API string
 
+	// APISource is the path to the root of the googleapis repository.
+	// When this is not specified, the googleapis repository is cloned
+	// automatically.
+	//
+	// APISource is used by generate, update-apis and configure
+	// commands.
+	//
+	// APISource is specified with the -api-source flag.
+	APISource string
+
 	// Build determines whether to build the generated library, and is only
 	// used in the generate command.
 	//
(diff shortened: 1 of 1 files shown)


internal/config/config.go
```diff
@@ -159,15 +159,15 @@ type Config struct {
...
 	// automatically.
 	//
-	// Source is used by generate, update-apis and configure
+	// APISource is used by generate, update-apis and configure
 	// commands.
 	//
-	// Source is specified with the -source flag.
-	Source string
+	// APISource is specified with the -api-source flag.
+	APISource string
```
zhumin8, line 170: nit: can you keep the alphabetical order of these configs?

internal/librarian/flags.go
```diff
@@ -54,8 +54,8 @@ func addFlagRepo(fs *flag.FlagSet, cfg *config.Config) {
 			"/path/to/repo`. Both absolute and relative paths are supported.")
 }
 
-func addFlagSource(fs *flag.FlagSet, cfg *config.Config) {
-	fs.StringVar(&cfg.Source, "source", "", "location of googleapis repository. If undefined, googleapis will be cloned to the output")
+func addFlagAPISource(fs *flag.FlagSet, cfg *config.Config) {
```
zhumin8, line 57: also move this up per alphabetic order

zhumin8 (commented):
LGTM. 
just a nit pick

internal/config/config.go
catchyzheng: Done

Issue #910, opened 2025-07-18
librarian: Support global file edits for generation


bisun-google:
See https://docs.google.com/document/d/1hS1YHaYd_OFqoKGV8xIhub28sBN47tvMmdFzc-SyHcg/edit?resourcekey=0-oq3vLClMKxKxeiClt5W4DQ&tab=t.0#heading=h.hq6c1oj48rte

Specifically: Some languages have global files like READMEs, dependency files, etc that may need to be updated as a result of generation. The current design never gives librarian access to the whole code repository to make such edits. This means we either need to draw a hard line that librarian will not make such edits or we need another lifecycle hook to make edits like this.

meltsufin:
Design ready for review: [go/librarian:global-file-edits](http://goto.google.com/librarian:global-file-edits).

PR #1002, opened 2025-07-28
docs: add an Excalidraw diagram for global file edits

meltsufin:
Diagram used in go/librarian:global-file-edits.

To view it in VS Code, install Excalidraw extension.

For #910


codyoss:
Need to update our header test to ignore this file, or maybe even the doc directory

meltsufin:
This is what it looks like, since GitHub doesn't have a preview.

image

Issue #938, opened 2025-07-22
librarian: generate should update LastGeneratedCommit in state.yaml


zhumin8:
After successfully generate, Librarian should update state.yaml record of LastGeneratedCommit per library.

Similar logic is present and can be borrowed from 0.1.0
https://github.com/googleapis/librarian/blob/dcf81d04e47df28c9ccfbba885b923946713a834/internal/librarian/updateapis.go#L240

PR #989, opened 2025-07-24
fix(internal/librarian): read configure response

JoeWang1127:
Fixes #985

PR #1017, opened 2025-07-29
fix(internal/librarian): update state file for each success 

codyoss:
Fixes: https://github.com/googleapis/librarian/issues/938

### DIFF
--- a/generate_e2e_test.go
+++ b/generate_e2e_test.go
@@ -28,6 +28,9 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/googleapis/librarian/internal/config"
+	"gopkg.in/yaml.v3"
 )
 
 func TestRunGenerate(t *testing.T) {
@@ -57,10 +60,10 @@ func TestRunGenerate(t *testing.T) {
 			workRoot := filepath.Join(t.TempDir())
 			repo := filepath.Join(workRoot, repo)
 			APISourceRepo := filepath.Join(workRoot, APISourceRepo)
-			if err := prepareTest(t, repo, workRoot, initialRepoStateDir); err != nil {
+			if err := initRepo(t, repo, initialRepoStateDir); err != nil {
 				t.Fatalf("languageRepo prepare test error = %v", err)
 			}
-			if err := prepareTest(t, APISourceRepo, workRoot, localAPISource); err != nil {
+			if err := initRepo(t, APISourceRepo, localAPISource); err != nil {
 				t.Fatalf("APISouceRepo prepare test error = %v", err)
 			}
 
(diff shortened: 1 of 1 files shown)


internal/librarian/generate.go
```diff
@@ -177,6 +183,10 @@ func (r *generateRunner) run(ctx context.Context) error {
 		}
 	}
 
+	if err := saveLibrarianState(r.repo.Dir, r.state); err != nil {
```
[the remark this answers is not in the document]
codyoss: yes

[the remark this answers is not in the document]
JoeWang1127: Ah, I saw the commit sha in the library is updated.

In [here](https://github.com/googleapis/librarian/pull/989/files#diff-77bcd9dfc5e4c1244e743a1d11816d30b2e4459536a445f41c76af1d32dfb898R244), I wrote a function to write the updated librarian state to `state.yaml` after reading `configure-response.json`.

I can create a follow up PR to remove that function after this PR is merged in.

internal/librarian/state.go
```diff
@@ -131,3 +131,12 @@ func findServiceConfigIn(contentLoader func(file string) ([]byte, error), path s
 
 	return "", errors.New("could not find service config in " + path)
 }
+
+func saveLibrarianState(repoDir string, state *config.LibrarianState) error {
```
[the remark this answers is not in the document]
JoeWang1127: Makes sense, thanks.

JoeWang1127:
I'm good with this PR but could you add more tests?

internal/librarian/generate.go
```diff
@@ -113,6 +114,10 @@ func newGenerateRunner(cfg *config.Config) (*generateRunner, error) {
 	if err != nil {
 		return nil, err
 	}
+	sourceRepo, err := gitrepo.NewRepository(&gitrepo.RepositoryOptions{Dir: cfg.APISource})
```
zhumin8, line 117: Looks like the default behavior "If undefined, googleapis is the default and will be cloned to the output" is never implemented? Do you mind add check for empty cfg.APISource and implement default behavior here?

Or, I can also create a separate issue to track the fix.

codyoss: I actually filed that exact bug this morning: #1512

internal/librarian/state.go
```diff
@@ -124,3 +124,12 @@ func findServiceConfigIn(contentLoader func(file string) ([]byte, error), path s
...
```
zhumin8, line 128: fyi - this seems duplicated/recreated in #989 writeLibrarianState

JoeWang1127: I didn't realize the commit sha needs to change so we should write librarian state here.

Could you remove the similar function in https://github.com/googleapis/librarian/blob/main/internal/librarian/generate.go#L522

JoeWang1127 (approved):

Issue #1044, opened 2025-07-30
`generate-request.json` should use snake case for keys to match `state.yaml` for consistency


parthea:
I expected the `generate-request.json` to have a similar schema as `state.yaml`. For example, if we use `id` for the library in `state.yaml`, we would also use `id` for the library in `generate-request`. It looks like `state.yaml` uses snake case but `generate-request.json` uses pascal case.

https://github.com/googleapis/librarian/blob/main/doc/state-schema.md

I expected a `generate-request.json` file to look like this

```
{
  "id": "google-cloud-storage-v1",
  "version": "1.2.3",
  "last_generated_commit": "123",
  "apis": [
    {
      "path": "google/storage/v1",
      "service_config": "storage.yaml"
    }
  ],
  "source_paths": [
    "src/google/cloud/storage",
    "test/google/cloud/storage"
  ],
  "preserve_regex": [
    "src/google/cloud/storage/generated-dir/HandWrittenFile.java"
  ],
  "remove_regex": [
    "src/google/cloud/storage/generated-dir"
  ]
}
```

but instead it looks like this

```
{
  "ID": "google-cloud-storage-v1",
  "Version": "1.2.3",
  "LastGeneratedCommit": "123",
  "APIs": [
    {
      "Path": "google/storage/v1",
      "ServiceConfig": "storage.yaml"
    }
  ],
  "SourcePaths": [
    "src/google/cloud/storage",
    "test/google/cloud/storage"
  ],
  "PreserveRegex": [
    "src/google/cloud/storage/generated-dir/HandWrittenFile.java"
  ],
  "RemoveRegex": [
    "src/google/cloud/storage/generated-dir"
  ]
}

```

PR #1045, opened 2025-07-30
fix: use snake_case in generate-request.json

parthea:
Fixes https://github.com/googleapis/librarian/issues/1044

### DIFF
--- a/doc/container-contract.md
+++ b/doc/container-contract.md
@@ -115,7 +115,7 @@ In order for the the container to have enough context on how and what to generat
 
 ### generate-request.json Schema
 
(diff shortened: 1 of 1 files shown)


doc/generate-request-schema.md
```diff
@@ -0,0 +1,51 @@
+# generate-request.json Schema
```
codyoss, line 1: Thank you for writing this! Would you mind moving it into the existing container-contract.md doc. I think it would be useful to see all that information together in one place.

codyoss (commented):
LGTM, just one suggestion.

doc/generate-request-schema.md
parthea: Done

PR #1505, opened 2025-07-30
feat: Add a yaml format check in librarian.yaml test step

catchyzheng:
Add a yaml format check in librarian.yaml test step.

If any yaml file is not in good format, the `Detect formatting changes` will failed, indicating which yaml file needs change. 

To fix, run `yamlfmt .` in the librarian root dir. 

fix: https://github.com/googleapis/librarian/issues/1502


.github/workflows/librarian.yaml
```diff
@@ -27,8 +27,13 @@ jobs:
         run: go version
       - name: Run tests
         run: go test -race -coverprofile=coverage.out -covermode=atomic ./...
-      - run: go fmt ./...
-      - name: Detect Changes
+      - name: Check Go formatting
+        run: go fmt ./...
+      - name: Check YAML formatting
+        run: |
+          go install github.com/google/yamlfmt/cmd/yamlfmt@latest
```
parthea, line 34: Is there a benefit to pin to a specific version of yamlfmt?

Issue #1512, opened 2025-07-31
librarian: api-source not cloning googleapis if flag not provided


codyoss:
From the flag docs: "location of googleapis repository. If undefined, googleapis will be cloned to the output"

catchyzheng:
@codyoss @parthea 

I think The [cloneOrOpenLanguageRepo](https://github.com/googleapis/librarian/blob/main/internal/librarian/command.go#L33) function can be used to clone googleapis repo. The function may need to be renamed.  

```
repo, err := cloneOrOpenLanguageRepo(workRoot, cfg.APISource, cfg.CI)
	if err != nil {
		return nil, err
	}
```
The default value of api-source will also be updated to `https://github.com/googleapis/googleapis`:
```
func addFlagAPISource(fs *flag.FlagSet, cfg *config.Config) {
	fs.StringVar(&cfg.APISource, "api-source", "https://github.com/googleapis/googleapis", "location of googleapis repository. If undefined, googleapis will be cloned to the output")
```

If you have any suggestion, please let me know.

PR #1521, opened 2025-07-31
fix: clone googleapis repo when api-source flag is unspecified

catchyzheng:
Set api-source value to `https://github.com/googleapis/googleapis` when the flag is unspecified.
Also clone googleapis repo.

fix: https://github.com/googleapis/librarian/issues/1512


internal/librarian/flags.go
```diff
@@ -26,7 +26,7 @@ func addFlagAPI(fs *flag.FlagSet, cfg *config.Config) {
 }
 
 func addFlagAPISource(fs *flag.FlagSet, cfg *config.Config) {
-	fs.StringVar(&cfg.APISource, "api-source", "", "location of googleapis repository. If undefined, googleapis will be cloned to the output")
+	fs.StringVar(&cfg.APISource, "api-source", "https://github.com/googleapis/googleapis", "location of googleapis repository. If undefined, googleapis will be cloned to the output")
```
codyoss, line 29: Having a URL default here I think will break other code as it assumes APISource is a path. Example: https://github.com/googleapis/librarian/blob/918507c568622dd4792da9ae175fee6239f33763/internal/librarian/state.go#L82

catchyzheng:
Too many merging conflicts, will create a new pr

PR #1526, opened 2025-08-01
chore(internal/config): remove `PushConfig` flag

### LINKED ISSUE #1517: librarian: remove push-config add push flag
See go/librarian:generate-command

This flag has been removed and replaced by a bool flag. There should be detection logic to pull the git user and email from the environment.

JoeWang1127:
Fixes #1517

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -128,18 +128,6 @@ type Config struct {
 	// Push is specified with the -push flag. No value is required.
 	Push bool
 
-	// PushConfig specifies the email address and display name used in Git commits,
-	// in the format "email,name".
-	//
-	// PushConfig is used in all commands that create commits in a language repository:
-	// create-release-pr, configure and update-apis.
-	//
-	// PushConfig is optional. If unspecified, commits will use a default name of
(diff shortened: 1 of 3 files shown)


internal/gitrepo/gitrepo_test.go
```diff
@@ -367,15 +367,33 @@ func TestAddAll(t *testing.T) {
 
 func TestCommit(t *testing.T) {
 	t.Parallel()
+	name, email := "tester", "tester@example.com"
```
codyoss, line 370: Can we verify in the test that the commit has these set as the author?

codyoss (commented):
Looks good overall, just one suggestion.

internal/gitrepo/gitrepo_test.go
JoeWang1127: It is verified in https://github.com/googleapis/librarian/blob/f0d67b59091f5718cfce5124e96be91346825be8/internal/gitrepo/gitrepo_test.go#L449

codyoss: oops, I did not expand the diff, thanks for the clarification!

codyoss (approved):

PR #1565, opened 2025-08-04
chore(internal/librarian): add error field to docker response

### LINKED ISSUE #1518: librarian: add error field to all docker responses
See go/librarian:generate-command

The response message from all container contracts has been updated to add an error field that can be used to signal back to librarian that an error from the container. This error should be logged and processed appropriately

JoeWang1127:
Fixes #1518

### DIFF
--- a/internal/config/state.go
+++ b/internal/config/state.go
@@ -99,6 +99,9 @@ type LibraryState struct {
 	// If not set, this defaults to the `source_roots`.
 	// A more specific `preserve_regex` takes precedence.
 	RemoveRegex []string `yaml:"remove_regex" json:"remove_regex"`
+	// The error message from the command response.
+	// This field is ignored when writing to state.yaml.
+	ErrorMessage string `yaml:"-" json:"error"`
 }
 
 var (
--- a/internal/config/state.go
+++ b/internal/config/state.go
@@ -101,7 +101,7 @@ type LibraryState struct {
 	RemoveRegex []string `yaml:"remove_regex" json:"remove_regex"`
 	// The error message from the command response.
 	// This field is ignored when writing to state.yaml.
-	ErrorMessage string `yaml:"-" json:"error"`
+	ErrorMessage string `yaml:"-" json:"error,omitempty"`
 }
 
 var (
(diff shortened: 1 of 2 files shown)


internal/librarian/generate_test.go
```diff
@@ -249,7 +304,8 @@ func TestRunConfigureCommand(t *testing.T) {
 				containerClient: test.container,
 			}
 
-			if test.name == "configures library" {
+			if test.name == "configures library successfully" ||
```
codyoss, line 307: nit: This code was already here so don't worry for now but I would suggest not using this pattern in the future where we branch a table test on its name. In such a case it is likely cleaner to just create another whole test.

internal/librarian/generate.go
```diff
@@ -76,6 +76,10 @@ local working tree for inspection.`,
 	},
 }
 
+var libraryStateLoader = func(data []byte, libraryState *config.LibraryState) error {
```
codyoss, line 79: What is the value of having this higher order function? Why not just inline this code directly? I think that would make the code more readable.

internal/librarian/generate_test.go
JoeWang1127: Sure, thanks for the suggestion.

internal/librarian/generate.go
JoeWang1127: Done.

internal/librarian/generate.go
```diff
@@ -250,7 +246,9 @@ func (r *generateRunner) runGenerateCommand(ctx context.Context, libraryID, outp
 
 	// Read the library state from the response.
 	if _, err := readLibraryState(
-		libraryStateLoader,
+		func(data []byte, libraryState *config.LibraryState) error {
```
codyoss, line 249: to clarify further. Why can't the json.Unmarshal just be inlined to readLibraryState instead of passing a function?

JoeWang1127: I use a func so that I can simulate an error in unmarshal to improve test coverage.

codyoss: ack if it is just for coverage of one line of code I am not sure it justifies the complexity in my mind. Don't worry about this for now, but may refactor this later.

codyoss (approved):

PR #1588, opened 2025-08-05
docs: a guide for those new to GitHub

coryan:


dbolduc (approved):

codyoss (commented):
I think this overlaps with some of the content in `CONTRIBUTING.md`. Maybe that doc can link to this one to avoid repetition and having to maintain two places.

coryan:
> I think this overlaps with some of the content in `CONTRIBUTING.md`. Maybe that doc can link to this one to avoid repetition and having to maintain two places.

Good idea, but more work than I am willing to put into this. Closing.

PR #1589, opened 2025-08-05
feat: clone googleapis if apisource is unspecified

catchyzheng:
Set api-source value to https://github.com/googleapis/googleapis when the flag is unspecified.
Also clone googleapis repo. 

fix: https://github.com/googleapis/librarian/issues/1512

### DIFF
--- a/internal/librarian/generate.go
+++ b/internal/librarian/generate.go
@@ -112,14 +112,13 @@ func newGenerateRunner(cfg *config.Config) (*generateRunner, error) {
 	if err != nil {
 		return nil, err
 	}
-	if cfg.APISource == "" {
-		if _, err := cloneOrOpenRepo(workRoot, "https://github.com/googleapis/googleapis", cfg.CI); err != nil {
-			return nil, err
-		}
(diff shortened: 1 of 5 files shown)


internal/librarian/generate.go
```diff
@@ -112,11 +112,21 @@ func newGenerateRunner(cfg *config.Config) (*generateRunner, error) {
 	if err != nil {
 		return nil, err
 	}
-	repo, err := cloneOrOpenLanguageRepo(workRoot, cfg.Repo, cfg.CI)
+	if cfg.APISource == "" {
```
codyoss, line 115: To avoid nested if and the else I would right this code as:

```go
apiSource := cfg.APISource
if apiSource == "" {
apiSource = "https://github.com/googleapis/googleapis"
}
if _, err := cloneOrOpenRepo(workRoot, apiSource, cfg.CI); err != nil {
return nil, err
}
```

codyoss (dismissed):

codyoss (approved):

Issue #1543, opened 2025-08-02
sidekick: increase code coverage to 80%


julieqiu:
At the moment, adding the sidekick codebase to librarian reduced overall code coverage from 82% to 65%.

We should improve code coverage for sidekick to meet baseline, then enable codecov for internal/sidekick. 

Codecov was disabled for sidekick in https://github.com/googleapis/librarian/pull/1533.

PR #1592, opened 2025-08-05
cleanup(sidekick): remove partial go codec

coryan:
We wrote the Go codec to demonstrate we could generate code for more
than one language. The `Dart` codec serves that purpose better, and
brining the Go codec to the required standards of code coverage is too
expensive.

Motivated by #1543


codyoss (approved):
Goodbye old friend

PR #1594, opened 2025-08-05
refactor(sidekick): make `createModel` testable

coryan:
Moving the function to the `parser` package where we can test it and use
it in other tests.

Part of the work for #1543

### DIFF
--- a/internal/sidekick/internal/parser/parser.go
+++ b/internal/sidekick/internal/parser/parser.go
@@ -33,6 +33,8 @@ func CreateModel(config *config.Config) (*api.API, error) {
 		model, err = ParseOpenAPI(config.General.SpecificationSource, config.General.ServiceConfig, config.Source)
 	case "protobuf":
 		model, err = ParseProtobuf(config.General.SpecificationSource, config.General.ServiceConfig, config.Source)
+	case "none":
+		return nil, nil
 	default:
 		return nil, fmt.Errorf("unknown parser %q", config.General.SpecificationFormat)
 	}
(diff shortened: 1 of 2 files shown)


internal/sidekick/refresh.go
```diff
@@ -61,55 +61,13 @@ func loadDir(rootConfig *config.Config, output string) (*api.API, *config.Config
...
-func createModel(config *config.Config) (*api.API, error) {
-	var err error
-	var model *api.API
-	switch config.General.SpecificationFormat {
-	case "openapi":
-		model, err = parser.ParseOpenAPI(config.General.SpecificationSource, config.General.ServiceConfig, config.Source)
-	case "protobuf":
-		model, err = parser.ParseProtobuf(config.General.SpecificationSource, config.General.ServiceConfig, config.Source)
-	case "none":
-		return nil, nil
```
dbolduc, line 80: I use this in storage control

https://github.com/googleapis/google-cloud-rust/blob/c329939e22fcaf342c3e58fc35cf5242feca0f2a/src/storage/src/control/generated/.sidekick.toml

dbolduc (changes requested):

internal/sidekick/refresh.go
coryan: Argh, I searched for any usage but my grep-fu was not strong today. Fixed.

coryan (commented):
PTAL

dbolduc (approved):

PR #1630, opened 2025-08-06
fix(librarian): add release related fields to state

### LINKED ISSUE #1006: librarian: [release] Update `state.yaml` Structure
Update `state.yaml` Structure: Modify the `state.yaml` parsing logic to include the new release-related fields: `version`, `tag_format`, and `release_exclude_paths`. Ensure backward compatibility with existing `state.yaml` files.

zhumin8:
add new fields related to release according to [go/librarian:release-command](http://goto.google.com/librarian:release-command).

Fixes #1006


doc/state-schema.md
```diff
@@ -24,6 +24,8 @@ Each object in the `libraries` list represents a single library and has the foll
 | `source_roots`          | list   | A list of directories in the language repository where Librarian contributes code.                                                                                    | Yes      | Must not be empty, and each path must be a valid directory path. |
 | `preserve_regex`        | list   | A list of regular expressions for files and directories to preserve during the copy and remove process.                                                                    | No       | Each entry must be a valid regular expression. |
 | `remove_regex`          | list   | A list of regular expressions for files and directories to remove before copying generated code. If not set, this defaults to the `source_roots`. A more specific `preserve_regex` takes precedence. | No       | Each entry must be a valid regular expression. |
+| `release_exclude_paths` | list   | A list of directories to exclude from the release.                                                                                                                    | No       | Each entry must be a valid directory path.     |
```
parthea, line 27:
```suggestion
| `release_exclude_libraries` | list   | A list of libraries to exclude from the release.                                                                                                                    | No       | Each entry must be a valid directory path.     |
```

Should this be libraries? If path is correct, please can you share an example in the comment to clarify when we would want to exclude specific paths, as opposed to libraries?

internal/config/state.go
```diff
@@ -99,12 +99,21 @@ type LibraryState struct {
 	// If not set, this defaults to the `source_roots`.
 	// A more specific `preserve_regex` takes precedence.
 	RemoveRegex []string `yaml:"remove_regex" json:"remove_regex"`
+	// Path of commits to be excluded from parsing while calculating library changes.
+	// If all files from commit belong to one of the paths it will be skipped.
+	ReleaseExcludePaths []string `yaml:"release_exclude_paths,omitempty" json:"release_exclude_paths,omitempty"`
```
parthea, line 104: Please can you share an example of how this will be used? Historically we would rely on the conventional commit message to decide if a release is needed.

doc/state-schema.md
zhumin8: It is `release_exclude_paths` according to [go/librarian:release-command](http://goto.google.com/librarian:release-command). I'd suggest port this discussion over to the doc for visibility.

JoeWang1127 (approved):

codyoss (approved):

PR #1659, opened 2025-08-07
chore(internal/librarian): extend configure contract to include version check

### LINKED ISSUE #1010: librarian: [release] Update Container Contract for configure

JoeWang1127:
Fixes #1010

### DIFF
--- a/internal/librarian/generate_test.go
+++ b/internal/librarian/generate_test.go
@@ -17,7 +17,6 @@ package librarian
 import (
 	"context"
 	"errors"
-	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -204,6 +203,7 @@ func TestRunConfigureCommand(t *testing.T) {
 		container          *mockContainerClient
 		wantConfigureCalls int
 		wantErr            bool
+		wantErrMsg         string
 	}{
 		{
 			name: "configures library successfully",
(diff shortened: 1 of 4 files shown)


internal/librarian/generate_test.go
```diff
@@ -1662,3 +1621,49 @@ func TestCleanAndCopyLibrary(t *testing.T) {
 		})
 	}
 }
+
+// newTestGitRepo creates a new git repository in a temporary directory.
+func newTestGitRepo(t *testing.T) gitrepo.Repository {
+	return newTestGitRepoWithState(t, true)
```
codyoss, line 1627: nit: I think you also want t.Helper() here too.

internal/librarian/state.go
```diff
@@ -103,7 +103,7 @@ func populateServiceConfigIfEmpty(state *config.LibrarianState, contentLoader fu
 func findServiceConfigIn(contentLoader func(file string) ([]byte, error), path string) (string, error) {
 	entries, err := os.ReadDir(path)
 	if err != nil {
-		return "", err
+		return "", errors.Join(fmt.Errorf("failed to read dir: %s", path), err)
```
codyoss, line 106: I would write this as:
```suggestion
		return "", fmt.Errorf("failed to read dir %q: %w", path, err)
```

codyoss (commented):
Just a couple of small nits

internal/librarian/state.go
JoeWang1127: Done.

codyoss (approved):

PR #1686, opened 2025-08-11
feat(librarian/docker): partially copy a repo

### LINKED ISSUE #1685: librarian: [release] only copy libraries and global files into /repo docker volume

JoeWang1127:
Fixes #1685


internal/docker/docker.go
```diff
@@ -118,21 +120,26 @@ type GenerateRequest struct {
...
-	// generate code
+	// Output specifies the empty output directory into which the command should
+	// generate code.
 	Output string
-	// RepoDir is the local root directory of the language repository.
-	RepoDir string
+	// PartialRepoDir is the local root directory of language repository contains
+	// files that make up libraries and global files.
+	// This is the directory that container can access.
+	PartialRepoDir string
```
codyoss, line 142: How are you imagining this field being set? I am wondering if this should be private to this package. I don't think anything outside of this package needs access to this, right?

internal/docker/docker.go
```diff
@@ -337,6 +348,101 @@ func (c *Docker) runCommand(cmdName string, args ...string) error {
...
+
+	return nil
+}
+
+func copyOneLibrary(dst, src string, library *config.LibraryState) error {
+	for _, srcRoot := range library.SourceRoots {
+		dstPath := filepath.Join(dst, srcRoot)
+		srcPath := filepath.Join(src, srcRoot)
+		if err := os.CopyFS(dstPath, os.DirFS(srcPath)); err != nil {
+			return fmt.Errorf("failed to copy %s to %s: %w", library.ID, dst, err)
```
codyoss, line 413: nit:
```suggestion
			return fmt.Errorf("failed to copy %s to %s: %w", library.ID, dstPath, err)
```

internal/docker/docker.go
```diff
@@ -337,6 +348,101 @@ func (c *Docker) runCommand(cmdName string, args ...string) error {
...
+		}
+	}
+
+	return nil
+}
+
+func copyFile(dst, src string) (err error) {
+	sourceFile, err := os.Open(src)
+	if err != nil {
+		return fmt.Errorf("failed to open file: %s", src)
```
codyoss, line 423: Same goes for the couple below:
```suggestion
		return fmt.Errorf("failed to open file %q: %w", src, err)
```

internal/docker/docker.go
```diff
@@ -337,6 +348,101 @@ func (c *Docker) runCommand(cmdName string, args ...string) error {
...
+
+	if err := os.MkdirAll(filepath.Dir(dst), 0755); err != nil {
+		return fmt.Errorf("failed to make directory: %s", src)
+	}
+	destinationFile, err := os.Create(dst)
+	if err != nil {
+		return fmt.Errorf("failed to create file: %s", dst)
+	}
+
+	defer func() {
```
codyoss, line 435: I think I would usually write this as just `defer destinationFile.Close()` and then have the copy return an error directly if needed.

internal/docker/docker.go
```diff
@@ -22,13 +22,13 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/googleapis/librarian/internal/config"
```
codyoss, line 25: CI appears to be broken, please run `go run golang.org/x/tools/cmd/goimports@latest -w . `

internal/docker/docker.go
```diff
@@ -337,6 +346,96 @@ func (c *Docker) runCommand(cmdName string, args ...string) error {
...
+//
+// 1. all directories that make up all libraries, or one library, if the library
+// ID is specified.
+//
+// 2. the .librarian directory.
+//
+// 3. global files declared in config.yaml.
+func setupPartialRepo(request *ReleaseRequest) error {
+	if request.partialRepoDir == "" {
+		request.partialRepoDir = filepath.Join(os.TempDir())
```
codyoss, line 360: can we use `filepath.Join(ctg.WorkRoot, "release-init")`

codyoss (commented):
Just a couple of small things or else looks good

internal/docker/docker.go

internal/docker/docker.go

PR #1715, opened 2025-08-13
feat: extends `configure-reques.json`

### LINKED ISSUE #1653: Update configure-request.json to list all libraries for global file edits
To support global file edits during library onboarding, the `configure` command needs to be updated so that the container will receive the context of all libraries in the repository to correctly regenerate manifests and dependency files.

* `configure-request.json` is modified to include a list of all libraries in the repository.  
* Each API in the request includes a `status` field to indicate if it is `"new"` or `"existing"`.  

Example:
```json
{
  "libraries": [
    {
      "id": "google-cloud-secretmanager",
      "apis": [
        {
          "path": "google/cloud/secretmanager/v1",
          "service_config": "secretmanager_v1.yaml",
          "status": "existing"
        },
        {
          "path": "google/cloud/secretmanager/v1beta1",
          "service_config": "secretmanager_v1beta1.yaml",
          "status": "new"
        }
      ],
      "source_roots": [ "secretmanager" ],
      ...
    },
    {
      "id": "google-cloud-pubsub-v1",
      "apis": [
        {
          "path": "google/cloud/pubsub/v1",
          "service_config": "pubsub_v1.yaml",
          "status": "existing"
        }
      ],
      "source_roots": [ "pubsub" ],
      ...
    }
  ]
}
```

See: [go/librarian:global-file-edits](http://goto.google.com/librarian:global-file-edits)

JoeWang1127:
Fixes #1653

### DIFF
--- a/internal/config/state.go
+++ b/internal/config/state.go
@@ -21,6 +21,11 @@ import (
 	"strings"
 )
 
+const (
+	statusNew      = "new"
+	statusExisting = "existing"
+)
+
 // LibrarianState defines the contract for the state.yaml file.
 type LibrarianState struct {
 	// The name and tag of the generator image to use. tag is required.
@@ -189,12 +194,17 @@ type API struct {
 	Path string `yaml:"path" json:"path"`
 	// The name of the service config file, relative to the API `path`.
 	ServiceConfig string `yaml:"service_config" json:"service_config"`
+	// The status of the API, one of "new" or "existing".
+	Status string `yaml:"status" json:"status"`
 }
 
 // Validate checks that the API is valid.
 func (a *API) Validate() error {
 	if !isValidDirPath(a.Path) {
 		return fmt.Errorf("invalid path: %q", a.Path)
+	}
+	if a.Status != statusNew && a.Status != statusExisting {
+		return fmt.Errorf("invalid status: %q", a.Status)
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


internal/config/state.go
```diff
@@ -189,13 +194,18 @@ type API struct {
 	Path string `yaml:"path" json:"path"`
 	// The name of the service config file, relative to the API `path`.
 	ServiceConfig string `yaml:"service_config" json:"service_config"`
+	// The status of the API, one of "new" or "existing".
+	Status string `yaml:"status" json:"status"`
```
codyoss, line 198: I think we might need to be start writing marshaller functions for our state file. These are fields we do want present in the request message in docker, but we don't want them committed to the state.yaml file of the repo. I think that leaves us with a couple of options.
 
- On marshalling we ignore these fields.
- We have some sort of cleaning function that strips fields like this
- open to other options too.

JoeWang1127: I think we can ignore this value when writing state to a yaml file.

internal/librarian/state_test.go
```diff
@@ -47,6 +47,7 @@ libraries:
     apis:
       - path: a/b/v1
         service_config: a/b/v1/service.yaml
+        status: existing
```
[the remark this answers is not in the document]
JoeWang1127: Done.

codyoss (dismissed):

codyoss (approved):

PR #1740, opened 2025-08-14
build: allow specifying organization for generate command

chingor13:
This will aid in testing. The default is set to `googleapis` on the Cloud Build trigger


ldetmer (commented):
this is fine to unbiock, but one feedback from review ws that we'll change this to accept the full repository url

ldetmer (approved):

PR #1747, opened 2025-08-14
test(internal/librarian): fix validation test

codyoss:
Also put some more details in the error for debugging.


codyoss:
@JoeWang1127 I should have mentioned this in the skeleton PR but you don't need to validate the config here. It is being done in `librarian.go` so by the time it gets to this method the config should be in good standing -- at least that is the hope.

Issue #1009, opened 2025-07-28
librarian: [release] Implement `release tag-and-release` Command


bisun-google:
Build the CLI for the `release tag-and-release` subcommand, including the logic to find merged PRs, create Git tags, create GitHub Releases, and update PR labels.

A release commit message structure depends on the tagging strategy configured for the repository.

#### Tag Per Repo

In this strategy, a single release pull request results in one release tag for the entire repository, which is suitable for languages like Java and C++. The Librarian will determine the version and other variables to fill in the template.

```
chore(main): release {version}

Librarian Version: {librarian_version}
Language Image: {language_image_name_and_digest}

<details><summary>{version}</summary>

[<version>](https://github.com/googleapis/{repo}/compare/{previous_tag}...{new_tag}) ({date})

### Features

* {commit_summary} ([{commit_hash}](https://github.com/googleapis/{repo}/commit/{commit_hash}))

### Bug Fixes

* {commit_summary} ([{commit_hash}](https://github.com/googleapis/{repo}/commit/{commit_hash}))

### Documentation

* {commit_summary} ([{commit_hash}](https://github.com/googleapis/{repo}/commit/{commit_hash}))

</details>
```

* [Real **release** PR example in Java](https://github.com/googleapis/google-cloud-java/pull/11665)

#### Tag Per Library

In this strategy, a single release pull request can generate multiple, distinct tags, one for each library that has changed. This approach is used by languages like Python and Go. A section for each library release will be wrapped into a `<details>` element, where the `<summary>` child element will contain the release tag.

```
chore: release main

Librarian Version: {librarian_version}
Language Image: {language_image_name_and_digest}

<details><summary>{library_name}: {version}</summary>

[{version}](https://github.com/googleapis/{repo}/compare/{library_name}-v{previous_version}...{library_name}-v{new_version}) ({date})

### Features

* {commit_summary} ([{commit_hash}](https://github.com/googleapis/{repo}/commit/{commit_hash}))

</details>

<details><summary>{another_library_name}: {version}</summary>

[{version}](https://github.com/googleapis/{repo}/compare/{another_library_name}-v<previous_version>...{another_library_name}-v{new_version}) ({date})

### Bug Fixes

* {commit_summary} ([{commit_hash}](https://github.com/googleapis/{repo}/commit/{commit_hash}))

</details>
```

* [Real **release** PR example in Python](https://github.com/googleapis/google-cloud-python/pull/14151) 

[...]

PR #1754, opened 2025-08-15
feat(internal/librarian): add logic to get pull requests

codyoss:
Updates: #1009

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -103,7 +103,7 @@ type Config struct {
 	// the tool is executing.
 	CI string
 
-	// commandName is the name of the command being executed.
+	// CommandName is the name of the command being executed.
 	//
 	// commandName is populated automatically after flag parsing. No user setup is
 	// expected.


internal/librarian/generate_test.go
```diff
@@ -1647,7 +1647,16 @@ func newTestGitRepoWithState(t *testing.T, writeState bool) gitrepo.Repository {
 			t.Fatalf("os.MkdirAll: %v", err)
 		}
 		stateFile := filepath.Join(stateDir, "state.yaml")
-		if err := os.WriteFile(stateFile, []byte(""), 0644); err != nil {
+		state := `
+image: gcr.io/test/image:v1.2.3
```
JoeWang1127, line 1651: FYI, I added similar test setup in #1746.

internal/librarian/tag_and_release.go
```diff
@@ -46,18 +52,104 @@ func init() {
...
+			slog.Error("failed to process pull request", "pr", p.GetNumber(), "error", err)
+			hadErrors = true
+			continue
+		}
+		slog.Info("processed pull request", "pr", p.GetNumber())
+	}
+	slog.Info("finished processing all pull requests")
+
+	if hadErrors {
+		return fmt.Errorf("failed to process some pull requests")
```
JoeWang1127, line 117: nit: `errors.New()` should be sufficient.

internal/librarian/tag_and_release_test.go
```diff
@@ -52,3 +58,147 @@ func TestNewTagAndReleaseRunner(t *testing.T) {
...
+			},
+			want: []*github.PullRequest{pr123},
+		},
+		{
+			name: "invalid pull request format",
+			cfg: &config.Config{
+				PullRequest: "invalid",
+			},
+			ghClient: &mockGitHubClient{},
+			wantErr:  true,
```
JoeWang1127, line 88: I think we should check the error message to make sure the error is correct.

PR #1757, opened 2025-08-15
ci: Cloud Build jobs to propagate _PUSH substitution variable

suztomo:
The Cloud Build generate-worker job to pass the _PUSH substitution variable to the next Cloud Build jobs. They invoke Librarian CLI with either '-push=true' or '-push=false' parameter. This controls whether the job creates a pull request or not.

This change should come with the Terraform change cl/795539979.


suztomo:
I believe the 80% code coverage target is for outside internal/automation.

chingor13 (approved):

PR #1763, opened 2025-08-15
docs: add documentation for generate CLI command

ldetmer:


suztomo (commented):
The file name `doc/container-commands.md` doesn't sound right to me. Shall we remove "container"?

I think the container commands are the communication between Librarian CLI and the language containers.

codyoss (approved):

PR #1773, opened 2025-08-18
fix: add -push in stage-release job

catchyzheng:
Ref: 
https://source.corp.google.com/piper///depot/google3/configs/cloud/gong/services/cloud_sdk/cloud_sdk_librarian/envs/prod/stage_release.tf;rcl=796442435;l=44

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -103,6 +103,12 @@ type Config struct {
 	// the tool is executing.
 	CI string
 
+	// CommandName is the name of the command being executed.
+	//
+	// commandName is populated automatically after flag parsing. No user setup is
+	// expected.
+	CommandName string
+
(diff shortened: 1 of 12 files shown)


infra/prod/stage-release.yaml
```diff
@@ -41,8 +41,7 @@ steps:
       - '/workspace/$_REPOSITORY'
       - '-library'
       - $_LIBRARY_ID
-      - '-api-source'
```
ldetmer, line 44: I don't think you should be getting rid of setting the api-source, is there a reason?

chingor13: @ldetmer I think I requested this. Does releasing need googleapis/googleapis?

PR #1775, opened 2025-08-18
feat: add ability to create branch and push to remote repository

### LINKED ISSUE #1762: librarian: `-push` doesn't push to GitHub
We're getting a 422 error on the CreatePullRequest API call to GitHub.

It appears we are creating a commit in the local repo, but not pushing to GitHub.

chingor13:
Fixes #1762

Introduces a new `GitUser` struct which represents all potential configuration data needed for git operations (username/password for auth, name/email for commit messages). This is stored on the LocalRepository when it is loaded or cloned. Currently, we only allow CLI-based configuration for the password (GitHub token) -- other values are hard-coded, but we could easily pipe through arguments from the command line if needed.

### DIFF
--- a/internal/librarian/command.go
+++ b/internal/librarian/command.go
@@ -46,7 +46,18 @@ func newCommandRunner(cfg *config.Config) (*commandRunner, error) {
 	if cfg.APISource == "" {
 		cfg.APISource = "https://github.com/googleapis/googleapis"
 	}
-	languageRepo, err := cloneOrOpenRepo(cfg.WorkRoot, cfg.Repo, cfg.CI)
+
+	var gitUser *gitrepo.GitUser
+	if cfg.GitHubToken != "" {
+		gitUser = &gitrepo.GitUser{
+			Username:    "cloud-sdk-librarian",
+			Password:    cfg.GitHubToken,
+			DisplayName: "Cloud SDK Librarian",
+			Email:       "cloud-sdk-librarian-robot@google.com",
+		}
+	}
+
+	languageRepo, err := cloneOrOpenRepo(cfg.WorkRoot, cfg.Repo, cfg.CI, gitUser)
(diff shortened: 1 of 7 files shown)


internal/librarian/command.go
```diff
@@ -204,17 +217,27 @@ func commitAndPush(ctx context.Context, r *generateRunner, commitMessage string)
...
 
+	datetimeNow := formatTimestamp(time.Now())
+	branch := fmt.Sprintf("librarian-%s", datetimeNow)
+	slog.Info("Creating branch", slog.String("branch", branch))
+	if err := r.repo.CreateBranchAndCheckout(branch); err != nil {
+		return err
+	}
+
 	// TODO: get correct language for message (https://github.com/googleapis/librarian/issues/885)
+	slog.Info("Committing", slog.String("message", commitMessage))
```
codyoss, line 228: nit: I would usually write this
```suggestion
	slog.Info("Committing", "message", commitMessage)
```

internal/librarian/command.go
```diff
@@ -46,15 +46,26 @@ func newCommandRunner(cfg *config.Config) (*commandRunner, error) {
 	if cfg.APISource == "" {
 		cfg.APISource = "https://github.com/googleapis/googleapis"
 	}
-	languageRepo, err := cloneOrOpenRepo(cfg.WorkRoot, cfg.Repo, cfg.CI)
+
+	var gitUser *gitrepo.GitUser
+	if cfg.GitHubToken != "" {
+		gitUser = &gitrepo.GitUser{
+			Username:    "cloud-sdk-librarian",
```
codyoss, line 53: For our CI environments can we just set the git config to have these values? I don't think we want this to be the username if people are running locally.

codyoss (approved):

PR #1783, opened 2025-08-19
fix: make config.yaml optional

codyoss:
This file should not be required.


chingor13 (approved):
It may cause a NPE in the future, but if it's not used, then we're ok for now.

Issue #1008, opened 2025-07-28
Implement `release init` command


bisun-google:
Build the CLI for the `release init` subcommand, including all specified flags, and integrate the `release-please lite` logic and container invocation.

PR #1793, opened 2025-08-21
feat: create a pull request in release init command

JoeWang1127:
Updates #1008


internal/librarian/command.go
```diff
@@ -239,25 +239,25 @@ func commitAndPush(ctx context.Context, r *generateRunner, commitMessage string)
...
-	if err := r.repo.Push(branch); err != nil {
+	if err := repo.Push(branch); err != nil {
 		return err
 	}
 
 	// Create a new branch, set title and message for the PR.
 	titlePrefix := "Librarian pull request"
 	title := fmt.Sprintf("%s: %s", titlePrefix, datetimeNow)
 	slog.Info("Creating pull request", slog.String("branch", branch), slog.String("title", title))
-	if _, err = r.ghClient.CreatePullRequest(ctx, gitHubRepo, branch, title, commitMessage); err != nil {
```
suztomo, line 260: It seems the logic was already there. Does this PR resolve something?

suztomo (approved):

internal/librarian/command.go
JoeWang1127: Refactor `commitAndPush` function so that it can be called in `release init` command.

suztomo: Nice.

codyoss (approved):

PR #1809, opened 2025-08-25
fix: release-init not to delete existing files

suztomo:


internal/librarian/command.go
```diff
@@ -212,6 +212,16 @@ func cleanAndCopyLibrary(state *config.LibrarianState, repoDir, libraryID, outpu
 	return nil
 }
 
+// copyLibrary copies the new files from outputDir to repoDir, without cleaning the existing files.
+func copyLibrary(repoDir, libraryID, outputDir string) error {
```
JoeWang1127, line 216: Could you pull the latest change? I think you don't need to change this function.

suztomo:
Closed.

PR #1822, opened 2025-08-25
fix: ignore version override if library flag is not specified

### LINKED ISSUE #1823: In release init command, version override should be ignored if `library` flag is not specified

JoeWang1127:
Fixes #1823

### DIFF
--- a/internal/librarian/release_init.go
+++ b/internal/librarian/release_init.go
@@ -125,7 +125,7 @@ func (r *initRunner) runInitCommand(ctx context.Context, outputDir string) error
 				continue
 			}
 			// Only update one library with the given library ID.
-			updatedLibrary, err := updateLibrary(r.repo, library, r.cfg.LibraryVersion, true)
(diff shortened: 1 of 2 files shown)


internal/librarian/release_init.go
```diff
@@ -188,28 +194,25 @@ func (r *initRunner) runInitCommand(ctx context.Context, outputDir string) error
...
-	library := state.Libraries[index]
-	updatedLibrary, err := getChangesOf(r.repo, library)
+// updateLibrary updates the given library in the following way:
+//
+// 1. Get the library's commit history in the given git repository.
+//
+// 2. Override the library version if overrideVersion is true.
+//
+// 3. Set the library's release trigger to true.
+func updateLibrary(repo gitrepo.Repository, library *config.LibraryState, libraryVersion string, overrideVersion bool) (*config.LibraryState, error) {
```
codyoss, line 204: I think this can be simplified to just pass `""` when `overrideVersion` would be false -- I don't think we need both fields.

Issue #1806, opened 2025-08-25
librarian: `release init` deletes all library code


parthea:
`release init` appears to be broken because all client library code is deleted after running this command. See log below

`librarian generate -api-source=/usr/local/google/home/partheniou/git/googleapis` works as expected for the same commit

```
partheniou@partheniou-vm-3:~/git/google-cloud-python$ librarian init -api-source=/usr/local/google/home/partheniou/git/googleapis
2025/08/25 14:50:05 INFO librarian arguments="[init -api-source=/usr/local/google/home/partheniou/git/googleapis]"
2025/08/25 14:50:05 INFO Temporary working directory dir=/tmp/librarian-20250825T145005Z
2025/08/25 14:50:05 INFO repo not specified, using current working directory as repo root path=/usr/local/google/home/partheniou/git/google-cloud-python
2025/08/25 14:50:05 INFO Opening repository dir=/usr/local/google/home/partheniou/git/google-cloud-python
2025/08/25 14:50:16 INFO Opening repository dir=/usr/local/google/home/partheniou/git/googleapis
2025/08/25 14:50:18 INFO Initiating a release dir=/tmp/librarian-20250825T145005Z/output
2025/08/25 14:50:18 WARN Invalid conventional commit message message=wip hash=c718942b427ab8401cb784da6796bca974a7295d
2025/08/25 14:50:18 WARN Invalid conventional commit message message=wip hash=d135170d8d8fe0d7f678816f3de70040959fb80b
2025/08/25 14:50:19 INFO === Docker start ===============================================================
2025/08/25 14:50:19 INFO /usr/bin/docker run --rm -v /tmp/librarian-20250825T145005Z/release-init/.librarian:/librarian -v /tmp/librarian-20250825T145005Z/release-init:/repo:ro -v /tmp/librarian-20250825T145005Z/output:/output --user 814163:89939 us-central1-docker.pkg.dev/cloud-sdk-librarian-prod/images-dev/python-librarian-generator:latest release-init --librarian=/librarian --repo=/repo --output=/output
2025/08/25 14:50:19 INFO --------------------------------------------------------------------------------
2025-08-25 14:50:19,970 synthtool [CRITICAL] > You are running the synthesis script directly, this will be disabled in a future release of Synthtool. Please use python3 -m synthtool instead.
2025/08/25 14:50:20 INFO === Docker end =================================================================
2025/08/25 14:50:20 INFO Clean destinations and copy generated results for library id=google-cloud-language
[...]

PR #1824, opened 2025-08-26
fix(internal/librarian): don't clean dir for release init copies

codyoss:
We should intentionally copy over only the files that are in the output directory of the release init container. We don't mandate today that all libraries files must be copied over, so we should selectively overwrite files when required.

Fixes: #1806

### DIFF
--- a/internal/librarian/command.go
+++ b/internal/librarian/command.go
@@ -208,6 +208,56 @@ func cleanAndCopyLibrary(state *config.LibrarianState, repoDir, libraryID, outpu
 	return copyLibrary(repoDir, outputDir, library)
 }
 
+func copyLibraryFiles(state *config.LibrarianState, dest, libraryID, src string) error {
+	library := findLibraryByID(state, libraryID)
+	if library == nil {
+		return fmt.Errorf("library %q not found", libraryID)
+	}
+	slog.Info("Copying library files", "id", library.ID, "destination", dest, "source", src)
+	for _, srcRoot := range library.SourceRoots {
+		dstPath := filepath.Join(dest, srcRoot)
+		srcPath := filepath.Join(src, srcRoot)
+		files, err := getDirectoryFilesnames(srcPath)
+		if err != nil {
+			return err
+		}
+		for _, file := range files {
+			slog.Info("Copying file", "file", file)
+			srcFile := filepath.Join(srcPath, file)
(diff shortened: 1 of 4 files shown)


codyoss:
@parthea would you mind checking if this patch resolves the issues you were facing? I don't have a good repo environment setup atm, but hope to later tmr if you don't get to this.

internal/librarian/command.go
```diff
@@ -208,6 +208,56 @@ func cleanAndCopyLibrary(state *config.LibrarianState, repoDir, libraryID, outpu
...
+			if err := copyFile(dstFile, srcFile); err != nil {
+				return fmt.Errorf("failed to copy file %q for library %s: %w", srcFile, library.ID, err)
+			}
+		}
+	}
+	return nil
+}
+
+func getDirectoryFilesnames(dir string) ([]string, error) {
+	if _, err := os.Stat(dir); os.IsNotExist(err) {
```
JoeWang1127, line 237: Does the function need to return early if `err` is another type of error?

internal/librarian/command.go
```diff
@@ -208,6 +208,56 @@ func cleanAndCopyLibrary(state *config.LibrarianState, repoDir, libraryID, outpu
 	return copyLibrary(repoDir, outputDir, library)
 }
 
+func copyLibraryFiles(state *config.LibrarianState, dest, libraryID, src string) error {
```
JoeWang1127, line 211: What's the difference between `copyLibraryFiles` and `copyLibrary`? Can we combine them?

internal/librarian/command_test.go
```diff
@@ -700,6 +700,142 @@ func TestCommitAndPush(t *testing.T) {
...
+					{
+						ID: "example-library",
+					},
+				},
+			},
+			wantErr:    true,
+			wantErrMsg: "not found",
+		},
+		{
+			name:      "source root not found",
```
JoeWang1127, line 774: This name suggests that this test should return an error but it actually doesn't, is this intentional?

parthea:
> @parthea would you mind checking if this patch resolves the issues you were facing? I don't have a good repo environment setup atm, but hope to later tmr if you don't get to this.

Yes, this resolved the issues. I'm not using the global file edits feature yet but I plan to do so for the top level changelog.

https://github.com/googleapis/google-cloud-python/blob/main/CHANGELOG.md

PR #1867, opened 2025-08-28
fix: get conventional commits from source repository

### LINKED ISSUE #1866: librarian: generation is looking for conventional commits in the wrong repo (I believe)
Having just updated to the latest version of the CLI, I'm getting errors like this when generating Go libraries:

> failed to fetch conventional commits for library, maps: failed to get commits for library maps: did not find commit 6914b3387f851ae8411231e4e57fefa5290eb9f8 when iterating

I believe the issue is in #1837, and I *think* it's for two reasons:

- It's looking in the language repo, not the googeapis repo
- It's looking for changes in the source roots of the library, not the source paths of the APIs

I may have misunderstood, of course.

Marking as p1 as I'd expect generation to always fail at the moment - the commit stored in state.last_generated_commit will *never* be in the language repo.

JoeWang1127:
Function `updateChangesSinceLastGeneration` may not be needed since we don't need to pass api changes to language container.

Fixes #1866

### DIFF
--- a/internal/librarian/generate_test.go
+++ b/internal/librarian/generate_test.go
@@ -17,16 +17,16 @@ package librarian
 import (
 	"context"
 	"errors"
+	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/google/go-cmp/cmp"
+	"github.com/googleapis/librarian/internal/config"
+	"github.com/googleapis/librarian/internal/gitrepo"
 	"os"
 	"path/filepath"
 	"regexp"
 	"sort"
 	"strings"
 	"testing"
-
-	"github.com/google/go-cmp/cmp"
-	"github.com/googleapis/librarian/internal/config"
-	"github.com/googleapis/librarian/internal/gitrepo"
 )
 
 func TestRunGenerateCommand(t *testing.T) {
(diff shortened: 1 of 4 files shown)


internal/librarian/generate_test.go
```diff
@@ -1468,3 +1481,139 @@ func TestCompileRegexps(t *testing.T) {
...
+					"api/another/path": {
+						{
+							Message: "fix: a bug fix\n\nThis is another body.\n\nPiperOrigin-RevId: 573342",
+							Hash:    hash2,
+						},
+					},
+				},
+				ChangedFilesInCommitValueByHash: map[string][]string{
+					hash1.String(): {
+						"api/one//path/file.txt",
```
jskeet, line 1535: Are the double forward slashes here and below deliberate?

internal/librarian/release_please_lite.go
```diff
@@ -39,9 +39,14 @@ func GetConventionalCommitsSinceLastRelease(repo gitrepo.Repository, library *co
...
-// the given library since the last generation.
+// all API paths in given library since the last generation.
 func GetConventionalCommitsSinceLastGeneration(repo gitrepo.Repository, library *config.LibraryState) ([]*conventionalcommits.ConventionalCommit, error) {
-	commits, err := repo.GetCommitsForPathsSinceCommit(library.SourceRoots, library.LastGeneratedCommit)
+	apiPaths := make([]string, 0)
+	for _, oneAPI := range library.APIs {
+		apiPaths = append(apiPaths, oneAPI.Path)
+	}
+
+	commits, err := repo.GetCommitsForPathsSinceCommit(apiPaths, library.LastGeneratedCommit)
```
jskeet, line 49: What would you expect to happen if LastGeneratedCommit is empty? (I'm expecting that to be common when onboarding a new library.) I think in that case we should just not bother looking for anything.

jskeet (commented):
One comment on the prod code - I'm not sure I'm qualfied to review the test code here.

internal/librarian/release_please_lite.go
JoeWang1127: It's a good point.

How about I file a new issue and fix it in another PR?

PR #1876, opened 2025-08-29
fix(internal/librarian): generate should fail with error if all libraries failed

### LINKED ISSUE #1787: librarian: should not attempt to open PR if all generation fails
Example: https://pantheon.corp.google.com/cloud-build/builds;region=global/c61222c7-576d-467c-9c78-38100c413b09?e=RetailSupportToolLaunch::RetailSupportToolEnabled,RetailDataQualityGaLaunch::RetailDataQualityGaEnabled&mods=allow_workbench_image_override&project=cloud-sdk-librarian-prod

In the above example, we attempt to generate all APIs for google-cloud-python. There is only a single library configured and generation fails.

The CLI should exit with non-zero code and should not open an empty PR.

zhumin8:
Fixes #1787

### DIFF
--- a/internal/librarian/generate.go
+++ b/internal/librarian/generate.go
@@ -141,12 +141,17 @@ func (r *generateRunner) run(ctx context.Context) error {
 		}
 		prBody += fmt.Sprintf("feat: generated %s\n", libraryID)
 	} else {
+		failedGenerations := 0
 		for _, library := range r.state.Libraries {
 			if err := r.generateSingleLibrary(ctx, library.ID, outputDir); err != nil {
 				// TODO(https://github.com/googleapis/librarian/issues/983): record failure and report in PR body when applicable
 				slog.Error("failed to generate library", "id", library.ID, "err", err)
 				prBody += fmt.Sprintf("%s failed to generate\n", library.ID)
+				failedGenerations++
 			}
+		}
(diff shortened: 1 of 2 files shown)


internal/librarian/generate.go
```diff
@@ -141,13 +141,18 @@ func (r *generateRunner) run(ctx context.Context) error {
...
+		failedGenerations := 0
 		for _, library := range r.state.Libraries {
 			if err := r.generateSingleLibrary(ctx, library.ID, outputDir); err != nil {
 				// TODO(https://github.com/googleapis/librarian/issues/983): record failure and report in PR body when applicable
 				slog.Error("failed to generate library", "id", library.ID, "err", err)
 				prBody += fmt.Sprintf("%s failed to generate\n", library.ID)
+				failedGenerations++
 			}
 		}
+		if failedGenerations > 0 && failedGenerations == len(r.state.Libraries) {
```
chingor13, line 153: nit: it might be more maintainable to keep track of # of successfully generated libraries > 0. We might have cases where we skip a library generation and don't rely on it actually failing. We could invert the check to only open the PR if at least 1 library successfully generates.

chingor13: Opened #1885 to follow up on this

chingor13 (approved):

PR #1940, opened 2025-09-05
fix: add --branch flags to repo clone to support branch checkout

ldetmer:


infra/prod/generate.yaml
```diff
@@ -21,6 +21,8 @@ steps:
     args:
       - 'clone'
       - '--depth=1'
+      - '--single-branch'
```
jskeet, line 24: I don't think this is needed here, as we're specifying `depth`. That said, it does no harm.

jskeet (dismissed):

jskeet (approved):

Issue #1832, opened 2025-08-26
librarian: consider combining copyLibraryFiles and copyLibrary


codyoss:
For details see thread: https://github.com/googleapis/librarian/pull/1824#discussion_r2301247514

These methods of different semantics, but we may need the former eventually as os.CopyFS does not handle symlinks.

suztomo:
> os.CopyFS does not handle symlinks.

It now handles symlinks now as we use Go 1.25.

suztomo:
Initial attempt failed due to the different handling of the existing files.

```
suztomo@suztomo2:~/librarian$ git diff
diff --git a/internal/librarian/command.go b/internal/librarian/command.go
index 361dd2fe..332b42e4 100644
--- a/internal/librarian/command.go
+++ b/internal/librarian/command.go
@@ -322,7 +322,7 @@ func copyLibrary(dst, src string, library *config.LibraryState) error {
                dstPath := filepath.Join(dst, srcRoot)
                srcPath := filepath.Join(src, srcRoot)
                if err := os.CopyFS(dstPath, os.DirFS(srcPath)); err != nil {
-                       return fmt.Errorf("failed to copy %s to %s: %w", library.ID, dstPath, err)
+                       return fmt.Errorf("os.CopyFS failed to copy %s to %s: %w", library.ID, dstPath, err)
                }
        }
 
diff --git a/internal/librarian/release_init.go b/internal/librarian/release_init.go
index 0fd643c6..a4f773de 100644
--- a/internal/librarian/release_init.go
+++ b/internal/librarian/release_init.go
@@ -183,7 +183,7 @@ func (r *initRunner) runInitCommand(ctx context.Context, outputDir string) error
                                continue
                        }
                        // Only copy one library to repository.
-                       if err := copyLibraryFiles(r.state, r.repo.GetDir(), r.cfg.Library, outputDir); err != nil {
+                       if err := copyLibrary(r.repo.GetDir(), outputDir, library); err != nil {
                                return err
                        }
 
@@ -191,7 +191,7 @@ func (r *initRunner) runInitCommand(ctx context.Context, outputDir string) error
                }
 
                // Copy all libraries to repository.
-               if err := copyLibraryFiles(r.state, r.repo.GetDir(), library.ID, outputDir); err != nil {
+               if err := copyLibrary(r.repo.GetDir(), outputDir, library); err != nil {
                        return err
                }
        }
[...]

PR #1941, opened 2025-09-05
chore: use copyLibraryFiles overe copyLibrary

suztomo:
CopyLibrary function does not override the existing files.

Updating the test about symbolic links. We should support symbolic
links in the library repositories.

Fixes https://github.com/googleapis/librarian/issues/1832.


suztomo:
Improving code coverage.

PR #1945, opened 2025-09-05
test(librarian): add e2e test for multi-library generation

### LINKED ISSUE #988: test(librarian): add test case to verify functions to generate multiple libraries
Followup to changes in #977, need to add integration test:
a test case in `generate_e2e_test.go` to verify the functions to generate multiple libraries

jinseopkim0:
Fixes #988


JoeWang1127:
Could you add more test cases to verify the behavior of
1. one of the library generation failed
2. all library generation failed

jinseopkim0:
> Could you add more test cases to verify the behavior of
> 
> 1. one of the library generation failed
> 2. all library generation failed

Thanks for the comments, I've added tests for each scenario.

JoeWang1127:
Could you consolidate the three test cases into a table test? Looks like the test setups are similar to each other, only the state.yaml is different.

Issue #1903, opened 2025-09-03
librarian: running generate without arguments should return an error


julieqiu:
The help text for `go run ./cmd/librarian generate -h` suggests that `-source` and `-api` are required arguments:

```
Usage:
  librarian generate -source=<api-root> -api=<api-path> [flags]
```

However, when I ran `librarian generate` with no argument, I got the following:

```
❯ go run ./cmd/librarian generate                                                                 
go: downloading go1.25.0 (darwin/arm64)
2025/09/02 21:31:41 INFO librarian arguments=[]
2025/09/02 21:31:41 INFO Temporary working directory dir=/var/folders/_j/n7nk6w3x6t3f0xjc0nz2chr80000gn/T/librarian-20250902T213141Z
2025/09/02 21:31:41 INFO repo not specified, using current working directory as repo root path=/Users/julieqiu/code/googleapis/librarian
2025/09/02 21:31:41 INFO Opening repository dir=/Users/julieqiu/code/googleapis/librarian
2025/09/02 21:31:41 INFO Checking for repository dir=/var/folders/_j/n7nk6w3x6t3f0xjc0nz2chr80000gn/T/librarian-20250902T213141Z/googleapis
2025/09/02 21:31:41 INFO Repository not found, executing clone
2025/09/02 21:31:41 INFO Cloning repository url=https://github.com/googleapis/googleapis dir=/var/folders/_j/n7nk6w3x6t3f0xjc0nz2chr80000gn/T/librarian-20250902T213141Z/googleapis
Enumerating objects: 146306, done.
Counting objects: 100% (137/137), done.
Compressing objects: 100% (80/80), done.
Total 146306 (delta 78), reused 60 (delta 57), pack-reused 146169 (from 4)
2025/09/02 21:32:29 INFO config.yaml not found, proceeding
2025/09/02 21:32:29 failed to get GitHub repo from remote: could not find an 'origin' remote pointing to a GitHub https URL
exit status 1
```

I expected the command to fail before running these systems with a message that `-source` and `-api` are required.

codyoss:
They are not required, we will update the documentation in the CLI as such to make this more clear.

Issue #1014, opened 2025-07-28
librarian: [release] Update Documentation for release command


bisun-google:
Update `doc/state-schema.md`, `doc/container-contract.md`, and the CLI's built-in help text with information about the new release commands and fields.

Issue #207, opened 2025-06-04
librarian: add top-level documentation for commands at doc.go 


julieqiu:
We should use the documentation from https://github.com/googleapis/librarian/issues/206 to generate a full doc.go in cmd/librarian, so that the content shows up at https://pkg.go.dev/github.com/googleapis/librarian/cmd/librarian.

See https://pkg.go.dev/cmd/go as an example. Code is here:
https://github.com/golang/go/blob/master/src/cmd/go/internal/help/help.go

Also run `go help modules` as an example for the type of content.

PR #1948, opened 2025-09-05
docs(librarian): add consistency and more examples for CLI help text

codyoss:
Updates: #207
Fixes: #1903
Fixes: #1014

### DIFF
--- a/internal/librarian/release_init.go
+++ b/internal/librarian/release_init.go
@@ -53,7 +53,9 @@ to the configured language-specific container.
 
 By default, 'release init' leaves the changes in your local working directory
 for inspection. Use the '--push' flag to automatically commit the changes to
-a new branch and create a pull request on GitHub.
+a new branch and create a pull request on GitHub. The '--commit' flag may be
+used to create a local commit without creating a pull request; this flag is
+ignored if '--push' is also specified.
 
 Examples:
   # Create a release PR for all libraries with pending changes.


internal/librarian/generate.go
```diff
@@ -34,33 +34,57 @@ const (
...
+
+# Regenerating existing libraries
+
+You can regenerate a single, existing library by specifying either the library
+ID or the API path. If no specific library or API is provided, Librarian will
+regenerate all libraries listed in '.librarian/state.yaml'.
+
+Examples:
+  # Regenerate a single library by its ID
+  librarian generate --library=secretmanager
```
lqiu96, line 65: qq if `-api` isn't specified, does this generate all the potential versions for the secretmanager path?

i.e. if there was something like `google/cloud/secretmanager/v1` and `google/cloud/secretmanager/v1beta`

internal/librarian/release_init.go
```diff
@@ -37,10 +37,33 @@ const (
...
+
+This command scans the git history since the last release, identifies changes
+(feat, fix, BREAKING CHANGE), and calculates the appropriate version bump
+according to semver rules. It then delegates all language-specific file
+modifications, such as updating a CHANGELOG.md or bumping the version in a pom.xml, 
+to the configured language-specific container.
+
+By default, 'release init' leaves the changes in your local working directory
+for inspection. Use the '--push' flag to automatically commit the changes to
+a new branch and create a pull request on GitHub.
```
lqiu96, line 56: I believe also the `-commit` flag for `release init`. Is there a use case for running `-commit` instead of `-push` and if so would it make sense to specify that?

internal/librarian/release_init.go
```diff
@@ -37,10 +37,33 @@ const (
...
+executing in such a directory the '--repo' flag must be provided.
+
+This command scans the git history since the last release, identifies changes
+(feat, fix, BREAKING CHANGE), and calculates the appropriate version bump
+according to semver rules. It then delegates all language-specific file
+modifications, such as updating a CHANGELOG.md or bumping the version in a pom.xml, 
+to the configured language-specific container.
+
+By default, 'release init' leaves the changes in your local working directory
+for inspection. Use the '--push' flag to automatically commit the changes to
```
JoeWang1127, line 55: I think we should also mention `--commit` flag here.

internal/librarian/generate.go
codyoss: That is correct

internal/librarian/release_init.go
codyoss: Updated, thanks

internal/librarian/release_init.go
codyoss: done

internal/librarian/generate.go
lqiu96: I think may be helpful to call out. I'm guessing that API a version and Library is a collection of APIs (versions)?

lqiu96 (approved):

PR #1954, opened 2025-09-05
fix(sidekick): Tests and commands run in Windows

amanda-tarafa:


internal/sidekick/downloads_cache.go
```diff
@@ -119,33 +119,30 @@ func downloadAttempt(target, source, expectedSha256 string) error {
 	if err != nil {
 		return err
 	}
-	defer os.Remove(tempFile.Name())
+	defer func() {
```
coryan, line 122: Ah, yes, of course.  I will keep this in mind for the future.

internal/sidekick/internal/gcloud/command_test.go
```diff
@@ -49,6 +49,9 @@ func TestCommandYAML(t *testing.T) {
 			if err != nil {
 				t.Fatalf("failed to read file: %v", err)
 			}
+			stringData := strings.ReplaceAll(string(data), "\r\n", "\n")
```
coryan, line 52: maybe:

```suggestion
			data = bytes.ReplaceAll(string(data), []byte("\r\n"), []byte("\n"))
```

Does that work?

internal/sidekick/internal/language/walk_templates_dir_test.go
```diff
@@ -23,6 +24,9 @@ import (
 func TestWalkDir(t *testing.T) {
 	// It should get the `*.md.mustache` files and skip `partial.mustache`
 	got := WalkTemplatesDir(templates, "testTemplates")
+	for i := range got {
+		got[i].OutputPath = strings.ReplaceAll(got[i].OutputPath, "\\", "/")
```
coryan, line 28: shouldn't we use `ToSlash()` here too?  If there is a reason why that won't work (types?) then maybe a comment?

internal/sidekick/downloads_cache.go
```diff
@@ -119,33 +119,30 @@ func downloadAttempt(target, source, expectedSha256 string) error {
...
+		tempFile.Close()
+		os.Remove(tempFile.Name())
+	}()
+
+	hasher := sha256.New()
+	writer := io.MultiWriter(tempFile, hasher)
```
coryan, line 128: TIL `MultiWriter`, thanks.

coryan (changes requested):
A few small nits, but otherwise LGTM. Thanks.

coryan (approved):

PR #1956, opened 2025-09-05
docs: start generating package godoc

codyoss:
Created a means for us to generate the doc.go file based off output from the help text. There is much more doc that still needs to be put into this file but wanted to land this as a proof of concept at least.

Updates: #207


codyoss:
Reopening as I can't reproduce this locally.

Issue #1973, opened 2025-09-08
Updates to generate docs


codyoss:
Should this also be updated since it's no longer for a single API? It can generate for all APIs + Libraries?

_Originally posted by @lqiu96 in https://github.com/googleapis/librarian/pull/1948#discussion_r2326057878_

Also

I think may be helpful to call out. I'm guessing that API a version and Library is a collection of APIs (versions)?

_Originally posted by @lqiu96 in https://github.com/googleapis/librarian/pull/1948#discussion_r2326058042_

PR #1974, opened 2025-09-08
docs(librarian): add some clarifications

codyoss:
Fixes: #1973

### DIFF
--- a/internal/librarian/generate.go
+++ b/internal/librarian/generate.go
@@ -33,7 +33,7 @@ const (
 )
 
 var cmdGenerate = &cli.Command{
-	Short:     "generate generates client library code for a single API",
+	Short:     "generate onboards and generates client library code",
 	UsageLine: "librarian generate [flags]",
 	Long: `The generate command is the primary tool for all code generation
 tasks. It handles both the initial setup of a new library (onboarding) and the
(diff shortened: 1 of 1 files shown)


codyoss:
@lqiu96 I think this addresses the clarifications you were looking for but lmk if I missed anything.

internal/librarian/generate.go
```diff
@@ -58,7 +58,8 @@ Example:
 
 You can regenerate a single, existing library by specifying either the library
 ID or the API path. If no specific library or API is provided, Librarian will
-regenerate all libraries listed in '.librarian/state.yaml'.
+regenerate all libraries listed in '.librarian/state.yaml'. If '--library' or
+'--api' is specified the whole library wil be regenerated.
```
suztomo, line 62:
```suggestion
'--api' is specified the whole library will be regenerated.
```

internal/librarian/generate.go
```diff
@@ -58,7 +58,8 @@ Example:
 
 You can regenerate a single, existing library by specifying either the library
 ID or the API path. If no specific library or API is provided, Librarian will
-regenerate all libraries listed in '.librarian/state.yaml'.
+regenerate all libraries listed in '.librarian/state.yaml'. If '--library' or
+'--api' is specified the whole library will be regenerated.
```
lqiu96, line 62: > If '--library' or '--api' is specified ...

To confirm: it can be either `--library` or `--api` for regeneration?

PR #1979, opened 2025-09-08
feat: log success in multiple library generation

### LINKED ISSUE #1885: librarian(cleanup): consider opening PR if >0 libraries generated
In #1787, we are checking to see if all library generation fails to skip opening the PR. Consider inverting this to open the PR if 1 or more libraries succeeded.

This might be easier to maintain for cases where we skip a library or some other reason other than failure.

JoeWang1127:
Fixes #1885

### DIFF
--- a/internal/librarian/generate.go
+++ b/internal/librarian/generate.go
@@ -166,6 +166,7 @@ func (r *generateRunner) run(ctx context.Context) error {
 		}
 		idToCommits[libraryID] = oldCommit
 	} else {
+		succeededGenerations := 0
 		failedGenerations := 0
 		for _, library := range r.state.Libraries {
 			oldCommit, err := r.generateSingleLibrary(ctx, library.ID, outputDir)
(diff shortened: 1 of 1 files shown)


chingor13 (changes requested):
Is this the behavior we want?

#1885 wasn't asking for a behavior change, but a refactor so that we count successful generations rather than failures (numSuccesses > 0 is easier to verify then numFailures >= numAttempts)

If all generation attempts fail, we probably do want to return a failure.

JoeWang1127:
> https://github.com/googleapis/librarian/issues/1885 wasn't asking for a behavior change, but a refactor so that we count successful generations rather than failures (numSuccesses > 0 is easier to verify then numFailures >= numAttempts)

Got it, I'll make the change.

internal/librarian/generate.go
```diff
@@ -179,7 +179,12 @@ func (r *generateRunner) run(ctx context.Context) error {
 				idToCommits[library.ID] = oldCommit
 			}
 		}
-		slog.Info("generation statistics", "all", len(r.state.Libraries), "failures", failedGenerations)
+
+		slog.Info(
+			"generation statistics",
+			"all", len(r.state.Libraries),
+			"success", len(r.state.Libraries)-failedGenerations,
```
chingor13, line 186: Can we collect successful runs rather than inferring this?

chingor13 (approved):

Issue #1971, opened 2025-09-08
librarian: release init creates release even if there are no conventional commits requiring a release


ldetmer:
If I run 

```release init -repo https://github.com/ldetmer/librarian```
 
It creates a release PR, even though the only changes since the [last release](https://github.com/ldetmer/librarian/releases/tag/v0.1.1) was a docs and an non conventional commit.

1) This should not have created a release since it was only a docs change outside of a generated code change
2) This should not have created a release PR because there were no releasable changes

parthea:
For Python we will raise an error if `librarian release init` is called and there is no version bump https://github.com/googleapis/google-cloud-python/pull/14381

jskeet:
https://github.com/googleapis/librarian/pull/2026 is a draft of what I'd expect we'd want to change. That PR is really only for discussion though - once we've agreed how we want it to behave, we can *at least* add tests, and also potentially refactor. (This was implemented at speed...)

jskeet:
I think we should clarify this issue, because there are two distinct matters:

- On a per-library basis, we shouldn't be calling the container `release` command for a library if it doesn't need to be released
- Overall, we shouldn't create a release PR if there are no libraries to be released

It's not entirely clear to me which of these was intended to be covered in this issue. We should clarify this issue, then create another for the other aspect - *or* make it clear that this issue covers both.

PR #2010, opened 2025-09-09
fix: Skip release if there are no releasable units

lqiu96:
Fixes: https://github.com/googleapis/librarian/issues/1971

### DIFF
--- a/internal/librarian/release_init.go
+++ b/internal/librarian/release_init.go
@@ -281,6 +281,7 @@ func (r *initRunner) updateLibrary(library *config.LibraryState) error {
 
 	if len(commits) == 0 {
 		slog.Info("Skip releasing library as there are no commits", "library", library.ID)
+		return nil
 	}
 
 	nextVersion, err := r.determineNextVersion(commits, library.Version, library.ID)
--- a/internal/librarian/release_init.go
+++ b/internal/librarian/release_init.go
@@ -136,6 +136,13 @@ func (r *initRunner) run(ctx context.Context) error {
 		return err
 	}
 
+	// No need to update the librarian state if there are no libraries
+	// that need to be released
+	if !hasLibrariesToRelease(r.state.Libraries) {
+		slog.Info("No release created; skipping the commit/PR")
+		return nil
+	}
+
 	if err := saveLibrarianState(r.repo.GetDir(), r.state); err != nil {
 		return err
 	}
@@ -163,6 +170,17 @@ func (r *initRunner) run(ctx context.Context) error {
 	return nil
 }
 
+// hasLibrariesToRelease searches through the state of each library and checks
+// that there is a single library configured to be triggered.
+func hasLibrariesToRelease(libraryStates []*config.LibraryState) bool {
+	for _, library := range libraryStates {
+		if library.ReleaseTriggered {
+			return true
+		}
+	}
+	return false
+}
+
 func (r *initRunner) runInitCommand(ctx context.Context, outputDir string) error {
 	dst := r.partialRepo
 	if err := os.MkdirAll(dst, 0755); err != nil {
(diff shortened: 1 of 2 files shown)


internal/librarian/release_init.go
```diff
@@ -117,7 +120,7 @@ func newInitRunner(cfg *config.Config) (*initRunner, error) {
 }
 
 func (r *initRunner) run(ctx context.Context) error {
-	outputDir := r.workRoot
+	outputDir := filepath.Join(r.workRoot, "output")
```
JoeWang1127, line 123: I think this line need to restore.

Context: https://github.com/googleapis/librarian/pull/2015/files

internal/librarian/release_init.go
```diff
@@ -126,19 +129,42 @@ func (r *initRunner) run(ctx context.Context) error {
...
 
-	commitInfo := &commitInfo{
+	var commits []*conventionalcommits.ConventionalCommit
+	for _, library := range r.state.Libraries {
+		commits = append(commits, library.Changes...)
+	}
+
+	// Only proceed with with the release if there are any releasable units
+	if getHighestChange(commits) == semver.None {
+		return nil
```
JoeWang1127, line 139: Add a log here so that we know what happened when no pull request is created.

internal/librarian/command_test.go
```diff
@@ -1400,152 +1370,95 @@ func TestCommitAndPush(t *testing.T) {
...
 			setupMockClient: func(t *testing.T) GitHubClient {
 				return nil
 			},
-			prType:         "generate",
 			push:           true,
 			wantErr:        true,
 			expectedErrMsg: "push error",
 		},
 		{
-			name: "Create PR body error",
```
JoeWang1127, line 1523: Is there an alternative to this test or why it's been removed?

internal/librarian/release_init.go
```diff
@@ -126,19 +129,42 @@ func (r *initRunner) run(ctx context.Context) error {
...
+	}
+
+	datetimeNow := formatTimestamp(time.Now())
+	branch := fmt.Sprintf("librarian-%s", datetimeNow)
+
+	commitInfo := &CommitInfo{
 		cfg:           r.cfg,
 		state:         r.state,
 		repo:          r.repo,
 		sourceRepo:    r.sourceRepo,
```
JoeWang1127, line 149: I don't think source repos is being used here, could we do not assign the value here?

JoeWang1127 (commented):
I think we should focus fixing the issue in this PR and raise another one for the refactor.

Also, since we don't have an end-to-end test for release, could you try to run `librarian release init ...` against this branch in your local env to make sure the librarian can create a pull request?

lqiu96:
> I think we should focus fixing the issue in this PR and raise another one for the refactor.
> 
> Also, since we don't have an end-to-end test for release, could you try to run `librarian release init ...` against this branch in your local env to make sure the librarian can create a pull request?

Yeah good idea. Original implementation required a few fields from across different structs which is where the refactor started. Now that it's simplified, I'll address the refactor in a separate PR. This should be simplified now.

JoeWang1127 (commented):
Looks good, just one comment.

internal/librarian/release_init_test.go
```diff
@@ -414,6 +439,143 @@ func TestInitRun(t *testing.T) {
 			wantErr:    true,
 			wantErrMsg: "failed to fetch conventional commits for library",
 		},
+		{
+			name: "single library has no releasable units, no state change",
```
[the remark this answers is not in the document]
lqiu96: Created a helper function to pass in a mockContainerClient that can be used to count the number of initCalls made (for this test case, it ended up being 0 or 1).

internal/librarian/release_init.go
```diff
@@ -262,16 +274,13 @@ func (r *initRunner) runInitCommand(ctx context.Context, outputDir string) error
...
-	library.PreviousVersion = library.Version
 	commits, err := GetConventionalCommitsSinceLastRelease(r.repo, library)
 	if err != nil {
 		return fmt.Errorf("failed to fetch conventional commits for library, %s: %w", library.ID, err)
 	}
 
-	library.Changes = commits
-	if len(library.Changes) == 0 {
-		slog.Info("Skip releasing library since no eligible change is found", "library", library.ID)
+	if len(commits) == 0 {
```
jskeet, line 282: We shouldn't skip this if we've been asked explicitly to release a single library. See https://github.com/googleapis/librarian/issues/2032

You could potentially pass a Boolean flag into this function as a "force even if there are no commits".

jskeet (commented):
I haven't reviewed the tests yet, as I expect they'll change with the requested production changes.

Other than the comments left, the production changes look good.

internal/librarian/release_init.go
lqiu96: Thanks with Jon offline and we're fine with making the changes in #2030 in a separate PR.

jskeet (approved):
I've only skimmed the tests rather than looking in *huge* amounts of detail - I'm relying on Joe for a more detailed level of test scrutiny.

As noted in chat, we can fix the "we should bump the version if a release was explicitly requested, even if there are no change" in a later PR.

PR #2026, opened 2025-09-10
feat(internal/librarian): DRAFT ONLY: selective releasing

jskeet:
- Only release libraries that have a changed version number
- Don't copy files for libraries that aren't being released
- Don't create a release commit/PR if nothing needs releasing
- Reduce redundancy in loops


jskeet:
Related to #1971 - and I *suspect* there's an issue for not creating a PR if there's nothing to release, but I haven't seen it.

PR #2115, opened 2025-09-12
feat: clean version override after use

### LINKED ISSUE #1704: update config.yaml to suport overriding next release version
For the rare cases where a generated change *should* result in a major version bump, Librarian will support a manual override mechanism. A library maintainer can specify the next desired version in the `.librarian/config.yaml` file.

For example:  
```
  libraries:
    - id: google-cloud-example
      next_version: 2.0.0
```
  During a release, if this configuration is present, Librarian will use `2.0.0` as the version for the `google-cloud-example` library. Librarian will ignore this value if the library's last-released version is already higher than the one specified.

  Alternatively, `librarian release init` command will allow explicitly specifying the release version, when asking to release a single library from the repository. The specified version will take precedence over the one in the `.librarian/config.yaml`.

Ref: go/librarian:commits

### LINKED ISSUE #2043: librarian: clean up `next_version` from config.yaml after usage

### LINKED ISSUE #2116: librarian: release PR updates yaml with non-conforming yaml format
Example: https://github.com/googleapis/librarian/pull/2108

The yaml state file updates do not pass the yamllint formatting tests

chingor13:
Fixes #2043
Fixes #1704
Fixes #2116

### DIFF
--- a/internal/librarian/state.go
+++ b/internal/librarian/state.go
@@ -153,8 +153,8 @@ func findServiceConfigIn(path string) (string, error) {
 
 func saveLibrarianConfig(repoDir string, librarianConfig *config.LibrarianConfig) error {
 	path := filepath.Join(repoDir, config.LibrarianDir, librarianConfigFile)
-	buffer := new(bytes.Buffer)
-	encoder := yaml.NewEncoder(buffer)
+	var buffer bytes.Buffer
+	encoder := yaml.NewEncoder(&buffer)
 	encoder.SetIndent(2)
 	err := encoder.Encode(librarianConfig)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


internal/librarian/state.go
```diff
@@ -150,13 +151,28 @@ func findServiceConfigIn(path string) (string, error) {
 	return "", nil
 }
 
+func saveLibrarianConfig(repoDir string, librarianConfig *config.LibrarianConfig) error {
```
jskeet, line 154: Drive-by comment: does this preserve formatting and comments? I suspect not.

Part of the point of separating state from config was that config would be hand-maintained, and *could* contain comments (which then wouldn't be wiped). I'm concerned that we may lose that as part of this change.

PR #2143, opened 2025-09-15
impl(sidekick): compute changed files

coryan:
Part of the work for #2127


internal/sidekick/internal/rust_release/changes.go
```diff
@@ -0,0 +1,50 @@
...
+// limitations under the License.
+
+package rustrelease
+
+import (
+	"os/exec"
+	"slices"
+	"strings"
+
+	"github.com/go-git/go-git/v5/plumbing/format/gitignore"
```
dbolduc, line 22: comment: good idea to use something we should all know for pattern matching

Issue #2080, opened 2025-09-11
release-init-release.json does not have a properly parsed piper_cl_number


lqiu96:
<img width="512" height="225" alt="Image" src="https://github.com/user-attachments/assets/e119aaf1-9549-41ce-9e63-8f8ad6240690" />

Image above is from one of the latest runs.

The Cl-Number has additional values like the Source-Link when it should be a separate footer value.

lqiu96:
Pending the decision made in https://github.com/googleapis/librarian/issues/2106

zhumin8:
Thanks for gathering these thoughts.

> 1. What splits the header/body/footer sections in a commit?

It is one blank line after the body, according to conventionalcommits [specs](https://www.conventionalcommits.org/en/v1.0.0/#:~:text=One%20or%20more%20footers%20MAY%20be%20provided%20one%20blank%20line%20after%20the%20body) 

> 4. Can we assume that there will only be at max one line break for a footer k/v pair? 

No. But this should not be an issue as the current logic in parseFooters [here](https://github.com/googleapis/librarian/blob/014b5f40a08846785660f2eb33dee6cd86d624df/internal/conventionalcommits/conventional_commits.go#L329) basically effectively collapses one or more empty lines between parts of a footer value.

> 3. Do we need to process all the footers? Or can we just search specifically for those two footers?
The implementation that parses the conventional commits tries to be generic and allow space for future enhancements if needed. If this becomes a problem, we could trim down and omit footers we are not interested in.

Question 2, 3, 5 around repeating footer keys. I will need to look into it a bit more.

I believe the first issue of "footer item not recognized due to a space missing" is fairly straight foward, will raise a PR to fix it first.

lqiu96:
I believe it was from this [commit](https://github.com/googleapis/google-cloud-python/commit/34a7916fa58727a01a7d32c6a1adff59a351fd54) which has generation changes for dlp.

>  I am guessing this is part of onboarding the library?

IIRC, I think dlp was one of the first libraries to be onboarded. In this case, it was generated with owlbot and released with librarian.

> If iiuc, this could be one time situation at the time of onboard, because if generation PR done by Librarian it'd done by template [here](https://github.com/googleapis/librarian/blob/2f52aa7da7d69e41eeb1a4d327b79ab215eae5f9/internal/librarian/release_notes.go#L99-L101). In theory, this can also be avoided if there is no pending merged owlbot pr to be released at time of onboarding?

[...]

PR #2145, opened 2025-09-15
fix: Parse footer key and values on different lines

lqiu96:
Fixes: https://github.com/googleapis/librarian/issues/2080


lqiu96:
Adding test cases to show current behavior fails for footers k/v maps on different lines

Issue #2142, opened 2025-09-15
investigate why the publish-release job takes 6 minutes to run


ldetmer:
This could potentially lead to a race condition where multiple jobs trying to create the same release.

We should optimize this job.

chingor13:
The dispatcher jobs currently clone the librarian repo and then run a `go run` command. This means we are pulling dependencies and compiling code before running the go script. The docker pull (golang image) + compilation takes > 4 minutes.

PR #2151, opened 2025-09-15
build: create dispatcher docker image that prefetches and prebuilds

chingor13:
Towards #2142

After publishing, we will need to update the cloudbuild.yaml files to use the dispatcher image instead of the golang base image.

### DIFF
--- a/cloudbuild-exitgate-dispatcher.yaml
+++ b/cloudbuild-exitgate-dispatcher.yaml
@@ -18,34 +18,16 @@
 # 'images-dev' repository, which serves as the entry point for the AR Exit Gate.
 # After passing the gate's security checks, the image is promoted and
 # published to the 'images-prod' repository.
-#
-# Local test:
-#   gcloud --project=<my-project> builds submit . --config=cloudbuild-test.yaml
 steps:
-  - id: 'build'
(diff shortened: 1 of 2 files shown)


cloudbuild-exitgate.yaml
```diff
@@ -28,11 +29,21 @@ steps:
...
   - id: structure-test
     name: gcr.io/gcp-runtimes/structure_test
+    waitFor: ['build']
     args:
       - '-i'
       - 'us-central1-docker.pkg.dev/cloud-sdk-librarian-prod/images-dev/librarian'
       - '--config'
       - '/workspace/container-structure-test.yaml'
       - '-v'
+  - id: build-dispatcher
```
ldetmer, line 39: yes unfortunately this needs to be  in a separate cloudbuild yaml, as AR exit gate louhi flow only takes 1 image.

ldetmer (changes requested):

cloudbuild-exitgate.yaml
chingor13: Moved to new cloudbuild.yaml

ldetmer (approved):

PR #2152, opened 2025-09-15
impl(sidekick): find cargo manifest files

coryan:
Part of the work for #2127

### DIFF
--- a/internal/sidekick/internal/rust_release/find_cargo_manifests_test.go
+++ b/internal/sidekick/internal/rust_release/find_cargo_manifests_test.go
@@ -31,9 +31,11 @@ func TestFindManifest(t *testing.T) {
 		"src/storage/src/generated/stub.rs",
 		"src/storage/src/generated/client.rs",
 		"src/generated/cloud/secretmanager/v1/src/stub/subdir/file.rs",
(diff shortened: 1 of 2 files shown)


internal/sidekick/internal/rust_release/find_cargo_manifests_test.go
```diff
@@ -0,0 +1,43 @@
...
+package rustrelease
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestFindManifest(t *testing.T) {
+	setupForVersionBump(t, "find-manifest-0.0.1")
```
dbolduc, line 24: consider a test for the case where we have multiple `Cargo.toml` on the path, and we stop at the outermost one.

I am specifically thinking of changes to `echo-server` which should not necessarily lead to a `gax-internal` version bump.

https://github.com/googleapis/google-cloud-rust/tree/main/src/gax-internal/echo-server

dbolduc: unnecessary if `release-plz` knows better than to bump the versions of `gax-internal` in that case

dbolduc (approved):

PR #2157, opened 2025-09-15
fix: add -branch flag to tag-and-release command which defaults to main

### LINKED ISSUE #2053: librarian: error in tag-and-release "remote branch is required when cloning"
Command line:

```sh
LIBRARIAN_GITHUB_TOKEN=$(gh auth token) librarian release tag-and-release -repo=https://github.com/jskeet/google-cloud-go -pr=https://github.com/jskeet/google-cloud-go/pull/7
```

Full log:

```text
025/09/11 08:10:06 INFO librarian arguments="[-repo=https://github.com/jskeet/google-cloud-go -pr=https://github.com/jskeet/google-cloud-go/pull/7]"
2025/09/11 08:10:06 INFO Temporary working directory dir=/tmp/librarian-20250911T081006Z
2025/09/11 08:10:06 INFO Checking for repository dir=/tmp/librarian-20250911T081006Z/google-cloud-go
2025/09/11 08:10:06 gitrepo: remote branch is required when cloning
```

Will investigate further.

chingor13:
Fixes #2053


chingor13:
> This will fix the specific _error_ from #2053, but I don't think it's the right long-term fix. The only thing we need is the pipeline state in the merged commit - and we only need that to get the tag format. If we include the tags to create within the PR itself (or the commit) then we can skip git entirely, and just use the github API.
> 
> If we _really_ want to avoid including the tag, we could fetch the state file that's in the release commit directly from GitHub. (I can find the code in v0.1.0 to do that if it would be helpful.)

Ah, I was trying to fix the error, rather than trying to rethink the implementation of tag and release. It is possible that we completely ignore the repo and just use the API. In that case, we still would want the `-branch` flag, but just use the API rather than a git clone.

chingor13:
Opened #2172 for the refactor

We can skip the -branch flag during this refactor as we'll know the branch from the HEAD branch of the merged PR

PR #2164, opened 2025-09-16
refactor(internal/github): remove error returns from NewClient

julieqiu:
NewClient and newClientWithHTTP previously always returned nil as an error. Remove error as a return value to simplify the API and remove unnecessary error handling throughout the codebase.

### DIFF
--- a/internal/github/github.go
+++ b/internal/github/github.go
@@ -52,16 +52,16 @@ type Client struct {
 }
 
 // NewClient creates a new Client to interact with GitHub.
-func NewClient(accessToken string, repo *Repository) (*Client, error) {
+func NewClient(accessToken string, repo *Repository) *Client {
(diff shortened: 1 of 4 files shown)


internal/github/github.go
```diff
@@ -52,16 +52,16 @@ type Client struct {
...
 }
 
-func newClientWithHTTP(accessToken string, repo *Repository, httpClient *http.Client) (*Client, error) {
+func newClientWithHTTP(accessToken string, repo *Repository, httpClient *http.Client) *Client {
 	return &Client{
 		Client:      github.NewClient(httpClient).WithAuthToken(accessToken),
 		accessToken: accessToken,
 		repo:        repo,
-	}, nil
+	}
```
suztomo, line 64: 👏

PR #2166, opened 2025-09-16
test(librarian): faster cloning

coryan:


jskeet:
Reasons for non-shallow clones:

- For googleapis, we want to get the commit messages since the last commit, when generating
- For language repos, we want to get the commit details to see whether we need to release (when releasing)

I believe there are cases where it *is* okay to do a shallow clone (I think the language repo should be fine for generate, for example) - it's a matter of whether we want to make it configurable and figure out how to configure it for each repo in each eventuality.

chingor13 (approved):

PR #2168, opened 2025-09-16
refactor(internal/librarian): inline logic from newCommandRunner

julieqiu:
newCommandRunner currently builds a shared base runner that is then consumed by initRunner, tagAndReleaseRunner, and generateRunner.

In practice, each of these runners use a different subset of that base. Inline the construction logic into each runner so that each runner declares exactly what it needs.

For https://github.com/googleapis/librarian/issues/2176

### DIFF
--- a/internal/librarian/command.go
+++ b/internal/librarian/command.go
@@ -92,68 +92,6 @@ type commitInfo struct {
 	failedGenerations int
 }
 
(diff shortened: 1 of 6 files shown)


codyoss:
@julieqiu FYI conflicts

PR #2173, opened 2025-09-16
chore: turn of patch report

codyoss:
We don't require this check and it super valuable. We will stil maintain our current coverage goals.


codyoss:
Talked with @julieqiu today about turning this off.

julieqiu:
> We don't require this check and it super valuable. We will stil maintain our current coverage goals.

Assuming this should say "it is not super valuable" :)

julieqiu (approved):

PR #2174, opened 2025-09-16
build: build go binary for automation command

chingor13:
Also pins the dispatcher image to the current version. We will need to update the command args when the new image is built.

Locally, the image size goes from 2.2Gb -> 900Mb

Towards #2142

### DIFF
--- a/infra/prod/generate-worker.yaml
+++ b/infra/prod/generate-worker.yaml
@@ -15,7 +15,7 @@
 # This runs the `librarian generate` command with a provided repository,
 # secret name, and optional library ID
 steps:
-  - name: us-central1-docker.pkg.dev/cloud-sdk-librarian-prod/images-prod/librarian-dispatcher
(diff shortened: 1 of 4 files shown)


infra/prod/generate-worker.yaml
```diff
@@ -15,7 +15,7 @@
 # This runs the `librarian generate` command with a provided repository,
 # secret name, and optional library ID
 steps:
-  - name: us-central1-docker.pkg.dev/cloud-sdk-librarian-prod/images-prod/librarian-dispatcher
+  - name: us-central1-docker.pkg.dev/cloud-sdk-librarian-prod/images-prod/librarian-dispatcher@sha256:00aa886a70e110e9e8cfb888a232a1c508031e58cfeee60870ef7a11ca3d5656
```
ldetmer, line 18: do we need to pin at this point? just to allow for us to validate changes and then update once we think its good?

chingor13: This is temporary until the new image is built. Otherwise, it'll be broken until the new image is published and we update the args in these files.

PR #2178, opened 2025-09-16
feat: Make the Dart class representing the service `final`

brianquinlan:
Dart classes are, by default, also interfaces. This means that adding a new method to a class will break any code that implements it.

The most conservative approach is to make the class `final` - we can relax that restriction and make it `base` later, if it makes sense.

### DIFF
--- a/internal/sidekick/internal/dart/templates/lib/service.mustache
+++ b/internal/sidekick/internal/dart/templates/lib/service.mustache
@@ -17,7 +17,7 @@ limitations under the License.
 {{#Codec.DocLines}}
 {{{.}}}
 {{/Codec.DocLines}}
-class {{Codec.Name}} {
+final class {{Codec.Name}} {
   static const String _host = '{{DefaultHost}}';
 
   final ServiceClient _client;


internal/sidekick/internal/dart/templates/lib/service.mustache
```diff
@@ -17,7 +17,7 @@ limitations under the License.
 {{#Codec.DocLines}}
 {{{.}}}
 {{/Codec.DocLines}}
-class {{Codec.Name}} {
+final class {{Codec.Name}} {
```
devoncarew, line 20: We'll want to figure out how users should test their code, but agree that we can lock things down more tightly for now, and relax later once we have a clear idea for how users will test.

brianquinlan: We can talk about it later but I was thinking that we should generate a set of mocks for each service. And you'd use it like:

```dart
final mock = MockGenerativeService(
  generateContent: (request) {
    if (contents == ...) {
      return Future.value(Candidates(candidates: [...]));
    }
    throw StateError('unexpected request: $request');
  }
)
```

Issue #2032, opened 2025-09-10
if release init command is run with library flag you should create a release regardless of the type of commits


ldetmer:
In the case that release init command is run with -library flag, a release should be created if there are ANY commits associated with that library

lqiu96:
> But we should fail with an error if we're not creating a release, rather than stopping silently.

Just to confirm: only for the this specific case (e.g. using `--library` and no `--version` flag when there are no changes)? Or for all cases when there isn't a release?

jskeet:
@lqiu96: This is definitely only in the scope of `--library` being specified. Within that, I see the following cases:

- `--version` is specified and is the same as the current version (or an earlier version): fail as user error
- `--version` is specified and is a later version: definitely create a release, whether or not there are changes
- `--version` is not specified and there are changes which would cause a SemVer bump of some kind: release
- `--version` is not specified and there are no changes which would create a SemVer bump. Options:
  - Release anyway - my preference
  - Don't release, but fail with an error - Cody's preference
  - Don't release, but don't fail either - I'm arguing this would be a bad thing

PR #2181, opened 2025-09-16
feat: Force release for a specific library even if there are no releasable changes

lqiu96:
Fixes: https://github.com/googleapis/librarian/issues/2032

Changes:
- Split updateLibrary's GetCommits functionality into a separate function for easier testing (inject commits directly without having to setup a repo)

Logic:
If a library is specified for release, force a release for it. Use the library version override if provided as the new version. If the override is not specified, report an error back to the user to use the version flag.

### DIFF
--- a/internal/librarian/release_init.go
+++ b/internal/librarian/release_init.go
@@ -227,7 +227,8 @@ func (r *initRunner) updateLibrary(library *config.LibraryState, commits []*conv
 		slog.Info("Library version override inputted", "currentVersion", library.Version, "inputVersion", r.libraryVersion)
 		nextVersion = semver.MaxVersion(library.Version, r.libraryVersion)
 		slog.Debug("Determined the library's next version from version input", "library", library.ID, "nextVersion", nextVersion)
-		// Inputted version is either equal or less than current version and cannot be used for release
+		// Currently, nextVersion is the max of current version or input version. If nextVersion is equal to the current version,
+		// then the input version is either equal or less than current version and cannot be used for release
 		if nextVersion == library.Version {
 			return fmt.Errorf("inputted version is not SemVer greater than the current version. Set a version SemVer greater than current than: %s", library.Version)
 		}
--- a/internal/librarian/release_init.go
+++ b/internal/librarian/release_init.go
@@ -232,20 +232,16 @@ func (r *initRunner) updateLibrary(library *config.LibraryState, commits []*conv
 	if nextVersion == library.Version {
 		// No library is specified for release. No releasable units found and will not continue with release
 		if r.library == "" {
-			slog.Info("Library does not have any releasable units since the last release and will not be released.",
+			slog.Info("Library does not have any releasable units and will not be released.",
 				"library", library.ID, "version", library.Version)
 			return nil
+		} else {
+			// Library was inputted for release, but a version was not specified or the version was not
+			// SemVer greater than the current version
+			slog.Error("Inputted library does not have a releasable unit and will not be released. Use the --version flag to set a specific version.",
+				"library", library.ID, "nextVersion", nextVersion)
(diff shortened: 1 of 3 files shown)


internal/librarian/release_init.go
```diff
@@ -203,55 +203,68 @@ func (r *initRunner) runInitCommand(ctx context.Context, outputDir string) error
...
+// 3. Set the library's release trigger to true.
+func (r *initRunner) updateLibrary(library *config.LibraryState, commits []*conventionalcommits.ConventionalCommit) error {
+	var nextVersion string
+	// If library version was explicitly set, attempt to use it. Otherwise, try to determine the version from the commits.
+	if r.libraryVersion != "" {
+		slog.Info("Library version override inputted", "currentVersion", library.Version, "inputVersion", r.libraryVersion)
+		nextVersion = semver.MaxVersion(library.Version, r.libraryVersion)
+		slog.Debug("Determined the library's next version from version input", "library", library.ID, "nextVersion", nextVersion)
+		// Inputted version is either equal or less than current version and cannot be used for release
+		if nextVersion == library.Version {
```
JoeWang1127, line 231: Does `==` also work for the `less than` case?

internal/librarian/release_init.go
```diff
@@ -203,55 +203,68 @@ func (r *initRunner) runInitCommand(ctx context.Context, outputDir string) error
...
+		slog.Debug("Determined the library's next version from commits", "library", library.ID, "nextVersion", nextVersion)
+		// Unable to find a releasable unit from the changes
+		if nextVersion == library.Version {
+			// No library was inputted for release. Skipping this library for release
+			if r.library == "" {
+				slog.Info("Library does not have any releasable units and will not be released.", "library", library.ID, "version", library.Version)
+				return nil
+			}
+			// Library was inputted for release, but does not contain a releasable unit
+			slog.Error("Inputted library does not have a releasable unit and will not be released. Use the version flag to force a release.",
```
JoeWang1127, line 250: Same as above, do we need this log?

lqiu96: Will remove, thanks!

internal/librarian/release_init.go
lqiu96: Yes I believe so. nextVersion is set to the max of the current version or the input version. If the input version is set to be less than the current version, then it nextVersion will be equal to `library.Version`

JoeWang1127: Got it, maybe add some comment can help understand the code.

internal/librarian/release_init.go
```diff
@@ -203,55 +203,65 @@ func (r *initRunner) runInitCommand(ctx context.Context, outputDir string) error
...
-//
-// 2. Get the library's commit history in the given git repository.
-//
-// 3. Override the library version if libraryVersion is not empty.
-//
-// 4. Set the library's release trigger to true.
-func (r *initRunner) updateLibrary(library *config.LibraryState) error {
+// processLibrary wrapper to process the library for release. Helps retrieve latest commits
+// since the last release and passing the changes to updateLibrary.
+func (r *initRunner) processLibrary(library *config.LibraryState) error {
```
JoeWang1127, line 208: This additional wrapper seems redundant.

What's the issue of getting commits inside `updateLibrary`?

lqiu96: The previous work to setup `UpdateLibrary`'s test requires creating a gitrepo, configuring the commits, creating files, configuring the source roots to match the file paths, and setting up a tag to retrieve the commits (i.e. Logic inside `GetConventionalCommitsSinceLastRelease()`. 

This small wrapper allows `UpdateLibrary`'s test to take in a list of commits without all the previous setup.

PR #2182, opened 2025-09-16
feat(sidekick): subcommand to publish Rust crates

coryan:


dbolduc (approved):
Only question is if we wanted to check that the branch is in sync with `upstream/main`

coryan:
I am making some more changes, but I think this time I got it right.

> Only question is if we wanted to check that the branch is in sync with `upstream/main`

Strictly "in sync", probably not. We want to allow people to continue merging PRs while the release is published (it can take hours, as you recall).  But "matches one commit in `upstream/main`" probably yes. Let me add that check too.

PR #2205, opened 2025-09-17
chore: librarian generate pull request: 20250917T145545Z

JoeWang1127:
No commit is found since last generation


.librarian/state.yaml
```diff
@@ -3,6 +3,9 @@ libraries:
   - id: librarian
     version: 0.1.1
     last_generated_commit: 97a83d76a09a7f6dcab43675c87bdfeb5bcf1cb5
+    apis: []
```
chingor13, line 6: There is a yaml `omitempty` option that we can set on the `LibrarianState` struct so it won't populate with the empty equivalent. https://pkg.go.dev/gopkg.in/yaml.v2#Marshal

PR #2207, opened 2025-09-17
refactor(librarian): allow dependency injection of GitHub client

jinseopkim0:
Refactors the command runners (generate, release init, tag-and-release) to allow a GitHub client to be injected as a parameter.

This change facilitates testing by enabling the use of mock clients, and involves updating the function signatures in the runners and their call sites in both the main application and the unit tests.

Note: 
I tried to increase code coverage for the codecov failure, but concluded this is not worth the effort. The challenge lies in testing the error paths for `github.ParseRemote` and `github.FetchGitHubRepoFromRemote`. These functions are called after `cloneOrOpenRepo`, which has its own logic for handling remote URLs and local paths that complicates testing the subsequent error conditions in isolation. To test the error path for `github.ParseRemote`, for example, we would need to create a test setup where `cloneOrOpenRepo` succeeds with a URL, but `ParseRemote` fails. This requires workarounds, like creating a local directory that mimics the structure of a cloned repository, which feels more complex than this refactoring warrants.

### DIFF
--- a/internal/librarian/command.go
+++ b/internal/librarian/command.go
@@ -98,7 +98,7 @@ type commandRunner struct {
 
 const defaultAPISourceBranch = "master"
 
-func newCommandRunner(cfg *config.Config) (*commandRunner, error) {
+func newCommandRunner(cfg *config.Config, ghClient GitHubClient) (*commandRunner, error) {
 	languageRepo, err := cloneOrOpenRepo(cfg.WorkRoot, cfg.Repo, cfg.Branch, cfg.CI, cfg.GitHubToken)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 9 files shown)


internal/librarian/command.go
```diff
@@ -98,7 +98,7 @@ type commandRunner struct {
 
 const defaultAPISourceBranch = "master"
 
-func newCommandRunner(cfg *config.Config) (*commandRunner, error) {
+func newCommandRunner(cfg *config.Config, ghClient GitHubClient) (*commandRunner, error) {
```
JoeWang1127, line 101: I don't think this is a good refactor.

Why not creating a command runner directly? You don't have to get a command runner via this function unless `newCommandRunner` is the function under test.

Issue #1013, opened 2025-07-28
librarian: [release] Expand E2E Testing Framework for Release command


bisun-google:
Add new E2E tests for the `release init` and `release tag-and-release` workflows, including verification against golden files.

jinseopkim0:
The current P1 issue will track the building the testing framework and including basic e2e test cases for `release init` and `generate`.

I've created P2 https://github.com/googleapis/librarian/issues/2250 to track building on top of the e2e testing framework to expand the coverage of the tests.

PR #2222, opened 2025-09-18
feat(testing): add e2e test for release tag-and-release

jinseopkim0:
Adds an end-to-end test for the 'release tag-and-release' command. This test uses a mock GitHub server to verify that the command makes the correct API calls for creating tags and releases.

Part of https://github.com/googleapis/librarian/issues/1013.


e2e_test.go
```diff
@@ -522,6 +526,134 @@ END_COMMIT_OVERRIDE
...
+			cmd := exec.Command("go", cmdArgs...)
+			cmd.Env = append(os.Environ(), "LIBRARIAN_GITHUB_TOKEN=fake-token")
+			cmd.Stderr = os.Stderr
+			cmd.Stdout = os.Stdout
+			if err := cmd.Run(); err != nil {
+				if !test.wantErr {
+					t.Fatalf("Failed to run release tag-and-release: %v", err)
+				}
+			}
+			if test.wantErr {
```
JoeWang1127, line 639: Maybe we can remove this part and add it back once we have more failed test cases.

JoeWang1127 (approved):

e2e_test.go
jinseopkim0: SG, removed.

PR #2226, opened 2025-09-18
refactor: Change `annotateModel.imports` into a set

brianquinlan:
The key of `annotateModel.imports` is not used so make the value a key and make the value a bool (as is Golang style for sets).


brianquinlan:
@devoncarew

devoncarew (approved):
> so make the value a key and make the value a bool (as is Golang style for sets).

👍  - good to make this more idiomatic

coryan (approved):

PR #2231, opened 2025-09-18
impl(rust): Add PathTemplate method for URL templates

westarle:
This commit introduces a PathTemplate() method to the pathBindingAnnotation in the Rust code generator. This method is necessary to generate low-cardinality URL templates for OpenTelemetry tracing (e.g. "GET /v1/{name}/databases")

The PathTemplate() method converts the internal path format (e.g., /v1/{name=projects/*/locations/*}/databases) into a format suitable for the url.template attribute in OpenTelemetry spans (e.g., /v1/{name}/databases), as per
  the semantic conventions. This involves replacing the complex segment matches with simple variable placeholders.

  This change is the first step in providing the necessary request-level information from the generated code to the google-cloud-rust HTTP client for enhanced tracing.

See also https://github.com/googleapis/google-cloud-rust/pull/3346
See also https://github.com/googleapis/google-cloud-rust/issues/3239

For #2212


internal/sidekick/internal/rust/annotate_test.go
```diff
@@ -1608,11 +1608,119 @@ func TestPathBindingAnnotations(t *testing.T) {
...
+		{
+			name: "Top-Level Double Wildcard",
+			binding: &pathBindingAnnotation{
+				PathFmt: "/{}",
+				Substitutions: []*bindingSubstitution{
+					{FieldName: "field"},
+				},
+			},
+			want: "/{field}",
+		},
```
dbolduc, line 1690: nit: I'd probably remove these. They don't hurt, but they are duplicates. Some differ only in name. :shrug:

dbolduc (approved):

internal/sidekick/internal/rust/annotate_test.go
westarle: Thank you!  I removed the worst one.  Leaving the others just in case we hit on some boundary condition.

coryan (approved):

PR #2233, opened 2025-09-18
feat: log request.json content at debug level

catchyzheng:


internal/docker/docker.go
```diff
@@ -193,6 +193,14 @@ func (c *Docker) Generate(ctx context.Context, request *GenerateRequest) error {
 		return err
 	}
 	defer func(name string) {
+		if filepath.Base(name) == config.GenerateRequest {
```
codyoss, line 196: Can we add something similar for the release commands too with their input? Also, if we are doing this for the requests I think we should be doing it for the response as well.

PR #2236, opened 2025-09-18
feat: allow configuring tag_format in config.yaml

### LINKED ISSUE #2177: librarian: consider moving `tag_format` to config.yaml
This currently exists in `state.yaml` but really feels like it belongs in `config.yaml`.

chingor13:
Fixes #2177

Moves tag formatting logic into `config` package as this logic is shared amongst other packages and only depends on types from `config`.
Adds the ability to load the librarian config file directly from GitHub


internal/config/librarian_config.go
```diff
@@ -28,14 +28,16 @@ const (
 type LibrarianConfig struct {
 	GlobalFilesAllowlist []*GlobalFile    `yaml:"global_files_allowlist"`
 	Libraries            []*LibraryConfig `yaml:"libraries"`
+	TagFormat            string           `yaml:"tag_format"`
```
JoeWang1127, line 31: Should we change it to `GlobalTagFormat`?

chingor13: I'm not sure: in this case, it's the default for the repository as the per-library config can override it. "Global" sounds like it might be the value no matter what

chingor13:
> Could you resolve the conflicts?

Conflicts are resolved

internal/librarian/commit_version_analyzer.go
```diff
@@ -26,13 +26,11 @@ import (
...
 )
 
-const defaultTagFormat = "{id}-{version}"
-
-// getConventionalCommitsSinceLastRelease returns all conventional commits for the given library since the
+// GetConventionalCommitsSinceLastRelease returns all conventional commits for the given library since the
 // version specified in the state file. The repo should be the language repo.
-func getConventionalCommitsSinceLastRelease(repo gitrepo.Repository, library *config.LibraryState) ([]*conventionalcommits.ConventionalCommit, error) {
-	tag := formatTag(library.TagFormat, library.ID, library.Version)
+func GetConventionalCommitsSinceLastRelease(repo gitrepo.Repository, library *config.LibraryState, tag string) ([]*conventionalcommits.ConventionalCommit, error) {
```
JoeWang1127, line 31:
```suggestion
func getConventionalCommitsSinceLastRelease(repo gitrepo.Repository, library *config.LibraryState, tag string) ([]*conventionalcommits.ConventionalCommit, error) {
```

Issue #1682, opened 2025-08-11
librarian: if one library fails in generation, should revert changes


zhumin8:
In generation logic, when generating all libraries, we allow failure for one library and continue:

https://github.com/googleapis/librarian/blob/4b35724fede5d9f202fdd0bb072580198ce38a70/internal/librarian/generate.go#L183-L186

For failure that occurred in `runGenerateCommand`, prior to "clean and copy" this should work as we expect.
https://github.com/googleapis/librarian/blob/4b35724fede5d9f202fdd0bb072580198ce38a70/internal/librarian/generate.go#L216-L219

However, if failure is in subsequent steps, e.g. in build. __We should expect the code change for the failed library reverted and not included in final worktree/PR.__ Currently, we do not have clean ups, I believe they will end-up in the changes produced.
https://github.com/googleapis/librarian/blob/4b35724fede5d9f202fdd0bb072580198ce38a70/internal/librarian/generate.go#L221-L223

PR #2237, opened 2025-09-18
feat: clean untracked files

JoeWang1127:
Updates #1682

### DIFF
--- a/internal/gitrepo/gitrepo.go
+++ b/internal/gitrepo/gitrepo.go
@@ -46,6 +46,7 @@ type Repository interface {
 	CreateBranchAndCheckout(name string) error
 	Push(branchName string) error
 	Restore(paths []string) error
+	CleanUntracked(paths []string) error
 	pushRefSpec(refSpec string) error
 }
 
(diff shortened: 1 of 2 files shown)


internal/gitrepo/gitrepo.go
```diff
@@ -479,3 +481,33 @@ func (r *LocalRepository) Restore(paths []string) error {
...
+	// Remove an untracked file if it lives within one of the given paths.
+	for _, path := range paths {
+		for file, fileStatue := range statue {
+			if !strings.Contains(file, path) || fileStatue.Worktree != git.Untracked {
+				continue
+			}
+
+			relPath := filepath.Join(r.Dir, file)
+			if err := os.Remove(relPath); err != nil {
+				return fmt.Errorf("failed to remove untracked file, %s: %q", relPath, err)
```
lqiu96, line 507: qq, what makes a file untracked in this case? Would it make sense to potentially leave an untracked file and proceed the rest of the flow?

internal/gitrepo/gitrepo.go
```diff
@@ -479,3 +481,33 @@ func (r *LocalRepository) Restore(paths []string) error {
...
+
+// CleanUntracked removes untracked files within the given paths.
+func (r *LocalRepository) CleanUntracked(paths []string) error {
+	slog.Info("Cleaning untracked files", "paths", strings.Join(paths, ","))
+	worktree, err := r.repo.Worktree()
+	if err != nil {
+		return err
+	}
+
+	statue, err := worktree.Status()
```
lqiu96, line 493: nit: s/statue/status

internal/gitrepo/gitrepo.go
```diff
@@ -479,3 +481,33 @@ func (r *LocalRepository) Restore(paths []string) error {
...
+	if err != nil {
+		return err
+	}
+
+	// Remove an untracked file if it lives within one of the given paths.
+	for _, path := range paths {
+		for file, fileStatue := range statue {
```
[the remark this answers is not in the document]
JoeWang1127: Done.

internal/gitrepo/gitrepo.go
lqiu96: Ok gotcha, I think I understand. Thanks for the clarification!

Last question: Does build run after all the libraries are generated? Or does it run after a single library is generated? Since I recall we do allow failures during generation, would erroring out when cleaning a single library prevent other libraries from generating? If so, is that intended?

Changes LGTM, asking questions for my curiosity. If this is intended, resolve this comment and merge :)

lqiu96 (approved):

Issue #2238, opened 2025-09-18
librarian: configure command trying to read api directory from the remote url.


ohmayr:
```
librarian generate -api=google/privacy/dlp/v2 -library=google-cloud-dlp
```

When running the command above, I'm get the following error where librarian tries to find the api directory in the remote url. I think this behaviour is incorrect and we probably want to open the temp folder that we clone to.

```
librarian generate -api=google/privacy/dlp/v2 -library=google-cloud-dlp
2025/09/18 23:36:05 INFO librarian arguments="[-api=google/privacy/dlp/v2 -library=google-cloud-dlp]"
2025/09/18 23:36:05 INFO Temporary working directory dir=/tmp/librarian-20250918T233605Z
2025/09/18 23:36:05 INFO repo not specified, using current working directory as repo root path=/usr/local/google/home/omairn/git/googleapis/google-cloud-python
2025/09/18 23:36:05 INFO Opening repository dir=/usr/local/google/home/omairn/git/googleapis/google-cloud-python
2025/09/18 23:37:00 INFO Checking for repository dir=/tmp/librarian-20250918T233605Z/googleapis
2025/09/18 23:37:00 INFO Repository not found, executing clone
2025/09/18 23:37:00 INFO Cloning repository url=https://github.com/googleapis/googleapis dir=/tmp/librarian-20250918T233605Z/googleapis
Enumerating objects: 147705, done.
Counting objects: 100% (325/325), done.
Compressing objects: 100% (175/175), done.
Total 147705 (delta 202), reused 150 (delta 150), pack-reused 147380 (from 2)
2025/09/18 23:37:48 INFO library not configured, start initial configuration library=google-cloud-dlp
%s &{us-central1-docker.pkg.dev/cloud-sdk-librarian-prod/images-prod/python-librarian-generator:latest
2025/09/18 23:37:48 failed to read dir "https:/github.com/googleapis/googleapis/google/privacy/dlp/v2": open https:/github.com/googleapis/googleapis/google/privacy/dlp/v2: no such file or directory
```

PR #2239, opened 2025-09-18
fix: read api dir from temp folder

ohmayr:
fixes: https://github.com/googleapis/librarian/issues/2238

### DIFF
--- a/internal/librarian/generate.go
+++ b/internal/librarian/generate.go
@@ -333,7 +333,7 @@ func (r *generateRunner) runBuildCommand(ctx context.Context, libraryID string) 
 // it returns an empty string and an error.
 func (r *generateRunner) runConfigureCommand(ctx context.Context) (string, error) {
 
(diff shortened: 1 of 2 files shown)


internal/librarian/generate.go
```diff
@@ -333,7 +333,7 @@ func (r *generateRunner) runBuildCommand(ctx context.Context, libraryID string)
 // it returns an empty string and an error.
 func (r *generateRunner) runConfigureCommand(ctx context.Context) (string, error) {
 
-	apiRoot, err := filepath.Abs(r.apiSource)
```
ldetmer, line 336: can you check if apiSource is being referenced anywhere else? if its not can you also please remove so there isn't confusion going forward?

ldetmer (changes requested):

internal/librarian/generate.go
ohmayr: addressed.

Issue #1850, opened 2025-01-03
Support ingesting discovery based API schemas


codyoss:
Currently our generator supports ingesting both Protobuf and OpenAPI based schemas to produce clients. There are a couple of core cloud services that are supported best through their discovery document schema -- storage and compute. We should consider if being able to create clients from discovery documents would better help us support such services in the future.

PR #2241, opened 2025-09-19
feat(sidekick): parse discovery doc path templates

coryan:
Discovery doc path templates are specified using URI templates, that is
RFC 6570 templates. This is not the same format as the HTTP annotations
for Protobuf.  Compute only uses a **subset** of the URI template spec,
which allows us to parse the strings and use (without change) the
`api.PathTemplate` structure.

If compute starts using more complex path templates we can grow the
data structures at that time.

Part of the work for #1850


internal/sidekick/internal/parser/discovery/uritemplate.go
```diff
@@ -0,0 +1,123 @@
...
+			segment, width, err = parseExpression(uriTemplate[pos:])
+		} else {
+			segment, width, err = parseLiteral(uriTemplate[pos:])
+		}
+		if err != nil {
+			return nil, err
+		}
+		template.Segments = append(template.Segments, *segment)
+		pos += width
+		if pos == len(uriTemplate) || uriTemplate[pos] != slash {
```
dbolduc, line 60: if it is not the end and not a slash, should we error?

internal/sidekick/internal/parser/discovery/uritemplate.go
```diff
@@ -0,0 +1,123 @@
...
+	return template, nil
+}
+
+func parseExpression(input string) (*api.PathSegment, int, error) {
+	if input == "" || input[0] != beginExpression {
+		return nil, 0, fmt.Errorf("missing `{` character in expression %q", input)
+	}
+	tail := input[1:]
+	if strings.IndexAny(tail, "+#") == 0 {
+		return nil, 0, fmt.Errorf("level 2 expressions unsupposed input=%q", input)
```
dbolduc, line 74: unsupported

internal/sidekick/internal/parser/discovery/uritemplate.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+	if strings.IndexAny(tail, "./?&") == 0 {
+		return nil, 0, fmt.Errorf("level 3 expressions unsupposed input=%q", input)
```
dbolduc, line 77: Same

internal/sidekick/internal/parser/discovery/uritemplate.go
```diff
@@ -0,0 +1,123 @@
...
+	if strings.IndexAny(tail, "=,!@|") == 0 {
+		return nil, 0, fmt.Errorf("reserved character on expression %q", input)
+	}
+	match := identifierRe.FindStringIndex(tail)
+	if match[0] != 0 {
+		return nil, 0, fmt.Errorf("no identifier found on expression %q", input)
+	}
+	id := tail[0:match[1]]
+	tail = tail[match[1]:]
+	if tail == "" || tail[0] != endExpression {
```
dbolduc, line 88: should we also check if `tail[1]` is a slash. We do that in `parseLiteral`.

Would the parser incorrectly accept input like `{foo}}bar` ?

Issue #897, opened 2025-07-17
librarian: design a means of blocking library generation/release


codyoss:
Sometimes, we want to limit the library generation for certain part of a repository while we want to keep the rest generated and released. For example, the language container may not be ready to process certain libraries in google-cloud-python yet.

With https://github.com/googleapis/librarian/pull/2039 _feat: add per-library override configurations to .librarian/config.yaml_ merged recently, we can add the configuration per library there.

This issue is not about having release freeze in the team.

## Cases

Let's call this is "blocked" state of a library.

- For "librarian generate" and "release init" without `--api` or `--library`, Librarian CLI should exclude the libraries with blocked status.
  - `librarian tag-and-release` does not take `--library` option.
- If the library is explicitly specified via `--api` or `--library`, Librarian CLI should generate (or release-init) the library regardless of the blocked state. This allows manual invocation of check while the main branch of the SDK repository still has the library in the blocked status.
- For a hybrid library, it should be possible to mark the library as "blocked" for "release init" only, while the "generate" flow works for it. The maintainer may want to release the library manually by invoking .

## Idea

We now have `id` and `next_version` for per-library configuration (https://github.com/googleapis/librarian/pull/2039). **Let's add `generate_blocked` and `release_blocked` fields (boolean; false by default) in the per-library configuration** to control the behavior.

As the new fields are better maintained by humans rather than CLI/automation, we put it in the `config.yaml` file, rather than the state file.

The logic is going to be implemented in around https://github.com/googleapis/librarian/blob/main/internal/librarian/generate.go#L91 and https://github.com/googleapis/librarian/blob/main/internal/librarian/release_init.go#L136.

### Document

Ensure we have the same level of documentation for the two new fields as we have `next_version`.

suztomo:
Revised the description:

>  Let's add generate_blocked and release_blocked fields (boolean)

suztomo:
Thank you.

PR #2243, opened 2025-09-19
feat: release_blocked field in per-library configuration

suztomo:
Fixes https://github.com/googleapis/librarian/issues/897.

This change implements the "release init" command part (`release_blocked`).

When library config has release_blocked field set to true, the "release init" command skips the library processing. If the library ID is explicitly specified, then `release_blocked` is ignored.

### DIFF
--- a/doc/config-schema.md
+++ b/doc/config-schema.md
@@ -29,8 +29,8 @@ Each object in the `libraries` list represents a single library and has the foll
 |-------------------------|--------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------|----------|-----------------------------------------------------------|
 | `id`           | string | A unique identifier for the library, in a language-specific format. It should not be empty and only contains alphanumeric characters, slashes, periods, underscores, and hyphens. | Yes      | Must be a valid library ID.                               |
 | `next_version` | string | The next released version of the library. Ignored unless it would increase the release version.                                                                                   | No       | Must be a valid semantic version, "v" prefix is optional. |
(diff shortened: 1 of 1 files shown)


doc/config-schema.md
```diff
@@ -30,6 +30,7 @@ Each object in the `libraries` list represents a single library and has the foll
 | `id`           | string | A unique identifier for the library, in a language-specific format. It should not be empty and only contains alphanumeric characters, slashes, periods, underscores, and hyphens. | Yes      | Must be a valid library ID.                               |
 | `next_version` | string | The next released version of the library. Ignored unless it would increase the release version.                                                                                   | No       | Must be a valid semantic version, "v" prefix is optional. |
 | `generate_blocked` | bool | Set this to `true` to skip the generation of this library. It's `false` by default. | No       |  |
+| `release_blocked`  | bool   | Set this to `true` to skip the release of this library. It's `false` by default. | No       |  |
```
lqiu96, line 33: Maybe we can note that this value is ignored if the `--version` flag is set by the user (`--version` flag has priority/ overrides)?

If this is already noted elsewhere that cli flags override the config.yaml file, then ignore this comment

lqiu96 (dismissed):

doc/config-schema.md
suztomo: It's `--library` flag.

suztomo: Updated. PTAL.

PR #2251, opened 2025-09-19
fix: use writeLibraryState in configure

ohmayr:
Fixes: https://github.com/googleapis/librarian/issues/2253

### DIFF
--- a/testdata/e2e_func.go
+++ b/testdata/e2e_func.go
@@ -345,22 +345,20 @@ func validateLibrarianDir(dir, requestFile string) error {
 // readConfigureRequest reads the configure request file and creates a librarianState
 // object.
 func readConfigureRequest(path string) (*librarianState, error) {
-	state := &librarianState{}
+	library := &libraryState{}
 	data, err := os.ReadFile(path)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


ldetmer:
> Let's get clarity from @codyoss on what should be included in the configure-request.json the full state or just the library
> 
> The [design](https://docs.google.com/document/d/1hS1YHaYd_OFqoKGV8xIhub28sBN47tvMmdFzc-SyHcg/edit?resourcekey=0-oq3vLClMKxKxeiClt5W4DQ&tab=t.0#heading=h.4lbhtkvm2hc3) has a single library But the spec has [all libraries](https://github.com/googleapis/librarian/blob/main/doc/language-onboarding.md#configure)

This design was changed in go/librarian:global-file-edits.  @ohmayr will follow up the conversation there to get additional information, but passing in full state file IS intended functionality

codyoss:
I was just about to comment on the linked issue. This is by design. You can check for the `status: new` field. in the json to disambiguate. If we are not properly setting that field that is a bug though. See: https://github.com/googleapis/librarian/blame/4d3c11859977d3e53172aa900fca3fe0b3aa7223/doc/language-onboarding.md#L96

PR #2252, opened 2025-09-19
feat(sidekick): Inject InstrumentationClientInfo for tracing

westarle:
- Add static INSTRUMENTATION_CLIENT_INFO to lib.rs.mustache.
- Use INSTRUMENTATION_CLIENT_INFO in transport.rs.mustache if tracing is enabled.

For #2212 see also https://github.com/googleapis/google-cloud-rust/pull/3347 and https://github.com/googleapis/google-cloud-rust/pull/3376

### DIFF
--- a/internal/sidekick/internal/rust/templates/crate/src/lib.rs.mustache
+++ b/internal/sidekick/internal/rust/templates/crate/src/lib.rs.mustache
@@ -109,6 +109,7 @@ const DEFAULT_HOST: &str = "https://{{Codec.DefaultHost}}/";
 pub(crate) mod info {
     const NAME: &str = env!("CARGO_PKG_NAME");
     const VERSION: &str = env!("CARGO_PKG_VERSION");
+
     lazy_static::lazy_static! {
         pub(crate) static ref X_GOOG_API_CLIENT_HEADER: String = {
             let ac = gaxi::api_header::XGoogApiClient{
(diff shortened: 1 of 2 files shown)


internal/sidekick/internal/rust/templates/crate/src/lib.rs.mustache
```diff
@@ -119,6 +120,13 @@ pub(crate) mod info {
             ac.rest_header_value()
         };
     }
+
+    pub(crate) static INSTRUMENTATION_CLIENT_INFO: gaxi::options::InstrumentationClientInfo = gaxi::options::InstrumentationClientInfo {
+        service_name: "{{Name}}",
+        client_version: VERSION,
+        client_artifact: NAME,
+        default_host: "{{Codec.DefaultHostShort}}",
+    };
```
coryan, line 129: Consider:


```suggestion
    pub(crate) fn instrumentation() -> Option<gaxi::options::InstrumentationClientInfo> {
        Some(gaxi::options::InstrumentationClientInfo {
            service_name: "{{Name}}",
            client_version: VERSION,
            client_artifact: NAME,
            default_host: "{{Codec.DefaultHostShort}}",
        })
    }
```

internal/sidekick/internal/rust/templates/crate/src/transport.rs.mustache
```diff
@@ -53,7 +53,13 @@ impl std::fmt::Debug for {{Codec.Name}} {
 {{/Codec.PerServiceFeatures}}
 impl {{Codec.Name}} {
     pub async fn new(config: gaxi::options::ClientConfig) -> gax::client_builder::Result<Self> {
+        let tracing_is_enabled = gaxi::options::tracing_enabled(&config);
         let inner = gaxi::http::ReqwestClient::new(config, crate::DEFAULT_HOST).await?;
+        let inner = if tracing_is_enabled {
+            inner.with_instrumentation(Some(crate::info::INSTRUMENTATION_CLIENT_INFO))
```
coryan, line 59: Then you can write:



```suggestion
            inner.with_instrumentation(crate::info::instrumentation())
```

As @dbolduc noted elsewhere: a function that receives `Option<T>` but always is called with `Some(x)` seems suspicious.  We need to fix that soon, because changing the API after a release is pain for no good reason.

PR #2259, opened 2025-09-19
impl(sidekick): bindings for discovery methods

coryan:
Parse the `path` field in a discovery method into the `PathInfo` field
of the `api.Method` objects.

Part of the work for #1850

### DIFF
--- a/internal/sidekick/internal/parser/discovery/methods.go
+++ b/internal/sidekick/internal/parser/discovery/methods.go
@@ -16,14 +16,15 @@ package discovery
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/googleapis/librarian/internal/sidekick/internal/api"
 )
 
-func makeServiceMethods(model *api.API, serviceID string, resource *resource) ([]*api.Method, error) {
+func makeServiceMethods(model *api.API, serviceID string, doc *document, resource *resource) ([]*api.Method, error) {
 	var methods []*api.Method
 	for _, input := range resource.Methods {
-		method, err := makeMethod(model, serviceID, input)
+		method, err := makeMethod(model, serviceID, doc, input)
 		if err != nil {
 			return nil, err
 		}
@@ -33,7 +34,7 @@ func makeServiceMethods(model *api.API, serviceID string, resource *resource) ([
 	return methods, nil
 }
 
-func makeMethod(model *api.API, serviceID string, input *method) (*api.Method, error) {
+func makeMethod(model *api.API, serviceID string, doc *document, input *method) (*api.Method, error) {
 	id := fmt.Sprintf("%s.%s", serviceID, input.Name)
 	if input.MediaUpload != nil {
 		return nil, fmt.Errorf("media upload methods are not supported, id=%s", id)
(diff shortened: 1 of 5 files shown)


coryan:
PTAL

internal/sidekick/internal/parser/discovery/services_test.go
```diff
@@ -44,13 +44,39 @@ func TestService(t *testing.T) {
...
 				ID:            "..zones.list",
 				Name:          "list",
 				Documentation: "Retrieves the list of Zone resources available to the specified project.",
 				InputTypeID:   ".google.protobuf.Empty",
 				OutputTypeID:  "..ZoneList",
+				PathInfo: &api.PathInfo{
+					Bindings: []*api.PathBinding{
+						{
+							Verb:         "GET",
+							PathTemplate: api.NewPathTemplate().WithLiteral("compute").WithLiteral("v1").WithLiteral("projects").WithVariable(api.NewPathVariable("project")).WithLiteral("zones"),
```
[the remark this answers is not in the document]
coryan: Done.

internal/sidekick/internal/parser/discovery/services_test.go
```diff
@@ -44,13 +44,50 @@ func TestService(t *testing.T) {
...
+							QueryParameters: map[string]bool{
+								"filter":               true,
+								"maxResults":           true,
+								"orderBy":              true,
+								"pageToken":            true,
+								"returnPartialSuccess": true,
+							},
+						},
+					},
+					BodyFieldPath: "*",
```
dbolduc, line 89: This seems counter to listing the query params. Isn't the body typically empty for GETs?

https://google.aip.dev/127

internal/sidekick/internal/parser/discovery/methods.go
```diff
@@ -46,6 +47,27 @@ func makeMethod(model *api.API, serviceID string, input *method) (*api.Method, e
...
+	if err != nil {
+		return nil, err
+	}
+	binding := &api.PathBinding{
+		Verb:            input.HTTPMethod,
+		PathTemplate:    path,
+		QueryParameters: map[string]bool{},
+	}
+	for _, p := range input.Parameters {
+		if p.Location != "path" {
```
dbolduc, line 67: if all the parameters belong in either the path or as query params, maybe we don't need a body? Not sure.

dbolduc (approved):
I am not sure what is the right thing to do with the Body. I think setting it to `*` can't be that bad.

internal/sidekick/internal/parser/discovery/services_test.go
coryan: I think we already have `*` as the body for GET requests. For example:

https://github.com/googleapis/librarian/blob/2835edabdba6002f8b4f091e456c113b7c3b452a/internal/sidekick/internal/sample/api.go#L167

https://github.com/googleapis/librarian/blob/2835edabdba6002f8b4f091e456c113b7c3b452a/internal/sidekick/internal/parser/protobuf_mixin_test.go#L230

I do not think we need to handle them specially here.

PR #2273, opened 2025-09-20
feat: restore files after build failure

### LINKED ISSUE #1682: librarian: if one library fails in generation, should revert changes

JoeWang1127:
Fixes #1682

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	cloud.google.com/go/iam v1.5.2
 	cloud.google.com/go/longrunning v0.6.7
 	github.com/cbroglie/mustache v1.4.0
-	github.com/go-git/go-git/v5 v5.16.2
+	github.com/go-git/go-git/v5 v5.16.0
 	github.com/google/go-cmp v0.7.0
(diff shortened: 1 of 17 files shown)


codyoss (approved):
Thanks!

PR #2282, opened 2025-09-20
impl(sidekick): capture query and path parameters

### LINKED ISSUE #2267: [sidekick/discovery]: capture request path and query parameters

### LINKED ISSUE #2275: [sidekick/discovery]: query parameters should be optional
Otherwise they are always set to some default value.

coryan:
Capture the query and path parameters of each method into a synthetic
request message.

Fixes #2275 and fixes #2267

### DIFF
--- a/internal/sidekick/internal/api/model.go
+++ b/internal/sidekick/internal/api/model.go
@@ -191,11 +191,30 @@ type Service struct {
 	DefaultHost string
 	// The Protobuf package this service belongs to.
 	Package string
+
+	// Synthetic messages for the parameters and body of each request.
+	//
+	// When using discovery docs or OpenAPI:
+	// 1. Not all methods have a request message.
+	// 2. The request message does not include the path and query parameters.
+	// 3. The request message may be shared, so it is risky to modify it.
+	//
+	// Sidekick creates synthetic messages for these specifications. The codecs
+	// need to generate these messages separate from "normal" messages. They
+	// should keep in mind that name clashes are possible.
+	Requests []*Message
+
 	// The model this service belongs to, mustache templates use this field to
 	// navigate the data structure.
 	Model *API
 	// Language specific annotations
 	Codec any
+}
+
+// HasRequests is used by mustache templates to optional emit code depending on
+// whether a service has synthetic request messages.
+func (service *Service) HasRequests() bool {
+	return len(service.Requests) != 0
 }
 
 // Method defines a RPC belonging to a Service.
--- a/internal/sidekick/internal/parser/discovery/methods.go
+++ b/internal/sidekick/internal/parser/discovery/methods.go
@@ -21,25 +21,33 @@ import (
 	"github.com/googleapis/librarian/internal/sidekick/internal/api"
 )
 
-func makeServiceMethods(model *api.API, serviceID string, doc *document, resource *resource) ([]*api.Method, error) {
+func makeServiceMethods(model *api.API, serviceName, serviceID string, doc *document, resource *resource) ([]*api.Method, []*api.Message, error) {
+	requests := &api.Message{
+		Name:          serviceName,
+		ID:            serviceID,
+		Package:       model.PackageName,
+		Documentation: fmt.Sprintf("Synthetic request messages for %s.", serviceID),
+	}
+	model.State.MessageByID[requests.ID] = requests
+
 	var methods []*api.Method
 	for _, input := range resource.Methods {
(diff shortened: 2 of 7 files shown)


internal/sidekick/internal/parser/discovery/methods.go
```diff
@@ -21,32 +21,68 @@ import (
 	"github.com/googleapis/librarian/internal/sidekick/internal/api"
 )
 
-func makeServiceMethods(model *api.API, serviceID string, doc *document, resource *resource) ([]*api.Method, error) {
+func makeServiceMethods(model *api.API, serviceName, serviceID string, doc *document, resource *resource) ([]*api.Method, []*api.Message, error) {
+	requests := &api.Message{
+		Name:          serviceName,
+		ID:            serviceID,
```
dbolduc, line 27: repeating IDs? :unamused: 

I know why we are doing it. It just gives me the ick.

internal/sidekick/internal/parser/discovery/methods.go
```diff
@@ -21,32 +21,68 @@ import (
 	"github.com/googleapis/librarian/internal/sidekick/internal/api"
 )
 
-func makeServiceMethods(model *api.API, serviceID string, doc *document, resource *resource) ([]*api.Method, error) {
+func makeServiceMethods(model *api.API, serviceName, serviceID string, doc *document, resource *resource) ([]*api.Method, []*api.Message, error) {
+	requests := &api.Message{
```
dbolduc, line 25: if we are going to go with this, can you add comments describing that
- this is a parent message. child message generation goes poorly if there is not a parent.
- it will not be generated; it is just for organization
- we use it so the synthetic requests for a service can be scoped to a module (i.e. `zones::ListRequest` vs. `ZonesListRequest`)

internal/sidekick/internal/api/model.go
```diff
@@ -191,13 +191,32 @@ type Service struct {
...
+	//
+	// When using discovery docs or OpenAPI:
+	// 1. Not all methods have a request message.
+	// 2. The request message does not include the path and query parameters.
+	// 3. The request message may be shared, so it is risky to modify it.
+	//
+	// Sidekick creates synthetic messages for these specifications. The codecs
+	// need to generate these messages separate from "normal" messages. They
+	// should keep in mind that name clashes are possible.
+	Requests []*Message
```
dbolduc, line 205: the alternative is to add a `Synthetic: bool` field to an `api.Message`. And then in generation, we could skip generating serde for any message where `m.Synthetic == true`.

And we can skip generating the message at all when `m.Synthetic == true && !m.Parent`. Or something.

internal/sidekick/internal/parser/discovery/methods.go
```diff
@@ -67,18 +104,29 @@ func makeMethod(model *api.API, serviceID string, doc *document, input *method)
...
-		// TODO(#1850) - handle deprecated methods
+		// TODO(#2268) - handle deprecated methods
 		// Deprecated: ...,
-		InputTypeID:  inputID,
+		InputTypeID:  requestMessage.ID,
 		OutputTypeID: outputID,
 		PathInfo: &api.PathInfo{
 			Bindings:      []*api.PathBinding{binding},
-			BodyFieldPath: "*",
+			BodyFieldPath: bodyPathField,
```
dbolduc, line 129: comment: This seems better.

internal/sidekick/internal/parser/discovery/methods_test.go
```diff
@@ -15,19 +15,32 @@
...
 		t.Fatal(err)
 	}
-
 	id := "..zones.get"
+	// The `model.Messages` is is automatically generated. We want the synthetic
+	// request messages to require ad-hoc generation. Each codec may need to
+	// do something special to avoid clashes.
+	index := slices.IndexFunc(model.Messages, func(a *api.Message) bool { return a.ID == id })
+	if index != -1 {
+		t.Errorf("synthetic request messages should not be in the `model.Messages` list")
```
dbolduc, line 37: I feel like they should be in the list, they should just be tagged as synthetic. :shrug:

dbolduc (commented):
FWIW, I agree with the output of the generation that you shared with me.

The way we are achieving it in the model seems hacky / too clever to me.

I don't want to block progress. I get that the specs don't line up and we are only going to do this hack once*. I think some comments explaining the strategy could help.

*: Well, the next language might get confused when they need to generate their `compute` library.

PR #2285, opened 2025-09-22
feat(internal/librarian): add flag to configure verbose logging

### LINKED ISSUE #2148: librarian: add a means to configure logging verbosity
We have logs at different levels today but don't expose a good hook to make more logs accessible. Maybe this is as easy as a `-v` flag to the CLI that sets the logging level to debug.

zhumin8:
Adds -v flag to the CLI that sets the logging level to debug for each command

Fixes #2148

### DIFF
--- a/internal/librarian/librarian.go
+++ b/internal/librarian/librarian.go
@@ -17,6 +17,8 @@ package librarian
 import (
 	"context"
 	"fmt"
+	"log/slog"
+	"os"
 
 	"github.com/googleapis/librarian/internal/cli"
 )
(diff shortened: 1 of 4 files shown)


internal/librarian/librarian.go
```diff
@@ -65,11 +77,14 @@ func newLibrarianCommand() *cli.Command {
 }
 
 func newCmdGenerate() *cli.Command {
+	var verbose bool
```
codyoss, line 80: Is there someplace more upstream we can make this change so we don't accidently forget to write this code as we add more commands in the future.

zhumin8: Yes, I had a first version done before newLibrarianCommand() here. But it happens before flag parsing and needs manual check on args.
https://github.com/googleapis/librarian/blob/4554b399378313d5dc802c26795ed66e52c9ef6d/internal/librarian/librarian.go#L27-L30

I find it a bit error prone and brittle that it works around the cli framework, and it still needs to add flag to each command for documentation purpose.

codyoss: ack

PR #2322, opened 2025-09-23
feat: allow "." as a valid `source_roots` entry

### LINKED ISSUE #2025: librarian: stop copying so much in release init?
Currently the `release init` command creates a copy of everything from all configured library source paths, into a new directory used as /repo. When we've got a lot of configured libraries, this is going to be a lot of copying.

It's unclear why this is necessary - can't we just mount the language repo as /repo (read-only)? If the idea is that the container mustn't depend on anything that's not in a source root (and not under .librarian) I'm not sure it's worth the constraint.

### LINKED ISSUE #2218: librarian: test behavior for "whole repo is a library"
For single-library repos, the whole repo is the source root. Need to check whether this works (for both release and generate).

chingor13:
Fixes #2218
Fixes #2025

Note that the language containers will need to be smart about where they decide to put files.

An example librarian state.yaml (for googleapis/gax-go):

```yaml
image: us-central1-docker.pkg.dev/cloud-sdk-librarian-prod/images-prod/librarian-go:latest
libraries:
  - id: gax
    version: 2.15.0
    last_generated_commit: f8776fec04e336527ba7279d960105533a1c4e21
    source_roots:
      - .
    tag_format: v{version}
```

### DIFF
--- a/internal/conventionalcommits/conventional_commits.go
+++ b/internal/conventionalcommits/conventional_commits.go
@@ -201,7 +201,6 @@ func parseSimpleCommit(commitPart commitPart, commit *gitrepo.Commit, libraryID 
 	if trimmedMessage == "" {
 		return nil, fmt.Errorf("empty commit message")
 	}
-	slog.Info("commit", "message", trimmedMessage)
 
 	lines := strings.Split(trimmedMessage, "\n")
 	bodyLines, footerLines := separateBodyAndFooters(lines)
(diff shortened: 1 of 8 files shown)


chingor13:
> Wait for [go/sdk-librarian:single-library-configuration](http://goto.google.com/sdk-librarian:single-library-configuration) approval

Design is approved, working to clean this up.

PR #2329, opened 2025-09-23
fix(internal/librarian): fix release_exclude_paths logic

### LINKED ISSUE #2307: librarian: release_exclude_paths logic will include commits incorrectly
The documentation around `release_exclude_paths` states:

> If all files from commit belong to one of the paths it will be skipped.

But that's not what we really need - we need them to not count towards a commit for the library in which they're excluded.

Concrete example state file:

```yaml
image: xyz
libraries:
  - id: libx
    source_roots:
      - x
    release_exclude_paths:
      - x/nested
```

Now consider a commit with files of:

- README.md
- x/nested/change.txt

That *will* be found when looking for commits for `libx`, because `x/nested/change.txt` is under `x`. It then *won't* be excluded by the current logic, because `README.md` isn't under `x/nested`.

The logic should be "are there any files in the commit which *are* under any source_root entries, but *aren't* under any release_exclude_paths entries".

zhumin8:
Change exclude logic from "If all files from commit belong to one of the paths it will be skipped." to "Files that belong to one of the paths in the exclusion list should not count towards a commit for the library in which they're excluded".

inverted shouldExclude() to shouldInclude() for clarity, added test case.
Some tests in release_init_test.go and release_notes_test.go needed changes because now SourceRoots are needed in exclude logic and it cannot be empty (should not be empty anyway according to schema), added to mocks.

Fixes #2307


internal/librarian/commit_version_analyzer.go
```diff
@@ -85,22 +85,38 @@ func convertToConventionalCommits(repo gitrepo.Repository, library *config.Libra
...
-func shouldExclude(files, excludePaths []string) bool {
+// shouldInclude determines if a commit should be included in a release.
+// It returns true if there is at least one file in the commit that is under a source_root
+// and not under a release_exclude_path.
+func shouldInclude(files, sourceRoots, excludePaths []string) bool {
 	for _, file := range files {
-		excluded := false
+		var isUnderSourceRoot bool
+		for _, sourceRoot := range sourceRoots {
+			if strings.HasPrefix(file, sourceRoot) {
```
jskeet, line 95: Drive-by comment: leading substrings can be tricky.

Imagine two libraries, one with a source root of "ai" and with a source root of "aiplatform". We don't want aiplatform commits to count as ai commits. Perhaps we should add a "/" to the source root here before testing the prefix?

It's unclear whether that should also be done for exclude paths. Maybe do it for both, but only if the value in question doesn't already end with a slash? It sort of depends on what we expect the values to be - for source roots, they're always directories, but I can imagine exclusion paths being something like foo/bar/go.mod... so maybe we can say that it's up to the state file populator (e.g. configure) to make sure that a simple prefix match works for exclusion paths, but for source roots it should just be the directory with no trailing slash?

jskeet: (Let me know if that isn't clear, and I can give more concrete examples and/or test cases. But we should document exactly what's expected, too.)

internal/config/state.go
```diff
@@ -127,7 +127,7 @@ type LibraryState struct {
 	// A more specific `preserve_regex` takes precedence.
 	RemoveRegex []string `yaml:"remove_regex" json:"remove_regex"`
 	// Path of commits to be excluded from parsing while calculating library changes.
-	// If all files from commit belong to one of the paths it will be skipped.
+	// Files matching these paths will not be considered part of a commit for this library.
```
jskeet, line 130: This is in doc/state-schema.md as well.

doc/state-schema.md
```diff
@@ -24,7 +24,7 @@ Each object in the `libraries` list represents a single library and has the foll
...
 | `preserve_regex`        | list   | A list of regular expressions for files and directories to preserve during the copy and remove process.                                                                    | No       | Each entry must be a valid regular expression. |
 | `remove_regex`          | list   | A list of regular expressions for files and directories to remove before copying generated code. If not set, this defaults to the `source_roots`. A more specific `preserve_regex` takes precedence. | No       | Each entry must be a valid regular expression. |
-| `release_exclude_paths` | list   | A list of directories to exclude from the release.                                                                                                                    | No       | Each entry must be a valid directory path.     |
+| `release_exclude_paths` | list   | A list of directories to exclude from the release. Files matching these paths will not be considered part of a commit for this library.                                                                                                                    | No       | Each entry must be a valid directory path.     |
```
jskeet, line 27: Perhaps change from "A list of directories" to "A list of paths" and remove the "Each entry must be a valid directory path" to "Each entry must be a valid directory file/path"?

(Basically I think we need it to be clear that it's okay to have a path to a file there. Most entries probably *will* be directories, but it's useful to have both as options.)

jskeet (commented):
A couple of clarifications to docs, but I'm pretty confident in the production code - thanks for all the tests :)

internal/librarian/commit_version_analyzer_test.go
```diff
@@ -27,49 +27,126 @@ import (
...
 )
 
-func TestShouldExclude(t *testing.T) {
+func TestShouldInclude(t *testing.T) {
 	t.Parallel()
 	for _, test := range []struct {
 		name         string
 		files        []string
+		sourceRoots  []string
 		excludePaths []string
```
JoeWang1127, line 36: Does it make sense to change `excludePaths` to `includePaths`?

internal/librarian/commit_version_analyzer_test.go
```diff
@@ -27,49 +27,126 @@ import (
...
 		want         bool
 	}{
 		{
-			name:         "no exclude paths",
+			name:         "file in source root, not excluded",
```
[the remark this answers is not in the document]
JoeWang1127: I prefer updating test names in an incremental manor. But if you feel against it, it's fine.

internal/librarian/commit_version_analyzer_test.go
```diff
@@ -27,49 +27,126 @@ import (
...
+			name:         "all files not in source root",
+			files:        []string{"x/y/c.go", "w/z/e.go"},
+			sourceRoots:  []string{"a"},
+			excludePaths: []string{},
 			want:         false,
 		},
 		{
-			name:         "one file in exclude path, one not",
-			files:        []string{"a/b/c.go", "d/e/f.go"},
+			name:         "some files not in source root",
```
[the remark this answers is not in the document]
zhumin8: updated in [53a3746](https://github.com/googleapis/librarian/pull/2329/commits/53a37469ffb536da4f3cef7c7ae3f4756900b466)

internal/librarian/commit_version_analyzer_test.go
```diff
@@ -27,49 +27,110 @@ import (
...
 		{
-			name:         "one file in exclude path, one not",
-			files:        []string{"a/b/c.go", "d/e/f.go"},
+			name:         "all files not in source root",
+			files:        []string{"x/y/c.go", "w/z/e.go"},
+			sourceRoots:  []string{"a"},
+			excludePaths: []string{},
+		},
+		{
+			name:         "a file not in source root and a file in exclude path",
```
JoeWang1127, line 85: Can you add a test of one file in source root but not in excluded path, one file in excluded path and one file outside of source root?

zhumin8: done [ac612f5](https://github.com/googleapis/librarian/pull/2329/commits/ac612f59ab917e645b4a014219db32f87e951ea6)

JoeWang1127 (approved):

PR #2336, opened 2025-09-24
cleanup(sidekick): OpenAPI uses synthetic mesages

### LINKED ISSUE #2281: [sidekick]: cleanup OpenAPI implementation to use synthetic requests
The implementation of `OpenAPI` adds synthetic fields to the request messages, creating the message if needed. Then each codec needs to do something funky to skip those fields in requests.  It is particularly ugly because sometimes the request message does not exist, and sometimes it is shared by two methods.

I am about to use a different approach in the discovery docs: generate synthetic requests, unique to each method.  The codec can generate these requests in a different namespace, without implementing serialization for them.

The templates and generated code are cleaner with the new approach.

coryan:
That is, the messages are labeled as synthetic. And it does not change
the existing messages: only new messages are created.

Fixes #2281

### DIFF
--- a/internal/sidekick/internal/parser/openapi_test.go
+++ b/internal/sidekick/internal/parser/openapi_test.go
@@ -523,7 +523,8 @@ func TestOpenAPI_MapInteger(t *testing.T) {
 	})
 }
 
-func TestOpenAPI_MakeAPI(t *testing.T) {
+func openapiSecretManagerAPI(t *testing.T) *api.API {
+	t.Helper()
 	contents, err := os.ReadFile("../../testdata/openapi/secretmanager_openapi_v1.json")
 	if err != nil {
 		t.Fatal(err)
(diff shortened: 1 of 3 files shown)


internal/sidekick/internal/parser/openapi_test.go
```diff
@@ -813,34 +838,76 @@ func TestOpenAPI_SyntheticMessageWithExistingRequest(t *testing.T) {
...
 	}
-	apitest.CheckMessage(t, setIamPolicyRequest, &api.Message{
-		Name:          "SetIamPolicyRequest",
-		ID:            "..SetIamPolicyRequest",
-		Documentation: "Request message for `SetIamPolicy` method.",
+	apitest.CheckMessage(t, got, want)
+
+	// Methods that share a body should create separate requests.
+	want = &api.Message{
+		Name:             "SetIamPolicyByProjectAndLocationAndSecretRequest",
```
dbolduc, line 856: oof

dbolduc (approved):

PR #2338, opened 2025-09-24
fix(config): race condition in createWorkRoot()

coryan:
Creating a temporary directory based on a timestamp is inherently racy.
Use the standard functions to create temporary directories, and relax
the tests to check for what matters.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -24,7 +24,6 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
-	"time"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


codyoss (approved):
Thank you!

jskeet:
I've personally found it useful to have timestamped work directories - it makes it easier to find which directory was associated with a run (I know it's logged, but that can be tricky to find with lots of logs), and easier to find the most recent one just with tab completion and inspection.

Not a blocker - just thought I'd note that this change will have a negative impact as well as any positive one.

PR #2355, opened 2025-09-25
refactor(sidekick): directly add fields to message

coryan:
We will want to add child types to a message, such as enums and inline
messages. Consuming the message object in `makemessageFields()` allows
us to add types and fields as we parse the fields.

Part of the work for #2269

### DIFF
--- a/internal/sidekick/internal/parser/discovery/message_fields.go
+++ b/internal/sidekick/internal/parser/discovery/message_fields.go
@@ -20,24 +20,23 @@ import (
 	"github.com/googleapis/librarian/internal/sidekick/internal/api"
 )
 
-func makeMessageFields(model *api.API, messageID string, schema *schema) ([]*api.Field, error) {
-	var fields []*api.Field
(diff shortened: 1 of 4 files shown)


internal/sidekick/internal/parser/discovery/message_fields.go
```diff
@@ -47,38 +46,33 @@ func makeField(model *api.API, messageID string, input *property) (*api.Field, e
...
-		JSONName:      input.Name, // OpenAPI field names are always camelCase
-		Documentation: input.Schema.Description,
-		Typez:         typez,
-		TypezID:       typezID,
-		Repeated:      true,
-		// TODO(#2268) - deprecated fields?
-	}, nil
+	field.Documentation = input.Schema.Description
+	field.Repeated = true
+	field.Optional = false
```
dbolduc, line 63: style comment: I think it is slightly cleaner as it was vs. using `makeScalarField` and having to override values.

No changes needed, just giving an opinion.

dbolduc (approved):

PR #2365, opened 2025-09-25
fix(sidekick): rustify accessors for bindings

coryan:
The accessors may need to be escaped if they are a rust keyword, and
may have undergone a style transformation.

Part of the work for #1850


dbolduc (approved):
Thanks!

PR #2368, opened 2025-09-25
feat(sidekick): Generate setter samples for repeated and map fields

amanda-tarafa:


amanda-tarafa:
Accompaniying PR: https://github.com/googleapis/google-cloud-rust/pull/3409

@coryan and reviewers, a couple of notes around maps:

- Samples for maps are as supported by protobuf, that is, keys are only strings, integers-like including bool. I'm not sure if the other specs that we are/ will be supporting allow for other key types. I'm happy to add more cases.
- I've added samples for a map with bool keys, that's unlikely to be ever used, but it's allowed, so...

coryan:
> 
> * Samples for maps are as supported by protobuf, that is, keys are only strings, integers-like including bool. I'm not sure if the other specs that we are/ will be supporting allow for other key types. I'm happy to add more cases.

The other specs are more restrictive than Protobuf, so far, only strings.

> * I've added samples for a map with bool keys, that's unlikely to be ever used, but it's allowed, so...

👍

amanda-tarafa (commented):
Change requests addressed on latest commit.

coryan (approved):

PR #2386, opened 2025-09-26
fix(sidekick): missing IDs in discovery doc fields

coryan:
Part of the work for #1850


internal/sidekick/internal/parser/discovery/message_fields.go
```diff
@@ -74,6 +74,7 @@ func makeScalarField(model *api.API, message *api.Message, name string, schema *
 	return &api.Field{
 		Name:          name,
 		JSONName:      name, // OpenAPI field names are always camelCase
```
dbolduc, line 76: nit: for another PR, remove references to OpenAPI in discovery parser

PR #2388, opened 2025-09-26
fix(internal/conventionalcommits): relax footer regex to allow line broke footers to be properly recognized

zhumin8:
The original logic expect footer to follow `key: value` (exactly one space before value) format as conventionalcommit specs. For multi-line footers, when parsing ([logic](https://github.com/googleapis/librarian/blob/9b28aa4a3619579dfa59c430afa612589b353598/internal/conventionalcommits/conventional_commits.go#L336)), the line content "Source-Link:" would fail the regex match and thus not properly recognized as footer, and gets appended to the value of the previous footer.

By relaxing the regex a bit to allow any number of spaces before "value" in footer, the multiline footer can be recognized. 

note: this format below from owlbot PR ([example](https://github.com/googleapis/google-cloud-python/commit/34a7916fa58727a01a7d32c6a1adff59a351fd54)) is technically not valid footer according to conventionalcommits.org spec [here](https://www.conventionalcommits.org/en/v1.0.0/#:~:text=Each%20footer%20MUST%20consist%20of%20a%20word%20token%2C%20followed%20by%20either%20a%20%3A%3Cspace%3E%20or%20%3Cspace%3E%23%20separator%2C%20followed%20by%20a%20string%20value). But we are relaxing to allow it for onboarding libraries.
```
Source-Link:
googleapis/googleapis@d300b15
```

This should solve the first issue causing this bug, see [comment](https://github.com/googleapis/librarian/issues/2080#issuecomment-3321551055)

For #2080

### DIFF
--- a/internal/conventionalcommits/conventional_commits_test.go
+++ b/internal/conventionalcommits/conventional_commits_test.go
@@ -674,19 +674,19 @@ func TestParseFooters(t *testing.T) {
 		{
 			name: "multiple footers with key on one line, value on the next",
 			footerLines: []string{
-				"PiperOrigin-RevId: 802200836",
+				"PiperOrigin-RevId: 123456",
 				"",
 				"Source-Link:",
 				"",
-				"googleapis/googleapis@d300b15",
+				"googleapis/googleapis@a12b345",
 				"",
 				"",
 				"Copy-Tag:",
 				"eyJwIjoic",
 			},
 			wantFooters: map[string]string{
-				"PiperOrigin-RevId": "802200836",
-				"Source-Link":       "\ngoogleapis/googleapis@d300b15",
+				"PiperOrigin-RevId": "123456",
+				"Source-Link":       "\ngoogleapis/googleapis@a12b345",
 				"Copy-Tag":          "\neyJwIjoic"},
 		},
 	} {
--- a/internal/conventionalcommits/conventional_commits_test.go
+++ b/internal/conventionalcommits/conventional_commits_test.go
@@ -672,7 +672,7 @@ func TestParseFooters(t *testing.T) {
 			},
 		},
 		{
-			name: "multiple footers with key on one line, value on the next",
+			name: "multi-line footers with key on one line, value on the next",
 			footerLines: []string{
 				"PiperOrigin-RevId: 123456",
 				"",
--- a/internal/conventionalcommits/conventional_commits_test.go
+++ b/internal/conventionalcommits/conventional_commits_test.go
@@ -621,7 +621,7 @@ func TestConventionalCommit_MarshalJSON(t *testing.T) {
 }
 
 func TestParseFooters(t *testing.T) {
-	for _, tt := range []struct {
+	for _, test := range []struct {
 		name           string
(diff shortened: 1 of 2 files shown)


internal/conventionalcommits/conventional_commits_test.go
```diff
@@ -619,3 +619,85 @@ func TestConventionalCommit_MarshalJSON(t *testing.T) {
...
+				"Reviewed-by":    "G. Gemini",
+				"Co-authored-by": "Another Person <another@person.com>",
+			},
+		},
+		{
+			name: "multi-line footers with key on one line, value on the next",
+			footerLines: []string{
+				"PiperOrigin-RevId: 123456",
+				"",
+				"Source-Link:",
```
codyoss, line 679: I don't think this is a valid trailer/footer. See https://git-scm.com/docs/git-interpret-trailers for examples of how to do multi line data if we really need to parse it.

zhumin8: > I don't think this is a valid trailer/footer.

Are you referring to this multi-line footer format? I agree with you if we go strict on definitions, however 
1. conventionalcommits.org and git trailers specs are slightly different. according to conventionalcommits.org [specs](https://www.conventionalcommits.org/en/v1.0.0/#:~:text=Each%20footer%20MUST%20consist%20of%20a%20word%20token%2C%20followed%20by%20either%20a%20%3A%3Cspace%3E%20or%20%3Cspace%3E%23%20separator%2C%20followed%20by%20a%20string%20value) `Each footer MUST consist of a word token, followed by either a :<space> or <space># separator, followed by a string value`
git page has slightly different requirements for trailer [here](https://git-scm.com/docs/git-interpret-trailers#:~:text=When%20reading%20trailers,newlines%20in%20it.) that is a bit more relax on the whitespace around colon but specifies for the <value> may be split over multiple lines with each subsequent line starting with at least one whitespace.

2. we are dealing with pass commit messages like [this one](https://github.com/googleapis/google-cloud-python/commit/176510830a231c3d19306218663d6f5d365f39be), that has footers in this format. My assumption so far is that we need to be able to parse these correctly even if they don't strictly follow conventionalcommits.
However, re-thinking now, and maybe if iiuc this situation is onetime and avoidable during onboarding only? But it still be annoying enough for the onboarding process for each library.  WDYT? https://github.com/googleapis/librarian/issues/2080#issuecomment-3339959040

zhumin8: I agree with you in general.
>  we should check that this can still happen in our librarian world

I believe they won't once *fully* onboarded to Librarian. But with per-library onboarding, this would be annoying if we don't support. I will leave a comment on this and so we can remove and get back to standard spec once all languages are onboarded.

> Copy tag

I don't think we care about it as long as it does not produce error. However, for "Source-Link:" we want to capture the first one from googleapis when there are multiple like [this owlbot commit](https://github.com/googleapis/google-cloud-python/commit/176510830a231c3d19306218663d6f5d365f39be). I will put up another PR for a quick fix.

Issue #2339, opened 2025-09-24
librarian: SSH remotes are not well supported


codyoss:
Yesterday I was trying to test the `librarian release init -push` command and it was failing with auth issues until a created a fresh HTTP git remote for the repo I was working in. I believe the culprit area of the code is here: https://github.com/googleapis/librarian/blob/f2d1a10c352f5fccde1c8fa92e0414b9abd37b85/internal/gitrepo/gitrepo.go#L449-L457

It would be great if we can better support SSH remotes.

lqiu96:
I believe this is similar to: https://github.com/googleapis/librarian/issues/1943. 

Last update for that was that it was WAI as it's documented, but I would love for proper support for SSH remotes.

codyoss:
I agree it is similar for sure, and maybe I just did not understand that issue well enough at the time. I think the difference here though is I am not specifying a repo flag and I expect my local remote to just work.

PR #2397, opened 2025-09-26
feat: Support git push via SSH

lqiu96:
Fixes: https://github.com/googleapis/librarian/issues/2339

## Changes
- Automatically determine if HTTPS or SSH should be used based on the remote URI
- SSH is configured via go-git transport using ssh-agent. Requires users to upload their private keys to ssh-agent so that it can be picked up. SSH-Agent is used so that the private key path + passphrase doesn't need to passed in via env vars or flags.

## Result from running cloudtop
Command used: `librarian release init -push`. Resulted in PR created: https://github.com/googleapis/librarian/pull/2449

Logs from invocation:
```
level=INFO msg="Authenticating with SSH"
level=INFO msg="Successfully pushed changes"
level=INFO [msg="Creating PR" branch=librarian-20250926T172344Z base=main title="chore: librarian release pull request: 20250926T172344Z"](https://github.com/googleapis/librarian/pull/2449)
level=INFO msg="PR created" url=https://github.com/googleapis/librarian/pull/2449
level=INFO msg="Labels added to issue" number=2449 labels=[release:pending]
```

### DIFF
--- a/internal/gitrepo/gitrepo.go
+++ b/internal/gitrepo/gitrepo.go
@@ -479,11 +479,7 @@ func (r *LocalRepository) pushRefSpec(refSpec string) error {
 		}
 	}
 
-	useSSH, err := canUseSSH(remoteURI)
-	if err != nil {
-		return err
-	}
-
+	useSSH := canUseSSH(remoteURI)
 	// While cloning a public repo does not require any authCreds, pushing
 	// to the repo requires authentication and verification of identity
 	auth, err := r.authCreds(useSSH)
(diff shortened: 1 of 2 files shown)


internal/gitrepo/gitrepo.go
```diff
@@ -87,6 +90,8 @@ type RepositoryOptions struct {
 	CI string
 	// GitPassword is used for HTTP basic auth.
 	GitPassword string
+	// UseSSH flag to determine if HTTPS (default) or SSH will be used for git operations
+	UseSSH bool
```
codyoss, line 94: Instead of adding configuration to do this how about we just detect it? I think we can make use of `Remotes()` and maybe have a private method to check what type of remote it is. If the URL looks like an ssh we setup that auth, otherwise do what we have been doing. Let me know if I am missing something though where that won't be possible.

lqiu96: Updated!

PR #2405, opened 2025-09-26
fix(internal/librarian): should not push to github when no push flag specified

### LINKED ISSUE #2390: librarian: -commit still tries to push to GitHub
If neither `-commit` nor `-push` are specified, the `commitAndPush` function (in command.go) correctly stops immediately.

However, if `-commit` is specified but not `-push`, it still tries to push to GitHub. It doesn't then try to create a PR, but it shouldn't even push IMO.

The documentation is somewhat ambiguous (and has a typo):

```go
// Commit determines whether to creat a commit for the release but not create
// a pull request.
```

Given that the flag for "push and create a PR" is `-push` rather than `-pr`, I really wouldn't expect it to push.

Assigning to Cody to validate that this really *isn't* the intended behavior; someone else can then fix, of course.

zhumin8:
now when commit = true and push = false, it does not push to GitHub. Updated test case to verify.

Fixes #2390

### DIFF
--- a/internal/librarian/command.go
+++ b/internal/librarian/command.go
@@ -356,13 +356,13 @@ func commitAndPush(ctx context.Context, info *commitInfo) error {
 		return err
 	}
 
-	if err := repo.Push(branch); err != nil {
-		return err
-	}
-
 	if !info.push {
 		slog.Info("Push flag is not specified, skipping pull request creation")
 		return nil
+	}
+
+	if err := repo.Push(branch); err != nil {
+		return err
 	}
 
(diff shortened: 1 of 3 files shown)


jskeet (approved):
Looks fine to me - the prod change is exactly as expected, and the test change looks okay to me.

I'll add a writePRBody call as per https://github.com/googleapis/librarian/commit/862c7d727fed8fa1637c808a8ab48e37e01521ef after this is merged, for the "commit but push" path... I *thought* I'd done that already, but apparently not...

PR #2410, opened 2025-09-27
feat(sidekick): enum serialization for non-proto sources

### LINKED ISSUE #2406: [sidekick/discovery]: serialize enums as strings

coryan:
Change the serialization of enums for non-proto sources (discovery docs
and OpenAPI) to use strings. The main motivation to use integers in the
case of Protobuf is roundstrips where we receive an unknown enum from
a Protobuf / gRPC response, copy that to a request, and want to
serialize it without loss. That scenario is not a consideration for
OpenAPI or Discovery docs, and serializing enums as integers just does
not work.

Fixes #2406

### DIFF
--- a/internal/sidekick/internal/rust/codec.go
+++ b/internal/sidekick/internal/rust/codec.go
@@ -53,25 +53,29 @@ var commentUrlRegex = regexp.MustCompile(
 
 func newCodec(protobufSource bool, options map[string]string) (*codec, error) {
 	var sysParams []systemParameter
+	var serializeEnumsAsStrings bool
 	if protobufSource {
 		sysParams = append(sysParams, systemParameter{
 			Name: "$alt", Value: "json;enum-encoding=int",
 		})
+		serializeEnumsAsStrings = false
 	} else {
 		sysParams = append(sysParams, systemParameter{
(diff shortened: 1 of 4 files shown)


internal/sidekick/internal/rust/codec.go
```diff
@@ -53,25 +53,29 @@ var commentUrlRegex = regexp.MustCompile(
...
-		releaseLevel:     "preview",
-		systemParameters: sysParams,
+		generationYear:          fmt.Sprintf("%04d", year),
+		modulePath:              "crate::model",
+		extraPackages:           []*packagez{},
+		packageMapping:          map[string]*packagez{},
+		version:                 "0.0.0",
+		releaseLevel:            "preview",
+		systemParameters:        sysParams,
+		serializeEnumsAsStrings: serializeEnumsAsStrings,
```
dbolduc, line 78: nit: could say `!protobufSource` here

dbolduc (approved):

internal/sidekick/internal/rust/codec.go
coryan: Done. PTAL.

Issue #2389, opened 2025-09-26
design: simplify maintainer flow for hand-edited release PRs


jskeet:
Currently the details for the release PR are only created with the `-push` flag.

That means if a library maintainer wants to edit the release notes, they either have to work out exactly what `librarian` *would* have done, or push a PR, then fetch the branch and modify it.

I don't think this needs to block migration, but we should have a simpler process soon.

Possible remedies:

- When `-push` isn't specified, log what the PR title and description *would* have been so that maintainers can copy/paste it
- Use the commit message instead of the PR title/description, so that maintainers can use the `-commit` flag, then create another commit afterwards

codyoss:
Let's just do it!

jskeet:
Sticking point: this is naturally a multi-line message. If I use slog, line breaks are escaped. Is there a preferred way of logging multiple lines so they're ready to copy/paste as a block?

PR #2430, opened 2025-09-29
fix(internal/librarian): write the PR body on "commit but no push"

jskeet:
Additional fix for #2389


jskeet:
Looking into the test failure.

PR #2432, opened 2025-09-29
feat(sidekick): support aggregated pagination

### LINKED ISSUE #2383: [sidekick/discovery]: aggregated lists need pagination markers
These are APIs that include a field of `map<K, V>` type instead of repeated elements.

coryan:
Fixes #2383


coryan (commented):
PTAL

PR #2441, opened 2025-09-29
feat(sidekick): override pagination items field

coryan:
In one API (SqlAdmin) the default field for pagination is completely
wrong:

https://github.com/googleapis/googleapis/blob/a4ded737527f57207f37ba9e9c9b44732a0eb70f/google/cloud/sql/v1/cloud_sql_instances.proto#L857-L861

We need a mechanism to override the default choice.

Part of the work for https://github.com/googleapis/google-cloud-rust/issues/3430


dbolduc (approved):
Thanks

PR #2463, opened 2025-10-01
fix(internal/librarian): handle slashes in library IDs

### LINKED ISSUE #2305: google-cloud-go: figure out how to handle multi-version modules
Currently we have two PubSub modules: cloud.google.com/go/pubsub/v2 and cloud.google.com/go/pubsub. These are separate modules, to be released separately - so need to have separate library IDs... but we use the library ID for generation etc.

(Filing as a bug due to it being an unexpected problem...)

jskeet:
This ensures that we don't create nested output directories during the generate command.

Fixes #2305

### DIFF
--- a/internal/librarian/generate.go
+++ b/internal/librarian/generate.go
@@ -21,6 +21,7 @@ import (
 	"log/slog"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/googleapis/librarian/internal/config"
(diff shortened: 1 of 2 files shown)


jskeet:
Note that as we generate `pubsub/v2` but not `pubsub`, this isn't *strictly* required for Go migration right now... but I think it's still worth doing to avoid a hard-to-diagnose later issue.

codyoss (approved):

PR #2474, opened 2025-10-02
feat: Generate methods that use server-side streaming

brianquinlan:
The generated methods look like this:

```dart
  /// Generates a [streamed response](https://ai.google.dev/gemini-api/docs/text-generation?lang=python#generate-a-text-stream)
  /// from the model given an input `GenerateContentRequest`.
  Stream<GenerateContentResponse> streamGenerateContent(
    GenerateContentRequest request,
  ) {
    final url = Uri.https(_host, '/v1/${request.model}:streamGenerateContent');
    return _client
        .postStreaming(url, body: request)
        .map(GenerateContentResponse.fromJson);
  }
```

### DIFF
--- a/internal/sidekick/internal/dart/annotate.go
+++ b/internal/sidekick/internal/dart/annotate.go
@@ -120,7 +120,7 @@ type methodAnnotation struct {
 	BodyMessageName     string
 	QueryLines          []string
 	IsLROGetOperation   bool
-	ServerSideStreaming bool
+	ServerSideStreaming bool // Whether the server supports streaming via server-sent events (SSE).
 }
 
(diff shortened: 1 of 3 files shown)


devoncarew:
>     final url = Uri.https(_host, '/v1/${request.model}:streamGenerateContent');

Can you confirm that this isn't instead:

```
final url = Uri.https(_host, '/v1/${request.model}:generateContent');
```

I.e., the same end-point as non-streaming, but expecting an SSE streaming response, or using a different endpoint in order to get the streaming response.

brianquinlan:
I filed https://github.com/googleapis/librarian/issues/2490 to track adding whitelist support.

PR #2484, opened 2025-10-03
fix(sidekick): real runs for `rust-publish`

coryan:
During testing I hard-coded `-dry-run` to true, eventually we need to
run for realsies.


dbolduc (approved):
lol

PR #2489, opened 2025-10-06
feat: the containers module to share data and functions

suztomo:
The language containers will use the containers module.


suztomo:
Test passed.

PR #2567, opened 2025-10-14
feat(sidekick/rust): disable some clippy warnings

coryan:
Clippy warns about the indentation of some markdown text. The *output*
of the markdown looks fine, it is just that the comments are less
readable with bad indentation. This PR will disable that warning for
all generated clients. Trying to fix the comments is not worth it.

### DIFF
--- a/internal/sidekick/internal/rust/annotate_model_test.go
+++ b/internal/sidekick/internal/rust/annotate_model_test.go
@@ -83,9 +83,9 @@ func TestRustdocWarnings(t *testing.T) {
 		},
 		{
 			Options: map[string]string{
-				"disabled-rustdoc-warnings": "",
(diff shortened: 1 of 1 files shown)


internal/sidekick/internal/rust/annotate_model_test.go
```diff
@@ -66,6 +66,76 @@ func TestDefaultFeatures(t *testing.T) {
...
+			Options: map[string]string{
+				"disabled-clippy-warnings": "",
+			},
+			Want: []string{},
+		},
+		{
+			Options: map[string]string{
+				"disabled-clippy-warnings": "",
+			},
+			Want: []string{},
```
dbolduc: ditto

coryan: Same.

PR #2586, opened 2025-10-15
feat(dart): adjust the service method exception documentation

devoncarew:
- minor tweak to the service method exception documentation

### DIFF
--- a/internal/sidekick/internal/dart/templates/lib/method.mustache
+++ b/internal/sidekick/internal/dart/templates/lib/method.mustache
@@ -19,9 +19,8 @@ limitations under the License.
 {{/Codec.DocLines}}
 ///
 /// Throws a [http.ClientException] if there were problems communicating with
-/// the API service. Throws a [StatusException] if the API failed with a
-/// [Status] message. Throws a [ServiceException] if the API failed for any
-/// other reason.
+/// the API service, a [StatusException] if the API failed with a [Status]
+/// message, and a [ServiceException] if the API failed for any other reason.
 {{#Codec.ServerSideStreaming}}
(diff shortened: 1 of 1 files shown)


internal/sidekick/internal/dart/templates/lib/method.mustache
```diff
@@ -19,9 +19,8 @@ limitations under the License.
 {{/Codec.DocLines}}
 ///
 /// Throws a [http.ClientException] if there were problems communicating with
-/// the API service. Throws a [StatusException] if the API failed with a
```
brianquinlan: @devoncarew I don't feel strongly about this but I phrased this the way that I did to match how it is done in Effective Dart: https://dart.dev/effective-dart/documentation#do-separate-the-first-sentence-of-a-doc-comment-into-its-own-paragraph

devoncarew: Gotcha; I think this reads a bit easier but am fine w/ either.

(re: the style guide - I'm not sure its intending to comment on the exception style in that entry so much as starting doc comments w/ a markdown sentence; cc @natebosch )

devoncarew: 👍 , I'll close this for now

natebosch: Oops, I never hit the comment button on my draft. I'm aligned with the conversation anyway, but here's what I would have said.

I don't have particularly strong feelings either way, but I think I lean towards the 1 exception per sentence style - the regular format makes the extra verbosity not a problem for me.

PR #2589, opened 2025-10-16
fix(sidekick): Fixes several issues with oneof main setter samples

amanda-tarafa:
- Uses the right import for the oneof group
- Takes boxing of the oneof message variants into account.
- Messages variants are now deprioritized for selection as a oneof example variant.
- Makes the values shown match the ones used in other samples

This also removes the sample generation for when the example one of varinat that's use on the main setter is repeated or map, as that needs bigger improvements that will be done in a subsequent PR. Note that the samples for each specific variant, including repeated and map, continue to be generated.

### DIFF
--- a/internal/sidekick/internal/rust/annotate.go
+++ b/internal/sidekick/internal/rust/annotate.go
@@ -396,11 +396,16 @@ type oneOfAnnotation struct {
 	QualifiedName string
 	// The fully qualified name, relative to `codec.modulePath`. Typically this
 	// is the `QualifiedName` with the `crate::model::` prefix removed.
+	// This is always relative to `ScopeInExamples`.
 	RelativeName string
 	// The Rust `struct` that contains this oneof, fully qualified
 	StructQualifiedName string
-	FieldType           string
-	DocLines            []string
+	// The scope to show in examples. For messages in external packages
+	// this is basically `QualifiedName`. For messages in the current package
+	// this includes `modelAnnotations.PackageName`.
+	ScopeInExamples string
+	FieldType       string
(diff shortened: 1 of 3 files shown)


amanda-tarafa:
Generation result in https://github.com/googleapis/google-cloud-rust/pull/3527

internal/sidekick/internal/rust/annotate.go
```diff
@@ -991,12 +996,22 @@ func (c *codec) annotateOneOf(oneof *api.OneOf, message *api.Message, model *api
...
+	if strings.HasPrefix(scope, c.modulePath+"::") {
+		scopeInExamples = strings.Replace(scope, c.modulePath, fmt.Sprintf("%s::model", c.packageNamespace(model)), 1)
+	}
 
 	bestField := slices.MaxFunc(oneof.Fields, func(f1 *api.Field, f2 *api.Field) int {
 		if f1.Deprecated == f2.Deprecated {
 			if f1.Map == f2.Map {
 				if f1.Repeated == f2.Repeated {
-					return 0
+					if f1.MessageType != nil && f2.MessageType == nil {
```
coryan: For a separate PR: this function is getting large enough that it needs unit tests.

coryan (approved):

internal/sidekick/internal/rust/annotate.go
amanda-tarafa: annotate_test.TestOneOfExampleFieldSelection is meant to test just this, if that's not enough I'm happy to add more cases there and/or extract this function to test it directly.

PR #2595, opened 2025-10-16
chore: add unit test to show libraries can have global files in src roots

### LINKED ISSUE #2483: librarian: consider global files when propose release for a library
The current release logic does not account changes outside of source-root for the library.

This stemmed from discussions in [go/sdk-librarian:java](http://goto.google.com/sdk-librarian:java).  In Java, if dependency upgrade happened to parent POM (e.g. shared dep [here](https://github.com/renovate-bot/google-cloud-java/blob/bcf30d15de5114a9425acbde2e43db5522e00430/google-cloud-pom-parent/pom.xml#L16-L19)), each library needs a release.

we do have the opposite of this feature: release_exclude_paths. On option is to add a release_include_paths to support something outside the roots.

JoeWang1127:
Fixes #2483

### DIFF
--- a/internal/librarian/release_init_test.go
+++ b/internal/librarian/release_init_test.go
@@ -1153,6 +1153,154 @@ func TestInitRun(t *testing.T) {
 	}
 }
 
+func TestRunInitCommand(t *testing.T) {
+	t.Parallel()
+	for _, test := range []struct {
+		name   string
+		state  *config.LibrarianState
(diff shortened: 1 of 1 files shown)


meltsufin (commented):
This might require a separate test, but can we verify that `release-init-request.json` includes changes to the global files under release notes for each of the libraries that reference that global file in their source roots?

meltsufin (approved):

PR #2612, opened 2025-10-17
feat: skip a GitHub release for a library

### LINKED ISSUE #2606: Support skipping GitHub release creation per library
Java and possibly other languages only need to create GitHub releases for the BOM, which will be configured as a library in Librarian. Other libraries can still have release git tags, but no GitHub releases are needed.
This will be configured using the optional `skip_github_release_creation` property in `state.yaml` at the library level.

This supersedes #1705 which proposed creation of a tag-per-repo mechanism.

JoeWang1127:
Fixes #2606

### DIFF
--- a/internal/librarian/tag_and_release.go
+++ b/internal/librarian/tag_and_release.go
@@ -181,7 +181,11 @@ func (r *tagAndReleaseRunner) processPullRequest(ctx context.Context, p *github.
 			return fmt.Errorf("library %s not found", release.Library)
 		}
 
-		libraryConfig := librarianConfig.LibraryConfigFor(release.Library)
+		var libraryConfig *config.LibraryConfig
+		if librarianConfig != nil {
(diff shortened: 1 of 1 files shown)


internal/config/state.go
```diff
@@ -137,6 +137,8 @@ type LibraryState struct {
 	// Whether including this library in a release.
 	// This field is ignored when writing to state.yaml.
 	ReleaseTriggered bool `yaml:"-" json:"release_triggered,omitempty"`
+	// Whether to create a GitHub release for this library.
+	SkipGitHubReleaseCreation bool `yaml:"skip_github_release_creation" json:"-"`
```
chingor13: Should this be in the config.yaml? vs state.yaml?

JoeWang1127: It should be in the `state.yaml` according to the issue.

chingor13 (dismissed):
LGTM, aside from the e2e test failing which is maybe unrelated

Issue #2628, opened 2025-10-21
librarian(bug): librarian release init fails for recently configured library


parthea:
I ran `librarian release init` on a branch with PR https://github.com/googleapis/google-cloud-python/pull/14791 however the command failed with error `failed to find tag google-cloud-gkerecommender-v0.0.0: tag not found`.

```
partheniou@partheniou-vm-3:~/git/google-cloud-python$ librarian release init
time=2025-10-21T09:28:27.913Z level=INFO msg="Temporary working directory" dir=/tmp/librarian-3061479323
time=2025-10-21T09:28:27.914Z level=INFO msg="repo not specified, using current working directory as repo root" path=/usr/local/google/home/partheniou/git/google-cloud-python
time=2025-10-21T09:28:27.914Z level=INFO msg="Opening repository" dir=/usr/local/google/home/partheniou/git/google-cloud-python
time=2025-10-21T09:28:35.963Z level=INFO msg="source not specified, skipping service config population"
time=2025-10-21T09:28:35.966Z level=INFO msg="Initiating a release" dir=/tmp/librarian-3061479323/output
time=2025-10-21T09:28:35.968Z level=INFO msg="failed to fetch conventional commits for library, google-cloud-gkerecommender: failed to get commits for library google-cloud-gkerecommender: failed to find tag google-cloud-gkerecommender-v0.0.0: tag not found"
```

A potential solution could be to skip searching for git tags when the library version is 0.0.0

PR #2631, opened 2025-10-21
fix(librarian): resolve issue where commits cannot be fetched for new library

parthea:
Fixes https://github.com/googleapis/librarian/issues/2628


codyoss (approved):
Thanks!

Issue #2623, opened 2025-10-20
librarian: container testing - implement new CLI command


zhumin8:
new standalone `test-container generate` command

PR #2656, opened 2025-10-22
feat(librarian): add test-container test logic

zhumin8:
Adding test logic according to [go/sdk-librarian:container-test](http://goto.google.com/sdk-librarian:container-test).

Tested locally with a couple of libraries in google-cloud-go.
Note: Removed changes for adding the command, will address in separate PR.

For #2623

### DIFF
--- a/internal/gitrepo/gitrepo.go
+++ b/internal/gitrepo/gitrepo.go
@@ -263,7 +263,7 @@ func (r *LocalRepository) ChangedFiles() ([]string, error) {
 
 // NewAndDeletedFiles returns a list of files that are new or deleted.
 func (r *LocalRepository) NewAndDeletedFiles() ([]string, error) {
-	slog.Debug("Getting new and deleted files")
+	slog.Debug("getting new and deleted files")
 	worktree, err := r.repo.Worktree()
 	if err != nil {
 		return nil, err
@@ -543,7 +543,7 @@ func (r *LocalRepository) CreateBranchAndCheckout(name string) error {
 // CheckoutCommitAndCreateBranch creates a new git branch from a specific commit hash
 // and checks out the branch in the local git repository.
 func (r *LocalRepository) CheckoutCommitAndCreateBranch(name, commitHash string) error {
-	slog.Debug("Creating branch from commit and checking out", "name", name, "commit", commitHash)
+	slog.Debug("creating branch from commit and checking out", "name", name, "commit", commitHash)
 	worktree, err := r.repo.Worktree()
 	if err != nil {
 		return err
@@ -741,7 +741,7 @@ func (r *LocalRepository) ResetHard() error {
 // DeleteLocalBranches deletes a list of local branches.
 // It returns an error if any branch deletion fails, or nil if all succeed.
 func (r *LocalRepository) DeleteLocalBranches(names []string) error {
-	slog.Debug("Starting batch deletion of local branches", "count", len(names))
+	slog.Debug("starting batch deletion of local branches", "count", len(names))
 	headRef, headErr := r.repo.Head()
 	if headErr != nil && !errors.Is(headErr, plumbing.ErrReferenceNotFound) {
 		return fmt.Errorf("failed to get HEAD to protect against its deletion: %w", headErr)


internal/librarian/test_container_generate.go
```diff
@@ -0,0 +1,326 @@
...
+	}
+
+	insertionLine := findProtoInsertionLine(lines)
+	if insertionLine == -1 {
+		// No suitable line found to inject the comment.
+		return "", nil
+	}
+
+	guid := uuid.New().String()
+	comment := "// test-change-" + guid
```
JoeWang1127: nit: use fmt.Sprintf

JoeWang1127:
Given the complexity of the implementation, e.g., additional git features, finding protos and insertion points, validation steps, write tests for testing functions, do you think this is better than alternative [option 1](https://docs.google.com/document/d/1WCPjbyqv_JKb4sCjmozJbUPzGcLrm10w4bUJhv5M40s/edit?resourcekey=0-rMTA-iK0KthvOE_f0zLFbg&tab=t.0#bookmark=id.bu6yztoh4pd9)?

internal/gitrepo/gitrepo.go
```diff
@@ -512,6 +535,21 @@ func (r *LocalRepository) CreateBranchAndCheckout(name string) error {
 	})
 }
 
+// CheckoutCommitAndCreateBranch creates a new git branch from a specific commit hash
+// and checks out the branch in the local git repository.
+func (r *LocalRepository) CheckoutCommitAndCreateBranch(name, commitHash string) error {
```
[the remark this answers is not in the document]
zhumin8: Yes, same functionality can be done with the combination. But it would require two separate git operations, leading to more I/O, which could be significant if this repeats for all libraries.
I think we can refactor this to do either Checkout and CreateBranchAndCheckout based on input though. I'd rather keep that as seperate followup.

zhumin8:
> Given the complexity of the implementation, e.g., additional git features, finding protos and insertion points, validation steps, write tests for testing functions, do you think this is better than alternative [option 1](https://docs.google.com/document/d/1WCPjbyqv_JKb4sCjmozJbUPzGcLrm10w4bUJhv5M40s/edit?resourcekey=0-rMTA-iK0KthvOE_f0zLFbg&tab=t.0#bookmark=id.bu6yztoh4pd9)?

One major drawback of that approach is the maintenance burden of the golden files and less tolerance to changes in container. e.g., change in generator or config files would need an update.
We can discuss further in [go/sdk-librarian:container-test](http://goto.google.com/sdk-librarian:container-test) if you have more doubts.

internal/gitrepo/gitrepo.go
```diff
@@ -257,6 +260,30 @@ func (r *LocalRepository) ChangedFiles() ([]string, error) {
 	return changedFiles, nil
 }
 
+// NewAndDeletedFiles returns a list of files that are new or deleted.
+func (r *LocalRepository) NewAndDeletedFiles() ([]string, error) {
+	slog.Debug("Getting new and deleted files")
```
codyoss: nit: I believe we do this at the Info level for other methods. Same for other new methods.

internal/gitrepo/gitrepo.go
```diff
@@ -512,6 +540,21 @@ func (r *LocalRepository) CreateBranchAndCheckout(name string) error {
 	})
 }
 
+// CheckoutCommitAndCreateBranch creates a new git branch from a specific commit hash
+// and checks out the branch in the local git repository.
+func (r *LocalRepository) CheckoutCommitAndCreateBranch(name, commitHash string) error {
+	slog.Debug("Creating branch from commit and checking out", "name", name, "commit", commitHash)
```
ldetmer: FYI we recently normalized logging messages to start with lowercase, can you please update your messages to follow this convention?

internal/gitrepo/gitrepo.go
```diff
@@ -257,6 +261,30 @@ func (r *LocalRepository) ChangedFiles() ([]string, error) {
 	return changedFiles, nil
 }
 
+// NewAndDeletedFiles returns a list of files that are new or deleted.
+func (r *LocalRepository) NewAndDeletedFiles() ([]string, error) {
+	slog.Debug("Getting new and deleted files")
```
ldetmer: lowercase

ldetmer (approved):

PR #2726, opened 2025-10-30
feat(dart): use proto presence info to generate null-safe code

devoncarew:
Use proto presence info to generate null-safe code:

- `annotateField` updated to use information from proto optional annotations
- toJson() implementations write values if they're nullable and have a non-null value, or if they're not nullable and have a non-default value for that type
- fromJson parses a value for parse a value for nullable fields if one is given; it will parse any value for non-null field or fall back to using the default value for that type
- properly dereference null values when referring to request fields for use in url paths
- enums definitions now get a default value (`EnumClass.$default`)
- removed the older annotate.requiredFields and tests (based on usage in service calls)
- closes https://github.com/googleapis/google-cloud-dart/issues/24

Tests are updated for `buildQueryLines()` to include optional fields. Tests were added for the `createFromJsonLine()` and `createToJsonLine()` methods; previously we got our testing here from being on the same repo as the dart code, and running the dart unit tests on the generator's putput.

In-lined from the PR:

>	- proto 3 fields default to implicit presence
>	- the 'optional' keyword changes a field to explicit presence
>	- types like lists (repeated) and maps are always implicit presence
>	
>	Explicit presence means that you can know whether the user set a value or
>	not. Implicit presence means you can always retrieve a value; if one had
>	not been set, you'll see the default value for that type.
>	
>	We translate explicit presence (a optional annotation) to using a nullable
>	type for that field. We translate implicit presence (always returning some
>	value) to a non-null type.
>	
>	Some short-hand:
>	  - optional == explicit == nullable
>	  - implicit == non-nullable
>	  - lists and maps == implicit == non-nullable
>	  - singular message == explicit == nullable

Sample output can be seen at https://github.com/googleapis/google-cloud-dart/pull/58.

### DIFF
--- a/internal/sidekick/internal/dart/annotate_test.go
+++ b/internal/sidekick/internal/dart/annotate_test.go
@@ -503,6 +503,12 @@ func TestBuildQueryLines(t *testing.T) {
 			[]string{"if (result.stringOpt != null) 'string': result.stringOpt!"},
 		},
 
+		// one ofs
+		{
+			&api.Field{Name: "bool", JSONName: "bool", Typez: api.BOOL_TYPE, IsOneOf: true},
+			[]string{"if (result.bool$ != null) 'bool': '${result.bool$}'"},
+		},
+
 		// repeated primitives
 		{
 			&api.Field{Name: "boolList", JSONName: "boolList", Typez: api.BOOL_TYPE, Repeated: true},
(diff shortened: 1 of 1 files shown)


internal/sidekick/internal/dart/annotate.go
```diff
@@ -741,6 +822,14 @@ func (annotate *annotateModel) createFromJsonLine(field *api.Field, state *api.A
 		return fmt.Sprintf("decodeInt64(%s)%s", data, bang)
 	case field.Typez == api.FLOAT_TYPE || field.Typez == api.DOUBLE_TYPE:
 		return fmt.Sprintf("decodeDouble(%s)%s", data, bang)
+	case field.Typez == api.INT32_TYPE || field.Typez == api.FIXED32_TYPE ||
```
brianquinlan: optional: you can join this with the next two case statements

brianquinlan (dismissed):
There are a few unconvered lines. Could you see if it is possible to add tests for them?

devoncarew:
> There are a few unconvered lines. Could you see if it is possible to add tests for them?

For sure (I haven't gotten into the habit of looking at the coverage output).

PR #2730, opened 2025-10-30
feat(java): release init to update pom.xml

suztomo:


internal/container/java/message/message.go
```diff
@@ -26,9 +26,21 @@ import (
 	"os"
 )
 
+// ReleaseInitRequest is the structure of the release-init-request.json file.
+type Request struct {
```
suztomo: I don't think this is needed.

PR #2733, opened 2025-10-31
chore: add a check to enforce golang version

### LINKED ISSUE #2737: all: make golang version consistent
We setup golang in multiple place, e.g., `go.mod` and `Dockerfile`. We should update the versions in one-go when needed.

Currently we don't have a rule/test to enforce golang version in the repo and we raised multiple PRs to update the version (#2724 and #2725).

JoeWang1127:
Add a test to enforce golang version in `go.mod` and Dockerfile.

Fixes #2737

### DIFF
--- a/all_test.go
+++ b/all_test.go
@@ -166,7 +166,7 @@ func hasValidHeader(path string, r io.Reader) (bool, error) {
 // TestConsistentGoVersions walks the directory tree and checks Dockerfiles and go.mod files for specified Go versions.
 // It ensures that only one unique Go version is specified across all found files to maintain consistency. The test
 // fails if multiple Go versions are detected.
-// TODO: remove this test once https://github.com/googleapis/librarian/issues/2739 is resolved.
+// TODO(https://github.com/googleapis/librarian/issues/2739): remove this test once is resolved.
 func TestConsistentGoVersions(t *testing.T) {
 	goVersions := make(map[string][]string)
 	sfs := os.DirFS(".")


all_test.go
```diff
@@ -159,6 +163,59 @@ func hasValidHeader(path string, r io.Reader) (bool, error) {
...
 
+func TestGolangVersion(t *testing.T) {
+	goVersions := make(map[string][]string)
+	sfs := os.DirFS(".")
+	err := fs.WalkDir(sfs, ".", func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if strings.HasSuffix(path, "Dockerfile") {
```
jskeet: Any reason for using a suffix here and below, rather than equality using `d.Name()`?

Basically I guess I'm asking whether we *want* to check files of "SittingOnTheDockerfile" and "letmego.mod"

all_test.go
```diff
@@ -159,6 +163,59 @@ func hasValidHeader(path string, r io.Reader) (bool, error) {
 	return headerRegex.Match(allBytes), nil
 }
 
+func TestGolangVersion(t *testing.T) {
```
jskeet: Possibly write a comment saying what it's actually testing, namely that the golang version is uniform across the repo? (I was wondering how we worked out which version to assert, and figured it out by the end, but a comment or possibly a rename would make that simpler.)

all_test.go
JoeWang1127: My assumption is we may have multiple Dockerfiles (we have 2 within `infra`) and the file name maybe `something.Dockerfile` or `Dockerfile`. In either case, the test can check the go version.

The same with `go.mod`.

jskeet:
@JoeWang1127 I'm done until Monday - I can look then, but it might be better to ask someone in the US to review.

all_test.go
```diff
@@ -159,6 +163,59 @@ func hasValidHeader(path string, r io.Reader) (bool, error) {
 	return headerRegex.Match(allBytes), nil
 }
 
+func TestConsistentGoVersions(t *testing.T) {
```
julieqiu: Add a comment about why we have this test. We should consider just moving the failing test to presubmit, in which case I don't think we will need this test. It's fine to merge this PR for now though.

julieqiu (dismissed):

Issue #2176, opened 2025-09-16
librarian: refactor CLI framework


julieqiu:
Refactor CLI framework to move command execution logic from internal/librarian into internal/cli, and move business logic to internal/librarian.

PR #2742, opened 2025-10-31
refactor(internal/cli): make Init idempotent and remove redundant calls

julieqiu:
The Init method now recursively initializes all subcommands and is idempotent, checking if Flags and Config are nil before creating them.

This allows calling Init once on the root command to initialize the entire command tree.

This change simplifies command initialization in newLibrarianCommand by removing redundant Init calls on cmdVersion and cmdRelease.

For https://github.com/googleapis/librarian/issues/2176


julieqiu:
Going to approach this differently. Closing for now.

PR #2765, opened 2025-11-03
fix: enforce removal before copying library files

### LINKED ISSUE #2593: librarian: file copying doesn't enforce previous file removal
When copying generated files (as opposed to updated files for versions etc) we should fail if the file already exists in the destination - because it means our removal regexes aren't removing everything they should.

At the moment, this doesn't happen. The file is just truncated, due to the use of `os.Create` (in command.go, `copyFile`). I can't immediately see a similar function that implements "create, failing if the file already exists". We could try to open the file first to see if it exists.

(In Librarian v0.1.0 we implemented this by simply using `os.CopyFS` for the whole output directory. I suspect that wouldn't work with symlinks though.)

Assigning to Cody to validate that this is a bug and consider the priority. Note that we'll need to work out when to require this and when not to - it's *probably* only in generate and maybe configure that this should happen.

JoeWang1127:
Fixes #2593

### DIFF
--- a/internal/librarian/command.go
+++ b/internal/librarian/command.go
@@ -330,7 +330,7 @@ func copyLibraryFiles(state *config.LibrarianState, dest, libraryID, src string,
 			slog.Debug("copying file", "file", file)
 			srcFile := filepath.Join(srcPath, file)
 			dstFile := filepath.Join(dstPath, file)
-			if _, err := os.Stat(dstFile); failOnExistingFile && !errors.Is(err, os.ErrNotExist) {
+			if _, err := os.Stat(dstFile); failOnExistingFile && err == nil {
 				return fmt.Errorf("file existed in destination: %s", dstFile)
 			}
 			if err := copyFile(dstFile, srcFile); err != nil {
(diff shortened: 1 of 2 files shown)


internal/librarian/command.go
```diff
@@ -299,15 +299,21 @@ func cleanAndCopyLibrary(state *config.LibrarianState, repoDir, libraryID, outpu
...
+// If `failOnExistingFile` is true, the function will check if a file already
+// exists at the destination. If it does, an error is returned immediately without copying.
+// If `failOnExistingFile` is false, it will overwrite any existing files.
+//
 // If there's no files in the library's SourceRoots under the src directory, no copy will happen.
+//
 // If a file is being copied to the library's SourceRoots in the dest folder but the folder does
 // not exist, the copy fails.
-func copyLibraryFiles(state *config.LibrarianState, dest, libraryID, src string) error {
+func copyLibraryFiles(state *config.LibrarianState, dest, libraryID, src string, failOnExistingFile bool) error {
```
miguelvelezsa: WDYT about using overwrite or overwriteFile instead of failOnExistingFile?

internal/librarian/generate_command.go
```diff
@@ -375,7 +375,7 @@ func (r *generateRunner) runConfigureCommand(ctx context.Context, outputDir stri
 		r.state.Libraries[i] = libraryState
 	}
 
-	if err := copyLibraryFiles(r.state, r.repo.GetDir(), libraryState.ID, outputDir); err != nil {
+	if err := copyLibraryFiles(r.state, r.repo.GetDir(), libraryState.ID, outputDir, false); err != nil {
```
miguelvelezsa: is there a test checking this?

PR #2770, opened 2025-11-04
docs: use consistent library id in flags and examples

### LINKED ISSUE #2704: doc: Update godoc for flags/examples to use a consistent library ID throughout
@julieqiu noticed: "In the [pkg.go.dev](https://www.google.com/url?q=http://pkg.go.dev&sa=D&source=docs&ust=1761745593842871&usg=AOvVaw2jdQ_mIwwqZHrFYDCsTCuA) docs it says that an example library id is google-cloud-secretmanager-v1, but for generate the library id is just secretmanager. Are these defined differently between these two commands?"

We should be consistent so this is not confusing.

JoeWang1127:
Use `secretmanager` as an example of library ID in flags and examples.

Fixes #2704

### DIFF
--- a/doc/language-onboarding.md
+++ b/doc/language-onboarding.md
@@ -100,16 +100,6 @@ The container is expected to produce up to two artifacts:
 {
   "libraries": [
     {
-      "id": "secretmanager",
-      "apis": [
-        {
-          "path": "google/cloud/secretmanager/v1",
-          "service_config": "secretmanager_v1.yaml",
-          "status": "new"
-        }
-      ],
-    },
-    {
       "id": "pubsub",
       "apis": [
         {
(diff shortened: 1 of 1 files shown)


doc/language-onboarding.md
```diff
@@ -100,7 +100,7 @@ The container is expected to produce up to two artifacts:
 {
   "libraries": [
     {
-      "id": "google-cloud-secretmanager",
+      "id": "secretmanager",
```
jskeet: Can you move this to after pubsub? We started keeping state.yaml in alphabetical order a while ago, and I suspect the request will conform to that.

jskeet (commented):
One nit (to address a pre-existing issue). Let me know if you'd rather tackle that separately.

doc/language-onboarding.md
JoeWang1127: Done.

Issue #2416, opened 2025-09-29
automation: add package docs


julieqiu:
The package documentation for cmd/automation currently looks like a stale copy of docs from cmd/librarian, and internal/automation doesn’t have any package docs at all.

See:
- https://github.com/googleapis/librarian/blob/0d0642df0789611287df68968bffdbcfab404867/cmd/automation/main.go#L15-L18
- https://pkg.go.dev/github.com/googleapis/librarian/cmd/automation
- https://pkg.go.dev/github.com/googleapis/librarian/internal/automation

We should update both to accurately describe what this tool is for and how it works. For reference, see the cmd/librarian documentation here: https://pkg.go.dev/github.com/googleapis/librarian/cmd/librarian.

See also https://go.dev/doc/comment how guidance on Go package docs.

ldetmer:
We should create something similar to  https://github.com/googleapis/librarian/blob/main/cmd/librarian/doc_generate.go for automation

JoeWang1127:
We have not set up the automation to run `update-image` command for automation CLI, per https://github.com/googleapis/librarian/pull/2908#pullrequestreview-3465184986.

All other commands now have proper docs.

PR #2810, opened 2025-11-06
chore(cmd): add doc.go in `cmd/automation`

JoeWang1127:
Add an empty doc.go in `cmd/automation`. 

In addition, modify `doc_generate.go` so that it can be used to generate docs for different packages.

For #2416


cmd/librarian/doc_generate.go
```diff
@@ -41,7 +42,7 @@ const docTemplate = `// Copyright 2025 Google LLC
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:generate go run -tags docgen doc_generate.go
+//go:generate go run -tags docgen doc_generate.go -cmd .
```
ldetmer: if this file will no longer be command specific, can you move it to the top level cmd folder?

ldetmer (approved):

PR #2814, opened 2025-11-06
feat(internal/automation): refactor `publish-release` command

JoeWang1127:
Add `publish-release` command in automation package so we can auto generate documentation.

For #2416

### DIFF
--- a/cmd/automation/doc.go
+++ b/cmd/automation/doc.go
@@ -39,7 +39,5 @@ Flags:
 
 	-project string
 	  	Google Cloud Platform project ID (default "cloud-sdk-librarian-prod")
-	-push
-	  	The _PUSH flag (true/false) to Librarian CLI's -push option
 */
 package main


internal/automation/publish_release.go
```diff
@@ -0,0 +1,121 @@
...
+	}, nil
+}
+
+func (r *publishRunner) run(ctx context.Context) error {
+	if shouldSkip(time.Now(), r.forceRun) {
+		slog.Info("skipping stage-release on an even week.")
+		return nil
+	}
+
+	errs := make([]error, 0)
```
ldetmer: for the same reason as above (having duplicate logic in 2 places), can we make the logic from line 79-93 as common logic and update trigger.go to point to this.  Or I guess can this logic just continue to live in trigger.go and we only have publish-release specific logic in this command?

JoeWang1127: Done.

cmd/automation/doc.go
```diff
@@ -25,5 +25,21 @@ Usage:
 	librarian <command> [arguments]
```
JoeWang1127: I'll raise another PR to change this part.

internal/config/config.go
```diff
@@ -194,6 +194,9 @@ type Config struct {
 	// Requires the --library flag to be specified.
 	LibraryVersion string
 
+	// Project is the ID of the Google Cloud project to use.
```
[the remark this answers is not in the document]
JoeWang1127: The flags are parsed from command line to `cli.Config` to align with the current functions, for example:
https://github.com/googleapis/librarian/blob/07874e73c4868d1d62c652153ee8236a8c947959/internal/librarian/generate_command.go#L60

I don't see this value stored in other places.

ldetmer (approved):

PR #2823, opened 2025-11-07
fix(internal/librarian): update-image should skip libraries where generate is blocked by config

### LINKED ISSUE #2773: librarian: update-image does not honor config if generate_blocked
When update-image is run, I think the libraries configured with `generate_blocked` should be skipped. For example, they can be handwritten and only in state for release.

In generate flow, generation is skipped if `generate_blocked` is true. This logic is inside [`shouldGenerate()`](https://github.com/googleapis/librarian/blob/87a10056f182928c4f76ba33567229033608b87d/internal/librarian/generate_command.go#L429) which has additional logic specific to generate.

zhumin8:
`update-image` command now skips a library if it is configured with `generate_blocked` in config.yaml.
Also done minor refactoring to reuse logic in identifying blocked libraries.

Fixes #2773

### DIFF
--- a/internal/librarian/update_image.go
+++ b/internal/librarian/update_image.go
@@ -148,8 +148,12 @@ func (r *updateImageRunner) run(ctx context.Context) error {
 	if len(failedGenerations) > 0 {
 		slog.Warn("failed generations", "num", len(failedGenerations))
 	}
-	slog.Info("skipped generations", "num", skippedGenerationsCount)
-	slog.Info("successful generations", "num", len(successfulGenerations))
+	slog.Info(
+		"generation statistics",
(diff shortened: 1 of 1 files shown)


ldetmer:
> Fix #2773

super nit: please update to use convention from [contributing guidelines](https://github.com/googleapis/librarian/blob/main/CONTRIBUTING.md#referencing-issues) to use "Fixes" instead of "Fix"

Also please include [1 line description](https://github.com/googleapis/librarian/blob/main/CONTRIBUTING.md#description) in complete sentence (it can just be a copy of the title but in full sentence format)

ldetmer (approved):

PR #2836, opened 2025-11-10
chore: update stage-release automation to run weekly instead of bi-weekly

ldetmer:
Originally we had set a bi-weekly cadence in anticipation of needing to support java and its release train.  Now that it seems that java will not be using a release train, let's make the releases weekly to be more timely and consistent. 

Note this also removes the forceRun flag which was used in the case you wanted automation to force run stage-release command on an even week.


jskeet (approved):
It's nice to be able to get rid of the conditional logic there :)

Issue #2753, opened 2025-11-03
all: update CODEOWNERS to avoid double-approval for sidekick changes


julieqiu:
PRs that modify both sidekick files and other repo files (even when very minor) currently require approvals from both the sidekick team and the Librarian team. This slows down development for cross-cutting changes.

We should consider updating CODEOWNERS and adding `@googleapis/cloud-sdk-librarian-team` as a co-owner on sidekick directories so either team can approve:

```
cmd/sidekick/ @googleapis/cloud-sdk-sidekick-team @googleapis/cloud-sdk-librarian-team
internal/sidekick/ @googleapis/cloud-sdk-sidekick-team @googleapis/cloud-sdk-librarian-team

* @googleapis/cloud-sdk-librarian-team
```

See https://github.com/googleapis/librarian/pull/2751 and https://github.com/googleapis/librarian/pull/2750 as examples.

julieqiu:
@coryan and I discussed this, and he's on board with the Librarian team being an approver on PRs for sidekick.

@ldetmer, how do you want to move forward?

ldetmer:
Sure i'm fine updating adding @googleapis/cloud-sdk-librarian-team as a co-owner on sidekick directories.  Will prioritize this for this sprint.

julieqiu:
We probably also want to reduce the scope of cloud-sdk-librarian-team. Right now there are 17 people in the group, and I don't think they need to be pinged on every PR.

PR #2845, opened 2025-11-10
chore(codeowners): adding cloud-sdk-librarian-team as a co-owner for sidekick

zhumin8:
Adding @googleapis/cloud-sdk-librarian-team as a co-owner on sidekick directories so either team can approve to speed up development for cross-cutting changes.

For #2753

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -5,5 +5,5 @@
 # The owners for all files in the repo.
 * @googleapis/cloud-sdk-librarian-team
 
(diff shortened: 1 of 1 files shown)


julieqiu:
> Fixes: #2753

Would you be able to add a commit message? See https://github.com/googleapis/librarian/blob/1289914cad447ffd1c7b0e2968ed8f6c7c3a8a84/CONTRIBUTING.md#commit-messages

Also we probably want to reduce the scope of cloud-sdk-librarian before marking this as fixed. Right now there are 19 ppl and I don't think they need to be pinged on every PR.

PR #2857, opened 2025-11-11
fix(librarian): use a test-output directory for tests in update-image

### LINKED ISSUE #2851: librarian: update-image -test fails due to reusing output directory
Steps to reproduce:

1. Clone https://github.com/googleapis/google-cloud-go
2. Edit the state file to remove all but the first library (just to keep it quick and the output short)
3. Commit the change
4. Run `librarian update-image -test`

Librarian finds the latest image (which is the same as the current image, that's fine), and regenerates the API - all fine so far. It then moves on to the test phase, so tweaks the protos and runs generate again... but reuses the existing output directory from before. That fails because the Go container assumes the output directory will be empty.

While the container contract at https://github.com/googleapis/librarian/blob/main/doc/language-onboarding.md isn't *explicit* about this, I think internal docs have always said "the output directory will be empty".

Assuming we agree as a team that the container *should* be able to assume the output directory will be empty, I think we should:

- Clarify that in the container contract
- Change the test code to use a new output directory with a "test" prefix or suffix. An alternative would be to delete the previous output directory first, but I think it would help diagnostics to keep the old one and use a separate test directory, so it's easy to compare the two.

Each library has a separate output directory. Looking at the code, I *think* it would be easiest to create a "test-output" subdirectory of the work root which has the same directory structure as the "output" subdirectory of the work root, rather than have output/foo, output/test-foo, output/bar, output/test-bar etc. (That will probably make diffing easier too.)

jskeet:
This avoids reusing existing directories under output, which will have been populated by the container prior to running tests.

Fixes #2851

### DIFF
--- a/internal/librarian/test_container_generate.go
+++ b/internal/librarian/test_container_generate.go
@@ -67,7 +67,10 @@ func (r *testGenerateRunner) run(ctx context.Context) error {
 // listing the failed libraries, preserving the generated files. If all tests
 // pass or are skipped, it cleans up the work directory.
 func (r *testGenerateRunner) runTests(ctx context.Context, sourceRepoHead string) error {
-	outputDir := filepath.Join(r.workRoot, "output")
+	outputDir := filepath.Join(r.workRoot, "test-output")
+	if err := os.Mkdir(outputDir, 0755); err != nil {
+		return fmt.Errorf("failed to make output directory, %s: %w", outputDir, err)
+	}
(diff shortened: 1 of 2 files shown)


zhumin8 (approved):
> @zhumin8: Let me know if you think it's worth writing a test for "we return an error if we can't create test-output" (which I expect to be the missing line from coverage).

I think it's fine to leave this one out.

PR #2869, opened 2025-11-12
all: group consecutive var declarations

julieqiu:
Consecutive var declarations are grouped into var blocks, which is more idiomatic in Go. See https://go.dev/doc/effective_go#variables.


miguelvelezsa (approved):
I think all variables should be const :)

PR #2878, opened 2025-11-12
fix(internal/gitrepo): surface error with more context at failure

zhumin8:
Provide more context in error message when getConventionalCommitsSinceLastRelease fails.

### DIFF
--- a/internal/librarian/commit_version_analyzer.go
+++ b/internal/librarian/commit_version_analyzer.go
@@ -31,7 +31,7 @@ func getConventionalCommitsSinceLastRelease(repo gitrepo.Repository, library *co
 	commits, err := repo.GetCommitsForPathsSinceTag(library.SourceRoots, tag)
 
 	if err != nil {
-		return nil, fmt.Errorf("failed to get commits for library %s with source roots %s at tag %s: %w", library.ID, library.SourceRoots, tag, err)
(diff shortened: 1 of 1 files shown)


internal/librarian/commit_version_analyzer.go
```diff
@@ -40,7 +40,11 @@ func getConventionalCommitsSinceLastRelease(repo gitrepo.Repository, library *co
 		return shouldIncludeForRelease(files, library.SourceRoots, library.ReleaseExcludePaths)
 	}
 
-	return convertToConventionalCommits(repo, library, commits, shouldIncludeFiles)
+	conventionalCommits, err := convertToConventionalCommits(repo, library, commits, shouldIncludeFiles)
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert commits to conventional commits for library %q: %w", library.ID, err)
```
JoeWang1127: Why not writing the error log inside `convertToConventionalCommits`? You can avoid an additional layer of unwrapping errors.

JoeWang1127: `generator: failed to get commits for library generator: did not find commit c64fa6012bad2381cab8a3c764c4eb4b8b14258d when iterating`

This part of the error log seems not easy to understand.

It's OK we can refactor later.

PR #2894, opened 2025-11-13
feat(internal/cli): add version command

JoeWang1127:
Refactor command initialization, make `version` command as a default command in each command set.

For https://github.com/googleapis/librarian/issues/2416


julieqiu:
I would move this to internal/cli so it gets added to every command, since this is exactly the same as librarian version by default:

https://github.com/googleapis/librarian/blob/bac2f0c78bcf9d2cceec56a35507991df2311435/internal/cli/cli.go#L115

julieqiu (approved):

julieqiu:
> Refactor command initialization, make `version` command as a default command in each command set.
> 
> For #2416

For clarity, I would make the PR title:

feat(internal/cli): add version command

ldetmer (approved):

PR #2907, opened 2025-11-14
feat(infra/prod): onboard 25 repositories to automation for publish-release

ohmayr:
This PR onboards all the remaining Python repositories for Librarian automation to support the `publish-release` command. 

We'll determine which repos require support for `generate` and/or `stage-release` and will address that separately.


ldetmer:
@ohmayr can you confirm all these repositories are already onboarded to Librarian? ex: they have a state.yaml file and are using Librarian to release?

PR #2932, opened 2025-11-17
impl(rust): generate gcp.resource.name using annotations

westarle:
This commit updates the Rust generator to populate the `gcp.resource.name`
tracing attribute in `RequestOptions` by inspecting `google.api.resource_reference`
annotations on request fields.

Key changes:
- Implements a revised heuristic in `annotate.go` that prioritizes annotated fields,
  including nested fields up to one level deep.
- Updates `transport.rs.mustache` to inject runtime logic that evaluates candidate
  fields in priority order (e.g., `name`, `parent`, `resource`).
- Adds `candidateField` struct and `findResourceNameFields` logic to support
  runtime resolution of the resource name.

This ensures that traces generated by the Rust client libraries contain the
correct resource name, improving observability and alignment with GCP standards.

### DIFF
--- a/internal/sidekick/rust/annotate.go
+++ b/internal/sidekick/rust/annotate.go
@@ -19,7 +19,6 @@ import (
 	"fmt"
 	"log/slog"
 	"slices"
-	"sort"
 	"strings"
 
 	"github.com/googleapis/librarian/internal/sidekick/api"
@@ -724,7 +723,7 @@ func makeChainAccessor(fields []*api.Field, parentAccessor string) string {
 			} else if field.Optional {
 				accessor = fmt.Sprintf("%s.and_then(|s| s.%s.as_ref())", accessor, fieldName)
 			} else {
-				accessor = fmt.Sprintf("%s.and_then(|s| Some(&s.%s))", accessor, fieldName)
+				accessor = fmt.Sprintf("%s.map(|s| &s.%s)", accessor, fieldName)
 			}
 		}
 	}
@@ -802,7 +801,7 @@ func (c *codec) findResourceNameFields(m *api.Method) []*resourceNameCandidateFi
 		return httpParams[fieldName]
 	}
 
-	sortResourceNameCandidates(candidates, isInPath)
+	slices.SortStableFunc(candidates, compareResourceNameCandidates(isInPath))
 
 	return candidates
 }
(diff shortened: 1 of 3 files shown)


internal/sidekick/rust/annotate_test.go
```diff
@@ -1963,3 +1963,329 @@ func TestOneOfExampleFieldSelection(t *testing.T) {
...
+				annotateModel(tt.method.Model, codec)
+			}
+
+			got := codec.findResourceNameFields(tt.method)
+
+			if len(tt.wantPaths) == 0 {
+				if len(got) != 0 {
+					t.Errorf("findResourceNameFields() = %v, want nil", got)
+				}
+				return
```
coryan: `continue` so the next test can run?

internal/sidekick/rust/annotate.go
```diff
@@ -690,6 +701,119 @@ func (c *codec) addFeatureAnnotations(model *api.API, ann *modelAnnotations) {
...
+		}
+		// 2. In-Path before Not-In-Path
+		inPathI := isInPath(candidates[i])
+		inPathJ := isInPath(candidates[j])
+		if inPathI != inPathJ {
+			return inPathI // true (in path) < false (not in path) -> we want true first
+		}
+		// 3. Stable sort preserves proto order
+		return i < j
+	})
```
coryan: The sorting function looks complicated enough to have a name and (hopefully smaller) unit tests?

internal/sidekick/rust/annotate.go
```diff
@@ -690,6 +701,119 @@ func (c *codec) addFeatureAnnotations(model *api.API, ann *modelAnnotations) {
 	}
 }
 
+// makeFieldAccessor generates the Rust accessor code for a field.
+// It handles optional fields and oneofs correctly.
+// parentAccessor is the accessor for the parent message (e.g. "req" or "s").
+func makeFieldAccessor(field *api.Field, parentAccessor string) string {
```
coryan: I cannot grok how this works for fields nested multiple levels.  I am sure you thought of that, I can see there are test cases for it.  I am a bit tired, and maybe "future us" will be too.  Maybe a comment in the mustache template?

internal/sidekick/rust/annotate.go
```diff
@@ -792,26 +802,30 @@ func (c *codec) findResourceNameFields(m *api.Method) []*resourceNameCandidateFi
...
+	sortResourceNameCandidates(candidates, isInPath)
+
+	return candidates
+}
+
+// sortResourceNameCandidates sorts candidates by priority:
+// 1. Top-level fields (IsNested == false).
+// 2. Fields in HTTP path (isInPath == true).
+// 3. Proto definition order (stable sort).
+func sortResourceNameCandidates(candidates []*resourceNameCandidateField, isInPath func(*resourceNameCandidateField) bool) {
```
coryan: I meant moving the comparison function and writing a test for it.  I did not realize you were using `sort.SliceStable`, which makes that hard. You could use [slices.SortStableFunc](https://pkg.go.dev/slices#SortStableFunc) which makes that kind of testing easier.

coryan (approved):
An optional simplification, I think you missed some `return -> continue` changes to make the testing more informative.

internal/sidekick/rust/annotate.go
westarle: I did this, `sortstablefunc` is much more familiar thank you.

westarle:
> An optional simplification, I think you missed some `return -> continue` changes to make the testing more informative.

I made a couple fixes to address linter issues with the result.

internal/sidekick/rust/annotate_test.go
coryan: D'oh, thanks.

coryan (approved):

PR #2935, opened 2025-11-17
chore: add a structure test for automation CLI

### LINKED ISSUE #2813: automation: add test to ensure correct Librarian image is used for publish-release.yaml
In https://github.com/googleapis/librarian/pull/2812, the change from `tag` to `tag-and-release` was reverted. 

We should add a test case to prevent future regressions where the Librarian image is upgraded when it should be.

JoeWang1127:
Add structure tests for automation CLI, the test will fail if the underlying librarian CLI change but automation CLI remains the same.

Fixes #2813


suztomo (approved):

suztomo:
I'm experimenting approval permission here.

Issue #2874, opened 2025-11-12
all: ensure urgent action on HEAD failures and other critical issues


julieqiu:
Tests have failed at HEAD several times recently. Because we aim to keep the main branch consistently green, we should have a system such that when tests fail at HEAD, the Librarian team will respond immediately.

We should also have a system to address [critical issues](https://github.com/googleapis/librarian/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22%3Arotating_light%3A%20critical%22), and downgrade the labels to `needs fix soon` if the issue is important but can be addressed during business hours.

jskeet:
As part of this, we should consider what "failed at HEAD" means. One recent issue - which was entirely my fault - ended up making HEAD "fail" due to a GCB job failing, which happened to be at HEAD.

In that particular case, it wasn't a problem that it was failing (I was diagnosing a known issue) and the problem was that I'd connected to this repo. But I can imagine other situations where we have jobs which are connected to this repo and fail... if they're production jobs, they should probably be looked at with the same urgency, but may already be filing a bug on failure, in which case we don't want to have a different mechanism that files a duplicate bug. I don't have a suggestion around this yet - it's just an observation.

julieqiu:
If there is a red X next to a commit on the main branch, it is considered a failure at HEAD.

There are many people working in this codebase, so the status signal needs to be unambiguous. We shouldn't have to spend time manually investigate a failing status to decide if it's "real" or just an experimental job. If we want to run experiments, we should have a setup such that it can be done on forks, not against a shared main branch.

In the rare case that we must test on the main branch of this repository, we should have a communication strategy (maybe a pinned issue) that details the experiment, what the expected failure are, and when the experiment will end.

ldetmer:
Currently there is no alert, @JoeWang1127 can you work to add one please? It should alert you, myself and Min.

I have updated the on call rotation to review high priority unassigned issues and assign them (go/bilge-pump-automation) row 27.

PR #2947, opened 2025-11-18
chore(all): add notifications to the issue created by check for failing tests at head action

JoeWang1127:
Add a comment to the issue created by check for failing tests at head action, so that the librarian team will receive a notification.

For #2874


zhumin8 (approved):
LGTM

PR #2960, opened 2025-11-19
chore(internal/config): update config docs with current commands

noahdietz:
The `type Config struct` documentation referred to commands that don't exist anymore, and was missing newer commands that do reference the properties in question.


ldetmer (commented):
Thanks Noah, we don't have a configure command anymore, so just left some updates

noahdietz:
>  we don't have a configure command anymore, so just left some updates

Awesome, thanks for that, merged all!

Issue #2966, opened 2025-11-20
sideflip: migrate Rust to the new Librarian CLI design


julieqiu:
We want to ensure that sidekick will work with the new librarian CLI design. Since the librarian CLI is still under active development / stabilization, use the sideflip command for this experiment.

PR #2968, opened 2025-11-20
feat(internal/command): add command execution package with logging

julieqiu:
An internal/command package is created to execute external commands and print them to stderr before execution, in order to provide progress reports to the user.

Existing call sites from internal/sidekick/external are migrated to use the new package and the old package is deleted.

### DIFF
--- a/internal/sidekick/sidekick/downloads_cache.go
+++ b/internal/sidekick/sidekick/downloads_cache.go
@@ -21,12 +21,12 @@ import (
 	"log/slog"
 	"net/http"
 	"os"
-	"os/exec"
 	"path"
 	"path/filepath"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


internal/sidekick/sidekick/downloads_cache.go
```diff
@@ -76,7 +75,11 @@ func makeSourceRoot(rootConfig *config.Config, configPrefix string) (string, err
 func extractTarball(source, destination string) error {
 	cmd := exec.Command("tar", "-zxf", source)
 	cmd.Dir = destination
```
dbolduc: nit: could maybe use `Run(...)`, if you add `"-C", destination` to the command, but no big deal

dbolduc (approved):

dbolduc (approved):

PR #3009, opened 2025-11-21
fix(dart): Generate team lints compliant output

brianquinlan:
- Remove generated "analysis_options.yaml
- Add `// ignore_for_file: ...` for acceptable lint errors. This allows hand-written code in the same package to follow all the rules
- Change generated code to use `=>` for single line functions
- Sort dev dependencies
- Use relative imports when possible
- Added lots of tests for `calculateImports`

### DIFF
--- a/internal/sidekick/dart/annotate.go
+++ b/internal/sidekick/dart/annotate.go
@@ -457,8 +457,8 @@ func calculateDependencies(packages map[string]bool, constraints map[string]stri
 // calculateImports generates Dart import statements given a set of imports.
 //
 // For example:
-// {"dart:io": true, "package:http/http.dart as http": true} to
-// ["import 'dart:io';", "", "import 'package:http/http.dart' as http;"]
+// `{"dart:io": true, "package:http/http.dart as http": true}` to
+// `{"import 'dart:io';", "", "import 'package:http/http.dart' as http;"}`.
 func calculateImports(imports map[string]bool, curPkgName string, curFileName string) []string {
 	var dartImports []string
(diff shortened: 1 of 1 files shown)


internal/sidekick/dart/templates/lib/main.dart.mustache
```diff
@@ -25,6 +25,14 @@ limitations under the License.
 {{/Codec.DocLines}}
 library;
 
+// ignore_for_file: argument_type_not_assignable {{! dynamic is used for protobuf messages }}
+// ignore_for_file: avoid_unused_constructor_parameters {{! `fromJson` may not use its arguments if the message is empty }}
+// ignore_for_file: camel_case_types {{! Nested messages have the parent/child separated by an underscore }}
+// ignore_for_file: comment_references {{! TODO(#25): improve the generated dartdocs }}
+// ignore_for_file: implementation_imports {{! imports of protobuf and rpc implementations }}
+// ignore_for_file: lines_longer_than_80_chars {{! TODO(#25): improve the generated dartdocs }}
```
natebosch: This one would be fine to ignore in the long-term in my opinion. Some of the lints don't matter much for generated code and I think this one qualifies.

natebosch (approved):

internal/sidekick/dart/templates/lib/main.dart.mustache
brianquinlan: I think that one will get fixed as part of correctly parsing the proto docs.

Issue #2755, opened 2025-11-03
automation: add vulnerability scanning to all images in cloud-sdk-librarian-prod


ldetmer:
We should add vulnerability scanning to all images in cloud-sdk-librarian-prod GCP in the images-prod  artifact registry.  If there is an issue we should open a ticket as assign to component "Cloud Platform > DevTools > Cloud SDK > APIs > Automation > Librarian"  

Please follow steps [here](https://g3doc.corp.google.com/company/teams/fedramp/index.md?cl=head).

ldetmer:
lets wait to add tag till monday: @ldetmer

PR #3010, opened 2025-11-21
chore: switch base image for Dockerfiles to debian12

### LINKED ISSUE #2983: librarian: update to use base images recommended by MOSS
To onboard to vulnerability scanning and get notified (#2755), there are requirements on base images used [here](http://go/autovm-scanner-activation#image-prerequisites). We use use MOSS-compliant base images similar to https://github.com/googleapis/google-cloud-go/blob/main/internal/librariangen/Dockerfile

zhumin8:
To properly turn on vulnerability scan and report, container images needs to meet [prerequisites](go/autovm-scanner-activation#image-prerequisites) and use MOSS complaint images.

Fixes #2983
For #2755

### DIFF
--- a/infra/imagebuilders/cli/Dockerfile
+++ b/infra/imagebuilders/cli/Dockerfile
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 # This stage builds the librariangen binary using the MOSS-compliant base image.
-FROM marketplace.gcr.io/google/debian12:latest AS builder
+FROM marketplace.gcr.io/google/debian12@sha256:326ccf35aa72f7cc8cbd25b69e819a81c5fd9ed675c6c9ffb51f3214a64f23cf AS builder
 
 # Set environment variables for tool versions for easy updates.
 ENV GO_VERSION=1.25.3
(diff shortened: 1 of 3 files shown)


zhumin8:
I wonder if it worth to pin the base image instead of latest? According to [image list](https://pantheon.corp.google.com/artifacts/docker/cloud-marketplace/us/gcr.io/google%2Fdebian12?e=RetailSupportToolLaunch::RetailSupportToolEnabled,RetailDataQualityGaLaunch::RetailDataQualityGaEnabled&mods=allow_workbench_image_override), there is no other tags other than latest. We would need to point to a SHA.

ldetmer (approved):

ldetmer:
> I wonder if it worth to pin the base image instead of latest? According to [image list](https://pantheon.corp.google.com/artifacts/docker/cloud-marketplace/us/gcr.io/google%2Fdebian12?e=RetailSupportToolLaunch::RetailSupportToolEnabled,RetailDataQualityGaLaunch::RetailDataQualityGaEnabled&mods=allow_workbench_image_override), there is no other tags other than latest. We would need to point to a SHA.

It looks like you did this, correct? I think that's the way to go so we don't get random updates.

PR #3016, opened 2025-11-23
refactor(internal/fetch): move tarball download logic from sidekick

julieqiu:
Logic related to downloading tarballs is moved from internal/sidekick/sidekick to internal/fetch to make it reusable across the codebase.


coryan (approved):
We need to improve the testing for errors, but that was bad already.

julieqiu:
I'll file an issue

Issue #2975, opened 2025-11-20
proposal: librarian: prefix existing librarian code and files with `legacy`


julieqiu:
To make it easier to tell legacy Librarian CLI code apart from the new redesign, we should prefix all existing code and files with legacy, and standardize on the `legacylibrarian` binary name everywhere it’s referenced. This means:

- [x] Renaming cmd/librarian to cmd/legacylibrarian
- [ ] Creating an internal/legacylibrarin folder, and moving all of the existing code inside that directory
- [x] Renaming subpackages with the legacy prefix (i.e. internal/config to internal/librarian/legacyconfig), so we introduce new packages with that name

As we find functionality that can be reused, we can move that code back to the top-level internal/ folder.

This will allow us to continue work under the librarian namespace, instead of a new one like [sideflip](https://github.com/googleapis/librarian/issues/2966).

PR #3017, opened 2025-11-23
refactor: rename commands

julieqiu:
Add a legacy prefix to legacy commands. cmd/automation is renamed to
cmd/legacyautomation and cmd/librarian is renamed to cmd/legacylibrarian.

Rename cmd/sideflip to cmd/librarian now that the namespace is available.

Also rename sideflip.go to librarian.go.

Fixes https://github.com/googleapis/librarian/issues/2975

### DIFF
--- a/cmd/librarian/main.go
+++ b/cmd/librarian/main.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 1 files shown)


julieqiu:
Let's leave them for now for reference. We can always delete them later.

PR #3025, opened 2025-11-23
feat(librarian): add generate command

julieqiu:
An initial version of the generate command is added to the librarian CLI, which generates client libraries based on configuration in librarian.yaml.

The command supports two modes:

    librarian generate <library-name>: generates a single library
    librarian generate --all: generates all libraries

A testhelper generator is provided in order to verify the generate command logic without calling real language-specific tooling.

Support for rust.Generate is also implemented.

### DIFF
--- a/internal/language/internal/rust/generate_test.go
+++ b/internal/language/internal/rust/generate_test.go
@@ -46,10 +46,9 @@ func TestGenerate(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	// Verify generated files and their content.
 	for _, test := range []struct {
-		path    string
-		content string
+		path string
+		want string
 	}{
 		{filepath.Join(outDir, "Cargo.toml"), "name"},
 		{filepath.Join(outDir, "Cargo.toml"), "secretmanager"},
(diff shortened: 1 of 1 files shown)


internal/language/internal/rust/generate_test.go
```diff
@@ -0,0 +1,46 @@
...
+		ServiceConfig: "google/cloud/secretmanager/v1/secretmanager_v1.yaml",
+		Channel:       "google/cloud/secretmanager/v1",
+		Version:       "0.1.0",
+		ReleaseLevel:  "preview",
+		CopyrightYear: "2025",
+	}
+	if err := Generate(t.Context(), library, googleapisDir); err != nil {
+		t.Fatal(err)
+	}
+}
```
coryan: Consider verifying that some files are generated (README.md, Cargo.toml, src/lib.rs ?), and maybe that they have the right copyright year?

I would avoid golden files because they generate more toil than capture problems.

ldetmer (commented):
@julieqiu it would be great to see a design for how the flow is going to work end to end, how is the output of this command getting used ?
Also it would be good to know what functionality we expect Librarian will provide and what functionality language will provide (contract).

Its hard to provide feedback without knowing these things.

For example, since I'm not sure what we are doing with the generated files, I'm not sure if we need a repository flag passed in to allow us to commit generated files there.  If that's an orthogonal process I think we need to design it out to make sure it works all together.

internal/librarian/generate.go
```diff
@@ -0,0 +1,93 @@
...
+	if all && libraryName != "" {
+		return errBothLibraryAndAllFlag
+	}
+
+	cfg, err := config.Read(librarianConfigPath)
+	if err != nil {
+		return err
+	}
+	if all {
+		var errs []error
```
coryan: Feels like this could use some refactoring? Maybe a local function?

internal/librarian/generate.go
```diff
@@ -0,0 +1,93 @@
...
+		for _, lib := range cfg.Libraries {
+			// TODO(https://github.com/googleapis/librarian/issues/2966): use
+			// fetch.RepoDir to fetch googleapisDir once implemented
```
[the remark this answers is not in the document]
julieqiu: updated the TODO

internal/librarian/generate.go
julieqiu: done

PR #3026, opened 2025-11-23
feat: add GitHub action for tests that require Rust tools and fix TestBumpVersions

julieqiu:
A new GitHub Actions workflow is added to test sidekick and language functionality that requires Rust tooling. The workflow installs cargo, taplo, and protoc to support tests like TestBumpVersions.

TestBumpVersions is also fixed to check for double quotes in Cargo.toml version fields, matching the actual TOML format.

### DIFF
--- a/.github/workflows/sidekick.yaml
+++ b/.github/workflows/sidekick.yaml
@@ -40,10 +40,7 @@ jobs:
       - name: Install protoc
         run: |
           set -e
-
-          curl -fsSL --retry 5 --retry-delay 15 -o /workspace/.bin/codecovcli https://github.com/getsentry/prevent-cli/releases/download/${_CODECOV_VERSION}/codecovcli_linux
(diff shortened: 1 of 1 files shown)


.github/workflows/sidekick.yaml
```diff
@@ -0,0 +1,52 @@
...
+      - uses: actions/checkout@v5
+      - uses: actions/setup-go@v6
+        with:
+          go-version-file: "go.mod"
+      - name: Display Go version
+        run: go version
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+      - name: Display Rust version
+        run: cargo --version
```
coryan: optional: this is the version of `cargo` (the package manager) not of Rust (the compiler), you need something like:

https://github.com/googleapis/google-cloud-rust/blob/efbcf8baee844946791db8a0012d585b5db63817/.github/workflows/sdk.yaml#L49-L52

Issue #3005, opened 2025-11-21
config: remove configuration from BUILD.bazel


jskeet:
BUILD.bazel files should not be the source of truth for any configuration.

They could *potentially* continue to contain configuration which has its source of truth elsewhere, although for language-specific information that will become increasingly difficult.

Instead:

- Language-neutral information such as `transports` and `rest_numeric_enums` should be part of the service config YAML file
- Language-specific information should live in the config for the language repo

GAPIC generators - and the Librarian code which runs both those and protoc - should be able to consume the relevant information from both of those sources of truth.

julieqiu:
Linking https://github.com/googleapis/librarian/pull/3030#pullrequestreview-3500744711 here for future reference

PR #3030, opened 2025-11-24
feat(internal/config/bazel): add BUILD.bazel parser

julieqiu:
Introduce the bazel package to parse BUILD.bazel files from the googleapis/googleapis repository.

Tests from the implementation at https://github.com/googleapis/google-cloud-go/tree/40e33f63550074964ffde0fc4f5b07308d3a9e5b/internal/librariangen/bazel are copied and modified to fit this package.

Currently, the parser extracts configuration from `go_gapic_library` and `go_proto_library` rules. Support for additional languages will be added over time.

This package will be used to migrate Python and Go off legacylibrarian. It will also support parsing BUILD.bazel files for other languages, with the long-term goal of storing language-specific configuration in `librarian.yaml` and language-neutral configuration in the service config YAML.

For https://github.com/googleapis/librarian/issues/3005

### DIFF
--- a/internal/config/bazel/parser.go
+++ b/internal/config/bazel/parser.go
@@ -18,7 +18,6 @@ package bazel
 import (
 	"fmt"
 	"os"
-	"path/filepath"
 	"strings"
 
 	"github.com/bazelbuild/buildtools/build"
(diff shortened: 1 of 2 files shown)


jskeet (approved):
After discussion: approving to unblock work on Go migration, but would recommend adding a comment saying we should either make this general-purpose or (ideally) remove it entirely.

(I suspect we'll need to make it general purpose for the sake of migrating all languages, *then* remove it when all the information is in librarian.yaml files.)

PR #3078, opened 2025-11-29
feat(sidekick): enhance sidekick model

quirogas:
## Overview
This PR enhances the Sidekick API model and Protobuf parser to explicitly capture Google Cloud Resource annotations (`google.api.resource`, `google.api.resource_definition`, and `google.api.resource_reference`). This data is essential for downstream tools like `surfer` to understand resource identities and relationships.

## Changes
- **Model:** Added `Resource` and `ResourceReference` structs to `internal/sidekick/api`, replacing the simple `IsResourceReference` boolean.
- **Parser:** Updated `internal/sidekick/parser/protobuf.go` to extract full resource metadata from Protobuf options.
- **Cleanup:** Removed duplicate fields and optimized struct layout.
- **Tests:** Added comprehensive tests for resource annotation parsing.

Updates #3079

### DIFF
--- a/internal/sidekick/parser/protobuf.go
+++ b/internal/sidekick/parser/protobuf.go
@@ -578,7 +578,9 @@ func processResourceDefinitions(f *descriptorpb.FileDescriptorProto, result *api
 	}
 }
 
-// TODO(https://github.com/googleapis/librarian/issues/3036): This function needs to be made more robust. For methods that operate on
+// TODO(https://github.com/googleapis/librarian/issues/3036): This function needs
+// to be made more robust. For methods that operate on
+//
 // collections (e.g., a `List` method), the `(google.api.resource_reference)`
 // annotation often uses the `type` field to refer to the *parent* resource
 // (e.g., `cloudresourcemanager.googleapis.com/Project`). The actual resource


internal/sidekick/api/model.go
```diff
@@ -747,8 +750,6 @@ type Field struct {
 	// - For OpenAPI, it is an optional field
 	// - For OpenAPI, it has format == "uuid"
 	AutoPopulated bool
-	// IsResourceReference is true if the field is annotated with google.api.resource_reference.
-	IsResourceReference bool
```
quirogas: This looks redundant now.  If we are going to add `ResourceReference` then we should add a helper:

```
func (f *field) IsResourceReference() bool {
    return f.ResourceReference != nil
}
```
-- @coryan

internal/sidekick/api/model.go
```diff
@@ -890,3 +894,46 @@ type OneOf struct {
...
 }
+
+// Resource is a fundamental building block of an API, representing an individually-named entity (a "noun").
+// Resources are typically organized into a hierarchy, where each node is either a simple resource or a collection of resources.
+// This definition is based on AIP-121 (https://google.aip.dev/121).
+type Resource struct {
+	// Type identifies the kind of resource (e.g., "cloudresourcemanager.googleapis.com/Project").
+	// This string is globally unique and identifies the type of resource across Google Cloud.
+	Type string
+	// Pattern is the resource name pattern.
```
quirogas: That comment is not very useful. Can you say something about where the namespace for these types? Are they always messages in the current data model? Could the be imported from a different data model?

-- @coryan

internal/sidekick/api/model.go
```diff
@@ -890,3 +894,46 @@ type OneOf struct {
...
+	Plural string
+	// Singular is the singular form of the resource name.
+	// For example, for a "Book" resource, Singular would be "book".
+	Singular string
+	// Self points to the Message that defines this resource.
+	// This creates a back-reference for navigating the API model,
+	// allowing a Resource definition to access its originating Message structure.
+	Self *Message
+	// Language specific annotations.
+	Codec any
```
quirogas: Leave room for the codecs to annotate these things. Add Codec any.

-- @coryan

internal/sidekick/api/model.go
```diff
@@ -890,3 +894,46 @@ type OneOf struct {
...
+}
+
+// ResourceReference describes a field's relationship to another resource type.
+// It acts as a foreign key, indicating that the field's value identifies an instance of another resource.
+// This relationship is established via the `google.api.resource_reference` annotation in Protobuf.
```
quirogas: As above, make these comments meaningful, not just "not empty".

-- @coryan

internal/sidekick/api/model.go
```diff
@@ -890,3 +894,49 @@ type OneOf struct {
...
+	// implies interaction with a specific child type (e.g., "Book" within that shelf).
+	ChildType string
+	// Language specific annotations.
+	Codec any
+}
+
+// IsResourceReference returns true if the field is annotated with google.api.resource_reference.
+func (f *Field) IsResourceReference() bool {
+	return f.ResourceReference != nil
+}
```
quirogas: This is very similar to `.ID`, why do we need both?

-- @coryan

internal/sidekick/parser/protobuf.go
```diff
@@ -533,6 +539,74 @@ func processMessage(state *api.APIState, m *descriptorpb.DescriptorProto, mFQN,
...
+	message.Resource = &api.Resource{
+		Type:     res.GetType(),
+		Pattern:  res.GetPattern(),
+		Plural:   res.GetPlural(),
+		Singular: res.GetSingular(),
+		Self:     message,
+	}
+}
+
+func processResourceDefinitions(f *descriptorpb.FileDescriptorProto, result *api.API) {
```
quirogas: These changes are missing tests in `protobuf_test.go`.

-- @coryan

quirogas (commented):
I've addressed all comments from the previous review, and I've also added those comments and tagged their original authors.

PR #3092, opened 2025-12-01
feat(internal/rust): run rustfmt and taplo fmt after generation

julieqiu:
Format generated Rust files with rustfmt and Cargo.toml with taplo fmt. This caught an issue with TestGenerate which is now fixed.

### DIFF
--- a/internal/librarian/internal/rust/generate_test.go
+++ b/internal/librarian/internal/rust/generate_test.go
@@ -33,10 +33,17 @@ func TestGenerate(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	outDir := t.TempDir()
+	// Change to testdata directory so cargo fmt can find Cargo.toml
+	workspaceDir, err := filepath.Abs("testdata")
+	if err != nil {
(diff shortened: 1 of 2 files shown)


internal/language/internal/rust/generate.go
```diff
@@ -47,7 +49,25 @@ func Generate(ctx context.Context, library *config.Library, sources *config.Sour
...
+		return err
+	}
+	if len(rsFiles) > 0 {
+		// rustfmt defaults to 2015 edition when run directly on files. Specify
+		// 2024 to match the edition in Cargo.toml.
+		args := append([]string{"--edition", "2024"}, rsFiles...)
+		if err := command.Run("rustfmt", args...); err != nil {
+			return err
+		}
+	}
```
coryan: Do we know what the package name is? If we do we can run this:

```suggestion
    if err := command.Run("cargo", "fmt", "-p", ${thing that has the Rust package name}); err != nil {
        return err
    }
```

coryan (approved):

PR #3096, opened 2025-12-01
feat(internal/serviceconfig): add Find to add default paths

julieqiu:
FindServiceConfig locates the service config file for an API path by scanning for YAML files containing `type: google.api.Service` in the first 5 lines, while skipping `_gapic.yaml` files.

The generate command now automatically populates the ServiceConfig field for APIs that don't have one explicitly configured. This reduces manual configuration when adding new libraries, when we can determine the service config path from the API path.


internal/serviceconfig/testdata/googleapis/google/cloud/speech/v1/speech_v1.yaml
coryan: Why do we need two copies of this testdata file? Can we share them?

PR #3109, opened 2025-12-02
feat(librarian): add repometadata package

jskeet:
A repometadata package is added to librarian, allowing .repo-metadata.json files to be generated.

The `DescriptionOverride` field that was previously part of `RustCrate` is moved to `Library` so it can be used by all languages (with appropriate changes to Rust tests).


internal/repometadata/repometadata_test.go
```diff
@@ -0,0 +1,318 @@
...
+		APIShortname:         "secretmanager",
+		APIDescription:       "Stores sensitive data such as API keys, passwords, and certificates.\nProvides convenience while improving security.",
+	}
+
+	if diff := cmp.Diff(want, got); diff != "" {
+		t.Errorf("mismatch (-want +got):\n%s", diff)
+	}
+}
+
+func TestGenerateRepoMetadata_WithOverrides(t *testing.T) {
```
jskeet: I took this from the librarianx repo, where there are two separate test functions - but I can refactor it to a table-driven test if you'd prefer.

internal/config/config.go
```diff
@@ -125,6 +125,13 @@ type Library struct {
 	// overrides Default.Transport.
 	Transport string `yaml:"transport,omitempty"`
 
+	// DescriptionOverride overrides the library description.
```
julieqiu: I would split this into a separate PR since it seems unrelated to repometadata

internal/repometadata/repometadata.go
```diff
@@ -0,0 +1,270 @@
...
+	if library.DescriptionOverride != "" {
+		// Use override from library configuration
+		metadata.APIDescription = library.DescriptionOverride
+	} else if svcCfg.GetDocumentation() != nil && svcCfg.GetDocumentation().GetSummary() != "" {
+		// Fall back to service YAML documentation
+		metadata.APIDescription = strings.TrimSpace(svcCfg.GetDocumentation().GetSummary())
+	}
+
+	// Set default release level if not specified
+	if metadata.ReleaseLevel == "" {
```
julieqiu: I think you should be able to use library.ReleaseLevel and that will always be accurate

internal/config/config.go
jskeet: It's used by repometadata. (The original version used a Python-specific description override; this is now using a general one.)

I could:

- Leave it as-is, with that extra context
- Change this PR to not use an override at all, and then introduce a second PR which moves it from being Rust-specific to being general, and use it in repometadata at that point
- Introduce a new PR which *just* moves it from being Rest-specific to being general, and make this the "second PR".

Thoughts?

internal/repometadata/repometadata.go
```diff
@@ -0,0 +1,270 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/googleapis/librarian/internal/config"
+	"github.com/googleapis/librarian/internal/serviceconfig"
+)
+
+// RepoMetadata represents the .repo-metadata.json file structure.
+type RepoMetadata struct {
```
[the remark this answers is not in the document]
jskeet: I think they're already in alphabetical order, aren't they? (Wondering what I'm missing now...)

I don't know of any ordering requirements, and I really hope there aren't any.

internal/repometadata/repometadata.go
```diff
@@ -0,0 +1,270 @@
...
+	// but with the ability to override.
+	defaultVersion := library.DefaultVersionOverride
+	if defaultVersion == "" {
+		defaultVersion = selectDefaultVersion(apiPaths)
+	}
+
+	// Build client documentation URL based on language
+	clientDocURL := buildClientDocURL(language, extractNameFromAPIID(svcCfg.GetName()))
+
+	// Create metadata
```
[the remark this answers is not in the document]
jskeet: Done.

internal/repometadata/repometadata.go
```diff
@@ -0,0 +1,270 @@
...
+		ClientDocumentation: clientDocURL,
+		ReleaseLevel:        library.ReleaseLevel,
+		Language:            language,
+		LibraryType:         "GAPIC_AUTO",
+		Repo:                repo,
+		DistributionName:    library.Name,
+		DefaultVersion:      defaultVersion,
+	}
+
+	// Add optional fields if available
```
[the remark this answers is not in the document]
jskeet: Done.

internal/repometadata/repometadata.go
```diff
@@ -0,0 +1,270 @@
...
+}
+
+// GenerateRepoMetadata generates the .repo-metadata.json file by parsing the service YAML.
+func GenerateRepoMetadata(library *config.Library, language, repo, serviceConfigPath, outdir string, apiPaths []string) error {
+	svcCfg, err := serviceconfig.Read(serviceConfigPath)
+	if err != nil {
+		return fmt.Errorf("failed to read service config: %w", err)
+	}
+
+	// Select the default version from all API paths, preferring stable versions,
```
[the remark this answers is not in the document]
jskeet: Indeed, I'm pretty sure it will be used in Python to work out the ordering of things.

Are you suggesting "leave the code here for now, merge this PR, but then add another PR to move it" or remove everything about default versions for this first PR?

internal/repometadata/repometadata.go
```diff
@@ -0,0 +1,270 @@
...
+
+// extractNameFromAPIID extracts the service name from the API ID.
+// Example: "secretmanager.googleapis.com" -> "secretmanager".
+func extractNameFromAPIID(apiID string) string {
+	parts := strings.SplitN(apiID, ".", 2)
+	if len(parts) > 0 {
+		return parts[0]
+	}
+	return apiID
+}
```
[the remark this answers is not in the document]
jskeet: Done.

internal/repometadata/repometadata.go
```diff
@@ -0,0 +1,270 @@
...
+	}
+	versionNum := strings.TrimPrefix(version, "v")
+	// Check if it contains only digits (stable) or has alpha/beta (not stable)
+	for _, r := range versionNum {
+		if r < '0' || r > '9' {
+			return false
+		}
+	}
+	return true
+}
```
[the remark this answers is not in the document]
jskeet: Done (in a slightly different way) and added a test - albeit a very odd test...

internal/config/config.go
julieqiu: I would do 2 or 3

internal/repometadata/repometadata.go
```diff
@@ -0,0 +1,262 @@
...
+	// Set API description from override or service YAML
+	if library.DescriptionOverride != "" {
+		// Use override from library configuration
+		metadata.APIDescription = library.DescriptionOverride
+	} else if svcCfg.GetDocumentation() != nil && svcCfg.GetDocumentation().GetSummary() != "" {
+		// Fall back to service YAML documentation
+		metadata.APIDescription = strings.TrimSpace(svcCfg.GetDocumentation().GetSummary())
+	}
+
+	// Write metadata file
```
julieqiu: this comment doesn't seem very useful, I would just remove it

internal/repometadata/repometadata_test.go
```diff
@@ -0,0 +1,378 @@
...
+	// Create a test service YAML
+	yamlContent := `type: google.api.Service
```
julieqiu: consider moving this and above into a testdata/ folder

we may want a general internal/testdata that is shared across different packages

jskeet (commented):
(Interim review comments - I still need to move some stuff to a testdata folder.)

jskeet:
> > This also moves the existing Rust-specific DescriptionOverride to Library (as we want it in repometadata).
> 
> There's a lot happening in this PR beyond just adding repometadata. I would describe the changes. See https://github.com/googleapis/librarian/blob/main/CONTRIBUTING.md#commit-messages

There's now *less* happening (as I've removed default versions) but I've updated the PR comment. (I think the only non-repometadata-related change is to DescriptionOverride...)

jskeet:
Note (as per chat) - I've only just noticed that https://github.com/googleapis/librarian/blob/main/internal/config/repometadata.go exists. (It didn't when I created the prototype that this PR is derived from.) Happy to take guidance about whether that should be deleted as part of this PR, or whether to remove the struct in my PR (but keep the code in its own package) etc.

Issue #3076, opened 2025-11-26
librarian: implement allowlist


julieqiu:
We don't need to list every API in librarian.yaml if it follows the standard naming and one_library_per conventions for that library. 

Instead, traverse googleapis/googleapis for every API that should be generated, and generate them with default settings.

For now if an API should not be generated, set generate.disabled = true in the librarian.yaml. In the future we will want to have some sort of global config since this likely applies to all languages.

julieqiu:
Some decisions on https://github.com/googleapis/librarian/pull/3112#issuecomment-3610076566

PR #3112, opened 2025-12-02
feat(internal/librarian): auto-discover APIs during generate --all

julieqiu:
When running `generate --all`, scan the googleapis directory for APIs (directories containing .proto files) and add them as library entries if not already covered by librarian.yaml. These libraries do not have an entry in the librarian.yaml, since their information can be derived based on standard conventions.

Service configs are automatically looked up for each API path if one is not provided in the librarian.yaml.

Fixes https://github.com/googleapis/librarian/issues/3076

### DIFF
--- a/internal/librarian/generate.go
+++ b/internal/librarian/generate.go
@@ -256,9 +256,12 @@ func containsProtoFiles(dir string) bool {
 // defaultRustLibraryName derives an API path from a Rust library name.
 // For example: google-cloud-foo-v1 -> google/cloud/foo/v1
 //
-// TODO(https://github.com/googleapis/librarian/issues/3058): See
+// TODO(https://github.com/googleapis/librarian/issues/3058): see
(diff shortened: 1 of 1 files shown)


internal/librarian/generate.go
```diff
@@ -169,3 +177,88 @@ func cleanOutput(dir string, keep []string) error {
...
+
+// defaultRustLibraryName derives an API path from a Rust library name.
+// For example: google-cloud-foo-v1 -> google/cloud/foo/v1
+//
+// TODO(https://github.com/googleapis/librarian/issues/3058): See
+// go/cloud-rust:on-crate-names for overrides and make sure they are handled
+// properly.
+func defaultRustLibraryName(name string) string {
+	return strings.ReplaceAll(name, "-", "/")
+}
```
coryan: Odd to have a Rust-specific thing here?  I thought this package was language-neutral?

Issue #2764, opened 2025-11-03
librarian: implement stable-relative versioning for preview track


noahdietz:
The `librarian release` command, when run on the `preview` branch, suggests a new SemVer bump as follows:

* a minor bump on the "stable" (i.e. `main` branch) SemVer value if higher than current "preview" (i.e. `preview` branch) version
* a prerelease iteration bump on the "preview" (i.e. `preview` branch SemVer value if high than current "stable" (i.e. `stable` branch) version

For example, if the latest "preview" release is `1.3.0-preview.1` and latest "stable" release is `1.2.3`, the suggested bump should be to `1.3.0-preview.2`. If the latest "preview" release is `1.3.0-preview.1` and latest "stable" release is `1.3.0`, the suggested bump should be to `1.4.0-preview.1`.

PR #3114, opened 2025-12-02
fix(internal/semver): change PrereleaseNumber to int

noahdietz:
Changes `Version.PrereleaseNumber` to be an `*int` instead of a `string` - this is easier to work with everywhere after `Parse` is called and is explicit in communicating the presence of a prerelease number. Moves `strconv.Atoi` calls from `incrementPrerelease` to `Parse` where it _should_ fail if not a valid number. Refactors test cases accordingly.

Refactors `DeriveNext` to pull change-level-based version bumping logic into an un-exported `Version` method named `bump` in preparation for its reuse in a future function related to #2764.

Adds a missing test case for `Version.Compare` that shows it will favor a prerelease with a higher version core than a non-prerelease version with a lower version core, as per the spec ([ref](https://semver.org/#spec-item-11)).

Note: I'm not entirely sure the `Version` methods (`Compare`, `String`, `bump`, `incrementPrerelease`) need to be using pointer receivers. These `Version` instances are not passed around and producing a new, updated copy of a version when calling a mutation method would be a bit easier to work with. If that sounds interesting, I can refactor in another PR.

### DIFF
--- a/internal/semver/semver.go
+++ b/internal/semver/semver.go
@@ -196,26 +196,21 @@ func (v *Version) bump(highestChange ChangeLevel) {
 	}
 
 	// Bump the version core.
-	if v.Major == 0 {
-		// breaking change and feat result in minor bump for pre-1.0.0
-		if highestChange == Major || highestChange == Minor {
-			v.Minor++
-			v.Patch = 0
-		} else {
-			v.Patch++
-		}
-	} else {
-		switch highestChange {
-		case Major:
-			v.Major++
-			v.Minor = 0
-			v.Patch = 0
-		case Minor:
-			v.Minor++
-			v.Patch = 0
-		case Patch:
-			v.Patch++
-		}
+	// Breaking changes and feat result in minor bump for pre-1.0.0 versions.
+	if (v.Major == 0 && highestChange == Major) || highestChange == Minor {
+		v.Minor++
+		v.Patch = 0
+		return
+	}
+	if highestChange == Patch {
+		v.Patch++
+		return
+	}
+	if highestChange == Major {
+		v.Major++
+		v.Minor = 0
+		v.Patch = 0
+		return
 	}
 }
 


internal/semver/semver.go
```diff
@@ -130,39 +142,76 @@ func (v *Version) Compare(other *Version) int {
...
 // String formats a Version struct into a string.
 func (v *Version) String() string {
 	version := fmt.Sprintf("%d.%d.%d", v.Major, v.Minor, v.Patch)
 	if v.Prerelease != "" {
-		version += "-" + v.Prerelease + v.PrereleaseSeparator + v.PrereleaseNumber
+		version += "-" + v.Prerelease
+		if v.PrereleaseNumber != nil {
+			version += v.PrereleaseSeparator + strconv.Itoa(*v.PrereleaseNumber)
+		}
 	}
```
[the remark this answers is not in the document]
noahdietz: These are not supported. Resolving.

noahdietz:
> However, I've identified a couple of bugs in the Compare and String methods that were introduced by these changes, particularly affecting versions with purely numeric prereleases (e.g., 1.2.3-1)

These types of prereleases are not supported, so we do not need to include them. This is already only a semi-compliant implementation, we only implemented what we needed.

internal/semver/semver.go
```diff
@@ -33,14 +33,18 @@ type Version struct {
 	// its version (e.g., ".").
 	PrereleaseSeparator string
 	// PrereleaseNumber is the numeric part of the pre-release string (e.g., "1", "21").
```
zhumin8: Can you also update comment

internal/semver/semver.go
```diff
@@ -130,39 +142,76 @@ func (v *Version) Compare(other *Version) int {
...
 // incrementPrerelease increments the pre-release version number, or appends
 // one if it doesn't exist.
-func (v *Version) incrementPrerelease() error {
-	if v.PrereleaseNumber == "" {
+func (v *Version) incrementPrerelease() {
+	if v.PrereleaseNumber == nil {
 		v.PrereleaseSeparator = "."
-		v.PrereleaseNumber = "1"
-		return nil
+		v.PrereleaseNumber = new(int)
```
zhumin8: nit: Perhaps add a comment to this block along the lines of  "If PrereleaseNumber is nil, it's initialized to 0 and then incremented to 1." It is was not immediate obvious to me.

zhumin8 (approved):
LGTM.

internal/semver/semver.go
```diff
@@ -130,39 +142,76 @@ func (v *Version) Compare(other *Version) int {
 		return 1
 	}
 	// prerelease number (e.g. "alpha1" vs "alpha2")
```
miguelvelezsa: nit: WDYT about creating a comparator function only for prereleasenumber and use it here? :)

internal/semver/semver.go
noahdietz: Good call. PTAL!

internal/semver/semver.go
noahdietz: Good call. PTAL!

internal/semver/semver.go
```diff
@@ -130,39 +142,76 @@ func (v *Version) Compare(other *Version) int {
...
+		v.PrereleaseNumber = new(int)
 	}
-	num, err := strconv.Atoi(v.PrereleaseNumber)
-	if err != nil {
-		// This should not happen if Parse is correct.
-		return fmt.Errorf("invalid prerelease version: %w", err)
+	*v.PrereleaseNumber++
+}
+
+func (v *Version) bump(highestChange ChangeLevel) {
```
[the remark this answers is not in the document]
julieqiu: ditto re: Reference in new issue" button (although in case I though you can just use that function instead of writing new code?)

internal/semver/semver.go
```diff
@@ -33,14 +33,18 @@ type Version struct {
...
 }
 
 // semverRegex defines format for semantic version.
 // Regex from https://semver.org/, with buildmetadata part removed.
 // It uses named capture groups for major, minor, patch, and prerelease.
 var semverRegex = regexp.MustCompile(`^(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?$`)
 
+// unsegmentedPrereleaseRegexp extracts the prerelease number, if present, in the
+// prerelease portion of the SemVer version parsed by [semverRegex].
+var unsegmentedPrereleaseRegexp = regexp.MustCompile(`^(.*?)(\d+)$`)
```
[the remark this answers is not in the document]
julieqiu: ack

internal/semver/semver.go
```diff
@@ -33,14 +33,18 @@ type Version struct {
...
+
 // Parse parses a version string into a Version struct.
 func Parse(versionString string) (*Version, error) {
 	matches := semverRegex.FindStringSubmatch(versionString)
```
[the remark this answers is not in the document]
noahdietz: Yes might be able to but I'd like to explore that in a separate issue/PR directed at "replace dupe logic with x/mod/semver" in general.

julieqiu (approved):

internal/semver/semver.go
noahdietz: I'd rather not refactor this further as it looks like we might be able to replace this with golang.org/x/mod/semver

From a style perspective, I'm not sure that encapsulating the prerelease comparison logic in another function is worthwhile. The logic won't be used elsewhere, and it will add a layer of indirection to the code that makes reading the Compare logic harder (need to click out to another function definition). It is clearer to keep it here, inline - the only place it is used - to make reading the SemVer comparison/precedence logic quite clear.

julieqiu (approved):

Issue #3142, opened 2025-12-03
serviceconfig: hardcode API config extensions


julieqiu:
There are a number of fields we eventually want to move into service config files. That migration will take time, so as an interim step, we should hardcode these values directly in the Librarian repository in order to create a temporary representation generators can read without requiring a full service-config migration.

This structure should be simple and easy to replace once we start work unified configs: https://github.com/googleapis/librarian/milestone/72

The Rust generator currently depends on these fields, which would ideally be the same across all languages:
- release_level
- service config when it is not in the channel directory (examples: https://github.com/googleapis/librarian/issues/3123)
- title-override
- list of APIs to generate

Python and Go additionally rely on:
- REST numeric enums (https://github.com/googleapis/librarian/issues/3005)
- transports
- `*_grpc_service_config.json` (https://github.com/googleapis/librarian/issues/3004)

PHP and Java additionally rely on:
- *`_gapic.yaml` (https://github.com/googleapis/librarian/issues/3041)

JoeWang1127:
We've moved some configurations from `BUILD.bazel` to `sdk.yaml` but we still use service config and gapic yaml.

PR #3148, opened 2025-12-03
feat(internal/serviceconfig): add GRPCServiceConfig map

julieqiu:
serviceconfig.GRPCServiceConfig is added, which provides a map of the channel to its *_grpc_service_config.json. In a future PR, we will create a function to derive the filename based on the channel if it follows a standard naming convention, so that only exceptions are listed.

For https://github.com/googleapis/librarian/issues/3142

### DIFF
--- a/internal/serviceconfig/grpcserviceconfig.go
+++ b/internal/serviceconfig/grpcserviceconfig.go
@@ -14,8 +14,8 @@
 
 package serviceconfig
 
-// grpcServiceConfigMap maps channels to their gRPC service config filename.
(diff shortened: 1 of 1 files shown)


julieqiu:
From conversation with @jskeet over chat - we don't need this because given an API path, "there should be exactly one file matching this glob" 

https://github.com/googleapis/google-cloud-dotnet/blob/main/tools/Google.Cloud.Tools.ReleaseManager/GenerateApisCommand.cs#L251 does not hardcode it today. Other languages likely should not either.

Closing this PR.

PR #3163, opened 2025-12-04
fix: add call to librarian tidy from migrate-sidekick command

ldetmer:
Invoke librarian runTidy from the migrate-sidekick command.  This exports the librarian function runTidy to give migrate-sidekick command access to it.  Additionally it renames the default librarian yaml file in migrate-sidekick to be librarian.yaml to align with default file in librarian.go.


devtools/cmd/migrate-sidekick/main_test.go
```diff
@@ -414,3 +417,47 @@ func TestBuildConfig(t *testing.T) {
 		})
 	}
 }
+
+func TestRunMigrateCommand(t *testing.T) {
+	t.Parallel()
```
julieqiu: I don't think you need this

devtools/cmd/migrate-sidekick/main_test.go
```diff
@@ -414,3 +417,47 @@ func TestBuildConfig(t *testing.T) {
...
+		wantErr string
+	}{
+		{
+			name: "success",
+			path: "testdata/root-sidekick/success",
+		},
+		{
+			name:    "tidy_command_fails",
+			path:    "testdata/tidy-fails",
+			wantErr: "librarian tidy command failed",
```
julieqiu: if you want to check for a specific error, I would use a named error (https://pkg.go.dev/errors#New), and check with https://pkg.go.dev/errors#Is, so that if the string changes tests don't break. Checking for error strings can be brittle

devtools/cmd/migrate-sidekick/main_test.go
```diff
@@ -414,3 +417,47 @@ func TestBuildConfig(t *testing.T) {
...
+		},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			t.Parallel()
```
julieqiu: I don't think you need this

devtools/cmd/migrate-sidekick/main_test.go
```diff
@@ -414,3 +417,47 @@ func TestBuildConfig(t *testing.T) {
...
+				"migrate-sidekick",
+				"-repo", test.path,
+			}
+
+			err := run(args)
+
+			if test.wantErr != "" && !strings.Contains(err.Error(), test.wantErr) {
+				t.Errorf("expected error %s, got: %v", test.wantErr, err)
+			}
+
```
julieqiu: Currently I think that if test.wantErr is set but err is nil, the code will panic. I also think that if err is not nil and test.wantErr is not set, the error will not be caught.

I would write it like this:

```
if err := run([]string{"migrate-sidekick", "-repo", test.path}); err != nil {
    if test.wantErr == "" {
        t.Fatal(err)
    }
    if !strings.Contains(err.Error(), test.wantErr) {
        t.Fatalf("expected error containing %q, got: %v", test.wantErr, err)
    }
}
if test.wantErr != "" {
   t.Fatalf("expected error containing %q, got nil", test.wantErr)
}
```

devtools/cmd/migrate-sidekick/main_test.go
ldetmer: removed for all tests

devtools/cmd/migrate-sidekick/main_test.go
ldetmer: thanks for the pointer, updated code

devtools/cmd/migrate-sidekick/main.go
```diff
@@ -115,6 +117,11 @@ func run(args []string) error {
 	}
 	slog.Info("Wrote config to output file", "path", outputPath)
 
+	if err := librarian.RunTidy(); err != nil {
+		slog.Error(errTidyFailed.Error(), "error", err)
```
JoeWang1127: Why do we need this error log?

PR #3166, opened 2025-12-04
fix(dart): Fix decoding of `repeated double` fields.

brianquinlan:
1. Replace having a bespoke `repeated` deserializer per primitive type, e.g. `decodeListBytes` with a list comprehension using the primitive decoder , e.g.
`decodeBytes`.

So the generated code looks like:

```dart
switch (json['foo')) {
  null => [],
  List<Object?> $1 => [for (final i in $1) decodeBytes(i)],
  _ => throw FormatException('"foo" is not a list')
}
```

2. Do the same with `map` except use a map comprehension and decode both the key and value (I think that key decoding is incorrect because non-string keys are supported by proto - but that issue existed previously as well)

3. Expect that all decoders, e.g. `decodeBytes` return the correct type and that is not nullable. That means that the `null` guards have to be placed in the generated code. But it also means `!` isn't needed in the above list/map comprehensions.

```diff
- decodeBytes(json['bytes']) ?? Uint8List(0)
+ switch (json['bytes']) { null => Uint8List(0), Object $1 => decodeBytes($1)}
```

4. Change the signature of all decoders to be `Object?` so that we don't rely on `dynamic` to allow calls to succeed, e.g.

```diff
- factory MyEnum.fromJson(String json) => MyEnum(json);
+ factory MyEnum.fromJson(Object? json) => MyEnum(json as String);
```

5. Remove `// ignore_for_file: argument_type_not_assignable` from generated code because it is no longer needed after 1-4.

### DIFF
--- a/internal/sidekick/dart/annotate_test.go
+++ b/internal/sidekick/dart/annotate_test.go
@@ -857,14 +857,41 @@ func TestCreateFromJsonLine(t *testing.T) {
 			&api.Field{Name: "bytes", JSONName: "bytes", Typez: api.BYTES_TYPE},
 			"switch (json['bytes']) { null => Uint8List(0), Object $1 => decodeBytes($1)}",
 		}, {
-			&api.Field{Name: "int32", JSONName: "int32", Typez: api.INT32_TYPE},
-			"switch (json['int32']) { null => 0, Object $1 => decodeInt($1)}",
+			&api.Field{Name: "double", JSONName: "double", Typez: api.DOUBLE_TYPE},
+			"switch (json['double']) { null => 0, Object $1 => decodeDouble($1)}",
 		}, {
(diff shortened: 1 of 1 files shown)


brianquinlan:
> > 7. Change the signature of all decoders to be `Object?` so that we don't rely on `dynamic` to allow calls to succeed, e.g.
> 
> What are the consequences if we change it to `Object` instead of `Object?`? As a general rule I find it's useful to push null checks as close to the source of the null as possible.

The problem is that there is (at least) one type that is non-null but has `null` as its serialized representation: [`NullValue`](https://protobuf.dev/reference/protobuf/google.protobuf/#null-value). 

I decided on this approach rather than special-casing the types that have a `null` representation.

natebosch (approved):

PR #3169, opened 2025-12-04
fix(internal/semver): refactor Version.Parse with x/mod/semver

### LINKED ISSUE #3127: internal/semver: use golang.org/x/mod/semver to replace regex parsing
I think you can use `golang.org/x/mod/semver` to replace this regex and simplify some of the logic below

_Originally posted by @julieqiu in https://github.com/googleapis/librarian/pull/3114#discussion_r2582846730_

noahdietz:
Replaces `Version.Parse` regexp parsing logic with some golang.org/x/mod/semver helpers, and hand spun parsing logic.

The `x/mod/semver` APIs do not provide sufficient capabilities to easily replace our previous parsing logic - there has to be a certain amount of handwritten code to parse out the segments we need. By comparison, the regexp used originally pulled the matches out directly so that we didn't have to do string spitting and trimming.

Notably, we support a SemVer v1 flavor of prerelease formatting which allows for a prerelease number to 1. have leading zeros and 2. be joined to the prerelease string (i.e. not separated by a `"."`) ([spec](https://semver.org/spec/v1.0.0.html#spec-item-4)). `x/mod/semver` targets SemVer 2.0.0. Thus, we must retain custom prerelease parsing logic. That said, C#, the primary user of SemVer 1.0.0 prerelease number format, will want to migrate to SemVer 2.0.0, so we wont need to maintain this forever.

Fixes #3127

### DIFF
--- a/internal/semver/semver.go
+++ b/internal/semver/semver.go
@@ -40,9 +40,9 @@ type Version struct {
 	PrereleaseNumber *int
 }
 
-// unsegmentedPrereleaseRegexp extracts the prerelease number, if present, in the
-// prerelease portion of the SemVer version parsed by [semverRegex].
-var unsegmentedPrereleaseRegexp = regexp.MustCompile(`^(.*?)(\d+)$`)
+// semverV1PrereleaseNumberRegexp extracts the prerelease number, if present, in
+// the prerelease portion of the SemVer 1.0.0 version string.
+var semverV1PrereleaseNumberRegexp = regexp.MustCompile(`^(.*?)(\d+)$`)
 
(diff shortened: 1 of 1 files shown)


noahdietz:
> I'm not sure that this refactor is worth doing. It seems low value given the SemVer spec version differences.

At some point C# will want to move onto SemVer 2.0.0 and we won't need to retain SemVer 1.0.0 prerelease number mgmt anymore. At that point we can clean this code up some more.

I do like the `semver.Canonical`  helper as well. starting to think that this is worth it.

julieqiu (approved):

Issue #3072, opened 2025-11-26
librarian: add create command for Rust


julieqiu:
Add a new `librarian create` command:

This should do the same thing as https://github.com/googleapis/librarian/blob/9a37edc3c97115c755010d6bbb3b1ac23c8e1c2d/internal/sidekick/sidekick/rust_generate.go#L49

The Rust specific command should live in internal/rust. Create will end up calling Generate:
https://github.com/googleapis/librarian/blob/9a37edc3c97115c755010d6bbb3b1ac23c8e1c2d/internal/language/internal/rust/generate.go#L32

PR #3244, opened 2025-12-12
feat(create): add default argument logic and match specs and adds entry to librarian.yaml

ldetmer:
1) Refactor arguments to match [rust spec]([go/sdk-librarian-cli](http://goto.google.com/sdk-librarian-cli))
This includes:
a) making library name an argument instead of a flag
b) setting default values for output, spec source - this is following generate format that contains default logic per language

2) Added functionality to add entry to Librarian yaml file.

3) moved language agnostic logic outside of rust switch

For #3072

### DIFF
--- a/internal/librarian/create.go
+++ b/internal/librarian/create.go
@@ -127,10 +127,9 @@ func deriveOutput(output string, cfg *config.Config, libraryName string, specSou
 			}
 			if specSource != "" {
 				return defaultOutput(language, specSource, cfg.Default.Output), nil
-			} else {
-				libOutputDir := strings.ReplaceAll(libraryName, "-", "/")
-				return defaultOutput(language, libOutputDir, cfg.Default.Output), nil
 			}
+			libOutputDir := strings.ReplaceAll(libraryName, "-", "/")
+			return defaultOutput(language, libOutputDir, cfg.Default.Output), nil
 		}
 	default:
 		return defaultOutput(language, specSource, cfg.Default.Output), nil


internal/librarian/create.go
```diff
@@ -80,36 +78,86 @@ func runCreate(ctx context.Context, name, specSource, serviceConfig, output, spe
...
+	}
+	return output, nil
+}
 
-		// TODO: port over sidekick rustGenerate logic to create a new librarian
-		slog.InfoContext(ctx, "Creating new Rust library", "name", name, "specSource", specSource, "serviceConfig", serviceConfig, "output", output, "specFormat", specFormat)
-		return nil
-	default:
-		return errUnsupportedLanguage
+func addLibraryToLibrarianConfig(rootConfig *config.Config, name string, output string, specificationSource string, serviceConfig string, specificationFormat string) error {
```
JoeWang1127: nit: only one string is needed in the func signature.

internal/librarian/create.go
```diff
@@ -80,36 +78,86 @@ func runCreate(ctx context.Context, name, specSource, serviceConfig, output, spe
...
+	lib := config.Library{
```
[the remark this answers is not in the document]
ldetmer: thanks for the pointer!

internal/librarian/create.go
```diff
@@ -80,36 +78,86 @@ func runCreate(ctx context.Context, name, specSource, serviceConfig, output, spe
...
-				return errOutputFlagRequired
+				return "", errOutputFlagRequired
+			}
+			if specSource != "" {
+				return path.Join(cfg.Default.Output, strings.TrimPrefix(specSource, googlePrefix)), nil
+			} else {
+				packageDir := strings.ReplaceAll(libraryName, "-", "/")
+				return path.Join(cfg.Default.Output, strings.TrimPrefix(packageDir, googlePrefix)), nil
 			}
-			output = rust.DefaultOutput(specSource, cfg.Default.Output)
```
zhumin8: I am not sure this is language agnostic. Maybe use this from generate so we centralize where to edit in the future?
https://github.com/googleapis/librarian/blob/0980271836316c9166a6d2b773133c7664850bf9/internal/librarian/generate.go#L177-L184
This is piece is also used in tidy command

ldetmer: thanks! refactored to use this function

Issue #3258, opened 2025-12-16
surfer: add support for  generic resources


quirogas:
The `ParseResourcePattern` function in `internal/sidekick/parser/httprule.go` currently fails to parse the generic wildcard pattern `*`. This pattern is used in `google.api.resource` annotations, such as in `google/cloud/kms/inventory/v1/key_tracking_service.proto`, to define resources that do not follow a specific hierarchical structure (e.g., `projects/{project}/locations/{location}`). Instead, these "generic" or "proxy" resources can have any valid resource name.

The issue was reported in [PR #3220](https://github.com/googleapis/librarian/pull/3220#discussion_r2620886016). When `ParseResourcePattern` encounters `*`, it fails because the underlying parser (`parseSegments`) expects either a variable (starting with `{`) or a literal. The `*` character is not considered a valid literal character by the parser, leading to the error: `invalid literal, expected at least one character: *`.

According to **AIP-123** (Resource Types), resources are typically hierarchical. However, **AIP-127** (HTTP and gRPC Transcoding) and general API usage (like Cloud Asset Inventory) allow for usage where a resource might be polymorphic or generic, represented by `pattern: "*"`. This acts as a match-all wildcard for the resource name.

Currently, `ParseResourcePattern` delegates directly to `parsePathTemplate` (after prepending a slash), which enforces strict HTTP path template rules that don't accommodate this top-level wildcard usage.

## Proposed Solution

While a temporary fix handles the `*` pattern by checking for it explicitly in `ParseResourcePattern`, a more holistic solution is needed to properly support the full range of resource pattern capabilities and separate concerns.

1.  **Separate Resource Parsing Logic:** Create a new package (e.g., `internal/sidekick/parser/resource`) dedicated to parsing `google.api.resource` patterns. This separation will allow the resource parser to evolve independently of the HTTP rule parser (`internal/sidekick/parser/httprule`), which is bound by different constraints and AIPs (e.g AIP-127).
2.  **Refactor Shared Functionality:** Identify common parsing logic between HTTP path templates and resource patterns (e.g., parsing literals, variables, and identifiers) and move it to a shared utility package or internal helper functions that can be consumed by both parsers.
3.  **Enhanced Model Support:** Update the `api.Resource` model to explicitly support "generic" or "wildcard" resources, potentially adding a field like `IsGeneric` or `Kind` to distinguish them from hierarchical resources.
4.  **Robust Parsing:** Implement a robust parser in the new package that can handle:
[...]

PR #3259, opened 2025-12-16
fix(internal/sidekick): parsing failures when parsing generic resource patterns

quirogas:
This pull request addresses an issue where `ParseResourcePattern` would fail to parse the generic wildcard pattern `*`, which is used in `google.api.resource` annotations to denote generic or "proxy" resources.

## Changes

-   Modified `ParseResourcePattern` in `internal/sidekick/parser/httprule/http_rule_parser.go` to explicitly check for the single segment wildcard `*` before delegating to the standard path template parser.
-   Added a `TODO` comment linking to issue #3258, noting that this is a temporary fix and a more robust solution for generic resources is needed.
-   Added a new test case `TestParseResourcePattern` in `internal/sidekick/parser/httprule/http_rule_parser_test.go` to verify correct parsing of both the generic wildcard `*` and standard hierarchical patterns.

## Related Issues

Updates #3258

### DIFF
--- a/internal/sidekick/parser/httprule/http_rule_parser.go
+++ b/internal/sidekick/parser/httprule/http_rule_parser.go
@@ -94,7 +94,7 @@ func ParseResourcePattern(pathTemplate string) (*api.PathTemplate, error) {
 	// TODO(https://github.com/googleapis/librarian/issues/3258): ParseResourcePattern
 	// should support parsing generic resources more robustly than just checking for a literal `*`.
 	if pathTemplate == api.SingleSegmentWildcard {
-		return api.NewPathTemplate().WithLiteral(api.SingleSegmentWildcard), nil
+		return api.NewPathTemplate(), nil
 	}
 	return parsePathTemplate("/" + pathTemplate)
 }
--- a/internal/sidekick/parser/httprule/http_rule_parser_test.go
+++ b/internal/sidekick/parser/httprule/http_rule_parser_test.go
@@ -157,7 +157,7 @@ func TestParseResourcePattern(t *testing.T) {
 		{
 			"single wildcard",
 			api.SingleSegmentWildcard,
-			api.NewPathTemplate().WithLiteral(api.SingleSegmentWildcard),
+			api.NewPathTemplate(),
 			false,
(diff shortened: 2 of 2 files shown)


internal/sidekick/parser/httprule/http_rule_parser.go
```diff
@@ -91,6 +91,11 @@ func ParseSegments(pathTemplate string) (*api.PathTemplate, error) {
 //
 // This is different from `ParseSegments` because it does not expect a leading `/`.
 func ParseResourcePattern(pathTemplate string) (*api.PathTemplate, error) {
+	// TODO(https://github.com/googleapis/librarian/issues/3258): ParseResourcePattern
+	// should support parsing generic resources more robustly than just checking for a literal `*`.
+	if pathTemplate == api.SingleSegmentWildcard {
+		return api.NewPathTemplate().WithLiteral(api.SingleSegmentWildcard), nil
```
dbolduc: This will only match a string that is exactly `"*"`, right? I don't think a literal makes sense.

I don't think there's a good representation only using path segments (which are all named variables).

You could try something like a:

```
{
  MatchAny: bool,
  Path: []PathSegment
}
```

... but then you have to make sure everything downstream of this knows what `MatchAny` means.

dbolduc (commented):
Q, because I am new here and lazy: did a `*` pattern ever work? (Like not just parsing it, but also the intended behavior of whatever is downstream of the parser)

If not, I am tempted to just approve this PR and unblock things.

If so, we should just revert the other PR. It is like two button clicks.

quirogas:
> Q, because I am new here and lazy: did a `*` pattern ever work? (Like not just parsing it, but also the intended behavior of whatever is downstream of the parser)
> 
> If not, I am tempted to just approve this PR and unblock things.
> 
> If so, we should just revert the other PR. It is like two button clicks.

@dbolduc I'm sorry I didn't address this earlier. AFAIK, the only downstream consumer of resources is Surfer CLI, which I am currently working on. Since Surfer doesn't parse any generics yet, this does not impact its current behavior.

PR #3265, opened 2025-12-16
fix(internal/semver): add semver v1 string support

### LINKED ISSUE #3239: internal/semver: support SemVer 1.0.0 prerelease number bumps
The previous iteration of librarian's `internal/semver` ([link](https://github.com/googleapis/librarian/blob/ce8acebd78009b38047cef57fba48d03efdf726d/internal/semver/semver.go)) did not correctly support SemVer 1.0.0 prerelease number bumps. For example, if the prerelease was `alpha01`, the `01` would be parsed out and converted to an `int`, thus losing the leading `0`, which would not be replaced after bumping or stringifying.

C# is the only language that uses SemVer 1.0.0 prerelease number format, all others are using/will use SemVer 2.0.0.

We should retain recognition that the `Version`'s prerelease number conforms to SemVer 1.0.0, and be sure to print leading zeros when stringifying.

cc @jskeet as fyi

noahdietz:
Adds support for printing SemVer 1.0.0 strings, specifically in how the prerelease numbers are `0` padded. C# is the only language that uses this and should eventually adopt SemVer 2.0.0 versioning for prereleases.

Adds constants for the two support SemVer spec versions, and a corresponding `version` property that is set during `parse` and defaults to SemVer spec 2.0.0.

Also refactors `parse` error mode testing into its own test (reduces logic in tests, https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html, separate singular, disjointed table tests http://shortn/_Poh7fSYhkO), and switches to use of sentinel errors, wrapping, and `errors.Is` checks in tests.

Fixes #3239

cc @jskeet just as fyi

### DIFF
--- a/internal/semver/semver.go
+++ b/internal/semver/semver.go
@@ -62,13 +62,13 @@ var (
 	// prerelease - https://semver.org/spec/v1.0.0.html#spec-item-4.
 	semverV1PrereleaseNumberRegexp = regexp.MustCompile(`^(.*?)(\d+)$`)
 
-	// ErrInvalidVersion is returned when the version string provided is invalid as
+	// errInvalidVersion is returned when the version string provided is invalid as
 	// per the SemVer spec - https://semver.org.
-	ErrInvalidVersion = errors.New("invalid version format")
+	errInvalidVersion = errors.New("invalid version format")
 
-	// ErrInvalidPrereleaseNumber is returned when the prerelease number of a
+	// errInvalidPrereleaseNumber is returned when the prerelease number of a
 	// version string is invalid.
-	ErrInvalidPrereleaseNumber = errors.New("invalid prerelease number")
+	errInvalidPrereleaseNumber = errors.New("invalid prerelease number")
 )
 
 // parse deconstructs the SemVer 1.0.0 or 2.0.0 version string into a [version]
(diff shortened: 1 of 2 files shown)


internal/semver/semver_test.go
```diff
@@ -383,21 +417,21 @@ func TestDeriveNextOptions_DeriveNext_Error(t *testing.T) {
...
+			wantErrPhrase:  "failed to parse version",
 		},
 	} {
 		t.Run(test.name, func(t *testing.T) {
 			_, err := DeriveNextOptions{}.DeriveNext(test.changeLevel, test.currentVersion)
 			if err == nil {
 				t.Errorf("DeriveNextOptions.DeriveNext(%v, %q) did not return an error as expected.", test.changeLevel, test.currentVersion)
-			} else if !strings.Contains(err.Error(), test.wantErr) {
-				t.Errorf("mismatch, got %q, wanted inclusion of %q", err, test.wantErr)
+			} else if !strings.Contains(err.Error(), test.wantErrPhrase) {
```
julieqiu: ditto re version string comment above

internal/semver/semver.go
```diff
@@ -39,28 +50,41 @@ type version struct {
...
+var (
+	// semverV1PrereleaseNumberRegexp extracts the prerelease number, if present, in
+	// the prerelease portion of the SemVer 1.0.0 version string. For example, a
+	// version string like "1.2.3-alpha01" is a SemVer 1.0.0. compliant, numbered
+	// prerelease - https://semver.org/spec/v1.0.0.html#spec-item-4.
+	semverV1PrereleaseNumberRegexp = regexp.MustCompile(`^(.*?)(\d+)$`)
+
+	// ErrInvalidVersion is returned when the version string provided is invalid as
+	// per the SemVer spec - https://semver.org.
+	ErrInvalidVersion = errors.New("invalid version format")
```
julieqiu: I don't think this is used outside of this package, in which case I would unexport it

julieqiu (approved):

julieqiu (approved):

Issue #3250, opened 2025-12-15
librarian: update command to be able to set env vars


ldetmer:
This is a clean up task, currently if we we need to set env vars for the command line we need to run 

```
command.Run(ctx, "env", "ENV_VAR=..")
```

We could simply this to include this as a feature in command class.

PR #3275, opened 2025-12-17
feat(internal/command): introduce RunWithEnv

zhumin8:
Introduce `RunWithEnv` to be able to set env vars for commands without  putting in `env` command directly.
Keep the original `command.Run` as convenience wrapper.

Fixes https://github.com/googleapis/librarian/issues/3250

### DIFF
--- a/internal/command/command.go
+++ b/internal/command/command.go
@@ -31,11 +31,13 @@ func Run(ctx context.Context, command string, arg ...string) error {
 // RunWithEnv executes a program (with arguments) and optional environment
 // variables and captures any error output. If env is nil or empty, the command
 // inherits the environment of the calling process.
-func RunWithEnv(ctx context.Context, env []string, command string, arg ...string) error {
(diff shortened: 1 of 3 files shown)


internal/command/command.go
```diff
@@ -22,9 +22,21 @@ import (
...
+// Run executes a program (with arguments) and captures any error output. It is a
+// convenience wrapper around RunWithEnv.
 func Run(ctx context.Context, command string, arg ...string) error {
+	return RunWithEnv(ctx, nil, command, arg...)
+}
+
+// RunWithEnv executes a program (with arguments) and optional environment
+// variables and captures any error output. If env is nil or empty, the command
+// inherits the environment of the calling process.
+func RunWithEnv(ctx context.Context, env []string, command string, arg ...string) error {
```
ldetmer: how would you feel about making this a map string:string and then creating the assignments?  Seems less error prone, but if you think its less usable, feel free to leave as is.

JoeWang1127 (approved):

PR #3286, opened 2025-12-17
feat(internal/librarian): add branch-based source updates

### LINKED ISSUE #3257: librarian update: support for updating sources based on non-default branch
The API specs for Preview client library generation are found in a `preview` branch of googleapis.

We need to be able to run something like `librarian update` for API sources while targeting a branch e.g. `googleapis/googleapis` @ `preview`.

It would be great if this was simply a flag or some flag input variant that allowed specification of a branch to target generically, and does not pin to `preview` specifically, in the event we want to be able to support generation of a client based on some ad hoc branch.

Note: Of the "known" API sources, only `googelapis` has a `preview` branch. `discovery-artifact-manager` does not have a branch.

noahdietz:
Adds new `Branch string` property  to `config.Source`. Enables library repositories to maintain a  medium-to-long-lived branch of their own based on a similar branch of the API sources via `librarian`. The library repository would create a `librarian.yaml` in their own branch and set the `config.Source.Branch` to the desired source repository branch. When `librarian update` is run from that non-default library branch with this config, the `update` command pulls the latest commit from the `config.Source.Branch` as opposed to the previously hardcoded `master` branch. The default when `config.Source.Branch` is unset is to use `master`.

To facilitate this, a `Branch string` property is added to `fetch.Repo`. Notably, `fetch.Repo` instances derived from the commit-based archive URL **do not** contain a `Branch` value as this is not derivable from the commit-based archive URL.

Finally, I did a little refactoring of `update.go` to pass the desired known `fetch.Repo` instance by value to avoid errantly modifying a global pointer. In general, I think we should be doing more pass by value (http://shortn/_yASqz4zQBZ) to avoid errant modifications of data _unless_ that data should be mutated throughout the process lifecycle. The `update` command errors are converted into internal sentinels for easier testing, and factored error testing into a separate test (http://shortn/_Poh7fSYhkO).

Fixes #3257

### DIFF
--- a/internal/librarian/update_test.go
+++ b/internal/librarian/update_test.go
@@ -30,33 +30,47 @@ import (
 )
 
 type updateTestSetup struct {
-	server           *httptest.Server
-	configPath       string
-	initialConfig    string
-	googleapisCommit string
-	googleapisSHA    string
-	discoveryCommit  string
-	discoverySHA     string
+	server     *httptest.Server
+	configPath string
 }
 
-func setupUpdateTest(t *testing.T) *updateTestSetup {
-	const (
-		googleapisCommit = "123456"
-		discoveryCommit  = "abcdef"
-	)
-	googleapisTarball := "googleapis-tarball-content"
-	discoveryTarball := "discovery-tarball-content"
+const (
+	googleapisTestCommit  = "123456"
+	discoveryTestCommit   = "abcdef"
+	googleapisTestTarball = "googleapis-tarball-content"
+	discoveryTestTarball  = "discovery-tarball-content"
+	testBranch            = "other"
+)
+
+var (
+	googleapisTestSHA = fmt.Sprintf("%x", sha256.Sum256([]byte(googleapisTestTarball)))
+	discoveryTestSHA  = fmt.Sprintf("%x", sha256.Sum256([]byte(discoveryTestTarball)))
+)
+
+func setupUpdateTest(t *testing.T, conf *config.Config) *updateTestSetup {
+	// Source.Branch can be empty in the config file. Update should default to
+	// using the branch configured in [sourceRepos], so we only setup the
+	// test server handlers with Source.Branch when it is explicitly set as it
+	// would be in the file on disk.
+	googleapisBranch := sourceRepos["googleapis"].Branch
+	if conf.Sources.Googleapis.Branch != "" {
+		googleapisBranch = conf.Sources.Googleapis.Branch
+	}
+	discoveryBranch := sourceRepos["discovery"].Branch
+	if conf.Sources.Discovery.Branch != "" {
+		discoveryBranch = conf.Sources.Discovery.Branch
+	}
 
 	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		switch r.URL.Path {
-		case "/repos/googleapis/googleapis/commits/master":
-			w.Write([]byte(googleapisCommit))
-		case "/repos/googleapis/discovery-artifact-manager/commits/master":
-			w.Write([]byte(discoveryCommit))
-		case "/googleapis/googleapis/archive/123456.tar.gz":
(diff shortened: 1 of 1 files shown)


internal/fetch/fetch.go
```diff
@@ -56,11 +56,16 @@ type Repo struct {
 
 	// Repo is the name of the repository, such as `googleapis` or `google-cloud-rust`.
 	Repo string
+
+	// Branch is the name of the repository branch, such as `master` or `preview`.
+	Branch string
```
julieqiu: nit: order alphabetically

internal/fetch/fetch.go
```diff
@@ -121,7 +126,12 @@ func latestSha(query string) (string, error) {
 
 // LatestCommitAndChecksum fetches the latest commit SHA and the SHA256 of the tarball for that
 // commit from the GitHub API for the given repository.
+// Defaults to [defaultBranchMaster] if the provided [Repo.Branch] is unset.
 func LatestCommitAndChecksum(endpoints *Endpoints, repo *Repo) (commit, sha256 string, err error) {
+	branch := defaultBranchMaster
```
julieqiu: can we just pass in branch as an argument? it seems strange to have a default branch

internal/fetch/fetch_test.go
```diff
@@ -540,27 +540,43 @@ func TestDownloadTarballEmptySha(t *testing.T) {
...
-	hasher.Write([]byte(expectedTarballContents))
+	hasher.Write([]byte(expectedMasterTarballContents))
 	expectedTarballSHA256 := fmt.Sprintf("%x", hasher.Sum(nil))
 
+	hasher.Reset()
+	hasher.Write([]byte(expectedBranchTarballContents))
+	expectedBranchTarballSHA256 := fmt.Sprintf("%x", hasher.Sum(nil))
+
 	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		switch r.URL.Path {
```
julieqiu: I would factor out `w.WriteHeader(http.StatusOK)` because it is common to every switch statement

internal/fetch/fetch.go
noahdietz: For sure. Done where I've made changes. Filed a bug to fix other places.

internal/librarian/update.go
```diff
@@ -28,9 +30,9 @@ import (
 var (
 	githubAPI      = "https://api.github.com"
 	githubDownload = "https://github.com"
-	sourceRepos    = map[string]*fetch.Repo{
-		"googleapis": {Org: "googleapis", Repo: "googleapis"},
-		"discovery":  {Org: "googleapis", Repo: "discovery-artifact-manager"},
+	sourceRepos    = map[string]fetch.Repo{
+		"googleapis": {Org: "googleapis", Repo: "googleapis", Branch: "master"},
```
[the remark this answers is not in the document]
noahdietz: Done.

internal/fetch/fetch.go
noahdietz: Ok, I think I've got all the necessary spots to "default" `repo.Branch` to `DefaultBranchMaster`.

internal/librarian/update.go
```diff
@@ -28,10 +31,14 @@ import (
 var (
 	githubAPI      = "https://api.github.com"
 	githubDownload = "https://github.com"
-	sourceRepos    = map[string]*fetch.Repo{
-		"googleapis": {Org: "googleapis", Repo: "googleapis"},
-		"discovery":  {Org: "googleapis", Repo: "discovery-artifact-manager"},
+	sourceRepos    = map[string]fetch.Repo{
+		"googleapis": {Org: "googleapis", Repo: "googleapis", Branch: fetch.DefaultBranchMaster},
+		"discovery":  {Org: "googleapis", Repo: "discovery-artifact-manager", Branch: fetch.DefaultBranchMaster},
```
coryan: Should we have a test with a non-default branch?

Issue #2817, opened 2025-11-07
surfer: generate gcloud command YAML files for parallelstore


julieqiu:
Use the existing parallelstore gcloud.yaml config file and other relevant inputs from sidekick to produce the gcloud command YAML.

There is testdata for currently inside these folders:

Input:

- https://github.com/googleapis/googleapis/tree/master/google/cloud/parallelstore
- https://github.com/googleapis/librarian/blob/main/internal/sidekick/internal/config/gcloudyaml/testdata/gcloud.yaml

Output:

- https://github.com/googleapis/librarian/tree/main/internal/sidekick/internal/gcloud/testdata/parallelstore/surface

Issue #3291, opened 2025-12-18
surfer: support Multiple Services in gcloud Generator


quirogas:
The current gcloud generator assumes a single service in the API model, relying on `model.Services[0]` for service-specific information in functions like `newPrimaryResourceParam` and `newResourceReferenceSpec`. This prevents the generator from correctly handling API models that define multiple services.

**Goal:** Refactor the generator to properly support API models containing multiple services, ensuring that commands are generated accurately for each service without conflicts.

**Acceptance Criteria:**
-   The generator successfully processes API models with multiple services.
-   Commands are generated correctly for each service, with accurate service-specific details (e.g., collection paths derived from the correct service's default host).
-   No regressions are introduced in existing command generation for single-service APIs.
-   Relevant unit and integration tests are updated or added to cover multi-service scenarios.

## Relevant files
- `internal/surfer/gcloud/builder.go`
- `internal/surfer/gcloud/generate.go`

Issue #3033, opened 2025-11-24
surfer: Improve default help text inference


quirogas:
The gcloud command generator in `internal/surfer/gcloud/generate.go` currently generates a very basic default help text for command arguments when no explicit help text is provided in the `gcloud.yaml` configuration.

## Implementation Considerations:
1.  **Using Proto Comments:** Leverage the comments associated with the field in the protobuf source file as the help text. This would provide the most accurate and descriptive information.
2.  **More Contextual Text:** If proto comments are not available, the generated text could be made more contextual. For example, it could include the type of the field (e.g., "The string value for the `request_id` field.").

This will improve the usability of the generated gcloud commands by providing more informative help text to the user.

3. Refactor `findHelpTextRule` and `findFieldHelpTextRule` to support this tiered lookup.
4. Ensure Proto comments (often Markdown) are sanitized or formatted correctly for gcloud help text.

## Relevant Files

- `internal/surfer/gcloud/builder.go`

PR #3294, opened 2025-12-18
feat(internal/surfer): surfer generator poc

quirogas:
This pull request restructures the generator implementation by separating concerns into `Loader`, `Builder`, and `Generator` components. Additionally, it updates the `gcloud` golden test files by inlining `resource_spec` to reflect our desired YAML output.

 stack: #3293 <- #3296  <- #3295 <- #3294 (this pr)
 
## Changes

- Extracts model loading to `loader.go`.
- Extracts command construction to `builder.go`.
- Refactors `generate.go` to serve as a clean orchestrator.
- Implements robust error handling and validation.
- Adds unit tests for generator logic in `generate_test.go` and `builder_test.go`.
- Addresses TODOs for release track inference and resource resolution.

## Related Issue

Updates: #2817

### DIFF
--- a/internal/surfer/gcloud/builder.go
+++ b/internal/surfer/gcloud/builder.go
@@ -81,6 +81,7 @@ func newArguments(method *api.Method, overrides *Config, model *api.API, service
 		return args, nil
 	}
 
+	// TODO(https://github.com/googleapis/librarian/issues/3412): Refactor to use a dispatch pattern (IsIgnored, IsResourceArg, IsArg) to handle field processing.
 	// We iterate over each field in the method's request message (e.g., `CreateInstanceRequest`).
 	for _, field := range method.InputType.Fields {
 		// The "parent" field is a special case. Its information is captured by the
@@ -101,6 +102,7 @@ func newArguments(method *api.Method, overrides *Config, model *api.API, service
 		// For all other fields, we generate a standard flag argument. If the field
 		// is a nested message, its fields will be "flattened" into top-level flags.
 		// For example, a field `instance.description` becomes the `--description` flag.
+		// TODO(https://github.com/googleapis/librarian/issues/3413): Improve error handling strategy (Error vs Skip) and messaging.
 		if err := addFlattenedParams(field, field.JSONName, &args, overrides, model, service); err != nil {
 			return Arguments{}, err
 		}
@@ -150,6 +152,7 @@ func addFlattenedParams(field *api.Field, prefix string, args *Arguments, overri
 // newParam creates a single command-line argument (a `Param` struct) from a proto field.
 func newParam(field *api.Field, apiField string, overrides *Config, model *api.API, service *api.Service) (Param, error) {
 	// We initialize the Param with the basic information derived from the field.
+	// TODO(https://github.com/googleapis/librarian/issues/3414): Abstract away casing logic in the model.
 	param := Param{
 		// The command-line flag name is the kebab-case version of the field name.
 		// Example: `requestId` -> `request-id`
@@ -217,6 +220,7 @@ func newPrimaryResourceParam(field *api.Field, method *api.Method, model *api.AP
 	// We first need to get the full resource definition for the method.
 	resource := utils.GetResourceForMethod(method, model)
 	var segments []api.PathSegment
+	// TODO(https://github.com/googleapis/librarian/issues/3415): Support multiple resource patterns and multitype resources.
 	if resource != nil && len(resource.Patterns) > 0 {
 		segments = resource.Patterns[0]
 	}
@@ -267,6 +271,7 @@ func newResourceReferenceSpec(field *api.Field, model *api.API, _ *Config, servi
 			if len(def.Patterns) == 0 {
 				return nil, fmt.Errorf("resource definition for %q has no patterns", def.Type)
 			}
+			// TODO(https://github.com/googleapis/librarian/issues/3415): Support multiple resource patterns and multitype resources.
 			segments := def.Patterns[0]
 
 			// We determine the plural name, using the explicit `plural` field if available,
@@ -288,9 +293,10 @@ func newResourceReferenceSpec(field *api.Field, model *api.API, _ *Config, servi
 
 			// We assemble and return the `ResourceSpec`.
 			return &ResourceSpec{
-				Name:                  name,
-				PluralName:            pluralName,
-				Collection:            fullCollectionPath,
+				Name:       name,
+				PluralName: pluralName,
+				Collection: fullCollectionPath,
+				// TODO(https://github.com/googleapis/librarian/issues/3416): Investigate and enable auto-completers for referenced resources.
 				DisableAutoCompleters: true,
 				Attributes:            newAttributesFromSegments(segments),
 			}, nil
@@ -307,36 +313,38 @@ func newAttributesFromSegments(segments []api.PathSegment) []Attribute {
 	// We iterate over the segments of the pattern.
 	for i, part := range segments {
 		// A variable segment is enclosed in curly braces.
-		if part.Variable != nil {
-			// The `attribute_name` is the name of the variable (e.g., "project").
-			if len(part.Variable.FieldPath) == 0 {
-				continue
-			}
-			name := part.Variable.FieldPath[len(part.Variable.FieldPath)-1]
-			var parameterName string
-
-			// The `parameter_name` is derived from the preceding literal segment
-			// (e.g., "projects" -> "projectsId"). This is a gcloud convention.
-			if i > 0 && segments[i-1].Literal != nil {
-				parameterName = *segments[i-1].Literal + "Id"
-			} else {
-				// This is a fallback for the unlikely case that a pattern starts with a variable.
-				parameterName = name + "sId"
-			}
-
-			attr := Attribute{
-				AttributeName: name,
-				ParameterName: parameterName,
-				Help:          fmt.Sprintf("The %s id of the {resource} resource.", name),
-			}
-
-			// If the attribute is a project, we add the standard gcloud property fallback,
-			// so users don't have to specify `--project` if it's already configured.
-			if name == "project" {
-				attr.Property = "core/project"
-			}
-			attributes = append(attributes, attr)
+		if part.Variable == nil {
+			continue
 		}
+
+		// The `attribute_name` is the name of the variable (e.g., "project").
+		if len(part.Variable.FieldPath) == 0 {
+			continue
+		}
+		name := part.Variable.FieldPath[len(part.Variable.FieldPath)-1]
+		var parameterName string
+
+		// The `parameter_name` is derived from the preceding literal segment
+		// (e.g., "projects" -> "projectsId"). This is a gcloud convention.
+		if i > 0 && segments[i-1].Literal != nil {
+			parameterName = *segments[i-1].Literal + "Id"
+		} else {
+			// This is a fallback for the unlikely case that a pattern starts with a variable.
+			parameterName = name + "sId"
+		}
+
+		attr := Attribute{
+			AttributeName: name,
+			ParameterName: parameterName,
+			Help:          fmt.Sprintf("The %s id of the {resource} resource.", name),
+		}
+
+		// If the attribute is a project, we add the standard gcloud property fallback,
+		// so users don't have to specify `--project` if it's already configured.
+		if name == "project" {
+			attr.Property = "core/project"
+		}
+		attributes = append(attributes, attr)
 	}
 	return attributes
 }
--- a/internal/surfer/gcloud/generate.go
+++ b/internal/surfer/gcloud/generate.go
@@ -26,6 +26,10 @@ import (
 	"github.com/googleapis/librarian/internal/yaml"
 )
 
+// partialsHeader is the directive that tells gcloud to look in the `_partials` directory
+// for command definitions. This allows for sharing definitions across release tracks.
+const partialsHeader = "_PARTIALS_: true\n"
+
 // Generate generates gcloud commands for a service.
 func Generate(_ context.Context, googleapis, gcloudconfig, output, includeList string) error {
 	overrides, err := readGcloudConfig(gcloudconfig)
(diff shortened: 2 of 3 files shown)


internal/surfer/gcloud/builder.go
```diff
@@ -0,0 +1,403 @@
...
+	} else {
+		// TODO(https://github.com/googleapis/librarian/issues/3033): improve default help text inference
+		param.HelpText = fmt.Sprintf("Value for the `%s` field.", strcase.ToKebab(field.Name))
+	}
+	return param, nil
+}
+
+// newPrimaryResourceParam creates the main positional resource argument for a command.
+// This is the argument that represents the resource being acted upon (e.g., the instance name).
+func newPrimaryResourceParam(field *api.Field, method *api.Method, model *api.API, _ *Config, service *api.Service) Param {
```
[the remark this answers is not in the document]
quirogas: I think it's important to keep the `overrides` parameter, as it is part of the logic that allows `gcloud` to perform overrides.

internal/surfer/gcloud/builder.go
```diff
@@ -0,0 +1,403 @@
...
+	rule := findHelpTextRule(method, overrides)
+
+	// We initialize the command with some default values.
+	cmd := &Command{
+		AutoGenerated: true,
+	}
+
+	if len(overrides.APIs) > 0 {
+		cmd.Hidden = overrides.APIs[0].RootIsHidden
+	} else {
```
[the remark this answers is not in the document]
quirogas: @sarahheacock I would like to hear your input on this. For the POC, I'm currently assuming there is only one API; however, since we'll need to support more in the future, I'd like your opinion on how we should handle multiple APIs. One question I have is:

Does every API in the config file's `apis` node reference the same service, with the only differences being the release track and the package channel? For example: 

``` yaml

config_version: v2.0
service_name: parallelstore.googleapis.com
# ...
apis:
- name: Parallelstore
  api_version: v1
  release_tracks:
  - GA
 # ...
- name: Parallelstore
  api_version: v1beta
  release_tracks:
  - BETA
# ...
...
```
Or will there ever be more than one API  in the `apis` node that is not part of the same service? I believe this relates to https://github.com/googleapis/librarian/issues/3291; otherwise I can open an issue to track this.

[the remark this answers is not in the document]
sarahheacock: One API is definitely fine for a POC. The api refers to the `service` annotation in the protos. Apis like IAM have more than one service annotation https://github.com/googleapis/googleapis/blob/master/google/iam/v1/iam_policy.proto#L59.

internal/surfer/gcloud/builder.go
```diff
@@ -0,0 +1,403 @@
...
+func newArguments(method *api.Method, overrides *Config, model *api.API, service *api.Service) (Arguments, error) {
+	args := Arguments{}
+	if method.InputType == nil {
+		return args, nil
+	}
+
+	// We iterate over each field in the method's request message (e.g., `CreateInstanceRequest`).
+	for _, field := range method.InputType.Fields {
+		// The "parent" field is a special case. Its information is captured by the
+		// primary resource argument, so we skip it here to avoid creating a redundant flag.
```
sarahheacock: Not totally sure I follow this logic. Why is `name` not also redundant?

internal/surfer/gcloud/builder.go
```diff
@@ -0,0 +1,403 @@
...
+		// For example, in a `CreateInstance` method, this would be the `instance_id` field.
+		if utils.IsPrimaryResource(field, method) {
+			// If it is the primary resource, we generate a special positional argument for it.
+			param := newPrimaryResourceParam(field, method, model, overrides, service)
+			args.Params = append(args.Params, param)
+			continue
+		}
+
+		// For all other fields, we generate a standard flag argument. If the field
+		// is a nested message, its fields will be "flattened" into top-level flags.
```
sarahheacock: Feature: we don't want to flatten to root level. We want them to be in an argument group. I don't think parallelstore has field messages so I think this is fine for now

internal/surfer/gcloud/builder.go
```diff
@@ -0,0 +1,403 @@
...
+	param, err := newParam(field, prefix, overrides, model, service)
+	if err != nil {
+		return err
+	}
+	args.Params = append(args.Params, param)
+	return nil
+}
+
+// newParam creates a single command-line argument (a `Param` struct) from a proto field.
+func newParam(field *api.Field, apiField string, overrides *Config, model *api.API, service *api.Service) (Param, error) {
```
sarahheacock: Nit: To avoid this function becoming a "mutation soup," we should eventually move toward a registry-dispatch pattern (or some logic that makes input, analysis, and output easy to debug and add to). It will become hard to track how the Param state changes across nested if-else blocks, which makes it fragile (e.g., changing a field for update commands might accidentally break create).

Checking with rust and golang team for advice would be good. For now, this is fine for a POC and is more important for phase 3

internal/surfer/gcloud/builder.go
```diff
@@ -0,0 +1,403 @@
...
+}
+
+// newResourceReferenceSpec creates a ResourceSpec for a field that references
+// another resource type (e.g., a `--network` flag).
+func newResourceReferenceSpec(field *api.Field, model *api.API, _ *Config, service *api.Service) (*ResourceSpec, error) {
+	// We iterate through all the resource definitions in the API model to find the
+	// one that matches the type of our resource reference.
+	for _, def := range model.ResourceDefinitions {
+		if def.Type == field.ResourceReference.Type {
+			if len(def.Patterns) == 0 {
```
sarahheacock: Nit: I hate pedantic coding rules generally but this rule of thumb I think is helpful. If you have more than two layers of nesting, it's usually best to use guard statement or helper functions. TOTT has a rule for it

internal/surfer/gcloud/builder.go
```diff
@@ -0,0 +1,403 @@
...
+				return nil, fmt.Errorf("resource definition for %q has no patterns", def.Type)
+			}
+			segments := def.Patterns[0]
```
sarahheacock: You will have more than one pattern for multitype

internal/surfer/gcloud/builder.go
```diff
@@ -0,0 +1,403 @@
...
+			shortServiceName := hostParts[0]
+			baseCollectionPath := utils.GetCollectionPathFromSegments(segments)
+			fullCollectionPath := fmt.Sprintf("%s.%s", shortServiceName, baseCollectionPath)
+
+			// We assemble and return the `ResourceSpec`.
+			return &ResourceSpec{
+				Name:                  name,
+				PluralName:            pluralName,
+				Collection:            fullCollectionPath,
+				DisableAutoCompleters: true,
```
sarahheacock: The hard coded `true` I don't think is quite right

internal/surfer/gcloud/builder.go
```diff
@@ -0,0 +1,403 @@
...
+// newAsync creates the `Async` part of the command definition for long-running operations.
+func newAsync(_ *api.Method, _ *Config) *Async {
+	return &Async{
+		// TODO(https://github.com/googleapis/librarian/issues/3290): The collection path is partially hardcoded.
+		Collection: []string{"parallelstore.projects.locations.operations"},
+	}
+}
+
+// findHelpTextRule finds the help text rule from the config that applies to the current method.
+func findHelpTextRule(method *api.Method, overrides *Config) *HelpTextRule {
```
sarahheacock: Nit: I feel like this does not belong in this file. There should be an abstraction of the config that has these helpers

internal/surfer/gcloud/utils/resource.go
```diff
@@ -0,0 +1,177 @@
...
+	"strings"
+
+	"github.com/googleapis/librarian/internal/sidekick/api"
+	"github.com/iancoleman/strcase"
+)
+
+// GetPluralFromSegments infers the plural name of a resource from its structured path segments.
+// Per AIP-122, the plural is the literal segment before the final variable segment.
+// Example: `.../instances/{instance}` -> "instances".
+func GetPluralFromSegments(segments []api.PathSegment) string {
```
sarahheacock: I like how these helpers are abstracted by type! We want to make they don't become redundant or unwieldy

internal/surfer/gcloud/utils/resource.go
```diff
@@ -0,0 +1,177 @@
...
+		if segments[i].Literal != nil && segments[i+1].Variable != nil {
+			collectionParts = append(collectionParts, *segments[i].Literal)
+		}
+	}
+	return strings.Join(collectionParts, ".")
+}
+
+// IsPrimaryResource determines if a field represents the primary resource of a method.
+func IsPrimaryResource(field *api.Field, method *api.Method) bool {
+	if method.InputType == nil {
```
sarahheacock: Looking at the http pattern might be easier here

internal/surfer/gcloud/generate.go
```diff
@@ -17,35 +17,154 @@ package gcloud
...
+}
+
+func generateService(service *api.Service, overrides *Config, model *api.API, output string) error {
+	// Determine short service name for directory structure.
+	// The `shortServiceName` is derived from `service.DefaultHost` (e.g., "parallelstore.googleapis.com" -> "parallelstore").
+	// `service.DefaultHost`  matches the name field in the service config file
+	// (e.g., `default_host` for parallelstore is derived from `parallelstore_v1.yaml` name field).
+	shortServiceName := ""
+	hostParts := strings.Split(service.DefaultHost, ".")
+	if len(hostParts) > 0 {
```
ianthehat: shortServiceName, found, _ := strings.Cut(service.DefaultHost, ".")
	if !found {
		return fmt.Errorf("failed to determine short service name for service %q: default_host is empty", service.Name)
	}

internal/surfer/gcloud/testdata/parallelstore/surface/instances/_partials/_cancel_ga.yaml
```diff
@@ -32,7 +32,21 @@
     params:
       - help_text: The name of the operation resource to be cancelled.
         is_positional: true
-        resource_spec: !REF googlecloudsdk.command_lib.parallelstore.default_resources:project_location_operation
+        resource_spec:
+          name: operation
```
sarahheacock: I don't think this command belongs here?

internal/surfer/gcloud/testdata/parallelstore/surface/instances/_partials/_create_ga.yaml
```diff
@@ -37,36 +37,62 @@
         is_positional: true
         is_primary_resource: true
         request_id_field: instanceId
-        resource_spec: !REF googlecloudsdk.command_lib.parallelstore.v1_resources:project_location_instance
+        resource_spec:
```
sarahheacock: Are there __init__.py command group files?

internal/surfer/gcloud/testdata/parallelstore/surface/locations/describe.yaml
```diff
@@ -35,14 +35,20 @@
       - help_text: |
           The name of the location resource.
         is_positional: true
-        resource_spec: !REF googlecloudsdk.command_lib.parallelstore.custom_resources:location
+        resource_spec:
```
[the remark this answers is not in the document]
quirogas: I wrote these by hand. Since we want to avoid using `!REF` and the previous YAML parser no longer supported `!REF` nodes, I expanded the `!REF` by checking the files they pointed to. This will allow us to run tests on the expected output in the future.

internal/surfer/gcloud/builder.go
quirogas: Since name is the primary resource and is handled differently for Get and Delete methods, we don't skip it initially, though we do skip it later. However, you are right; this logic is confusing. I've created an issue to track the refactoring: https://github.com/googleapis/librarian/issues/3412. We might need to split this into different segments, but we can continue that discussion within the issue.

internal/surfer/gcloud/builder.go
quirogas: https://github.com/googleapis/librarian/issues/3412

internal/surfer/gcloud/builder.go
quirogas: https://github.com/googleapis/librarian/issues/3416

internal/surfer/gcloud/builder.go
quirogas: Yes, I agree that this clutters the builder. I do intend to enhance this function eventually, but I haven't been able to dedicate significant time to it yet. I've added a todo with a related issue I had created a while ago. 

https://github.com/googleapis/librarian/issues/3033

internal/surfer/gcloud/generate.go
```diff
@@ -17,35 +17,154 @@ package gcloud
...
+			methodsByResource[collectionID] = append(methodsByResource[collectionID], method)
+		}
 	}
 
-	// TODO(https://github.com/googleapis/librarian/issues/2817): implement
-	// gcloud command generation logic
-	_, _ = model, cfg
+	// Now that we have grouped the methods by resource, we can generate the
+	// command files for each resource.
+	for collectionID, methods := range methodsByResource {
```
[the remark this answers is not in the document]
quirogas: https://github.com/googleapis/librarian/issues/3417

internal/surfer/gcloud/builder.go
```diff
@@ -0,0 +1,403 @@
...
+
+	"github.com/googleapis/librarian/internal/sidekick/api"
+	"github.com/googleapis/librarian/internal/surfer/gcloud/utils"
+	"github.com/iancoleman/strcase"
+)
+
+// NewCommand constructs a single gcloud command definition from an API method.
+// This function assembles all the necessary pieces: help text, arguments,
+// request details, and async configuration.
+func NewCommand(method *api.Method, overrides *Config, model *api.API, service *api.Service) (*Command, error) {
```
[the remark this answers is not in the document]
quirogas: I've added this to our meeting topics.

PR #3295, opened 2025-12-18
feat(internal/surfer): surfer schema

quirogas:
Updates the `Command` struct schema to use strongly-typed fields and updates test data accordingly.

 stack: #3293 <- #3296 <- #3295 (this pr) <- #3294
 
## Changes

- Refactors `command.go` to use specific `*ResourceSpec` types instead of generic types.
- Updates golden files in `testdata/parallelstore` to match the new schema, removing legacy `!REF` tags.
- Ensures `TestCommandYAML` passes with the new schema.

## Related Issues 
Updates: #2817


internal/surfer/gcloud/utils/field.go
```diff
@@ -0,0 +1,36 @@
...
+
+// Package utils provides utility functions for the gcloud generator.
+package utils
+
+import (
+	"github.com/googleapis/librarian/internal/sidekick/api"
+)
+
+// GetGcloudType maps a proto data type to its corresponding gcloud type.
+func GetGcloudType(t api.Typez) string {
```
ianthehat: I don't think this should be a string, but hard to tell without seeing how it is used. It probably ought to be an enum that knows how to stringify.
We can do this for now and refactor later though.

PR #3296, opened 2025-12-18
feat(internal/surfer): add utils package

quirogas:
This pull request extracts reusable logic into the `internal/surfer/gcloud/utils` package to decouple low-level parsing from generation logic. 

 stack: #3293 <- #3296 (this pr) <- #3295 <- #3294

## Changes

- Creates `utils` package with `field.go`, `method.go`, `service.go`, `resource.go`.
- Adds comprehensive unit tests for all utilities.
- Replaces custom string manipulation with `strcase`.
- Implements segment-based pattern parsing.

## Related Issues

Updates: #2817

### DIFF
--- a/internal/surfer/gcloud/utils/field.go
+++ b/internal/surfer/gcloud/utils/field.go
@@ -17,6 +17,7 @@ package utils
 
 import (
 	"fmt"
+
 	"github.com/googleapis/librarian/internal/sidekick/api"
 )
 
--- a/internal/surfer/gcloud/utils/method.go
+++ b/internal/surfer/gcloud/utils/method.go
@@ -43,8 +43,8 @@ func GetVerb(methodName string) (string, error) {
 	}
 }
 
-// TODO(https://github.com/googleapis/librarian/issues/3362): implement a robust AIP-compliant method idetification.
 // IsCreate determines if the method is a Create method.
+// TODO(https://github.com/googleapis/librarian/issues/3362): implement a robust AIP-compliant method idetification.
 func IsCreate(methodName string) bool {
 	return strings.HasPrefix(methodName, "Create")
 }


internal/surfer/gcloud/utils/field.go
```diff
@@ -0,0 +1,36 @@
...
+	case api.STRING_TYPE:
+		return "" // Default is string
+	case api.INT32_TYPE, api.INT64_TYPE, api.UINT32_TYPE, api.UINT64_TYPE:
+		return "long"
+	case api.BOOL_TYPE:
+		return "boolean"
+	case api.FLOAT_TYPE, api.DOUBLE_TYPE:
+		return "float"
+	default:
+		return ""
```
ianthehat: This feels like a problem waiting to happen, as it returns the same value as string?
If the switch is meant to be exhastive (it probably should be for an internal enum) then the default case should panic.

internal/surfer/gcloud/utils/resource.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// IsPrimaryResource determines if a field represents the primary resource of a method.
+func IsPrimaryResource(field *api.Field, method *api.Method) bool {
+	if method.InputType == nil {
+		return false
+	}
+	// For `Create` methods, the primary resource is identified by a field named
+	// in the format "{resource}_id" (e.g., "instance_id").
+	if strings.HasPrefix(method.Name, "Create") {
```
ianthehat: We should have just one place that owns the logic of determining the type from the method name, if we decide HasPrefix is not enough, then we don't want to have to find them all.

internal/surfer/gcloud/utils/resource.go
```diff
@@ -0,0 +1,177 @@
...
+// This is a crucial function for linking a method to the resource it operates on.
+func GetResourceForMethod(method *api.Method, model *api.API) *api.Resource {
+	if method.InputType == nil {
+		return nil
+	}
+
+	// Strategy 1: For `Create` and `Update`, the request message usually contains
+	// a field that *is* the resource message. This message is annotated with `(google.api.resource)`.
+	for _, f := range method.InputType.Fields {
+		if msg := f.MessageType; msg != nil && msg.Resource != nil {
```
ianthehat: I think the msg local makes this code less clear

internal/surfer/gcloud/utils/resource.go
```diff
@@ -0,0 +1,177 @@
...
+			return msg.Resource
+		}
+	}
+
+	// Strategy 2: For `Get`, `Delete`, and `List`, the request message has a `name`
+	// or `parent` field with a `(google.api.resource_reference)`.
+	var resourceType string
```
ianthehat: This whole block of code is hard to read as is, I would probably extract it to a helper function, then you can just return values which would read much nicer than a local plus breaks.

internal/surfer/gcloud/utils/method.go
```diff
@@ -0,0 +1,44 @@
...
+		return "list", nil
+	case strings.HasPrefix(methodName, "Create"):
+		return "create", nil
+	case strings.HasPrefix(methodName, "Update"):
+		return "update", nil
+	case strings.HasPrefix(methodName, "Delete"):
+		return "delete", nil
+	default:
+		// For non-standard methods, we just use the snake_case version of the method name.
+		return strcase.ToSnake(methodName), nil
```
sarahheacock: I think this might be derived from the http method `:verb`. For a POC this is fine but good to note

internal/surfer/gcloud/utils/field.go
quirogas: Thank you! For now, I've mapped all `api.Typez` to an enum from the command schema. This is definitely something we should examine more closely in the future as we start working with more surfaces, though.

internal/surfer/gcloud/utils/resource.go
quirogas: I created #3363 to simplify the logic in this block, and I've added more comments to the code to help clarify how it works.

internal/surfer/gcloud/utils/method.go
```diff
@@ -0,0 +1,44 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/iancoleman/strcase"
+)
+
+// GetVerb maps an API method name to a standard gcloud command verb.
+func GetVerb(methodName string) (string, error) {
+	if methodName == "" {
+		return "", fmt.Errorf("method name cannot be empty")
```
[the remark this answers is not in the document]
ianthehat: My feeling is that the error itself adds no information, as it is always the same, and it is not reasonable to have an empty string return with a non nil error, so splitting it adds no extra information to the caller than testing an empty string return would give them. Thus my assertion that it is simpler to return just a string, and no harder outside to test for the empty string than the non nil error. You could argue that returning the error makes it more obvious if the calling code is forgetting to check it, but the calling code is ours as well, we are not going to forget, and it will (or should be) covered by tests anyway.

PR #3312, opened 2025-12-19
feat(sidekick/rust): Generate samples for get RPCs

amanda-tarafa:


internal/sidekick/api/model.go
```diff
@@ -369,6 +369,48 @@ func (m *Method) IsSimple() bool {
...
 
+// IsStandardMethod returns true if the method is a standard method.
+func (m *Method) IsStandardMethod() bool {
+	return m.StandardGetInfo() != nil
+}
+
+// StandardGetInfo contains information relevant to get operations as defined by AIP-131.
+type StandardGetInfo struct {
+	// ResourceNameRequestField is the field in the method input that contains the resource name
+	// of the resource that the get operation should fetc.
```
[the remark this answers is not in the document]
amanda-tarafa: Done

internal/sidekick/rust/templates/common/client_method_samples/parameters.mustache
```diff
@@ -13,4 +13,11 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
 }}
-///    client: &{{Service.Codec.Name}} 
+{{#StandardGetInfo}}
+///    client: &{{Service.Codec.Name}},
+///    resource_name: &str,
+{{/StandardGetInfo}}
+{{^IsStandardMethod}}
+///    client: &{{Service.Codec.Name}},
+///    request: {{InputType.Codec.Name}},
+{{/IsStandardMethod}}
```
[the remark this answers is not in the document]
amanda-tarafa: More blocks are coming, and the only common parameter is the service. But I think it's clearer to have all the parameters for each case on a specific block.

amanda-tarafa:
This is ready for review. Generation results can be seen in https://github.com/googleapis/google-cloud-rust/pull/4095

internal/sidekick/rust/templates/common/client_method_samples/builder_fields.mustache
```diff
@@ -13,4 +13,9 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 }}
-///         /* set fields */
+{{#StandardGetInfo}}
+///         .set_{{ResourceNameRequestField.Codec.SetterName}}(resource_name)
+{{/StandardGetInfo}}
+{{^IsStandardMethod}}
+///         .with_request(request)
```
coryan: Hmmm.. probably not the direction we want to steer folks into. Yes, it works, but it is the least desirable way to initialize a request.  Can we leave these complex requests as `/* set fields */`  for now?

internal/sidekick/api/model.go
```diff
@@ -369,6 +369,48 @@ func (m *Method) IsSimple() bool {
 		m.OperationInfo == nil && m.DiscoveryLro == nil
 }
 
+// IsStandardMethod returns true if the method is a standard method.
+func (m *Method) IsStandardMethod() bool {
```
coryan: Right now, it would be more appropriate to call this `IsGetResourceMethod()`.  I assume we will have more methods that are "standard"?

Also: the comment is not super useful.  Maybe say something about why the method exists:

```
// IsStandardMethod simplifies the writing of mustache templates
```

Though I would think `HasStandardGetInfo` is closer to what happens here.

coryan (changes requested):

internal/sidekick/api/model.go
amanda-tarafa: I'll fix the comment.

> I assume we will have more methods that are "standard" 

Yes, I'm seding PRs one by one (I have them lined up), but we have one similar for delete, create, etc. So, this `IsStandardMethod`  will cover more than just get.

If I change the Standard bit to AIPStandard, we will have: `AIPStandardGetInfo()`, `AIPStandardDeleteInfo()`, `AIPStandardCreateInfo()` ... and `IsAIPStandard()` that returns true if the method "has any of the AIP standard info". Does that sound better?

coryan: > I'll fix the comment.
> 
> > I assume we will have more methods that are "standard"
> 
> Yes, I'm seding PRs one by one (I have them lined up), but we have one similar for delete, create, etc. So, this `IsStandardMethod` will cover more than just get.

Ack. Sending one by one SGTM.

> If I change the Standard bit to AIPStandard, we will have: `AIPStandardGetInfo()`, `AIPStandardDeleteInfo()`, `AIPStandardCreateInfo()` ... and `IsAIPStandard()` that returns true if the method "has any of the AIP standard info". Does that sound better?

Mentioning the AIP SGTM.  If it gets too annoying we can remove it.

amanda-tarafa:
I've pushed the changes and updated https://github.com/googleapis/google-cloud-rust/pull/4095. Thanks!

coryan (approved):

internal/sidekick/api/model.go
```diff
@@ -363,12 +363,56 @@ func (m *Method) HasAutoPopulatedFields() bool {
...
+	if strings.ToLower(m.Name) != fmt.Sprintf("get%s", singular) ||
+		strings.ToLower(m.InputType.Name) != fmt.Sprintf("get%srequest", singular) {
+		return nil
+	}
+
+	// The request needs to have a field for the resource name of the resource to obtain.
+	nameFieldIndex := slices.IndexFunc(m.InputType.Fields, func(f *Field) bool {
+		return f.IsResourceReference() &&
+			f.ResourceReference.Type == m.OutputType.Resource.Type
+	})
```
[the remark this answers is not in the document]
amanda-tarafa: Yes, I've added this to a small document I have for tracking improvements.

For this same reason I though about identifying the methods in annotations instead of in the model, but I believe this will be useful for all languages and that seemed more important. Also, we should be searching on request messages, that in general should have very few fields. But yes, definetely we can make it happen in a x-ref phase.

PR #3342, opened 2025-12-23
chore: update generate function used in release

ldetmer:
We should call the generate function that includes formatting when releasing otherwise diffs will appear in PR.

### DIFF
--- a/internal/librarian/release.go
+++ b/internal/librarian/release.go
@@ -20,12 +20,17 @@ import (
 	"fmt"
 
 	"github.com/googleapis/librarian/internal/config"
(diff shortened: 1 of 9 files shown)


zhumin8 (approved):

internal/librarian/release.go
```diff
@@ -33,8 +33,8 @@ var (
 )
 
 var (
-	librarianGenerateLibrary = generateLibrary
 	rustReleaseLibrary       = rust.ReleaseLibrary
+	librarianGenerateLibrary = runGenerate
```
julieqiu: what's the reason for having the alias? can you just call runGenerate directly?

PR #3370, opened 2025-12-29
chore: refactor to remove service config from librarian yaml if it can find it in googleapis

### LINKED ISSUE #3358: librarian: tidy command should use googleapis/googleapis for validation
Currently librarian tidy uses naming convention heuristics to determine if values in librarian.yaml are derivable and can be omitted from the configuration file. The command should instead use the googleapis/googleapis repository to ensure that values are accurate.

ldetmer:
Update logic in RunTidy, to only remove the service config entry from librarian.yaml if it can find it in googleapis repo. (Safer implementation).  

This includes a refactor to make FetchSource available outside of generate command.

[Generated librarian yaml file](https://paste.googleplex.com/5634686921605120)

Fixes #3358

### DIFF
--- a/internal/librarian/tidy_test.go
+++ b/internal/librarian/tidy_test.go
@@ -398,7 +398,9 @@ libraries:
 		t.Fatal(err)
 	}
 	err := Run(t.Context(), "librarian", "tidy")
-
+	if err == nil {
+		t.Fatalf("expected error, got %v", nil)
+	}
 	if !errors.Is(err, errNoGoogleapiSourceInfo) {
 		t.Errorf("mismatch error want %v got %v", errNoGoogleapiSourceInfo, err)
 	}


internal/librarian/tidy_test.go
```diff
@@ -367,3 +382,24 @@ libraries:
...
+libraries:
+  - name: google-cloud-storage-v1
+    version: "1.0.0"
+  - name: google-cloud-bigquery-v1
+    version: "2.0.0"
+`
+	if err := os.WriteFile(configPath, []byte(configContent), 0644); err != nil {
+		t.Fatal(err)
+	}
+	err := Run(t.Context(), "librarian", "tidy")
```
julieqiu: fail if the error is not nil

ldetmer: added check to fail if error IS nil, because we're expecting error for this test case.

julieqiu (approved):

internal/librarian/source.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/googleapis/librarian/internal/config"
+	"github.com/googleapis/librarian/internal/fetch"
+)
+
+// FetchSource fetches a repository source.
+func FetchSource(ctx context.Context, source *config.Source, repo string) (string, error) {
```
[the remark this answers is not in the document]
julieqiu: no preference, a new file seems fine

Issue #3232, opened 2025-12-11
librarian: run release successfully on google-cloud-rust


julieqiu:
Confirm this replaces `sidekick rust-bump-versions`.

ldetmer:
@dbolduc @coryan What is the process for tagging releases in google-cloud-rust git repository? I see a [bump versions PR](https://github.com/googleapis/google-cloud-rust/commit/c6091874ae1a0f9dd52349654af60b0e35cbeba8) was merged today, but it looks like the last [git release] (https://github.com/googleapis/google-cloud-rust/releases/tag/release-20251210) was a month ago.  

I believe sidekick (and librarian) rely on the git release tag to determine what versions should be bumped.

PR #3447, opened 2026-01-05
chore: add missing logic to release command

ldetmer:
Added missing logic in the release command:
1) Take into account skipPublish config  
2) Take into account Release.FilesToIgnore config.  As part of this change reused existing FilesChangedSince helper function to get all files changed since last git release at one time.  
3) Updated migrate command to populate the ignored-files field in librarian.yaml
4) Added logic for rust releases to check if cargo version had already been bumped since last release tag

For #3232

### DIFF
--- a/internal/librarian/release.go
+++ b/internal/librarian/release.go
@@ -123,7 +123,7 @@ func releaseAll(ctx context.Context, cfg *config.Config, lastTag, gitExe string)
 		if err != nil {
 			return err
 		}
-		if shouldRelease(library, cfg, filesChanged, srcPath) {
+		if shouldRelease(library, filesChanged, srcPath) {
 			if err := releaseLibrary(ctx, cfg, library, srcPath, lastTag, gitExe); err != nil {
 				return err
 			}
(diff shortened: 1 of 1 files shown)


internal/librarian/release.go
```diff
@@ -104,18 +113,32 @@ func runRelease(ctx context.Context, cmd *cli.Command) error {
...
+		pathWithTrailingSlash := srcPath
+		if !strings.HasSuffix(pathWithTrailingSlash, "/") {
+			pathWithTrailingSlash = pathWithTrailingSlash + "/"
+		}
+		shouldRelease := false
+		for _, path := range filesChanged {
+			if strings.Contains(path, pathWithTrailingSlash) {
+				shouldRelease = true
+				break
+			}
```
zhumin8: nit: This logic seems a bit verbose and distracting within the loop. Can you re-introduce shouldReleaseLibrary as helper?

ldetmer: thanks good call, pulled out extra logic

PR #3465, opened 2026-01-06
fix(dart): Do generate code for omitted deps

brianquinlan:
The previous implementation would calculate the fields for messages, even if they did not result in generated code. This meant that they could produce dart `imports`, for fields that were never generated.

The symptom of this problem was caused by this message:

```
message Operation {
  oneof result {
    // The error result of the operation in case of failure or cancellation.
    google.rpc.Status error = 4;
}
```

The message `Operation` is omitted based on an allowlist but it's `error` field still results in an import of the `google.rpc` package, which is not used.

Note that the `map` type is also omitted but you must still process its fields because you may need to do an import tor the `map`'s value type.

### DIFF
--- a/internal/sidekick/dart/annotate_test.go
+++ b/internal/sidekick/dart/annotate_test.go
@@ -519,8 +519,8 @@ func TestAnnotateMessage_ToString(t *testing.T) {
 	}
 }
 
-// Tests that messages that are whitelisted as not being generated are, in fact, not generated.
-func TestAnnotateMessage_OmitGeneration_Whitelisted(t *testing.T) {
+// Tests that messages that are allowlisted as not being generated are, in fact, not generated.
+func TestAnnotateMessage_OmitGeneration_Allowlisted(t *testing.T) {
 	status := &api.Message{
 		Name:    "Status",
 		ID:      ".google.rpc.Status",


internal/sidekick/dart/annotate.go
```diff
@@ -585,6 +585,16 @@ func (annotate *annotateModel) annotateService(s *api.Service) {
 }
 
 func (annotate *annotateModel) annotateMessage(m *api.Message) {
+	if _, omit := omitGeneration[m.ID]; omit && !m.IsMap {
+		// If the message is whitelisted as omitted, and it's not a map,
```
natebosch: [nit] 


```suggestion
		// If the message is allowlisted as omitted, and it's not a map,
```

internal/sidekick/dart/annotate_test.go
```diff
@@ -519,6 +519,86 @@ func TestAnnotateMessageToString(t *testing.T) {
 	}
 }
 
+// Tests that messages that are whitelisted as not being generated are, in fact, not generated.
+func TestAnnotateMessage_OmitGeneration_Whitelisted(t *testing.T) {
```
natebosch: [nit]


```suggestion
// Tests that messages that are allowlisted as not being generated are, in fact, not generated.
func TestAnnotateMessage_OmitGeneration_Allowlisted(t *testing.T) {
```

natebosch (approved):

PR #3471, opened 2026-01-06
fix: veneers skipGenerate should be set to true

### LINKED ISSUE #3469: librarian[rust/generate] fails for veneers
Steps to Reproduce:

Install Librarian from the latest commit:

In the librarian repo
```
git checkout 90e5ac8ddee1a844519916ddda74f0ec4c89a170
go install ./cmd/librarian
go install ./devtools/cmd/librarian
```

Run migration and generation from the google-cloud-rust-repo:

```
migrate .
librarian generate --all
```

Fails with 
```
librarian command failed err="veneer \"google-cloud-auth\" has no modules defined"
```

If I comment this out it fails on another veneer:

```
librarian command failed err="veneer \"google-cloud-base\" has no modules defined"
```

ldetmer:
It looks like the inverse value was being set for skipGenerate for veneers.  If we want to skip generation, this value need to be true.

Fixes #3469

### DIFF
--- a/internal/librarian/tidy.go
+++ b/internal/librarian/tidy.go
@@ -75,8 +75,8 @@ func tidyLibrary(cfg *config.Config, lib *config.Library, googleapisDir string) 
 		lib.Output = ""
 	}
 	if lib.Veneer {
-		// Veneers are never generated, so ensure skip_generate is false.
-		lib.SkipGenerate = false
+		// Veneers are never generated, so ensure skip_generate is true.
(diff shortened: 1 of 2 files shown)


julieqiu:
If `lib.Veneer` is true, we should always skip generate. I think the fix is here: https://github.com/googleapis/librarian/blob/90e5ac8ddee1a844519916ddda74f0ec4c89a170/internal/librarian/generate.go#L136

Issue #3500, opened 2026-01-08
surfer: Incorrect usage of request_id_field in non-Create commands


quirogas:
The generated YAML populates `request_id_field` for all primary resource arguments (Delete, Get, List), whereas the official surface only uses it for `Create` commands.

## Context & Rationale

- **Proto Source:**
    - `CreateInstanceRequest` has `string instance_id`. The user provides an ID (e.g., "my-instance"), and gcloud must know to put this value into the `instance_id` field of the request.
    - `DeleteInstanceRequest` has `string name`. The user provides an ID or Name, and gcloud constructs the full resource name to put into the `name` field.
- **gcloud Schema:** Defines `request_id_field` as "The field in the create request that should hold the name of the resource being created".

## Discrepancies (Parallelstore )

**Generated (`_delete_ga.yaml`):**
```yaml
request_id_field: name
```

**Official (`_delete_ga.yaml`):**
(Field is absent)

## Proposed Solution

Modify `newPrimaryResourceParam` in `internal/surfer/gcloud/builder.go` to only set `RequestIDField` when the field name is NOT `name` (or strictly when it is `{resource}_id`), which aligns with `Create` method patterns where the ID is a separate parameter.

## Relevant Files

- `internal/surfer/gcloud/builder.go`

PR #3501, opened 2026-01-08
fix(internal/surfer): remove request_id_field in non-Create commands

quirogas:
Address discrepancies in the generated gcloud command surface by refining the logic for mapping Protobuf fields to CLI arguments.

 ### Key Changes
   - Fix `request_id_field` Logic: Conditioned the population of request_id_field in the primary resource parameter to only apply for Create methods. This prevents redundant or incorrect field mapping in
     Delete, Get, and List commands.
   - Improved Type Mapping: Updated the generator to correctly map Protobuf int64 and uint64 types to the long argparse type, ensuring compatibility with gcloud framework conventions.

  Fixes: https://github.com/googleapis/librarian/issues/3500

### DIFF
--- a/internal/surfer/gcloud/utils/field.go
+++ b/internal/surfer/gcloud/utils/field.go
@@ -28,7 +28,7 @@ func GetGcloudType(t api.Typez) string {
 		return "float"
 	case api.INT64_TYPE, api.UINT64_TYPE, api.FIXED64_TYPE, api.SFIXED64_TYPE, api.SINT64_TYPE:
 		return "long"
-	case api.INT32_TYPE, api.UINT32_TYPE, api.FIXED32_TYPE, api.SFIXED32_TYPE, api.SINT32_TYPE:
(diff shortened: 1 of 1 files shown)


internal/surfer/gcloud/utils/field.go
```diff
@@ -26,9 +26,9 @@ func GetGcloudType(t api.Typez) string {
 	switch t {
 	case api.DOUBLE_TYPE, api.FLOAT_TYPE:
 		return "float"
-	case api.INT64_TYPE, api.UINT64_TYPE, api.INT32_TYPE, api.UINT32_TYPE,
-		api.FIXED64_TYPE, api.FIXED32_TYPE, api.SFIXED32_TYPE, api.SFIXED64_TYPE,
-		api.SINT32_TYPE, api.SINT64_TYPE:
+	case api.INT64_TYPE, api.UINT64_TYPE, api.FIXED64_TYPE, api.SFIXED64_TYPE, api.SINT64_TYPE:
+		return "long"
```
sarahheacock: Is this intentional? I don't think we have a long type?

sarahheacock: I stand corrected. Good catch!

PR #3509, opened 2026-01-09
chore: fix source of truth for generate when doing releases

ldetmer:
Currently when running generate after bumping versions, it was reading directly from librarian.yaml (which wasn't updated with the latest version) so cargo.toml and readme files were not getting updated.  Instead of writing to disk every time for every library, I updated the logic to use the generate function that takes in the config object.  However generate updates the config object, so we need to pass in a clone so the original object doesn't get modified, as we write it back to librarian.yaml at the end of release with updated versions.

I do think this can be reworked but in order to be ready for tuesday's release I am proposing this fix and will open separate ticket to refactor when time allows.


JoeWang1127 (approved):
We will add unit tests for new functions in a separate issue.

PR #3546, opened 2026-01-13
chore: configify testhelper setup

### LINKED ISSUE #3522: internal/testhelper: Consider making Setup APIs take a param struct for easier test setup
Good call. Ultimately, I think our `testhelper.Setup*` APIs should take some param struct so all of them can be configured to do the same things. I'll file a tracking issue.

_Originally posted by @noahdietz in https://github.com/googleapis/librarian/pull/3515#discussion_r2683420804_

noahdietz:
Adds `Setup` and `SetupOptions` in `testhelper` to config-ify the heavier unit test setup for `librarian` functionality like `release`. This simplifies test logic by moving various common test environment setup actions into the setup flow its, and out from the test logic themselves ([relevant TotT](https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html)).

Adds `internal/sample` package with `Config` generation function that produces a unique instance of `librarian.yaml` `*config.Config` populated with static values also defined in `internal/sample`. This simplifies & unifies test data creation, and allows test cases to more directly modify the properties they are exercising without bloating the test case schema or logic with piecemeal properties. The test data/config mutation is much more closely coupled with the test case itself now, making them more declarative and easier to extend.

Refactors `release_test.go` to use new `testhelper` APIs. Makes it easier to write tests for #2764.

Fixes #3522

### DIFF
--- a/internal/librarian/release_test.go
+++ b/internal/librarian/release_test.go
@@ -264,7 +264,7 @@ func TestReleaseLibrary(t *testing.T) {
 			testhelper.CloneRepository(t, remoteDir)
 
 			targetLibCfg := test.cfg.Libraries[0]
-			// Unused string params: lastTag, googleapisDir.
+			// Unused string param: lastTag.
 			err := releaseLibrary(t.Context(), test.cfg, targetLibCfg, testUnusedStringParam, "git")
 			if err != nil {
 				t.Fatalf("releaseLibrary() error = %v", err)


internal/librarian/release_test.go
```diff
@@ -28,105 +30,94 @@ import (
 	"github.com/googleapis/librarian/internal/yaml"
 )
 
+const (
+	// testUnusedStringParam is used to fill the spot of a string parameter that
+	// won't be provided in the test, because the test does not exercise the
+	// functionality related to said parameter. It is an intentional signal
+	// rather than an ambiguous empty string.
+	testUnusedStringParam = ""
+)
```
julieqiu: nit: put consts on one line when there is only 1

// testUnusedStringParam is used to fill the spot of a string parameter that
// won't be provided in the test, because the test does not exercise the
// functionality related to said parameter. It is an intentional signal
// rather than an ambiguous empty string.
const testUnusedStringParam = ""

internal/testhelper/testhelper.go
```diff
@@ -66,8 +66,59 @@ version = "1.0.0"
 
 	// testRemoteURL is the URL set for the [TestRemote] in the test repository.
 	testRemoteURL = "https://example.com/git.git"
+
+	// TestLib1 is the name of the first library added to the [FakeConfig].
+	TestLib1 = "google-cloud-storage"
```
julieqiu: It seems like these are only ever used in librarian, in which case I would put them in a *_test.go file in that package

internal/testhelper/testhelper.go
```diff
@@ -66,8 +66,59 @@ version = "1.0.0"
...
+	// TestLib1SrcPath is the [config.Library] Output path of [TestLib1] added
+	// to the [FakeConfig].
+	TestLib1SrcPath = "src/storage"
```
julieqiu: Use "Output" instead of "SrcPath" since SrcPath is kind of meaningless in the context of this repo

internal/testhelper/testhelper.go
```diff
@@ -66,8 +66,59 @@ version = "1.0.0"
...
+	TestInitialVersion = "1.0.0"
+	// TestNextVersion is the next version typically assigned to libraries
+	// starting from [TestInitialVersion].
+	TestNextVersion = "1.1.0"
 )
 
+// FakeConfig produces a [config.Config] instance populated with most of the
+// properties necessary for testing. It produces a unique instance each time so
+// that individual test cases may modify their own instance as needed.
+func FakeConfig() *config.Config {
```
julieqiu: We have a /Users/julieqiu/code/googleapis/librarian/internal/sidekick/sample package with some sample data. 

Maybe move all of this to /Users/julieqiu/code/googleapis/librarian/internal/sample?

internal/testhelper/testhelper.go
```diff
@@ -159,17 +210,73 @@ func SetupRepo(t *testing.T) string {
...
+// Setup is a configurable test setup function that starts by creating a
+// fresh test repository via [SetupRepo], to which it then applies the
+// configured [SetupOptions].
+func Setup(t *testing.T, opts SetupOptions) string {
 	t.Helper()
-	remoteDir := SetupRepo(t)
-	AddLibrarianConfig(t, cfg)
-	return remoteDir
+	dir := SetupRepo(t)
+
```
julieqiu: remove unnecessary newline

internal/testhelper/testhelper.go
```diff
@@ -159,17 +210,73 @@ func SetupRepo(t *testing.T) string {
...
+	setup(t, opts)
+
+	return dir
+}
+
+func setup(t *testing.T, opts SetupOptions) {
```
julieqiu: what's the reason for having Setup and setup? Why not just one Setup function?

julieqiu (approved):

internal/testhelper/testhelper.go
noahdietz: It's an artifact of other changes for my preview branch work where I wanted to call `setup` "recursively" for setting up multiple branches at once (e.g. main and preview). I'll consolidate in this PR and propose that change in the corresponding feature PR.

internal/testhelper/testhelper.go
noahdietz: done

internal/testhelper/testhelper.go
noahdietz: Done!

julieqiu (approved):

PR #3557, opened 2026-01-14
fix(internal/librarian): move create logic to generate and fix ordering

### LINKED ISSUE #3545: refactor(librarian): move inside fakeGenerate to avoid multiple switch statements
I would move this inside fakeGenerate and fakeCreate, so that we don't have multiple switch statements

_Originally posted by @julieqiu in https://github.com/googleapis/librarian/pull/3498#discussion_r2687899879_

### LINKED ISSUE #3576: librarian: rust.Create does not add to Cargo.lock properly
for `librarian generate google-cloud-accessapproval-v1`, this is what produced,

```
[[package]]
name = "v1"
version = "0.1.0"
```
instead, expecting original 
https://github.com/googleapis/google-cloud-rust/blob/d51e1ca2011f03618e692d9dbf5cda03048ac17d/Cargo.lock#L892-L908

This is likely missed in the original implementation in https://github.com/googleapis/librarian/pull/3248, as the test run did not delete Cargo.lock entry at start: https://github.com/ldetmer/google-cloud-rust/pull/1/files#diff-13ee4b2252c9e516a0547f2891aa2105c3ca71c6d7a1e682c69be97998dfc87e

zhumin8:
Moves create new library skeleton logic inside `rust.Generate` to avoid an extra language switch and fix sidekickrust.Generate should run as part of Create for new libraries.
A few changes worth noting:
- `cleanOutput` now does not report error if dir does not exist.
- In `rust.Generate`, now sidekickrust.Generate is correctly called as part of `Create` so the post create steps are run after generation. Thus properly populating Cargo.lock. (Fix for #3576)
- removed `git add` commands in `Create` which are not necessary. After generate, it leaves the work dir with unstaged changes anyway. 
- removed sidekick rust-generate. This allows refactoring and separate `Create` into `create` that creates skeleton before generate and `validate` that formats and do other post processing changes to cargo files.

Verified manually in `google-cloud-rust` with ` librarian generate google-cloud-accessapproval-v1` (deleting original content and Cargo.lock), does not produce diff in code.

Fix #3545
Fix #3576

### DIFF
--- a/internal/librarian/rust/create.go
+++ b/internal/librarian/rust/create.go
@@ -22,8 +22,8 @@ import (
 	"github.com/googleapis/librarian/internal/command"
 )
 
-// Create creates a cargo workspace skeleton.
-func Create(ctx context.Context, outputDir string) error {
+// create creates a cargo workspace skeleton.
+func create(ctx context.Context, outputDir string) error {
 	if err := command.Run(ctx, "cargo", "--version"); err != nil {
 		return fmt.Errorf("got an error trying to run `cargo --version`, the instructions on https://www.rust-lang.org/learn/get-started may solve this problem: %w", err)
 	}
--- a/internal/librarian/rust/generate.go
+++ b/internal/librarian/rust/generate.go
@@ -64,7 +64,7 @@ func Generate(ctx context.Context, library *config.Library, sources *Sources) er
 		exists = false
 	}
 	if !exists {
-		if err := Create(ctx, library.Output); err != nil {
+		if err := create(ctx, library.Output); err != nil {
 			return err
 		}
 	}
--- a/internal/librarian/rust/create_test.go
+++ b/internal/librarian/rust/create_test.go
@@ -43,7 +43,7 @@ members = []
 	}
 
 	const libName = "secretmanager"
-	if err := Create(t.Context(), libName); err != nil {
+	if err := create(t.Context(), libName); err != nil {
 		t.Fatal(err)
 	}
 


internal/librarian/rust/generate.go
```diff
@@ -55,8 +56,28 @@ func Generate(ctx context.Context, library *config.Library, sources *Sources) er
...
+// If the directory already exists, it directly executes the generateFn.
+func createAndGenerate(ctx context.Context, outputDir string, generateFn func(context.Context) error) error {
+	if _, err := os.Stat(outputDir); err != nil {
+		// Create if the output path does not already exist
+		if !os.IsNotExist(err) {
+			return fmt.Errorf("failed to stat output directory %q: %w", outputDir, err)
+		}
+		if err := Create(ctx, outputDir,
+			func(ctx context.Context) error {
+				return generateFn(ctx)
```
julieqiu: Now that Rust isn't using sidekick, we don't need generateFn anymore

internal/librarian/rust/create.go
```diff
@@ -22,46 +22,28 @@ import (
...
 	}
 	if err := command.Run(ctx, "cargo", "test", "--manifest-path", manifestPath); err != nil {
 		return err
 	}
-	if err := command.RunWithEnv(ctx, map[string]string{"RUSTDOCFLAGS": "-D warnings"}, "cargo", "doc", "--manifest-path", manifestPath, "--no-deps"); err != nil {
-		return err
-	}
-	if err := command.Run(ctx, "cargo", "clippy", "--manifest-path", manifestPath, "--", "--deny", "warnings"); err != nil {
-		return err
-	}
```
coryan: `clippy` both lints and compiles the code.  If the new library exposes a bug in the generator (e.g. not dealing with reserved words correctly, or emitting bad code) this was the only step that would detect it.  I guess CI would catch it too, if you wanted to wait that long.

coryan (approved):

Issue #3561, opened 2026-01-15
librarian: inconsistent behavior in rust generation


coryan:
This command sometimes works well:

```bash
go run github.com/googleapis/librarian/cmd/librarian@${V} generate google-cloud-wkt
✓ Successfully generated google-cloud-wkt
Running: /Users/coryan/.cargo/bin/taplo fmt src/wkt/Cargo.toml
Running: /Users/coryan/.cargo/bin/cargo fmt -p google-cloud-wkt
```

But sometimes fails:

```bash
$ go run github.com/googleapis/librarian/cmd/librarian@${V} generate google-cloud-wkt

[...]

PR #3565, opened 2026-01-15
fix(internal/librarian/rust): find service config in all sources

JoeWang1127:
Find service config in all sources, not just googleapis dir.

For #3561


coryan:
FWIW, the error I reported in #3561 is for `google-cloud-wkt` which I think does not use a service config file?

PR #3566, opened 2026-01-15
fix(librarian/rust): Find sources in all roots

amanda-tarafa:


amanda-tarafa:
Closing now as this is being adressed by #3565

PR #3615, opened 2026-01-17
fix(internal/librarian): use a derived context in parallel generation

JoeWang1127:
Before the fix, the context is shared between parallel generation and sequence formatting. However, the context is cancelled after the `g.Wait` call. The formatting will fail because it still used the shared context.

Use the derived context for generation can fix this issue.


julieqiu (approved):
nice find!

PR #3653, opened 2026-01-20
feat(legacylibrarian): add -api-source-branch flag

### LINKED ISSUE #3651: legacylibrarian: consider adding -api-source-branch flag
`legacylibrarian` currently accepts a local directory path or clone-able repo URL for the `-api-source` flag, but defaults to `https://github.com/googleapis/googleapis` on `master`. The local `-api-source` variant is good enough for executing `generate` with a local `googleapis` clone, but requires that the local copy / branch be up to date. If we add an `-api-source-branch` flag, we could control the branch checked out post cloning and not need to provide `-api-source`. Such a flag would _not_ work when a local directory path is provided to `-api-source`.

noahdietz:
Adds a dedicated flag `-api-source-branch` for specifying the `-api-source` branch to checkout when cloning a remote API source repo.

Fixes #3651

### DIFF
--- a/internal/legacylibrarian/legacylibrarian/generate_command_test.go
+++ b/internal/legacylibrarian/legacylibrarian/generate_command_test.go
@@ -88,14 +88,15 @@ func TestNewGenerateRunner(t *testing.T) {
 		{
 			name: "empty API source",
 			cfg: &legacyconfig.Config{
-				API:            "some/api",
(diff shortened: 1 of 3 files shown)


noahdietz:
Fixed tests and added flag to `update-image` to be safe, since it may run `generate` (though IDK if that command will be used anytime soon)

PR #3676, opened 2026-01-21
fix(librarian): bump command only modifies versions in librarian.yaml

### LINKED ISSUE #3668: librarian: config mutation is reflected in librarian.yaml
It's possible my understanding is incorrect here, but basically I believe various commands go through a process of:

1. Load librarian.yaml
2. Fill in values that can be inferred, so that language-specific code doesn't have to
3. Pass the config (or parts of it, eg. `*config.Library` parameters) to language-specific code, which may modify it (and indeed that could be part of the purpose of it)
4. Save librarian.yaml with changes

That means we get all the inferred values present in librarian.yaml, which we don't want.

This is part of the cause of #3652. Even if we hadn't seen a failure, `librarian bump --all` would have updated `librarian.yaml` in ways we don't want.

We should figure out a design such that:

- Language specific code doesn't need to duplicate logic to infer values
- Only changes *intended* to be persisted are actually persisted

jskeet:
First instance of a pattern we may need elsewhere: if we need to write out a new version of librarian.yaml at the end of a command, and the command uses code which needs defaults filling in (as per prepareLibrary) or might make other changes we don't want to persist, clone the config first and only copy the desired changes at the end.

Fixes #3668

### DIFF
--- a/internal/librarian/bump.go
+++ b/internal/librarian/bump.go
@@ -412,12 +412,14 @@ func copyLibraryVersions(dirtyCfg, pristineCfg *config.Config) (*config.Config, 
 	if len(dirtyCfg.Libraries) != len(pristineCfg.Libraries) {
 		return nil, fmt.Errorf("mismatched library count after bump: %d != %d", len(dirtyCfg.Libraries), len(pristineCfg.Libraries))
 	}
-	// We don't care about whether or not the libraries are in the same order,
-	// so we just find them by name.
+	pristineLibraries := make(map[string]*config.Library, len(pristineCfg.Libraries))
+	for _, library := range pristineCfg.Libraries {
+		pristineLibraries[library.Name] = library
+	}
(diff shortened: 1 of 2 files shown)


jskeet:
(The coverage is tricky to improve as I'd need to figure out a way of making configuration formatting/parsing fail...)

jskeet:
@julieqiu As I commented on #3678, I suspect we *will* want `prepareLibrary` inside `bump`.

Basically I suspect that everywhere that language-specific code will need to do anything non-trivial (e.g. modify a file), it will benefit from being passed a "normalized" config, where it can know immediately where to find the source code etc. I don't think we want to bake "apply defaults to figure stuff out" code in each language's functions.

In fact, even the existing Rust code for `bump` assumes that `library.Output` has been populated, so I'm not sure how that would work right now for libraries which don't specify the output in `librarian.yaml`. It's possible that I'm missing something - or it's possible that #3678 broke `bump`.

Issue #3547, opened 2026-01-13
libarian: report error detail when failed to publish


zhumin8:
When publish failed halfway, librarian should report error better for debugging.

currently it just show command failed without any hint on why:
```
➜  google-cloud-rust git:(main) go run github.com/googleapis/librarian/cmd/librarian@${V} publish --skip-semver-checks
go: github.com/googleapis/librarian@v0.7.1-0.20260112221555-2ac5eec86b21 requires go >= 1.25.5; switching to go1.25.5
Running: /usr/bin/git --version
Running: /usr/bin/git remote get-url upstream
Running: /usr/local/google/home/aviebrantz/.cargo/bin/cargo --version
2026/01/13 19:44:25 INFO computing publication plan with: cargo workspaces plan
2026/01/13 19:45:22 INFO there are 193 crates in need of publishing, summary=[google-cloud-api-servicemanagement-v1 google-cloud-appengine-v1 google-cloud-eventarc-publishing-v1 google-cloud-longrunning google-cloud-monitoring-v3 google-cloud-spanner-admin-database-v1 google-cloud-assuredworkloads-v1 google-cloud-datastream-v1 google-cloud-notebooks-v2 google-cloud-build-v2 google-cloud-storagetransfer-v1 google-cloud-auth google-cloud-discoveryengine-v1 google-cloud-documentai-v1 google-cloud-metastore-v1 google-cloud-orgpolicy-v2 google-cloud-video-livestream-v1 google-cloud-beyondcorp-appconnectors-v1 google-cloud-confidentialcomputing-v1 google-cloud-oslogin-v1]
2026/01/13 19:45:22 INFO publishing crates with: cargo workspaces publish --skip-published ...
exit status 1
exit status 1
```
In this particular run, last successes published service was google-cloud-financialservices-v1 v1.4.0 26 minutes ago and `cargo workspaces plan --skip-published` shows [list](https://paste.googleplex.com/6296021506326528) not published.

A more verbose print on what succeeded and/or failed would help.

PR #3688, opened 2026-01-22
feat(internal/command): add -v flag

julieqiu:
librarian now supports a -v flag. When -v is set, the command is printed to stdout.

Fixes https://github.com/googleapis/librarian/issues/3547

### DIFF
--- a/internal/command/command.go
+++ b/internal/command/command.go
@@ -22,6 +22,12 @@ import (
 	"os/exec"
 )
 
+// Verbose controls whether commands are printed to stderr before execution.
+//
+// TODO(https://github.com/googleapis/librarian/issues/3687): pass in as
+// config.
+var Verbose bool
+
 // Run executes a program (with arguments) and captures any error output. It is a
 // convenience wrapper around RunWithEnv.
 func Run(ctx context.Context, command string, arg ...string) error {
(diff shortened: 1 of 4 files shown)


internal/command/command.go
```diff
@@ -39,6 +45,9 @@ func RunWithEnv(ctx context.Context, env map[string]string, command string, arg
 			cmd.Env = append(cmd.Env, k+"="+v)
 		}
 	}
+	if Verbose {
+		fmt.Fprintf(os.Stdout, "%s\n", cmd.String())
+	}
 	if output, err := cmd.CombinedOutput(); err != nil {
 		return fmt.Errorf("%v: %v\n%s", cmd, err, output)
```
zhumin8: Looks like to me that when error happens, it should reported back with the cmd. Wonder why it only reported `exit status 1` in #3547?

internal/command/command.go
```diff
@@ -39,6 +45,9 @@ func RunWithEnv(ctx context.Context, env map[string]string, command string, arg
 			cmd.Env = append(cmd.Env, k+"="+v)
 		}
 	}
+	if Verbose {
+		fmt.Fprintf(os.Stdout, "%s\n", cmd.String())
```
zhumin8: Do you think it worth to steam the output of the command to show real-time progress in verbose mode?

internal/command/command.go
julieqiu: Hmm I'm not sure - I will investigate as a follow up

Issue #3580, opened 2026-01-15
tool/cmd/migrate: support Dart migration


julieqiu:
https://github.com/googleapis/google-cloud-dart currently uses sidekick. We need to update tool/cmd/migrate (and the corresponding internal/config) to support anything that is used.

PR #3760, opened 2026-01-29
feat(tool/cmd/migrate): parse Dart packages from root `.sidekick.toml`

JoeWang1127:
Parse Dart package dependency information from root `.sidekick.toml`.

I think the migration script needs lots of modification so I'll split them into small ones for quick review.

For #3580

### DIFF
--- a/tool/cmd/migrate/main.go
+++ b/tool/cmd/migrate/main.go
@@ -722,7 +722,6 @@ func parseDartPackages(codec map[string]string) map[string]string {
 		}
 		packages[key] = value
 	}
-
 	return packages
 }
 


tool/cmd/migrate/main_test.go
```diff
@@ -889,3 +889,33 @@ func TestRunMigrateCommand(t *testing.T) {
...
+				"package:http":               "^1.3.0",
+			},
+			want: map[string]string{
+				"package:googleapis_auth": "^2.0.0",
+				"package:http":            "^1.3.0",
+			},
+		},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			t.Parallel()
```
julieqiu: does t.Parallel actually improve performance? I don't think it would in this case, in which case I'd leave it out 

see https://stackoverflow.com/questions/28058640/what-is-the-practical-benefit-to-putting-t-parallel-at-the-top-of-my-tests

PR #3766, opened 2026-01-29
fix: avoid hardcoding repometadata language

### LINKED ISSUE #3159: librarian: make full config available to generators
Currently the language-specific generation code is provided with a library and sources, but not the root config which contains the language and repo name. These need to be provided to GenerateRepoMetadata, so the root config should probably be available everywhere.

jameslynnwu:
The config will be used to avoid hardcoding language in repometadata. It should also be used for repo, but currently the librarian configs do not have repo set.

Fixes #3159

### DIFF
--- a/internal/librarian/python/generate_test.go
+++ b/internal/librarian/python/generate_test.go
@@ -366,7 +366,7 @@ func TestGenerate(t *testing.T) {
 			},
 		},
 	}
-	if err := Generate(t.Context(), &config.Config{}, library, googleapisDir); err != nil {
+	if err := Generate(t.Context(), &config.Config{Language: "python"}, library, googleapisDir); err != nil {
 		t.Fatal(err)
 	}
 	if _, err := os.Stat(filepath.Join(outdir, ".repo-metadata.json")); err != nil {


internal/librarian/dart/generate.go
```diff
@@ -51,7 +51,7 @@ func Format(ctx context.Context, library *config.Library) error {
 	return nil
 }
 
-func toSidekickConfig(library *config.Library, ch *config.API, googleapisDir string) (*sidekickconfig.Config, error) {
+func toSidekickConfig(cfg *config.Config, library *config.Library, ch *config.API, googleapisDir string) (*sidekickconfig.Config, error) {
```
JoeWang1127: Dart doesn't need to generate repo metadata, why we need to change this function?

I think if a language has a special param, we can use a function to saturate the value to the library before running this function, for example, `library = mergeDefaults()`.

internal/librarian/rust/generate_test.go
```diff
@@ -110,7 +110,7 @@ func TestSkipGenerateVeneer(t *testing.T) {
 	sources := &Sources{
 		Googleapis: googleapisDir,
 	}
-	if err := Generate(t.Context(), library, sources); err != nil {
+	if err := Generate(t.Context(), &config.Config{}, library, sources); err != nil {
```
julieqiu: Instead of `&config.Config{}`, it is more idiomatic for the function to accept `nil` as input and to pass that in when a value is not used

julieqiu (changes requested):

internal/librarian/dart/generate.go
jameslynnwu: Will revert changes since dart does not need to generate repo metadata.

However, I'm not sure about adding fields to library. Our code is already doing a lot of struct to struct transformations. we can discuss later on.

internal/sidekick/config/config.go
```diff
@@ -60,6 +61,8 @@ type Config struct {
 	CommentOverrides    []DocumentationOverride `toml:"documentation-overrides,omitempty"`
 	PaginationOverrides []PaginationOverride    `toml:"pagination-overrides,omitempty"`
 	Release             *Release                `toml:"release,omitempty"`
+	// Add config until we can migrate to it
+	Config *config.Config `toml:"-"`
```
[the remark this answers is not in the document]
jameslynnwu: responded to this above.

internal/librarian/python/generate.go
```diff
@@ -29,7 +29,7 @@ import (
 )
 
 // Generate generates a Python client library.
-func Generate(ctx context.Context, library *config.Library, googleapisDir string) error {
+func Generate(ctx context.Context, cfg *config.Config, library *config.Library, googleapisDir string) error {
```
julieqiu: It looks like the only value you need is language, in which case I would make the signature:

```
func Generate(ctx context.Context, cfg *config.Config, library *config.Library, googleapisDir, language string) error {
```

internal/librarian/dart/generate.go
```diff
@@ -28,8 +28,8 @@ import (
 )
 
 // Generate generates a Dart client library.
-func Generate(ctx context.Context, library *config.Library, googleapisDir string) error {
-	sidekickConfig, err := toSidekickConfig(library, library.APIs[0], googleapisDir)
+func Generate(ctx context.Context, cfg *config.Config, library *config.Library, googleapisDir string) error {
+	sidekickConfig, err := toSidekickConfig(cfg, library, library.APIs[0], googleapisDir)
```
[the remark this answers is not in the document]
julieqiu: The way I would approach this is to create a parser.Config that contains just the values used by parser.CreateModel, https://github.com/googleapis/librarian/blob/3dbe74ff9c635eaeb04b44387d4b21123f3c9468/internal/sidekick/parser/parser.go#L27, and same for other places. That way the config we are parser to these functions contain only the values needed by that function instead of a big grab bag

PR #3774, opened 2026-01-30
chore(tool/cmd/migrate): delete unused Rust migration code

JoeWang1127:
Remove unused functions in the migration tool.

Now that we have migrated Rust to librarian, delete the unused migration code. `readCargoConfig` is not removed because it requires additional logic to parse the library name, which will be added in follow-up PRs.

For #3580


julieqiu (approved):
Approved and disabled auto-merge for you to fix the title and commit message before merging.

PR #3792, opened 2026-01-31
feat: Use repometadata.GenerateRepoMetadata for generating .repo-metadata.json

### LINKED ISSUE #3606: rust: use repometadata.GenerateRepoMetadata for generating .repo-metdata.json
We should use the same method to generate the .repo-metdata.json file throughout the codebase. Consolidate on `repometadata.GenerateRepoMetadata`.

jameslynnwu:
Deletes the repo-metadata.json mustache template.

Fixes #3606


internal/librarian/rust/generate.go
```diff
@@ -71,6 +72,18 @@ func Generate(ctx context.Context, library *config.Library, sources *Sources) er
...
+	// TODO(https://github.com/googleapis/librarian/issues/3146):
+	// Remove the default version fudget here, as Generate should
+	// compute it. For now, use the last component of the first api path as
+	// the default version.
+	defaultVersion := filepath.Base(library.APIs[0].Path)
+
+	absoluteServiceConfig := filepath.Join(sources.Googleapis, sidekickConfig.General.ServiceConfig)
+	if err := repometadata.Generate(library, cfg.Language, "googleapis/google-cloud-python", absoluteServiceConfig, defaultVersion, library.Output); err != nil {
+		return fmt.Errorf("failed to generate .repo-metadata.json: %w", err)
+	}
```
coryan: Drive-by: this is going to generate a `.repo-metadata.json` file for **all** templates, but previously only the `crate` template generated it. 

For example, these directories contain generated files, but it does not have, and we do not want to have a `.repo-metadata.json` file:

https://github.com/googleapis/google-cloud-rust/tree/main/src/wkt/src/generated
https://github.com/googleapis/google-cloud-rust/tree/main/src/storage/src/generated/gapic

Consider a test to (a) verify the `.repo-metadata.json` is generated for a "crate"-type library, and (b) verify it is **not** generated for a "mod"-type library.

coryan: I think I follow, maybe?  Sounds like you have this under control.

If you have time, follow the instructions to test this in `google-cloud-rust`:

https://github.com/googleapis/google-cloud-rust/blob/main/doc/contributor/howto-guide-generated-code-maintenance.md#making-changes-to-librarian

That should be an integration test, but we are not quite there yet.

coryan (approved):

Issue #3627, opened 2026-01-19
serviceconfig: validate google-cloud-rust/librarian.yaml against APIs


julieqiu:
Verify serviceconfig/api.go includes all APIs in librarian.yaml, then update serviceconfig.Find to only allow APIs listed in the APIs arrary.

julieqiu:
The work involved should be to delete https://github.com/googleapis/librarian/blob/737e0d380ecf4a0dd91aeddeac181945b117b209/internal/serviceconfig/serviceconfig.go#L101 and confirm that generate will still work in google-cloud rust.

Issue #3617, opened 2026-01-17
golang: add generate command


julieqiu:
Add generate support for Go.

PR #3804, opened 2026-02-02
feat(internal/serviceconfig): add FindGRPCServiceConfig

julieqiu:
Add FindGRPCServiceConfig to discover gRPC service config files (*_grpc_service_config.json) in API directories, which are relied upon by various language generators, such as Python and Go.

Updates the Python generator to use this function.

For https://github.com/googleapis/librarian/issues/3617
For https://github.com/googleapis/librarian/issues/3627


jskeet (approved):
There's still the aspect about the directory being missing to consider, but I don't think that should block merging.

julieqiu:
I filed https://github.com/googleapis/librarian/issues/3814 as a follow up.

PR #3822, opened 2026-02-02
feat(rust): implement `internal-builder` visibility option

### LINKED ISSUE #3784: rust: remove pubsub internal-only types from public api
This issue serves https://github.com/googleapis/google-cloud-rust/issues/3959.

Currently, we are exposing builders with the tag `#[doc(hidden)]` as a workaround for doctest failures.
We should fix this via a Sidekick configuration to remove these types from the public API entirely, rather than just hiding them from the documentation. 

This requires:
- modifying the Rust templates to respect these visibility settings
- adding an configuration option to specify affected modules.

haphungw:
Implement a new config option `internal_builder` for Rust modules. 
When set to true, this option restricts the visibility of generated request builders and client constructors from `pub` to `pub(crate)`, and remove the doctests which would fail as the builders are no longer `pub`.
By making these internal builders internal to the crate, we prevent them from leaking into the public API while still allowing them to be used by the veneer.
Fix #3784

### DIFF
--- a/internal/config/language.go
+++ b/internal/config/language.go
@@ -79,6 +79,9 @@ type RustModule struct {
 	// IncludeList is a list of proto files to include (e.g., "date.proto,expr.proto").
 	IncludeList string `yaml:"include_list,omitempty"`
 
+	// InternalBuilder indicates whether generated builders should be internal to the crate.
+	InternalBuilder bool `yaml:"internal_builder,omitempty"`
+
 	// Language can be used to select a variation of the Rust generator.
 	// For example, `rust_storage` enables special handling for the storage client.
 	Language string `yaml:"language,omitempty"`
--- a/internal/config/language.go
+++ b/internal/config/language.go
@@ -79,8 +79,8 @@ type RustModule struct {
 	// IncludeList is a list of proto files to include (e.g., "date.proto,expr.proto").
 	IncludeList string `yaml:"include_list,omitempty"`
 
-	// InternalBuilder indicates whether generated builders should be internal to the crate.
-	InternalBuilder bool `yaml:"internal_builder,omitempty"`
+	// InternalBuilders indicates whether generated builders should be internal to the crate.
+	InternalBuilders bool `yaml:"internal_builders,omitempty"`
 
 	// Language can be used to select a variation of the Rust generator.
 	// For example, `rust_storage` enables special handling for the storage client.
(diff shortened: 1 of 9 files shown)


internal/config/language.go
```diff
@@ -79,6 +79,9 @@ type RustModule struct {
 	// IncludeList is a list of proto files to include (e.g., "date.proto,expr.proto").
 	IncludeList string `yaml:"include_list,omitempty"`
 
+	// InternalBuilder indicates whether generated builders should be internal to the crate.
+	InternalBuilder bool `yaml:"internal_builder,omitempty"`
```
coryan: nit: that suggest the name should be `InternalBuilders` ? Or maybe `CrateBuilders`.  But I am terrible at naming.

coryan (changes requested):
The naming nits are just that: nits.

The mustache templates are hard to follow, see my suggestions below.

internal/config/language.go
haphungw: would that give the impression that `InternalBuilders` expects a list of string? 
we have `InternalTypes` expect a list of strings.

coryan: As I said, I am terrible at naming.  But since this makes multiple `*Builder` types internal I figured it should be plural.  Don't get hang up on this. We can always change this if we change our minds. Talk to somebody with stronger opinions on naming if you want.

internal/sidekick/rust/annotate_model_test.go
```diff
@@ -136,6 +136,54 @@ func TestClippyWarnings(t *testing.T) {
...
+			t.Errorf("mismatch in InternalBuilders, want=%v, got=%v", test.Want, got.InternalBuilders)
+		}
+		svcAnn := model.Services[0].Codec.(*serviceAnnotations)
+		if svcAnn.InternalBuilders != test.Want {
+			t.Errorf("mismatch in service InternalBuilders, want=%v, got=%v", test.Want, svcAnn.InternalBuilders)
+		}
+		wantVisibility := "pub"
+		if test.Want {
+			wantVisibility = "pub(crate)"
+		}
```
coryan: nit: conditional code in tests is always suspect, consider using an extra field in the test structure.

coryan (approved):

coryan (approved):
:shipit:

PR #3845, opened 2026-02-04
Draft removal of toSidekickConfig

jameslynnwu:


jameslynnwu:
wanted a spot check to see if its on the right path.

Issue #3662, opened 2026-01-21
rust: delete toSidekickConfig


julieqiu:
Delete sidekick.Config and translation layer between librarian config and sidekick config.

julieqiu:
Other places to replace this:

https://github.com/googleapis/librarian/blob/3dbe74ff9c635eaeb04b44387d4b21123f3c9468/internal/sidekick/rust/generate.go#L31

https://github.com/googleapis/librarian/blob/3dbe74ff9c635eaeb04b44387d4b21123f3c9468/internal/sidekick/rust_prost/generate.go#L35

https://github.com/googleapis/librarian/blob/3dbe74ff9c635eaeb04b44387d4b21123f3c9468/internal/sidekick/dart/generate.go#L31

PR #3870, opened 2026-02-04
refactor(librarian/rust): extract model overrides from source

jameslynnwu:
Config.Sources contains a mix of data about the files sources that are used in client library generation and overrides for the model.

This change removes the model overrides from Config.Sources and moves them to an modelOverrides struct, which will be used to hold other pieces of Config related to model overrides such as pagination and documentation overrides.

There are some callers to NewModelOverridesFromSource, which generates model overrides from sidekick configs that still populate source with non-source information. We'll clean those up later on.

Part of #3662

### DIFF
--- a/internal/sidekick/codec_sample/generate_test.go
+++ b/internal/sidekick/codec_sample/generate_test.go
@@ -53,7 +53,7 @@ func TestFromProtobuf(t *testing.T) {
 			"proto:google.cloud.location": "package:google_cloud_location/location.dart",
 		},
 	}
-	model, err := parser.CreateModel(cfg, parser.NewModelOverridesFromSource(cfg.Source))
+	model, err := parser.CreateModel(cfg, &parser.ModelOverrides{})
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/internal/sidekick/dart/generate_test.go
+++ b/internal/sidekick/dart/generate_test.go
@@ -61,7 +61,7 @@ func TestFromProtobuf(t *testing.T) {
 			"proto:google.cloud.location":    "package:google_cloud_location/location.dart",
 		},
 	}
-	model, err := parser.CreateModel(cfg, parser.NewModelOverridesFromSource(cfg.Source))
+	model, err := parser.CreateModel(cfg, &parser.ModelOverrides{})
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/internal/sidekick/rust/generate_test.go
+++ b/internal/sidekick/rust/generate_test.go
@@ -98,7 +98,7 @@ func TestCodecError(t *testing.T) {
 			"--invalid--": "--invalid--",
 		},
 	}
-	model, err := parser.CreateModel(errorConfig, parser.NewModelOverridesFromSource(errorConfig.Source))
+	model, err := parser.CreateModel(errorConfig, &parser.ModelOverrides{})
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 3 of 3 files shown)


internal/sidekick/parser/protobuf.go
```diff
@@ -66,13 +66,13 @@ func newCodeGeneratorRequest(source string, options map[string]string) (_ *plugi
...
 
-	files, err := protobuf.DetermineInputFiles(source, options)
+	files, err := protobuf.DetermineInputFiles(source, sourceRoots, overrides.IncludeList, overrides.ExcludeList)
 	if err != nil {
 		return nil, err
 	}
 
 	// Call protoc with the given arguments.
-	contents, err := protoc(tempFile.Name(), files, options)
+	contents, err := protoc(tempFile.Name(), files, sourceRoots)
```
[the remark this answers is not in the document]
jameslynnwu: This is fine, since include-list is only used to allowlist files

internal/sidekick/protobuf/input_files.go
```diff
@@ -88,24 +86,22 @@ func findFiles(files map[string]bool, source string) error {
...
+func applyIncludeList(files map[string]bool, sourceDirectory string, list []string) {
+	if len(list) == 0 {
 		return
 	}
 	// Ignore any discovered paths, only the paths from the include list apply.
 	clear(files)
-	for _, p := range strings.Split(list, ",") {
+	for _, p := range list {
 		files[filepath.ToSlash(path.Join(sourceDirectory, p))] = true
 	}
```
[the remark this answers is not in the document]
jameslynnwu: This is fine, since include-list is only used to allowlist files

internal/sidekick/codec_sample/generate_test.go
```diff
@@ -53,7 +53,7 @@ func TestFromProtobuf(t *testing.T) {
 			"proto:google.cloud.location": "package:google_cloud_location/location.dart",
 		},
 	}
-	model, err := parser.CreateModel(cfg)
+	model, err := parser.CreateModel(cfg, &parser.ModelOverrides{})
```
jameslynnwu: is it best to update CreateModel to create an empty ModelOverride if nil is passed in instead?

internal/sidekick/parser/parser.go
```diff
@@ -24,21 +24,22 @@ import (
 // CreateModel parses the service specification referenced in `config`,
 // cross-references the model, and applies any transformations or overrides
 // required by the configuration.
-func CreateModel(config *config.Config) (*api.API, error) {
+func CreateModel(config *config.Config, overrides *ModelOverrides) (*api.API, error) {
```
julieqiu: I don't think I quite understand this change. I would have expected something like

```
func CreateModel(overrides *ModelOverrides) (*api.API, error) { ... }
```

where overrides has just the fields from config that CreateModel uses, since the goal is to get rid of config.Config

julieqiu: this is what I think the struct would be:

```
  type ModelConfig struct {
      // Source configuration
      SpecificationFormat  string
      SpecificationSource  string
      SourceRoots          map[string]string

      // Model overrides
      Name        string
      Title       string
      Description string
      SkippedIDs  []string
      IncludedIDs []string

      // Documentation/pagination overrides
      CommentOverrides    []config.DocumentationOverride
      PaginationOverrides []config.PaginationOverride
  }
  ```

Then `CreateModel(overrides *ModelOverrides)` would be self-contained.

PR #3876, opened 2026-02-05
fix(internal/librarian): remove default specification_format in tidy

julieqiu:
The tidy command now removes `specification_format: protobuf` from librarian.yaml since "protobuf" is the default value.


internal/librarian/tidy.go
```diff
@@ -76,6 +76,9 @@ func tidyLibrary(cfg *config.Config, lib *config.Library) error {
 	if len(lib.Roots) == 1 && lib.Roots[0] == "googleapis" {
 		lib.Roots = nil
 	}
+	if lib.SpecificationFormat == "protobuf" {
```
jameslynnwu: i wish this was a enum

Issue #3633, opened 2026-01-19
sidekick: delete CLI once google-cloud-rust and google-cloud-dart have fully migrated


julieqiu:
Once both google-cloud-rust and google-cloud-dart have fully migrated to the new pipeline, the CLI will no longer be needed and should be removed.

PR #3920, opened 2026-02-06
refactor(internal/sidekick): merge testdata directories

julieqiu:
Move internal/sidekick/testdata into internal/testdata and internal/sidekick/rust/testdata to reduce duplication and simplify the overall testdata structure.

For #3633


internal/sidekick/parser/discovery/discovery_test.go
```diff
@@ -29,7 +29,7 @@ import (
 	"github.com/googleapis/librarian/internal/sidekick/sample"
 )
 
-const computeDiscoveryFile = "../../testdata/disco/compute.v1.json"
+const computeDiscoveryFile = "../../../testdata/discovery/compute.v1.json"
```
coryan: But we can sing ABBA songs if it is called `disco` ... sad.

I don't know any ABBA songs, double sad.

julieqiu: 💀

Issue #3911, opened 2026-02-05
librarianops: add upgrade command


julieqiu:
Add command

   librarianops upgrade

This will:

1. Get the latest version of librarian
2. Update librarian.yaml
3. Run `librarian generate --all`

PR #3926, opened 2026-02-06
build(internal/librarianops/upgrade): add version config file updater

miguelvelezsa:
[2 of 4]

Create librarian_config_updater file. File consist in the function to update a librarian.yaml config file.
Second step in https://github.com/googleapis/librarian/issues/3911.


internal/librarianops/upgrade/librarian_config_updater.go
```diff
@@ -0,0 +1,49 @@
...
+func getConfigFile(configPath string) (*config.Config, error) {
+	if _, err := os.Stat(configPath); os.IsNotExist(err) {
+		return nil, fmt.Errorf("config file does not exist at path: %s", configPath)
+	}
+	return yaml.Read[config.Config](configPath)
+}
+
+// UpdateLibrarianVersion updates the version field in the librarian.yaml config file with the provided version.
+// If the file does not exist, returns an error.
+func UpdateLibrarianVersion(version, configPath string) error {
```
JoeWang1127: I think you can just use `repoPath` as an input and remove `GenerateLibrarianConfigPath` and `getConfigFile`, since they seem very straightforward.

internal/librarianops/upgrade/librarian_config_updater.go
```diff
@@ -0,0 +1,49 @@
+// Copyright 2026 Google LLC
```
julieqiu: ditto re: put this in internal/librarianops/upgrade.go as opposed to a separate package

internal/librarianops/upgrade/librarian_config_updater.go
miguelvelezsa: > [librarian/internal/librarianops/generate.go](https://github.com/googleapis/librarian/blob/e20ee8818e0b971e21e74ac05f4c2a59f6df012c/internal/librarianops/generate.go#L222)
> 
> Line 222 in [e20ee88](/googleapis/librarian/commit/e20ee8818e0b971e21e74ac05f4c2a59f6df012c)
> 
>  func updateLibrarianVersion(version, repoDir string) error { 
> already exists, so you can reuse that

If I use that function the upgrade.go will depend on generate.go. I prefer to have a separated file with the update logic so we can easily test it and modify it. Also that logic from generate.go will be deleted in a follow up PR.

miguelvelezsa: > I think you can just use `repoPath` as an input and remove `GenerateLibrarianConfigPath` and `getConfigFile`, since they seem very straightforward.

I think is better to keep them. Even though the logic seems straightforward, now we have a few functions doing one thing, we can fully and easily test them and if something change we can directly go the function and change it.

internal/librarianops/librarian_config_updater.go
```diff
@@ -0,0 +1,49 @@
...
+func getConfigFile(configPath string) (*config.Config, error) {
+	if _, err := os.Stat(configPath); os.IsNotExist(err) {
+		return nil, fmt.Errorf("config file does not exist at path: %s", configPath)
+	}
+	return yaml.Read[config.Config](configPath)
+}
+
+// UpdateLibrarianVersionInConfigFile updates the version field in the librarian.yaml config file with the provided
+// version. If the file does not exist, returns an error.
+func UpdateLibrarianVersionInConfigFile(version, configPath string) error {
```
julieqiu: This looks like a duplicate of https://github.com/googleapis/librarian/blob/43b900a12d48f699ffee781990826c93087b8f6f/internal/librarianops/generate.go#L222

PR #3934, opened 2026-02-06
prototype of configurable parallelism

jskeet:


internal/config/config.go
```diff
@@ -41,11 +41,30 @@ type Config struct {
...
 	// special handling, and differ from default settings.
 	Libraries []*Library `yaml:"libraries,omitempty"`
 }
 
+// Parallelization indicates the degree of parallelization that should be used
+// for various operations. Any negative value represents unlimited
+// parallelization; a positive number represents a specific limit of
+// parallelization; 0 is treated as equivalent to 1 (for the sake of
+// convenience, as a default value).
+type Parallelization struct {
```
julieqiu: I don't think we should put this in the librarian.yaml. We should just hardcode it in internal/librarian/{language} since it is an implementation detail of the function.

jskeet:
@julieqiu Fair enough. My guess is that the simplest way to do that is to change the sort of interface we've got with the languages to pass in the whole slice of libraries to clean (generate, format) and let the language decide whether to do it in series or parallel.

That means a bit of duplication (looping, basically) but is probably simpler than a sort of "parallelism negotiation" where the agnostic code asks the language code whether or not to parallelize - but you may have had something else in mind.

If you're happy with the "pass the slice in" approach I'll create a new PR tomorrow.

jskeet:
Closing in favor of #4005 (which still may not be what we want).

PR #3938, opened 2026-02-06
feat(internal/librarian/python): first part of bump for python

jskeet:
Adds version bumping for gapic_version.py files. There will be other files (in other formats) to bump over time, but this gets us a bit closer.


internal/librarian/python/bump.go
```diff
@@ -0,0 +1,85 @@
...
+
+// bumpGapicVersion finds all gapic_version.py files under output. For each
+// file, it expect to find a line starting "__version__ = ", and changes that
+// line to end with the specified version (in quotes, with no escaping applied -
+// we don't expect version numbers to require quoting). Any existing text after
+// "__version__ = " (on the same line) is omitted. All other lines are
+// preserved.
+func bumpGapicVersions(output, version string) error {
+	return filepath.WalkDir(output, func(path string, d fs.DirEntry, err error) error {
+		if d.Name() == gapicVersionFile {
```
JoeWang1127: Could you change to
```go
if d.Name() != gapicVersionFile {
  return nil
}
```

See https://github.com/googleapis/librarian/blob/main/doc/howwewritego.md#avoid-unnecessary-else

JoeWang1127 (approved):

Issue #3618, opened 2026-01-17
tool/cmd/migrate: support Go


julieqiu:
Ensure that a valid librarian.yaml can be created for Go.

PR #3958, opened 2026-02-09
fix(internal/serviceconfig): set transport at API level

JoeWang1127:
Set transport of `google/cloud/aiplatform/v1` to grpc.

The value is non-default and is different than another api (`google/cloud/aiplatform/v1beta1`) in the library. Therefore, set it in API level, rather than library level.

For #3618

### DIFF
--- a/internal/librarian/golang/generate.go
+++ b/internal/librarian/golang/generate.go
@@ -375,7 +375,7 @@ func updateSnippetMetadata(library *config.Library, output string) error {
 // The default value is serviceconfig.GRPCRest.
 func getTransport(sc *serviceconfig.API) string {
 	if sc != nil {
-		return sc.GetTransport("go")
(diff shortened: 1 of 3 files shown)


internal/librarian/golang/generate.go
```diff
@@ -373,3 +369,13 @@ func updateSnippetMetadata(library *config.Library, output string) error {
 		return nil
 	})
 }
+
+// getTransport get transport from serviceconfig.API for language Go.
+//
+// The default value is serviceconfig.GRPCRest.
+func getTransport(sc *serviceconfig.API) string {
+	if sc != nil {
+		return sc.Transport("go")
```
jameslynnwu: we should probably have a task to create an enum for language

jameslynnwu (approved):

PR #3961, opened 2026-02-09
fix(internal/librarian/golang): get release level for an api based on api path and library version

JoeWang1127:
For Go libraries, get release level based on the api path and library version.

Some library, `aiplatform` for example, has different release level in APIs.

For #3617

### DIFF
--- a/internal/librarian/golang/generate_test.go
+++ b/internal/librarian/golang/generate_test.go
@@ -394,6 +394,12 @@ func TestReleaseLevel_Success(t *testing.T) {
 			version: "",
 			want:    "alpha",
 		},
+		{
+			name:    "empty version with beta path",
+			apiPath: "google/cloud/alphabet/v1beta1",
+			version: "",
+			want:    "beta",
+		},
 	} {
 		t.Run(test.name, func(t *testing.T) {
 			got, err := releaseLevel(test.apiPath, test.version)


internal/serviceconfig/api.go
```diff
@@ -59,15 +60,20 @@ type API struct {
...
 
+	// Path is the proto directory path in github.com/googleapis/googleapis.
+	// If ServiceConfig is empty, the service config is assumed to live at this path.
+	Path string
+
+	// ReleaseLevel is the release level, such as "stable" or "preview". The value is different among
+	// languages.
+	// The key is the language name, e.g., "go", "python".
+	// Optional. If omitted, the value is empty.
+	ReleaseLevels map[string]string
```
zhumin8: What is the reason putting into API, I suppose because this needed to be per API?

For long run, we want to keep this API for configs that are language agnostic.

internal/librarian/golang/generate.go
```diff
@@ -373,3 +373,23 @@ func updateSnippetMetadata(library *config.Library, output string) error {
...
+// getReleaseLevel get release level for an API for language Go.
+//
+// The preference is as follows:
+//
+// 1. the value found in serviceconfig.API.
+//
+// 2. the value found in config.Library.
+//
+// 3. empty value.
+func getReleaseLevel(library *config.Library, sc *serviceconfig.API) string {
```
zhumin8: I see there is a comparison with val from service config. Can we fill this value in serviceconfig.Find similar to title?

nit: per go convention, getter should usually omit "get" in name

JoeWang1127:
> IIUC, go publishes libraries as bundling versions like v1, v1beta1 together in one module, similar to java. Then how can they have different release levels within? Or is this release level not used for versioning?

I found this in the protoc command in different api version, the source of truth is `BUILD.bazel`.

internal/librarian/golang/generate.go
zhumin8: As part of #3960, serviceconfig.Find will know the language. Does this change things for you?

noahdietz:
> I'll infer the release level from the api path.

yeah the only ones that need to be explicitly set (I believe) are `vN` that are "soaking" (just recently released and waiting to go 1.x). might want to double check that though

Python's "preview" mechanism may be different though!

internal/serviceconfig/api.go
JoeWang1127: This is irrelevant now.

JoeWang1127:
> yeah the only ones that need to be explicitly set (I believe) are vN that are "soaking" (just recently released and waiting to go 1.x). might want to double check that though

Do you manually change the release level from `beta` to `ga` after the soaking time elapsed?

JoeWang1127:
> @JoeWang1127 keep in mind, idk how python or other languages use this

This is PR is go specific so it doesn't affect Python.

> I'm not sure...the first version of a new client is currently 0.1.0 set to, which, with this logic, would result in the releaseLevel being beta (even for vN api packages), only switching to ga when a 1.0.0 is forced e.g. via --version. Typing that out, it probably makes sense as an algorithm to start with.

Since the release level is only used in `doc.go` and repo metadata, we can regenerate these files when bump the version to `1.0.0`.

internal/librarian/golang/generate.go
```diff
@@ -370,10 +379,26 @@ func updateSnippetMetadata(library *config.Library, output string) error {
 	})
 }
 
-// getTransport get transport from serviceconfig.API for language Go.
+// releaseLevel get release level for an API for language Go.
```
jskeet:
```suggestion
// releaseLevel determines the release level for an API based on the API path and the library's current version.
```

internal/librarian/golang/generate.go
```diff
@@ -370,10 +379,26 @@ func updateSnippetMetadata(library *config.Library, output string) error {
...
+func releaseLevel(apiPath, version string) (string, error) {
+	if strings.Contains(apiPath, releaseLevelAlpha) {
```
jskeet: I think we should take the last part of the API path and check in there - if we had "google/alphabet/v1" that shouldn't count as an alpha API.

internal/librarian/golang/generate.go
```diff
@@ -370,10 +379,26 @@ func updateSnippetMetadata(library *config.Library, output string) error {
...
+		return releaseLevelAlpha, nil
+	}
+	semverVer, err := semver.NewVersion(version)
+	if err != nil {
+		return "", err
+	}
+	if strings.Contains(apiPath, releaseLevelBeta) || semverVer.Major() < 1 {
```
jskeet: I'd be tempted to use the semver version to start with: if it's a stable semver version, it's GA. Otherwise, use the path to figure out alpha or beta.

Caveats for this: I'm hoping we can move to "always start at 1.0.0-alpha.1, 1.0.0-beta.1 or 1.0.0" rather than using 0.x versions... which means we can't just use the semver major version. But it would be okay to have a TODO for that for later.

internal/librarian/golang/generate_test.go
```diff
@@ -352,6 +353,50 @@ func TestBuildGAPICImportPath(t *testing.T) {
...
+		},
+		{
+			name:    "alpha",
+			apiPath: "google/cloud/secretmanager/v1alpha1",
+			want:    "alpha",
+		},
+		{
+			name:    "beta",
+			apiPath: "google/cloud/secretmanager/v1beta2",
+			version: "1.0.0",
```
jskeet: Why are we releasing a 1.0.0 of a beta API at all? That feels wrong. Is it because it could be a module that contains both v1 and v1beta? (I've never been happy with doing that, but...)

internal/librarian/golang/generate.go
```diff
@@ -181,10 +188,12 @@ func buildGAPICOpts(apiPath string, library *config.Library, goAPI *config.GoAPI
 	// TODO(https://github.com/googleapis/librarian/issues/3775): assuming
 	// transport is library-wide for now, until we have figured out the config
 	// for transports.
-	opts = append(opts, "transport="+getTransport(sc))
-	if library.ReleaseLevel != "" {
-		opts = append(opts, "release-level="+library.ReleaseLevel)
+	opts = append(opts, "transport="+transport(sc))
+	level, err := releaseLevel(apiPath, library.Version)
```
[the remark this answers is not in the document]
JoeWang1127: Yes, this is the plan, unless we remove the soak time.

internal/librarian/golang/generate.go

internal/semver/semver.go
```diff
@@ -71,20 +71,20 @@ var (
 	errInvalidPrereleaseNumber = errors.New("invalid prerelease number")
 )
 
-// parse deconstructs the SemVer 1.0.0 or 2.0.0 version string into a [version]
+// Parse deconstructs the SemVer 1.0.0 or 2.0.0 version string into a [version]
```
noahdietz:
```suggestion
// Parse deconstructs the SemVer 1.0.0 or 2.0.0 version string into a [Version]
```

Issue #3775, opened 2026-01-30
librarian: config assumes a single transport for a whole library


jskeet:
Transports can vary by API path - or at least, they do in the BUILD.bazel files. This is relatively rare, but any language that bundles multiple API paths together might see this.

I'm seeing it in Python, for the google-cloud-iam library.

Running this in googleapis:

```sh
$ grep transport $(find google/iam -name BUILD.bazel) | sort | uniq
```

gives this output:

```text
google/iam/admin/v1/BUILD.bazel:    transport = "grpc",
google/iam/admin/v1/BUILD.bazel:    transport = "grpc+rest",
google/iam/BUILD.bazel:    transport = "grpc+rest",
google/iam/credentials/BUILD.bazel:    transport = "grpc+rest",
google/iam/credentials/v1/BUILD.bazel:    transport = "grpc+rest",
google/iam/v1beta/BUILD.bazel:    transport = "grpc",
google/iam/v1beta/BUILD.bazel:    transport = "grpc+rest",
google/iam/v1/BUILD.bazel:    transport = "grpc+rest",
google/iam/v1/logging/BUILD.bazel:    transport = "grpc",
google/iam/v2beta/BUILD.bazel:    transport = "grpc",
google/iam/v2beta/BUILD.bazel:    transport = "grpc+rest",
google/iam/v2/BUILD.bazel:    transport = "grpc+rest",
google/iam/v3beta/BUILD.bazel:    transport = "grpc+rest",
google/iam/v3/BUILD.bazel:    transport = "grpc+rest",
```

That's problematic in terms of:

- Having different values for the same APIs in some cases for different languages. (This suggests to me that it probably supports grpc+rest for everything...)
- Being inconsistent overall
- Logging doesn't have any services anyway, so there's nothing to use a transport...

Next steps:

- Figure out whether we can actually use grpc+rest for all of these
- See if there are other clustered APIs with the same inconsistency
- Support this in the config only if it's actually a problem

Assigning to @zhumin8 as she's heading up config work.

zhumin8:
Good call. 
Recording how Java does it today for future ref:
- for each variant, generation looks for in that folder e.g. v1/BUILD.bazel, and finds transport in the first java_gapic_library target
- on library level, it grabs **one** value for .repo-metadata.json and README from the highest stable version in the library
- there is a config override, but only overrides for  .repo-metadata.json and README, not actual code.

[...]

PR #3964, opened 2026-02-09
fix(serviceconfig): populate transport in api.go

zhumin8:
Updated script `tool/cmd/importconfigs/update_transports.go` to use exported lang const in `api.go`, then populate transport in api.go with 
`go run ./tool/cmd/importconfigs update-transports --googleapis ../googleapis`

For #3775


internal/serviceconfig/api.go
```diff
@@ -17,10 +17,26 @@
 package serviceconfig
 
 const (
-	langDart   = "dart"
-	langGo     = "go"
-	langPython = "python"
-	langRust   = "rust"
+	// LangAll is the identifier for all languages.
```
jameslynnwu: This is duplicated with the lang definitions in https://github.com/googleapis/librarian/blob/main/internal/librarian/librarian.go#L38

We should have a single place to define all languages? Also define as  an enum?

jameslynnwu (approved):

JoeWang1127 (approved):

PR #3971, opened 2026-02-10
fix(internal/librarian): clean Go library directory

JoeWang1127:
Clean Go library directory and snippet directory.

For #3617

### DIFF
--- a/internal/sidekick/parser/parser.go
+++ b/internal/sidekick/parser/parser.go
@@ -51,13 +51,9 @@ func NewModelConfigFromSidekickConfig(cfg *config.Config) ModelConfig {
 	if cfg == nil {
 		return ModelConfig{}
 	}
-	specFormat := cfg.General.SpecificationFormat
-	if specFormat == "disco" {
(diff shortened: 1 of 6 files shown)


internal/librarian/generate.go
```diff
@@ -205,10 +205,19 @@ func prepareLibrary(language string, lib *config.Library, defaults *config.Defau
 	switch language {
 	case languageFake:
 		// No cleaning needed.
-	case languageDart, languageGo, languagePython:
+	case languageDart, languagePython:
 		if err := cleanOutput(library.Output, library.Keep); err != nil {
 			return nil, err
 		}
+	case languageGo:
```
jskeet: Just as a heads-up:

- I strongly suspect we'll need more specific logic for Go, as we can't easily have a keep list for every file.
- I'm *hoping* that as part of migration, we can move the snippets into the main output directory for the library
- Currently this PR expects internal/generated/snippets to be under library.Output - it isn't, it's under the root directory

PR #3975, opened 2026-02-10
feat(internal/serviceconfig): add more overrides to API

jskeet:
Adds fields which are normally populated from the service config publishing section, but which are absent from some older APIs which don't have publishing sections at all.

These new fields are then used when generating .repo-metadata.json files. The service config is still used directory in repometadata.go, but only for fields we really expect to be present in every service config. Currently, repometadata can only be generated for libraries with APIs, and where the first listed API has a service config.


internal/repometadata/repometadata.go
```diff
@@ -98,16 +114,14 @@ func Generate(library *config.Library, language, repo, serviceConfigPath, defaul
...
-			metadata.Name = publishing.GetApiShortName()
-		}
+	if api.DocumentationURI != "" {
+		metadata.ProductDocumentation = extractBaseProductURL(api.DocumentationURI)
+	}
+	if api.NewIssueURI != "" {
+		metadata.IssueTracker = api.NewIssueURI
 	}
+	metadata.APIShortname = api.APIShortName
+	metadata.Name = api.APIShortName
```
zhumin8: why are these assigned unconditionally unlike api.DocumentationURI and api.NewIssueURI? Is it because you expect  api.APIShortName to always be present? what if no publishing section?

jskeet: I'll change to unconditionally assign the others as well - checking that `extractBaseProductURL` can handle an empty string.

PR #3976, opened 2026-02-10
chore(internal/serviceconfig): rewrite API allowlist as YAML file

jskeet:
Adds an api.yaml file with the equivalent content to APIs, and embeds it within librarian. It is unmarshaled when librarian is run.

After #3975 is merged, we'll want to add more data to the allowlist, for all the APIs without publishing entries in the service config. That will be:

- Painful in terms of adding all the values
- Painful in terms of an ever-increasing amount of data in source code

### DIFF
--- a/doc/api-allowlist-schema.md
+++ b/doc/api-allowlist-schema.md
@@ -4,16 +4,16 @@ This document describes the schema for the API Allowlist.
 
 ## API Configuration
 
-[Link to code](../internal/serviceconfig/api.go#L53)
+[Link to code](../internal/serviceconfig/api.go#L50)
 | Field | Type | Description |
 | :--- | :--- | :--- |
-| `Path` | string | Path is the proto directory path in github.com/googleapis/googleapis. If ServiceConfig is empty, the service config is assumed to live at this path. |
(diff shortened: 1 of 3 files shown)


jskeet:
Note: this is still a draft, as I'd like to get #3975 merged first, and then I'll need to modify this to add YAML configuration for the new fields. But I thought it worth getting early feedback.

@julieqiu To check we're on the same page, am I right in saying that something like this was always the plan, and it was just a matter of working out when to do it? Assuming that's the case, I personally feel that adding three more fields to dozens of APIs to effectively compensate for missing publishing sections from service configs is a tipping point here. Feel free to disagree though :)

doc/api-allowlist-schema.md
```diff
@@ -4,13 +4,13 @@ This document describes the schema for the API Allowlist.
...
-| `Discovery` | string | Discovery is the file path to a discovery document in github.com/googleapis/discovery-artifact-manager. Used by sidekick languages (Rust, Dart) as an alternative to proto files. |
-| `OpenAPI` | string | OpenAPI is the file path to an OpenAPI spec, currently in internal/testdata. This is not an official spec yet and exists only for Rust to validate OpenAPI support. |
-| `ServiceConfig` | string | ServiceConfig is the service config file path override. If empty, the service config is discovered in the directory specified by Path. |
-| `Title` | string | Title overrides the API title from the service config. |
-| `Transports` | map[string]Transport | Transports defines the supported transports per language. Map key is the language name (e.g., "python", "rust"). Optional. If omitted, all languages use GRPCRest by default. |
+| `path` | string | Path is the proto directory path in github.com/googleapis/googleapis. If ServiceConfig is empty, the service config is assumed to live at this path. |
```
jskeet: I suppose when the allow-list is edited as a file, it makes sense to document it in terms of the YAML configuration...

PR #3997, opened 2026-02-10
chore(internal/serviceconfig, internal/repometadata): refactor testdata to separate packages

JoeWang1127:
The unit tests in internal/serviceconfig and internal/repometadata all use test data from `testdata/`. This PR moves the testdata to package directory for better maintainability.

### DIFF
--- a/internal/sidekick/api/model.go
+++ b/internal/sidekick/api/model.go
@@ -390,11 +390,6 @@ func (m *Method) IsLRO() bool {
 	return m.OperationInfo != nil
 }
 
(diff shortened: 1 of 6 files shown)


JoeWang1127:
> For the reasons described here, I would propose we stick with internal/testdata as opposed to package level testdata folder.

@julieqiu I think the deduplication makes sense. However, this does create a problem if an implementation changed in one package affects another. I also observed that each package has its own testdata in other packages so I assume this is the standard way (at least for this repo).

Of course, if you still prefer the current way, I can discard this PR.

PR #3999, opened 2026-02-11
fix(internal/librarian): clean Go library directories

JoeWang1127:
Create a dedicated function to clean Go library directories.

This implementation simply removes everything in the library directory, we may iterate on this in follow ups if we understand some files are not easy to regenerate, `go.mod`, for example.

Move clean function to a separate file as we may have more language-specific clean logic.

For #3617

### DIFF
--- a/internal/librarianops/flags.go
+++ b/internal/librarianops/flags.go
@@ -1,0 +1,39 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package librarianops
+
+import (
+	"fmt"
+	"path/filepath"
+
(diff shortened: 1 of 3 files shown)


internal/librarian/clean.go
```diff
@@ -0,0 +1,86 @@
+// Copyright 2025 Google LLC
```
julieqiu: 2026

internal/librarian/clean_test.go
```diff
@@ -0,0 +1,235 @@
...
+	"github.com/googleapis/librarian/internal/config"
+)
+
+func TestCleanOutput(t *testing.T) {
+	for _, test := range []struct {
+		name    string
+		files   []string
+		keep    []string
+		want    []string
+		wantErr bool
```
julieqiu: separate error cases: https://github.com/googleapis/librarian/blob/main/doc/howwewritego.md#separate-error-tests

internal/librarian/generate.go
```diff
@@ -266,52 +265,3 @@ func formatLibrary(ctx context.Context, language string, library *config.Library
 	}
 	return fmt.Errorf("language %q does not support formatting", language)
 }
-
```
julieqiu: nit: in the future I would move this in a separate PR from the one where you add new logic

julieqiu (dismissed):

internal/librarian/clean.go
JoeWang1127: Done.

internal/librarian/clean_test.go
JoeWang1127: Done.

zhumin8 (approved):

Issue #4008, opened 2026-02-11
librarian: remove version check and -f flag


julieqiu:
The current version check is creating more friction than value. We should remove both the version check and the -f flag to simplify the workflow.

PR #4012, opened 2026-02-11
fix(internal/librarian): remove version check and -f flag

julieqiu:
The version check was creating more friction than value. Remove the version check and the -f/--force flag used to skip it.

Fixes https://github.com/googleapis/librarian/issues/4008


coryan (approved):
LGTM, the bots and the tests are telling to sort some imports.

Issue #3389, opened 2025-12-30
python: add bump command


julieqiu:
Add support for librarian release for Python.

hj690:
Python will stay with release-please for now

PR #4030, opened 2026-02-12
feat(internal/librarian/python): bump snippet metadata versions

jskeet:
Towards #3389


jskeet:
@JoeWang1127 I'd expect you to be able to use basically the same code in the golang bump.

PR #4040, opened 2026-02-12
refactor(sidekick): introduce ModelOverride for configuration consistency

jameslynnwu:
- Replaces legacy map-based overrides (name, title, description, elements) with a structured ModelOverride struct.
- "include-list" remains in ModelConfig.Source because its related to which sources are included. We may reconsider this in the future.

### DIFF
--- a/internal/librarian/dart/codec_test.go
+++ b/internal/librarian/dart/codec_test.go
@@ -21,7 +21,6 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/googleapis/librarian/internal/config"
 	"github.com/googleapis/librarian/internal/sidekick/api"
-	sidekickapi "github.com/googleapis/librarian/internal/sidekick/api"
 	"github.com/googleapis/librarian/internal/sidekick/parser"
 	"github.com/googleapis/librarian/internal/sidekick/source"
 )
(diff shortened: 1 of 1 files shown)


internal/librarian/dart/codec_test.go
```diff
@@ -20,6 +20,7 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/googleapis/librarian/internal/config"
+	sidekickapi "github.com/googleapis/librarian/internal/sidekick/api"
```
julieqiu: why not just use api?

internal/librarian/dart/codec_test.go
```diff
@@ -301,6 +302,9 @@ func TestToModelConfig(t *testing.T) {
 					"roots":           "googleapis",
 				},
 				Codec: map[string]string{},
+				Override: sidekickapi.ModelOverride{
+					Title: "",
```
julieqiu: omit the value if it is an empty string

internal/librarian/dart/codec_test.go
```diff
@@ -317,11 +321,14 @@ func TestToModelConfig(t *testing.T) {
...
-					"googleapis-root":      googleapisDir,
-					"description-override": "this is a description override",
-					"roots":                "googleapis",
+					"googleapis-root": googleapisDir,
+					"roots":           "googleapis",
 				},
 				Codec: map[string]string{},
+				Override: sidekickapi.ModelOverride{
+					Description: "this is a description override",
+					Title:       "",
```
julieqiu: ditto: omit the value if it is an empty string

Issue #4049, opened 2026-02-13
librarian: remove --all flag from update


julieqiu:
The librarian update --all flag is misleading, since it only updates googleapis, or googleapis + discovery in the case of Rust and doesn't actually mean "all".

We should remove --all entirely and require users to explicitly specify what source they want to update.

PR #4050, opened 2026-02-13
feat(internal/librarian): remove --all from update command

julieqiu:
The update command now accepts multiple sources as positional arguments instead of using the --all flag.

```
Usage: librarian update <sources...>

Examples:
    librarian update googleapis
    librarian update discovery googleapis
```

The librarianops generate command now reads librarian.yaml to determine which sources are configured and only calls update for those sources.

Fixes https://github.com/googleapis/librarian/issues/4049

### DIFF
--- a/internal/librarianops/generate_test.go
+++ b/internal/librarianops/generate_test.go
@@ -191,13 +191,6 @@ func TestSourcesToUpdate(t *testing.T) {
 		},
 		{
 			name: "no sources configured",
-			cfg: &config.Config{
-				Sources: &config.Sources{},
-			},
-			want: nil,
-		},
-		{
-			name: "nil sources",
 			cfg:  &config.Config{},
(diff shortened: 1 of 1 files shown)


cmd/librarianops/doc.go
```diff
@@ -46,7 +46,7 @@ DESCRIPTION:
 	  1. Clone the repository to a temporary directory (or use existing directory with -C)
 	  2. Create a branch: librarianops-generateall-YYYY-MM-DD
 	  3. Run librarian tidy
-	  4. Run librarian update --all
+	  4. Run librarian update for configured sources (discovery, googleapis)
```
jameslynnwu: IIUC, discovery, googleapis is an example, since its read from the librarian.yaml

If so (e.g. discovery, googleapis)

PR #4065, opened 2026-02-13
fix(internal/librarian): rust to add extra dependencies when librarian add

### LINKED ISSUE #3905: Automate dependency management in librarian
## Objective

We want to `librarian add ${thing}` and automatically populate the dependency information in the package manifest (the `Cargo.toml` or its moral equivalent for other languages.

## Problem

The mapping between "protobuf package dependency" to "${LANG} package that provides that protobuf package" is manually curated. When a new dependency enters the mono-repo we need to manually add this to the `librarian.yaml` file.

## Background

This problem is only applicable for Protobuf-based client libraries. OpenAPI and Discovery doc specifications are self-contained and do not have dependencies beyond `gax`, `auth` and similar "core" libraries.

`sidekick` has full dependency information for the Protobuf **packages** used by a library. If `google.cloud.secretmanager.v1` depends on `google.iam.v1` then Protobuf knows that dependency exists. In fact, it already computes all the direct dependencies of a library based on used messages, enums, including LRO metadata and responses.

`sidekick` assumes "one package generates one library", so anything that is not in the "current package" is assumed to be an external library. That is not true in all generators, but we can generalize to "these packages are being generated, everything else is external".

The `librarian.yaml` configuration describes a mapping between Protobuf packages to Rust packages. Presumably this could be expanded to a mapping between Protobuf packages to ${LANG} packages, which could be many-to-one.

## Outline

Maybe we could have a `librarian resolve` command to automatically update the mapping between Protobuf packages to ${LANG} packages, and update the `librarian.yaml` file accordingly?  That could simplify the steps to `librarian add` a new file.

zhumin8:
Introduces automated Protobuf dependency resolution for Rust when running `librarian add`. Previously, external dependencies had to be manually added in librarian.yaml after librarian add and before librarian generate.
  This new logic analyzes the Protobuf model to identify external imports and automatically maps them to the corresponding Rust crates within the workspace.

Fix #3905


internal/librarian/add.go
```diff
@@ -84,13 +102,13 @@ func deriveLibraryName(language, api string) string {
 	}
 }
 
-func addLibrary(cfg *config.Config, apis ...string) (*config.Config, error) {
+func addLibrary(cfg *config.Config, apis ...string) (string, *config.Config, error) {
```
JoeWang1127: Could you add a comment on what are the return values?

internal/librarian/rust/resolve.go
```diff
@@ -0,0 +1,162 @@
...
+	modelConfig, err := libraryToModelConfig(lib, lib.APIs[0], sources)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create model config: %w", err)
+	}
+	model, err := parser.CreateModel(modelConfig)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create model: %w", err)
+	}
+	// Identify the packages owned by the current library.
+	ownedPackages := map[string]bool{}
```
JoeWang1127:
```suggestion
	ownedPackages := make(map[string]bool)
```

JoeWang1127 (approved):
Need to resolve the merging conflicts.

internal/librarian/rust/resolve.go
```diff
@@ -0,0 +1,159 @@
...
+		return cfg, nil
+	}
+	externalPackages, err := findExternalPackages(lib, sources)
+	if err != nil {
+		return nil, err
+	}
+	return resolveExternalPackages(cfg, lib, externalPackages), nil
+}
+
+func findExternalPackages(lib *config.Library, sources *source.Sources) (map[string]bool, error) {
```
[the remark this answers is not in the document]
zhumin8: done

JoeWang1127 (approved):

PR #4084, opened 2026-02-17
feat(internal/librarian/golang): skip cleaning nested module in Go library

### LINKED ISSUE #4001: Move `cleanGo` to `internal/librarian/golang` when the logic is deviate from standard clean function
add a TODO that this should eventually be moved inside internal/librarian/golang

_Originally posted by @julieqiu in https://github.com/googleapis/librarian/pull/3999#discussion_r2791071679_

Note that we can reuse `cleanOutput` (even if we make it public) in internal/librarian/golang due to circular dependency.

JoeWang1127:
Skip cleaning nested module in a Go library.

Some libraries, e.g., bigquery, pubsub, have nested major versions. We should skip cleaning these directories because they are not part of the library.

A more intuitive approach is moving the nested module to the outer directory, but it is rejected by the Go team.

We need further refactor to move the generic check and clean function to a different package to reduce code duplication.

For #3617
Fixes #4001

### DIFF
--- a/internal/librarian/rust/codec.go
+++ b/internal/librarian/rust/codec.go
@@ -80,9 +80,9 @@ func libraryToModelConfig(library *config.Library, ch *config.API, sources *sour
 			}
 		}
 		if len(library.Rust.PaginationOverrides) > 0 {
-			modelCfg.PaginationOverrides = make([]sidekickconfig.PaginationOverride, len(library.Rust.PaginationOverrides))
+			modelCfg.PaginationOverrides = make([]api.PaginationOverride, len(library.Rust.PaginationOverrides))
 			for i, override := range library.Rust.PaginationOverrides {
-				modelCfg.PaginationOverrides[i] = sidekickconfig.PaginationOverride{
+				modelCfg.PaginationOverrides[i] = api.PaginationOverride{
 					ID:        override.ID,
 					ItemField: override.ItemField,
 				}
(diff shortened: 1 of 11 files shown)


internal/librarian/golang/clean.go
```diff
@@ -0,0 +1,82 @@
...
+	return filepath.WalkDir(dir, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if d.IsDir() {
+			if {
+				return fs.SkipDir
+			}
+			return nil
+		}
```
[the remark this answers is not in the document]

internal/librarian/golang/clean.go
```diff
@@ -0,0 +1,105 @@
...
+	if err := clean(snippetDir, nil); err != nil {
+		return nil, err
+	}
+	return library, nil
+}
+
+// check validates the given directory and returns a set of files to keep.
+// It ensures that the provided directory exists and is a directory.
+// It also verifies that all files specified in 'keep' exist within 'dir'.
+func check(dir string, keep []string) (map[string]bool, error) {
```
jskeet: Rather than copying this code here, would it be worth having a `clean` package? That way both the librarian and language-specific packages can call code within it. (I have a change coming up for Python which *might* be able to reuse existing code, although I'm not sure.)

For `clean` to be parameterized like that, we might need a `func(string) bool` to be passed in as a flexible bit of extra configuration - I'm not sure.

internal/librarian/golang/clean.go
```diff
@@ -0,0 +1,105 @@
...
+
+	"github.com/googleapis/librarian/internal/config"
+)
+
+var (
+	nestedVersion = regexp.MustCompile(`^v(\d+)$`)
+)
+
+// CleanGo cleans up a Go library and its associated snippets.
+func CleanGo(library *config.Library) (*config.Library, error) {
```
jskeet: As a side-note, I'd rename this to CleanLibrary - it's already within the go package, so the go part is implicit, I'd say.

internal/librarian/golang/clean.go
JoeWang1127: > We need further refactor to move the generic check and clean function to a different package to reduce code duplication.

I had a similar thoughts so I put this in the description. I plan to do the refactor in a follow up PR.

PR #4091, opened 2026-02-18
refactor: extract load sources logic and export

zhumin8:
Extracts load sources logic from generateLibraries into exported `LoadSources` so it can be reused. 

context: https://github.com/googleapis/librarian/pull/4065#discussion_r2819172279

For https://github.com/googleapis/librarian/issues/3905

### DIFF
--- a/internal/librarian/generate.go
+++ b/internal/librarian/generate.go
@@ -80,7 +80,7 @@ func runGenerate(ctx context.Context, cfg *config.Config, all bool, libraryName 
 }
 
 func generateLibraries(ctx context.Context, all bool, cfg *config.Config, libraryName string) error {
-	googleapisDir, rustSources, err := LoadSources(ctx, cfg)
+	googleapisDir, rustDartSources, err := LoadSources(ctx, cfg)
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


internal/librarian/generate.go
```diff
@@ -150,6 +130,34 @@ func generateLibraries(ctx context.Context, all bool, cfg *config.Config, librar
...
+		googleapisDir = cfg.Sources.Googleapis.Dir
+	} else {
+		dir, err := fetch.RepoDir(ctx, googleapisRepo, cfg.Sources.Googleapis.Commit, cfg.Sources.Googleapis.SHA256)
+		if err != nil {
+			return "", nil, fmt.Errorf("failed to fetch %s: %w", googleapisRepo, err)
+		}
+		googleapisDir = dir
+	}
+
+	var rustSources *source.Sources
```
jameslynnwu: rustSources doesn't align with FetchRustDartSources

I think it might be best to have a unified struct that represents fetched sources. For non-rust/dart, only googleapis will be populated.

JoeWang1127 (approved):

PR #4096, opened 2026-02-18
doc: remove link to code from config docs

zhumin8:
Removes logic to generate "[Link to code]" links in config docs. It causes more friction than being helpful right now, as both config are frequently edited and causing merging conflicts. We can bring it back later if it becomes helpful.


jameslynnwu (approved):

JoeWang1127 (approved):
As a reminder, could you file an issue to add it back, we don't need to specify the date.

PR #4114, opened 2026-02-19
refactor(internal/repometadata): create language-specific repo metadata

JoeWang1127:
Though client in different languages has a repo metedata, they have different key-values.

Refactor metadata generation functions to create language specific metadata file.

For example, metadata in Python client:
```json
{
  "api_description": "enables controlling access to your organization's data by Google personnel.",
  "api_id": "accessapproval.googleapis.com",
  "api_shortname": "accessapproval",
  "client_documentation": "https://cloud.google.com/python/docs/reference/accessapproval/latest",
  "default_version": "v1",
  "distribution_name": "google-cloud-access-approval",
  "issue_tracker": "",
  "language": "python",
  "library_type": "GAPIC_AUTO",
  "name": "accessapproval",
  "name_pretty": "Access Approval",
  "product_documentation": "https://cloud.google.com/access-approval",
  "release_level": "stable",
  "repo": "googleapis/google-cloud-python"
}
```
where as in Go client:
```json
{
  "api_shortname": "cloudasset",
  "client_documentation": "https://cloud.google.com/go/docs/reference/cloud.google.com/go/asset/latest/apiv1",
  "client_library_type": "generated",
  "description": "Cloud Asset API",
  "distribution_name": "cloud.google.com/go/asset/apiv1",
  "language": "go",
  "library_type": "GAPIC_AUTO",
  "release_level": "stable"
}

```

For #3617

### DIFF
--- a/internal/librarian/golang/generate.go
+++ b/internal/librarian/golang/generate.go
@@ -363,6 +363,7 @@ func generateREADME(library *config.Library, api *serviceconfig.API, moduleRoot 
 }
 
 // updateSnippetMetadata updates the snippet metadata files with the correct library version.
+// Skip nested module if exists.
 func updateSnippetMetadata(library *config.Library, output string) error {
 	baseDir := filepath.Join(output, "internal", "generated", "snippets", library.Name)
 	return filepath.WalkDir(baseDir, func(path string, d fs.DirEntry, err error) error {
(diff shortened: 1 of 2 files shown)


JoeWang1127:
Pause the change for now, I'll discuss the approach with @jskeet next week.

JoeWang1127:
@jskeet I agree that repo metadata should not have variant among different languages or, in the best scenario, be removed entirely, but we have a long way to go.

At this stage, the metadata _is_ different in Go, Python and Rust. My solution here is to use library and api info to build common part and language-specific info in `repometadata` package. To make the structure clear, I put functions to build language-specific info to `language.go` but we can extend it to `golang.go` or `python.go` later.

>  think a simpler solution would be to put the language code in more control, and keep as much language-specific code in the language repo as possible

In this approach, what functions will be in `repometadata` package? Do we expect this package only define metadata struct and a write function, everything else lives in `internal/librarian/{language}` packages?

PR #4115, opened 2026-02-19
fix(internal/librarian/golang): skip updating snippet metadata in nested module

JoeWang1127:
Skip updating snippet metadata in nested module because it belongs to another Go library.

For #3617


jskeet (approved):
I think this is okay for now, but an alternative would be to call into the snippetmetadata package, but only for directories you know about based on the API version (rather than the library version).

If this is simpler for now, let's stick with it - partly because if we can move the snippets alongside the library code, all of this can definitely go away and nested modules won't be a problem (because `pubsub/v2/internal/generated/snippets` isn't nested in `pubsub/internal/generated/snippets` for example).

PR #4117, opened 2026-02-20
chore(serviceconfig): convert data in api.go into a YAML file

### LINKED ISSUE #4081: librarian: convert the data in api.go to a YAML file
Currently we have a lot of data in source code, in https://github.com/googleapis/librarian/blob/main/internal/serviceconfig/api.go

That makes some changes easy - but it makes it harder to perform bulk edits. (The one I particularly want to do is adding a load of "overrides" of things like issue URI and product docs where there's no publishing section in the service config.) It's really easy to write one-off code that imports the new data, adds it to the in-memory implementation, and writes it out as a YAML file - it's harder to make that write out source code.

This commit is massively out of date by now, but shows roughly what I think the change should be: https://github.com/googleapis/librarian/compare/main...jskeet:librarian:api-yaml

I'm happy to do the work when we're all agreed that it's time, but I don't want to jump the gun.

@zhumin8 please feel free to assign this to yourself, or to me, or leave it unassigned - and add any suitable labels etc.

jskeet:
Changes the data in the APIs field to be loaded from an embedded sdk.yaml file instead of being initialized as source code. This makes it easier to update the data programmatically - at the cost of any significant schema refactoring being slightly harder. We're assuming now that schema changes will be at least reasonably incremental. (Adding a new field etc is easy enough - it's only structural changes to existing data that are made harder by this, and not *that* much harder.)

The apis.yaml file was created by changing cmd/librarian/main.go to:

```go
yaml.Write("internal/serviceconfig/sdk.yaml", serviceconfig.APIs)
```

and then running librarian. A similar line of code could be used to regenerate the file after programmatic one-off changes, e.g. importing override data from other sources, where there's no service config. The file need to be reformatted (`go tool yamlfmt .`) after this.

Fixes #4081

### DIFF
--- a/doc/config-schema.md
+++ b/doc/config-schema.md
@@ -124,6 +124,7 @@ This document describes the schema for the librarian.yaml.
 | `disable_gapic` | bool | DisableGAPIC determines whether to generate the GAPIC client. |
 | `import_path` | string | ImportPath is the Go import path for the API. |
 | `nested_protos` | list of string | NestedProtos is a list of nested proto files. |
+| `no_metadata` | bool | NoMetadata indicates whether to skip generating gapic_metadata.json. This is typically false. |
 | `no_rest_numeric_enums` | bool | NoRESTNumericEnums determines whether to use numeric enums in REST requests. The "No" prefix is used because the default behavior (when this field is `false` or omitted) is to generate numeric enums |
 | `path` | string | Path is the source path. |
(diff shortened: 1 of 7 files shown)


zhumin8 (commented):
note to self: will need to followup and update [tool/cmd/importconfigs](https://github.com/googleapis/librarian/tree/main/tool/cmd/importconfigs) to update yaml file.

zhumin8 (approved):

PR #4149, opened 2026-02-23
chore(repometadata): refactor repometadata package for flexibility

jskeet:
Changes .repo-metadata.json file generation from "one-shot, everything must be right in a single function call" (which ends up with language-specific code in the repometadata package) there are now separate functions for populating from a library, writing metadata, and reading metadata.

The Python code has been updated with the changes to populate language-specific aspects. This is the only language which previously used the repometadata package. The language-specific code for other languages (e.g. building the documentation URL) should be reinstated within the relevant language package at the point when the repo metadata is created for that language.

Note that this change requires the overall Librarian config to be passed to GenerateLibraries. This has currently only been done for Python, but can easily be done for other languages as and when they need it.

### DIFF
--- a/internal/repometadata/repometadata_test.go
+++ b/internal/repometadata/repometadata_test.go
@@ -25,7 +25,7 @@ import (
 	"github.com/googleapis/librarian/internal/serviceconfig"
 )
 
-func TestGenerate(t *testing.T) {
+func TestFromLibrary(t *testing.T) {
 	for _, test := range []struct {
 		name    string
 		library *config.Library
@@ -99,7 +99,7 @@ func TestGenerate(t *testing.T) {
 	}
 }
 
-func TestGenerate_Error(t *testing.T) {
+func TestFromLibrary_Error(t *testing.T) {
 	for _, test := range []struct {
 		name    string
 		library *config.Library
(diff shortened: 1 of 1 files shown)


internal/repometadata/repometadata.go
```diff
@@ -78,76 +82,59 @@ type RepoMetadata struct {
...
 	}
-	return FromAPI(api, library, language, repo, defaultVersion, outdir)
+	return fromAPI(config, api, library), nil
 }
 
-// FromAPI generates the .repo-metadata.json file from a serviceconfig.API and library information.
-func FromAPI(api *serviceconfig.API, library *config.Library, language, repo, defaultVersion, outputDir string) error {
-	clientDocURL := buildClientDocURL(api, library, language)
+// fromAPI generates the .repo-metadata.json file from a serviceconfig.API and library information.
+func fromAPI(config *config.Config, api *serviceconfig.API, library *config.Library) *RepoMetadata {
```
JoeWang1127: Why not passing repo and language as string or adding language and repo to metadata in language package.

I feel it's too much info in config and most of them are irrelevant.

jskeet: > I did consider having two functions - one to populate the repo-level metadata (repo, language) and one to populate the library-level metadata (api short name etc). Would you prefer that?

I've now noticed that the library-level function would need the language anyway, so it's probably simpler to stick to doing it all in one go. (And the fact that the language is needed when looking up the API is another reason to populate the Language field within the repometadata code.)

We *could* change this back to passing the repo and the language as separate parameters into the repometadata package, but I'd still want to pass the config into the python code's generate, rather than the previous hard-coding of the language and repo in there. (Hard-coding the language isn't *so* bad, at least if we use the constant... it's not like the python package should ever be used for non-Python code - but hard-coding the repo when it's already in the config feels really bad.)

PR #4168, opened 2026-02-24
fix(internal/librarian/golang): fix client version.go directory for nested major version client

JoeWang1127:
Fix client version.go directory for nested major version client.
For `bigquery/v2`, the client version.go should be in `bigquery/v2/apiv2`. It was `bigquery/v2/bigquery/apiv2` before the fix.

For #3617


jskeet (approved):
I hope once the migration is done, we can simplify all of this a bit - but this is fine for now.

PR #4184, opened 2026-02-25
feat(sidekick/rust): expose resource name components in annotate.go

### LINKED ISSUE #4182: sidekick/rust: expose resource name components to codec
Inject resource name template and arguments from `TargetResource` into `annotation`.

haphungw:
Add `ResourceNameTemplate` and `ResourceNameArgs` to `methodAnnotation`. 

This is the first step to generate `resource_name()` helper method for Rust clients.

Fixes #4182

### DIFF
--- a/internal/sidekick/rust/annotate.go
+++ b/internal/sidekick/rust/annotate.go
@@ -19,7 +19,6 @@ import (
 	"cmp"
 	"fmt"
 	"log/slog"
-	"regexp"
 	"slices"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


internal/sidekick/rust/annotate.go
```diff
@@ -1566,6 +1574,38 @@ func (c *codec) annotateEnumValue(ev *api.EnumValue, model *api.API, full bool)
...
+				annotation.HasResourceNameGeneration = true
+				break
+			}
+		}
+	}
+	return nil
+}
+
+// resourceNameVarPattern matches variable segments in a resource name template (e.g., "{project}", "{name=*}").
+var resourceNameVarPattern = regexp.MustCompile(`\{[^}]+\}`)
```
coryan: This seems overly complicated.  We have a parsed data structure, then use that to generate a string, then parse the string with a regular expression to generate a different string?  Why not generate the string from the original data structure?

PR #4197, opened 2026-02-25
fix(internal/librarian/golang): update clean regex for snippets

JoeWang1127:
Update regex to match snippet directory in Go SDK.

We always run the librarian command at the repository directory, which makes the snippet directory doesn't have a prefix. In this case, the current regex doesn't match snippet directory.

For #3617

### DIFF
--- a/internal/librarian/golang/clean.go
+++ b/internal/librarian/golang/clean.go
@@ -30,6 +30,8 @@ var (
 	// These patterns are essential to ensure the 'Clean' operation only removes
 	// files that are known to be generated, protecting handwritten code or
 	// configuration that may reside in the same directory.
+	// TODO(https://github.com/googleapis/librarian/issues/4217): document each regex about
+	// what are matched and why it is necessary.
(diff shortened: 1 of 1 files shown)


internal/librarian/golang/clean.go
julieqiu: what's the reason for all of this regex?

internal/librarian/golang/clean_test.go
```diff
@@ -84,17 +88,25 @@ func TestClean_Success(t *testing.T) {
...
+			name:         "removes all snippets in current directory",
+			snippetFiles: []string{"snippet1.go", "snippet2.go", "README.md"},
+			keep:         []string{},
+			libraryRoot:  ".",
+			wantOutput:   []string{},
+		},
 	} {
 		t.Run(test.name, func(t *testing.T) {
 			t.Parallel()
-			root := t.TempDir()
```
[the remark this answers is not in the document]
JoeWang1127: Done.

internal/librarian/golang/clean.go

Issue #4018, opened 2026-02-11
sidekick: implement resource identification heuristic


dbolduc:
Googlers can see: go/client-libraries:destination-resource-name

And @westarle prototyped the change in: https://paste.googleplex.com/5554688184418304#l=112

---

It looks like Rust picks up the resource here:

https://github.com/googleapis/librarian/blob/c0735c0f8f4ab39fcf2d4e05cc93183dfa73d87b/internal/sidekick/rust/templates/crate/src/transport.rs.mustache#L167-L173

PR #4199, opened 2026-02-25
fix(sidekick): correct heuristic template indexing

haphungw:
When generating the `Template` for resource names, the parser was incorrectly capturing preceding elements in the HTTP URL (e.g., `/v1`). 

Adding `firstIndex` tracking allows the parser to identify when the valid `collection/id` chain begin.

The next PRs will add the necessary blocks in `transport.rs.mustache`. The generated Rust snippets should look like something below:
```
#[cfg(google_cloud_unstable_resource_name)]
let options = if let Some(resource_name) = (|| {
    Some(format!(
        "//compute.googleapis.com/projects/{}/zones/{}/acceleratorTypes/{}",
        Some(&req)
            .map(|m| &m.project)
            .map(|s| s.as_str())
            .filter(|s| !s.is_empty())?,
        Some(&req)
            .map(|m| &m.zone)
            .map(|s| s.as_str())
            .filter(|s| !s.is_empty())?,
        Some(&req)
            .map(|m| &m.accelerator_type)
            .map(|s| s.as_str())
            .filter(|s| !s.is_empty())?,
    ))
})() {
    use google_cloud_gax::options::internal::{RequestOptionsExt, ResourceName};
    options.insert_extension(ResourceName(resource_name))
} else {
    options
};
```
Staged at: https://github.com/googleapis/google-cloud-rust/pull/4826

For #4018

### DIFF
--- a/internal/sidekick/api/resource_identification.go
+++ b/internal/sidekick/api/resource_identification.go
@@ -83,6 +83,7 @@ func identifyHeuristicTarget(method *Method, binding *PathBinding, vocabulary ma
 	}
 
 	var fieldPaths [][]string
+	var firstIndex = -1
 	var lastIndex int
 	segments := binding.PathTemplate.Segments
 
@@ -108,6 +109,9 @@ func identifyHeuristicTarget(method *Method, binding *PathBinding, vocabulary ma
 		if err != nil {
 			return nil, err
 		}
+		if firstIndex == -1 {
+			firstIndex = i - 1
+		}
 		fieldPaths = append(fieldPaths, fieldPath)
 		lastIndex = i + 1
 	}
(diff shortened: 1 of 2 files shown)


coryan (commented):
Compute sets these things:

```
  "basePath": "/compute/v1/",
  "baseUrl": "https://compute.googleapis.com/compute/v1/",
```

Meanwhile DNS sets these things:

```
  "basePath": "",
  "baseUrl": "https://dns.googleapis.com/",
```

Will this change work for DNS?

Should we use the `basePath` or `baseUrl` to decide how much we strip?

haphungw:
> Will this change work for DNS?

This should work for both since the logic dynamically looks for the first valid (collection, variable) pair (e.g., `projects/{project}`).

> Should we use the `basePath` or `baseUrl` to decide how much we strip?

I think `basePath` can be more inconsistent, as you pointed out with `/compute/v1` and `""`. 
I haven't seen many of our APIs, but if a random literal is injected after a `basePath` (e.g., `/v1/admin/projects/{project}`) then we might have more cases to take care of.

coryan (approved):
Thanks for the extra context.

PR #4200, opened 2026-02-26
refactor(sidekick): Simplify sample info representation

amanda-tarafa:
This will help with:
- showing resource name formatting in samples. We can pass the resource name segments as parameters and initialize the resoure name field by combining them in the right format.
- generating samples with initialized requests. The templates don't need to know about any specific fields, only about different types of fields.

### DIFF
--- a/internal/sidekick/rust/templates/common/client_method_samples/builder_fields.mustache
+++ b/internal/sidekick/rust/templates/common/client_method_samples/builder_fields.mustache
@@ -13,32 +13,24 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 }}
-{{#AIPStandardGetInfo}}
-///         .set_{{ResourceNameRequestField.Codec.SetterName}}(resource_name)
-{{/AIPStandardGetInfo}}
-{{#AIPStandardDeleteInfo}}
-///         .set_{{ResourceNameRequestField.Codec.SetterName}}(resource_name)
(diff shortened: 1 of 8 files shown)


internal/sidekick/rust/templates/common/client_method_samples/builder_fields.mustache
```diff
@@ -13,32 +13,24 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
-///         .set_{{ResourceNameRequestField.Codec.SetterName}}(resource_name)
-{{/AIPStandardDeleteInfo}}
-{{#AIPStandardUndeleteInfo}}
-///         .set_{{ResourceNameRequestField.Codec.SetterName}}(resource_name)
-{{/AIPStandardUndeleteInfo}}
-{{#AIPStandardCreateInfo}}
-///         .set_{{ParentRequestField.Codec.SetterName}}(parent){{#ResourceIDRequestField}}.set_{{Codec.SetterName}}("{{Codec.SetterName}}_value"){{/ResourceIDRequestField}}
-///         .set_{{ResourceRequestField.Codec.SetterName}}(
-///             {{ResourceRequestField.MessageType.Codec.Name}}::new()/* set fields */
+{{#IsAIPStandard}}
```
coryan: I see how this is nice for Rust, still a bit wary of how this will work for $NEXT_LANG

amanda-tarafa: I think I've replied above, so resolving this to keep the conversation in a single place, but let me know if I'm missing something specifically here.

PR #4249, opened 2026-02-27
fix(java): add header to generated proto- and grpc- modules

### LINKED ISSUE #4129: java: fix header for grpc and proto modules
raw generation output does not include headers for grpc and proto module files (except *Name.java resource files that are generated by gapic generator but relocated into proto module).

zhumin8:
Adds header to generated proto- and grpc- modules when not present.
Similar logic to synthtool [here](https://github.com/googleapis/sdk-platform-java/blob/4206e6ed7546d6b8330dcb8dc29179e59677ccd3/hermetic_build/library_generation/owlbot/synthtool/languages/java.py#L97), except not dealing with resource name classes malformed license headers.

Fix #4129

### DIFF
--- a/internal/librarian/java/generate.go
+++ b/internal/librarian/java/generate.go
@@ -21,7 +21,6 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"regexp"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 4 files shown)


internal/librarian/java/generate.go
```diff
@@ -145,6 +154,38 @@ func postProcess(ctx context.Context, outdir, libraryName, version, googleapisDi
 	return nil
 }
 
+var headerRegex = regexp.MustCompile(`\* Copyright \d{4} Google LLC`)
```
julieqiu: use https://github.com/googleapis/librarian/blob/c36c53a20c7dc9ec3b8706684da4e9745cee135a/internal/license/license.go#L21

zhumin8: license.Header is used for generating header, this `headerRegex` was used to identify if a header already exists (mimic existing java logic, I believe reason is some `*ResourceName.java` classes inside proto- module are generated by gapic-generator thus already has header, we do not want to append extra to them) . 

Moved this logic `HasHeader` to `license` package and checking for "Licensed under the Apache License, Version 2.0" line so no need for regex.

julieqiu (approved):

PR #4251, opened 2026-02-27
refactor(internal/librarian/golang): simplify Go client path resolution

### LINKED ISSUE #4252: librarian: simplify Go client import path resolution to avoid complex string manipulation
Currently, when the Go librarian generates client libraries, it relies on path manipulation to determine the correct client package, import path, and output directory. Functions like `resolveClientPath` and `extractMiddleDir` split strings and attempt to locate the library name and client directory within the API path dynamically.

This approach is prone to errors, hard to maintain, and does not gracefully handle edge cases where the path structure does not align perfectly with standard naming conventions (e.g., custom client directories, nested major versions). It also unnecessarily duplicates logic since the GoAPI configuration already specifies the intended ImportPath and ClientDirectory.

JoeWang1127:
This change simplifies how client paths, import paths, and distribution names are resolved when generating Go client libraries.

Previously, these values were computed dynamically using `resolveClientPath`, `clientDirectory`, and `extractMiddleDir` based on the API path. 
Now, the generator uses the `ImportPath` and `ClientDirectory` fields directly from the `GoAPI` configuration, which are more reliable.

**Key changes:**
* Updated `buildGAPICImportPath` to use `goAPI.ImportPath` and `goAPI.ClientDirectory` directly.
* Updated `generateRepoMetadata`, `clientDocURL`, and `distributionName` to rely on the configured import path rather than calculating it.
* Removed the unused `resolveClientPath`, `clientDirectory`, and `extractMiddleDir` functions from `version.go`.
* Updated corresponding unit tests to reflect the new expected behavior.

Fixes #4252

### DIFF
--- a/internal/librarian/golang/generate.go
+++ b/internal/librarian/golang/generate.go
@@ -41,10 +41,8 @@ const (
 
 var (
 	//go:embed template/_README.md.txt
-	readmeTmpl string
-
+	readmeTmpl       string
 	readmeTmplParsed = template.Must(template.New("readme").Parse(readmeTmpl))
-
 	errGoAPINotFound = errors.New("go API not found")
 )
 


internal/librarian/golang/generate.go
```diff
@@ -44,6 +44,8 @@ var (
 	readmeTmpl string
 
 	readmeTmplParsed = template.Must(template.New("readme").Parse(readmeTmpl))
+
```
julieqiu: nit: you don't need all of these line breaks

Issue #4276, opened 2026-03-01
config: remove veneer field from librarian.yaml


julieqiu:
The veneer field on a library in librarian.yaml is a top-level boolean that indicates a library has handwritten code wrapping generated code. In practice, only Rust uses this field at runtime, and every Rust veneer library already has rust.modules configured, so this field is redundant.

We should replace the veneer field with a derived method IsVeneer() that returns true when language-specific module config is present (i.e., lib.Rust != nil && len(lib.Rust.Modules) > 0)

PR #4277, opened 2026-03-01
refactor(rust): remove use of veneer field

julieqiu:
The veneer field on Library in librarian.yaml is a top-level boolean that is redundant because every Rust veneer library already has rust.modules configured. This replaces the field with an IsVeneer function for Rust, as a step towards getting rid of that field.

For https://github.com/googleapis/librarian/issues/4276
Fixes https://github.com/googleapis/librarian/issues/4280

### DIFF
--- a/.github/workflows/rust.yaml
+++ b/.github/workflows/rust.yaml
@@ -46,6 +46,7 @@ jobs:
         run: go test -race ./internal/librarian/rust
   integration:
     runs-on: ubuntu-24.04
+    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
     steps:
       - uses: actions/checkout@v5
       - uses: actions/setup-go@v6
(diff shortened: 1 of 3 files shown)


jskeet:
> I’m not sure what the right solution is here, but from what I can tell, Python does not use the `veneer` field in librarian.yaml, so I don’t think this blocks the current PR. Is there something I might be overlooking?

The Python generation code doesn't use the `veneer` field, but the language-agnostic code does, to determine whether or not to infer an API when the API list is empty. That's why the Python *migration* code explicitly sets the `veneer` field to true when it migrates a library with no APIs - to *prevent* the language-agnostic code from auto-populating the API list when it's empty. If we could remove that auto-population, Python wouldn't need to know or care about veneers at all - it would generate code when there's an API to generate, and not generate anything when there aren't any APIs.

julieqiu:
I think as a next step we need to add some more test cases for each of Python, Go, and Rust that capture the current behavior and edge cases before making further changes, since the assumptions being made by each is not immediately clear. 

This way we'll have a clear picture of what the current behavior is and can more easily see where changes might break things. I'll file issues.

JoeWang1127 (approved):

PR #4287, opened 2026-03-02
refactor(config): migrate language string literals to a new `config.Language` type with defined constants

### LINKED ISSUE #4010: librarian: use enum for language
Also dedupe duplicated language constants in codebase.

Similar to #4009, we should define a language type for languages so we are not passing strings around the codebase.

This guarantees that once a language has been parsed from a config, it is a valid language.

jameslynnwu:
Migrate language string literals to a new `config.Language` type with defined constants.

Fixes #4010

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -21,7 +21,7 @@ package config
 // Config represents a librarian.yaml configuration file.
 type Config struct {
 	// Language is the language for this workspace (go, python, rust).
-	Language Language `yaml:"language"`
+	Language string `yaml:"language"`
 
 	// Version is the librarian tool version to use.
 	Version string `yaml:"version,omitempty"`
--- a/internal/config/language.go
+++ b/internal/config/language.go
@@ -18,36 +18,33 @@ import (
 	"github.com/googleapis/librarian/internal/yaml"
 )
 
-// Language is a programming language supported by Librarian.
-type Language = string
-
 const (
 	// LanguageUnknown represents an unsupported or unspecified language.
-	LanguageUnknown Language = "unknown"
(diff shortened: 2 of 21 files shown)


julieqiu:
I think this may be introducing more complexity than we need. A few concerns:

 - The existing string constants (e.g. languageRust = "rust") already provide compile-time safety and map directly to the YAML values, so no custom marshaling logic is required.
 - The int + iota approach is harder to debug. You see 5 instead of "go", and the error handling is inconsistent. UnmarshalText silently returns LanguageUnknown for invalid input, while String() panics on unknown values.
 - The user-facing schema docs now say list of config.Language instead of list of string, which is confusing since users are still writing YAML strings.

If the goal is to consolidate the language constants from serviceconfig and librarian into one place, I think we can do that by putting the shared string constants in config. That achieves the same consolidation with much less code.

jameslynnwu:
> The existing string constants (e.g. languageRust = "rust") already provide compile-time safety and map directly to the YAML values, so no custom marshaling logic is required.
I think using a type definition for enums provides more safety.

> The int + iota approach is harder to debug. You see 5 instead of "go", and the error handling is inconsistent. UnmarshalText silently returns LanguageUnknown for invalid input, while String() panics on unknown values.

I went with int+iota to define the enum, rather than a typedef with string as the underlying type, based on discussion with the team. I agree it does add complexity. IIUC, Stringer should make it display as "go" in most cases.

> The user-facing schema docs now say list of config.Language instead of list of string, which is confusing since users are still writing YAML strings.
Right, we could update the doc generators to support displaying string for simplicity.

PR #4293, opened 2026-03-02
refactor(internal/librarian/golang): refactor library cleaning logic

### LINKED ISSUE #4294: golang: refactor library cleaning logic to be more explicit
The current implementation of the Clean operation for Go libraries in `internal/librarian/golang/clean.go` relies on a recursive directory traversal and a complex set of regular expressions (generatedRegex) to identify generated files for removal.

This approach has several drawbacks:
- The generatedRegex is difficult to maintain and reason about as the generator evolves.
- Broad regex matching increases the risk of accidentally deleting handwritten files or veneers if they match the patterns.
- Recursive walking of the entire library directory is less efficient than targeted cleaning of known generated locations.

JoeWang1127:
This change refactors the Go library Clean operation to use a more explicit and targeted approach. 

Previously, the logic relied on a complex, recursive walk with a broad regex to identify generated files. 
The new implementation splits this into two focused functions, improving readability and reducing the risk of accidental deletion.

Changes:
- Refactored `Clean` logic: Replaced the generic clean function and generatedRegex with specialized helpers.
- Added `cleanRootFiles`: Specifically manages the removal of standard root files like `README.md` and `internal/version.go`, respecting the keep list.
- Added `cleanClientDirectory`: Iterates through API directories to remove generated Go client files (e.g., `_client.go`, `.pb.go`) based on explicit suffix matching.
- Improved Snippet Cleaning: Updated snippet directory cleanup to be more direct.
- Updated Tests: Refactored clean_test.go to align with the new logic, including more descriptive test cases and updated expectations for nested modules and API paths.

For #3617 
Fixes #4294

### DIFF
--- a/doc/config-schema.md
+++ b/doc/config-schema.md
@@ -121,7 +121,7 @@ This document describes the schema for the librarian.yaml.
 | Field | Type | Description |
 | :--- | :--- | :--- |
 | `client_package` | string | Is the package name of the generated client. |
-| `disable_gapic` | bool | Determines whether to generate the GAPIC client. Also known as proto-only client, which does not define a service in the proto files. TODO(https://github.com/googleapis/librarian/issues/4312): rename this config. |
+| `proto_only` | bool | Determines whether to generate a Proto-only client. A proto-only client does not define a service in the proto files. |
 | `enabled_generator_features` | list of string | Provides a mechanism for enabling generator features at the API level. |
 | `has_diregapic` | bool | TODO(https://github.com/googleapis/librarian/issues/4311): rename this config. HasDiregapic indicates whether generation uses DIREGAPIC (Discovery REST GAPICs). This is typically false. Used for the GCE (compute) client. |
 | `import_path` | string | Is the Go import path for the API. |
--- a/internal/config/language.go
+++ b/internal/config/language.go
@@ -32,10 +32,9 @@ type GoModule struct {
 type GoAPI struct {
 	// ClientPackage is the package name of the generated client.
 	ClientPackage string `yaml:"client_package,omitempty"`
-	// DisableGAPIC determines whether to generate the GAPIC client.
-	// Also known as proto-only client, which does not define a service in the proto files.
-	// TODO(https://github.com/googleapis/librarian/issues/4312): rename this config.
-	DisableGAPIC bool `yaml:"disable_gapic,omitempty"`
+	// ProtoOnly determines whether to generate a Proto-only client.
+	// A proto-only client does not define a service in the proto files.
+	ProtoOnly bool `yaml:"proto_only,omitempty"`
 	// EnabledGeneratorFeatures provides a mechanism for enabling generator features
 	// at the API level.
 	EnabledGeneratorFeatures []string `yaml:"enabled_generator_features,omitempty"`
--- a/internal/librarian/golang/generate.go
+++ b/internal/librarian/golang/generate.go
@@ -162,7 +162,7 @@ func generateAPI(ctx context.Context, api *config.API, library *config.Library, 
 		"--go-grpc_out=" + outdir,
 		"--go-grpc_opt=require_unimplemented_servers=false",
 	}
-	if !goAPI.DisableGAPIC {
+	if !goAPI.ProtoOnly {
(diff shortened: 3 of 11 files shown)


internal/librarian/golang/clean.go
```diff
@@ -102,36 +94,64 @@ func check(dir string, keep []string) (map[string]bool, error) {
...
+		// Skip cleaning these files.
+		if keepSet[rootFile] {
+			continue
 		}
-		if d.IsDir() {
-			if d.Name() == nestedModule {
-				return fs.SkipDir
-			}
-			return nil
+		rootFile := filepath.Join(libraryDir, rootFile)
```
zhumin8: the loop variable and this constructed path variable share the same name, is this intended?

internal/librarian/golang/clean.go
```diff
@@ -102,36 +94,64 @@ func check(dir string, keep []string) (map[string]bool, error) {
...
+			for _, file := range generatedClientFiles {
+				if !strings.HasSuffix(filepath.Base(path), file) {
+					continue
+				}
+				return os.Remove(path)
+			}
 			return nil
+		}); err != nil {
+			return err
+		}
```
miguelvelezsa: this can be in a new function :)

internal/librarian/golang/clean.go
```diff
@@ -102,36 +94,67 @@ func check(dir string, keep []string) (map[string]bool, error) {
...
+		if keepSet[rootFile] {
+			continue
+		}
+		rootFilePath := filepath.Join(libraryDir, rootFile)
+		// Some library may not have the root file, README.md for example, this is rare,
+		// but we should not fail the clean in this case.
+		if _, err := os.Stat(rootFilePath); os.IsNotExist(err) {
+			continue
+		}
+		if err := os.Remove(rootFilePath); err != nil {
```
jskeet: Can we not just use os.Remove and check any error for IsNotExist, rather than using Stat earlier?

internal/librarian/golang/clean.go
```diff
@@ -20,32 +20,28 @@ import (
...
-			regexp.MustCompile(prefix + `.*_client\.go$`),
-			regexp.MustCompile(prefix + `.*_client_example_go123_test\.go$`),
-			regexp.MustCompile(prefix + `.*_client_example_test\.go$`),
-			regexp.MustCompile(prefix + `gapic_metadata\.json$`),
-			regexp.MustCompile(prefix + `helpers\.go$`),
-			regexp.MustCompile(`.*pb/.*\.pb\.go$`),
-			regexp.MustCompile(`(^|.*/)internal/generated/snippets/.*$`),
-		}
-	}()
+	generatedClientFiles = []string{
```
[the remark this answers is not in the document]

internal/librarian/golang/clean.go
```diff
@@ -102,36 +94,67 @@ func check(dir string, keep []string) (map[string]bool, error) {
...
+		if err := os.RemoveAll(snippetDir); err != nil {
+			return err
 		}
-		rel, err := filepath.Rel(dir, path)
+	}
+	return nil
+}
+
+func cleanGeneratedClientFiles(clientPath string) error {
+	return filepath.WalkDir(clientPath, func(path string, d fs.DirEntry, err error) error {
```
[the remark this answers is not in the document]

internal/librarian/golang/clean.go
```diff
@@ -87,13 +84,9 @@ func check(dir string, keep []string) (map[string]bool, error) {
...
 		if _, err := os.Stat(path); errors.Is(err, fs.ErrNotExist) {
-			return nil, fmt.Errorf("keep file %q does not exist", k)
+			return nil, fmt.Errorf("error keeping %s: %w", k, err)
 		}
-		// Effectively get a canonical relative path. While in most cases
-		// this will be equal to k, it might not be - in particular,
-		// on Windows the directory separator in paths returned by Rel
-		// will be a backslash.
-		rel, err := filepath.Rel(dir, path)
+		rel, err := filepath.Abs(path)
```
jskeet: At the very least, the variable shouldn't be called "rel" if it's absolute - but why are we doing this at all? I think it makes more sense for the keep set to be names relative to the repo root, and then use those when comparing in the clean calls.

I think either way it would be worth modifying the `check` comment to make it clear whether the keep set is relative (and if so, to what) or absolute.

PR #4305, opened 2026-03-03
chore(tool/cmd/migrate): parse disable GAPIC from `BUILD.bazel`

JoeWang1127:
Parse disable GAPIC from `BUILD.bazel`.

For #3618


zhumin8 (approved):
question: does disable GAPIC means [SkipGenerate](https://github.com/googleapis/librarian/blob/f23571dbf84279da5dcccf643ea1b4c17e93501c/internal/config/config.go#L189)?

Issue #4183, opened 2026-02-25
sidekick/rust: implement helper to generate resource name


haphungw:
Generate resource name from `TargetResource.Template` specifically for Rust.
Modify `.mustache` to inject resource name into Rust code.

PR #4320, opened 2026-03-04
refactor(sidekick): build TargetResource.Template from PathTemplate.Segments

haphungw:
Update the internal representation of `TargetResource.Template` from a flat string to a structured `[]PathSegment` slice. This change aligns the resource name logic more closely with `PathTemplate` and avoids brittle string-parsing/regex-replacements.

For #4183

### DIFF
--- a/internal/sidekick/api/test.go
+++ b/internal/sidekick/api/test.go
@@ -278,7 +278,10 @@ func ParseTemplateForTest(template string) []PathSegment {
 	var segments []PathSegment
 	parts := strings.Split(strings.TrimPrefix(template, "//"), "/")
 
-	host := "//" + parts[0]
+	host := parts[0]
+	if strings.HasPrefix(template, "//") {
+		host = "//" + parts[0]
+	}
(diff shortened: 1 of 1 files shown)


internal/sidekick/api/test.go
```diff
@@ -272,3 +272,23 @@ func (r *Resource) WithSingular(singular string) *Resource {
 	r.Singular = singular
 	return r
 }
+
+// ParseTemplateForTest converts a string literal into a []PathSegment slice for testing purposes.
+func ParseTemplateForTest(template string) []PathSegment {
+	var segments []PathSegment
+	parts := strings.Split(strings.TrimPrefix(template, "//"), "/")
+
+	host := "//" + parts[0]
```
coryan: If the string was missing the `//` prefix this would mask the problem, no?

Maybe:
```suggestion
	host := parts[0]
	if strings.HasPrefix(template, "//") {
	    host = "//" + parts[0]
	 }
```

Issue #4062, opened 2026-02-13
java: add generate command


julieqiu:
Add generate command for Java.

PR #4341, opened 2026-03-05
feat(internal/librarian/java): add support for NoRestNumericEnums in Java generation

zhumin8:
Java generation to respect `no_rest_numeric_enums` set in librarian config (parsed from BUILD.bazel in [migrate](https://github.com/googleapis/librarian/pull/4335)), and pass down argument to java generator.

For #4062

### DIFF
--- a/internal/librarian/java/generate.go
+++ b/internal/librarian/java/generate.go
@@ -88,17 +88,7 @@ func generateAPI(ctx context.Context, api *config.API, library *config.Library, 
 			return fmt.Errorf("failed to create directory %s: %w", dir, err)
 		}
 	}
-
-	var javaAPI *config.JavaAPI
-	if library.Java != nil {
-		for _, ja := range library.Java.JavaAPIs {
-			if ja.Path == api.Path {
-				javaAPI = ja
-				break
-			}
-		}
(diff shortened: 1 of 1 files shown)


internal/librarian/java/generate.go
```diff
@@ -215,21 +216,22 @@ func createProtocOptions(api *config.API, library *config.Library, googleapisDir
...
 	}
 	// transport specifies whether to generate gRPC, REST, or both types of clients.
 	gapicOpts = append(gapicOpts, gapicOpt("transport", transport))
+
 	// rest-numeric-enums ensures that enums in REST requests are encoded as numbers
 	// rather than strings.
-	// TODO(https://github.com/googleapis/librarian/issues/4130):
-	// assign this according to config
-	gapicOpts = append(gapicOpts, "rest-numeric-enums")
+	if javaAPI == nil || !javaAPI.NoRestNumericEnums {
```
JoeWang1127: nit: you may consider fill JavaAPI for every API path of a library before running generate function. Therefore, return an error for nil JavaAPI and remove all check.

JoeWang1127 (approved):

PR #4346, opened 2026-03-05
feat(sidekick/rust): extract variables in rust template

haphungw:
Extract segment substitutions into variable bindings before string formatting, making the variables reusable for both `path` and `_resource_name`.

For #4183

### DIFF
--- a/internal/sidekick/rust/templates/crate/src/transport.rs.mustache
+++ b/internal/sidekick/rust/templates/crate/src/transport.rs.mustache
@@ -104,10 +104,13 @@ impl super::stub::{{Codec.Name}} for {{Codec.Name}} {
         {{#PathInfo.Bindings}}
         .or_else(|| {
             {{#Codec.HasVariablePath}}
+            {{#Codec.Substitutions}}
+            let {{VariableName}} = try_match({{{FieldAccessor}}}, {{{TemplateAsArray}}})?;
+            {{/Codec.Substitutions}}
             let path = format!(
                 "{{Codec.PathFmt}}",
                 {{#Codec.Substitutions}}
-                try_match({{{FieldAccessor}}}, {{{TemplateAsArray}}})?,
+                {{VariableName}},
                 {{/Codec.Substitutions}}
             );
             {{/Codec.HasVariablePath}}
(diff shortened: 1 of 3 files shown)


internal/sidekick/rust/templates/crate/src/transport.rs.mustache
```diff
@@ -104,10 +104,13 @@ impl super::stub::{{Codec.Name}} for {{Codec.Name}} {
         {{#PathInfo.Bindings}}
         .or_else(|| {
             {{#Codec.HasVariablePath}}
+            {{#Codec.Substitutions}}
+            let {{VariableName}} = try_match({{{FieldAccessor}}}, {{{TemplateAsArray}}})?;
```
westarle: maybe `var_{{VariableName}}` or `{{VariableName}}_id` in case the variable name is path.

haphungw: > maybe `var_{{VariableName}}` or `{{VariableName}}_id` in case the variable name is path.

This seems simple at first, but it breaks down on reserved keywords. If the API has a parameter named `type` (which is the case for `dns`, our `VariableName()` helper evaluates to `r#type` to safely escape it for Rust. If we blindly prepend or append in Mustache, it generates `var_r#type` or `r#type_id`, which is a syntax error.

I think we can prevent shadowing by isolating the variable extraction and string formatting inside its block scope.

```rust
let (path, _resource_name) = {
    let project = try_match(Some(&req).map(|m| &m.project)...)?;
    let r#type = try_match(Some(&req).map(|m| &m.r#type)...)?;

    let url_path = format!("v1/projects/{}/types/{}", project, r#type);
    let rn = format!("//compute.googleapis.com/projects/{}/types/{}", project, r#type);
    (url_path, rn)
};
```
@coryan Do you think this would look okay?

coryan: `VariableName()` could use `fmt.Sprintf("var_%s", toSnakeNoMangling(thing))` the `*NoMangling` function which never appends a `r#` and is intended for names that get a suffix / prefix like this.

That assumes there are no keywords that start with `var_`, but that seems reasonable?

haphungw: thanks, that's great! updated accordingly

coryan (approved):

Issue #4395, opened 2026-03-07
tool/cmd/migrate: migrate configuration for .NET


julieqiu:
Update tool/cmd/migrate to migrate configuration for .NET to librarian.yaml.

See go/sdk-librarian-dotnet for details.

PR #4414, opened 2026-03-08
feat(tool/cmd/migrate): add .NET migration

julieqiu:
The migrate tool now supports google-cloud-dotnet. It reads generator-input/apis.json and converts each API entry into a librarian.yaml library config. The migration handles generated libraries (micro and proto generators), handwritten veneers, non-default transports, filtered dependencies, preview release levels, blocked releases, and package groups.

The output of `tool/cmd/migrate` for dotnet is at https://gist.github.com/julieqiu/8dc44f2c18696f196a5bc0476229de50. 

For https://github.com/googleapis/librarian/issues/4395

### DIFF
--- a/tool/cmd/migrate/dotnet.go
+++ b/tool/cmd/migrate/dotnet.go
@@ -28,33 +28,31 @@ import (
 	"github.com/googleapis/librarian/internal/librarian"
 )
 
-// ApisJSON represents the root of the apis.json file.
-type ApisJSON struct {
-	APIs          []APIEntry     `json:"apis"`
-	PackageGroups []PackageGroup `json:"packageGroups"`
+// DotnetAPIsJSON represents the root of the apis.json file.
+type DotnetAPIsJSON struct {
+	APIs          []DotnetAPIEntry     `json:"apis"`
+	PackageGroups []DotnetPackageGroup `json:"packageGroups"`
 }
 
-// APIEntry represents a single API entry in apis.json.
-type APIEntry struct {
-	ID            string            `json:"id"`
-	Version       string            `json:"version"`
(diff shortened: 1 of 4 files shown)


tool/cmd/migrate/dotnet.go
```diff
@@ -0,0 +1,204 @@
...
+	var libs []*config.Library
+	for _, api := range apisJSON.APIs {
+		lib := &config.Library{
+			Name:    api.ID,
+			Version: api.Version,
+		}
+
+		// Set APIs from protoPath for generated libraries.
+		isHandwritten := api.Generator == "None"
+		if !isHandwritten && api.ProtoPath != "" {
```
jskeet: I'd *expect* api.ProtoPath to always be specified when there's a generator. Anything else is probably the sign of a problem, so maybe something like

```go
if !isHandwritten {
  if api.ProtoPath == "" 
    // return some error here
  }
  lib.Apis = // code as below
}

Comment (julieqiu) [reply tool/cmd/migrate/dotnet.go]: fixed

File: tool/cmd/migrate/dotnet_test.go [hunk 2]
```diff
@@ -0,0 +1,418 @@
...
+					{
+						ID: "Google.Cloud.Datastore",
+						PackageIDs: []string{
+							"Google.Cloud.Datastore.V1",
+							"Google.Cloud.Datastore.Admin.V1",
+						},
+					},
+				},
+			},
+			want: wantConfig([]*config.Library{
```
jskeet: I'm not at all sure this is really what we'll want, but I guess it's okay until we've actually decided what to do about package groups?

Issue #4404, opened 2026-03-08
nodejs: implement Node.js generation logic


julieqiu:
This issue tracks the implementation of the Node.js code generation pipeline in librarian generate.

We need to create the internal/librarian/nodejs package and integrate it into internal/librarian/generate.go.

The generation logic should :

  1. Clean the target directory, preserving files listed in keep (e.g., CHANGELOG.md, handwritten .ts files).
  2. Invoke gapic-generator-typescript, writing output to owl-bot-staging/.
  3. Run synthtool (node_mono_repo.owlbot_main) to move files from staging to packages/<library>/, apply owlbot.py patches, and format with
  gts.

  See go/sdk-librarian-nodejs for details.

suztomo:
The `generate` command works for NodeJS.


<details>

<summary>How to run the generate command for the google-cloud-agentregistry library</summary>

```
suztomo@suztomo:~/librarian-2026/google-cloud-node$ /tmp/librarian generate google-cloud-agentregistry
suztomo@suztomo:~/librarian-2026/google-cloud-node$ git diff |head -30
diff --git a/packages/google-cloud-agentregistry/protos/protos.js b/packages/google-cloud-agentregistry/protos/protos.js
index e8aaa3ea7c..6154bd8bc0 100644
--- a/packages/google-cloud-agentregistry/protos/protos.js
+++ b/packages/google-cloud-agentregistry/protos/protos.js
@@ -28,7 +28,7 @@
     var $Reader = $protobuf.Reader, $Writer = $protobuf.Writer, $util = $protobuf.util;
     
     // Exported root namespace
-    var $root = $protobuf.roots._google_cloud_agentregistry_protos || ($protobuf.roots._google_cloud_agentregistry_protos = {});
+    var $root = $protobuf.roots["_google_cloud_agentregistry_protos"] || ($protobuf.roots["_google_cloud_agentregistry_protos"] = {});
     
     $root.google = (function() {
     
@@ -230,9 +230,13 @@
                          * @param {$protobuf.Writer} [writer] Writer to encode to
                          * @returns {$protobuf.Writer} Writer
                          */
-                        Agent.encode = function encode(message, writer) {
+                        Agent.encode = function encode(message, writer, q) {
                             if (!writer)
                                 writer = $Writer.create();
+                            if (q === undefined)
+                                q = 0;
+                            if (q > $util.recursionLimit)
+                                throw Error("max depth exceeded");
[...]

PR #4442, opened 2026-03-09
feat(internal/librarian/nodejs): implement dual-path post-processing

julieqiu:
The Node.js generator is updated to support both the legacy synthtool workflow and the new standalone script workflow for post-processing.

Template generation and exclusions for Node.js are now handled at the generator level. The runPostProcessor logic checks for the existence of owlbot.py. If present, it executes it using python. If not, it runs the new standalone post-processing scripts: librarian.js (unrelated to the Librarian CLI tool) and generates partial README updates using .readme-partials.yaml via npx gapic-node-processing generate-readme.

The .github/workflows/nodejs.yaml file is also updated to include the Python setup and synthtool installation steps required for the legacy owlbot.py fallback path.

For https://github.com/googleapis/librarian/issues/4404

### DIFF
--- a/internal/librarian/nodejs/generate_test.go
+++ b/internal/librarian/nodejs/generate_test.go
@@ -289,8 +289,7 @@ func TestBuildGeneratorArgs(t *testing.T) {
 			name: "grpc+rest transport is default and not passed",
 			api:  &config.API{Path: "google/cloud/secretmanager/v1"},
 			library: &config.Library{
-				Name:      "google-cloud-secretmanager",
-				Transport: "grpc+rest",
+				Name: "google-cloud-secretmanager",
 			},
 			want: []string{
 				"gapic-generator-typescript",


internal/librarian/nodejs/generate.go
```diff
@@ -206,3 +285,51 @@ func derivePackageNameFromLibraryName(name string) string {
...
+		if err != nil {
+			return err
+		}
+		return os.WriteFile(target, data, info.Mode())
+	})
+}
+
+// restoreNonConflicting copies files from backup into dst that do not already
+// exist in dst.
+func restoreNonConflicting(backup, dst string) error {
```
JoeWang1127: Maybe we can also move this function to filesystem package?

julieqiu: I actually think this approach is too complicated, and went with something simpler: https://github.com/googleapis/librarian/pull/4442/commits/224dcfae4b7ef05ea83dbe9b35a4f073f3336f7f

internal/librarian/nodejs/generate.go
```diff
@@ -153,6 +163,34 @@ func buildGeneratorArgs(api *config.API, library *config.Library, googleapisDir,
...
+
+	// combine-library wipes the destination directory before writing generated
+	// files (src/, protos/). Back up existing files so non-generated files like
+	// librarian.js, README.md, and .readme-partials.yaml survive.
+	backupDir, err := os.MkdirTemp("", "librarian-backup-*")
+	if err != nil {
+		return fmt.Errorf("failed to create backup dir: %w", err)
+	}
+	defer os.RemoveAll(backupDir)
+	if err := copyDir(outDir, backupDir); err != nil {
```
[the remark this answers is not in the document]
julieqiu: ditto re: something simpler: https://github.com/googleapis/librarian/pull/4442/commits/224dcfae4b7ef05ea83dbe9b35a4f073f3336f7f

PR #4447, opened 2026-03-09
chore(internal/librarian): disable API path derivation for Python

JoeWang1127:
This PR extracts the logic for determining whether an API path can be derived into a new `canDeriveAPIPath` helper function. It also updates the behavior to skip API path derivation for Python (`config.LanguagePython`), in addition to Go.

This change ensures that Librarian does not attempt to automatically derive API paths for Python libraries, as the library name may not contain the relevant information needed for derivation.

### DIFF
--- a/internal/librarian/tidy_test.go
+++ b/internal/librarian/tidy_test.go
@@ -277,12 +277,32 @@ func TestTidy_DerivableFields(t *testing.T) {
 			wantNumLibs:  1,
 			wantNumChnls: 0,
 		},
+		{
+			name: "do not derive api path for Python library",
+			config: &config.Config{
+				Language: config.LanguagePython,
+				Sources:  googleapisSource,
+				Libraries: []*config.Library{
+					{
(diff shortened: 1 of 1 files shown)


internal/librarian/library.go
```diff
@@ -191,7 +191,7 @@ func libraryOutput(language string, lib *config.Library, defaults *config.Defaul
 // applyDefaults applies language-specific derivations and fills defaults.
 func applyDefaults(language string, lib *config.Library, defaults *config.Default) (*config.Library, error) {
 	if !isVeneer(language, lib) {
-		if len(lib.APIs) == 0 && language != config.LanguageGo {
+		if len(lib.APIs) == 0 && canDeriveAPIPath(language) {
```
jskeet: I think we should change tidy.go's `isDerivableAPIPath` code to use this as well, otherwise `tidy` might remove APIs that it thinks will be derived, but then this code won't do so.

PR #4480, opened 2026-03-10
feat(librarianops): support running librarian in docker

### LINKED ISSUE #4463: librarianops: add --docker flag
This should derive a docker image/tag to use, and invoke librarian by using "docker run" with that image.

### LINKED ISSUE #4478: librarianops: declare that google-cloud-go and google-cloud-python are supported repos

jskeet:
Adds the --docker flag to librarianops, initially just to the generate command. This uses the version and language from the librarian config to derive a docker image to run whenever librarian should be invoked.

Fixes very minor issues at the same time:
- Makes librarianops fail clearly when there's no version in the config
- Allows a relative path to be specified when running librarianops with `-C`; in particular, `librarianops -C .` really should work
- Removes an unnecessary `filepath.Join` call when reading the config
- Adds google-cloud-python and google-cloud-go to the list of supported repos

Fixes #4463
Fixes #4478


cmd/librarian/build-docker-images.sh
JoeWang1127: Can we use Go for this file?

julieqiu (approved):

Issue #4473, opened 2026-03-10
librarian: add --push-script flag to tag command


jskeet:
The tag command creates tags in the local repository, but does not push them. We should add a --push-script flag; if specified, which assumes a remote of "origin" to push to.

This will simplify operations, while the pushing is performed by the Platform team.

Once automation has been completed, we *might* want to remove this flag. We *might* also want to add the ability to determine the remote automatically, but the playbook instructions will clone a fresh repo anyway, with a default remote of origin.

codyoss:
Closing as we will revisit when we design publish.

PR #4501, opened 2026-03-11
feat(librarian): adds --push-remote flag to librarian tag command

jskeet:
Adds an optional flag to the tag command, invoking "git push" after creating all tags locally, to push the tags to the specified remote.

This is an alternative approach to #4473.


JoeWang1127 (commented):
Do we need to consider permission to run this command in googleapis repo?

JoeWang1127 (approved):

PR #4504, opened 2026-03-11
feat(internal/librarian): add support for Java fill defaults and update migration

zhumin8:
Support Java-specific defaults in librarian.yaml and update the migration tool to populate these values.


zhumin8:
filling default for lib.Java.LibraryTypeOverride is a bit weird. Will re-think on how to handle it.

PR #4519, opened 2026-03-11
fix(internal/librarian/rust): remove wrong argument in semverCheck

miguelvelezsa:
The `--frozen` argument was incorrectly added to `semverCheck` in I wrongly added it in https://github.com/googleapis/librarian/commit/60b64baccb223381382c11bcab62f6b459a12a1b without validation, and the flag does not actually exist.

See https://crates.io/crates/cargo-semver-checks#what-features-does-cargo-semver-checks-enable-in-the-tested-crates for details.


JoeWang1127:
Do you test this command locally?

Issue #4306, opened 2026-03-03
java: migrate configurations to librarian.yaml


zhumin8:
Add to `tool/cmd/migrate` to support configuration migration from google-cloud-java repo for Java. 
This should include: 
- migrate configurations from generation_config.yaml to librarian.yaml
- migrate configurations parsed from BUILD.bazel
- migrate configurations from .OwlBot-hermetic.yaml (for files/globs to keep)

PR #4553, opened 2026-03-12
feat(tool/cmd/migrate): migrate Java OwlBot preserve regex to librarian.yaml

zhumin8:
The migration tool for Java now parses .OwlBot-hermetic.yaml files in each library directory to extract deep-preserve-regex patterns. These patterns are added to the keep field of the corresponding library in librarian.yaml.

This assumes that the "regex" in .OwlBot-hermetic.yaml files are actually full file or dir paths, which is true for most of the cases out of the 33 in google-cloud-java. Will deal with outliers in followups.

For #4306


zhumin8:
#4559 to re-evaluate clean logic implemented. For now, merging this change can unblock more libraries to generate.

Issue #4566, opened 2026-03-13
migrate: support selective, incremental migration


jskeet:
We should be able to migrate one library at a time, or a selected set of libraries, rather than just "everything".
If a librarian@latest configuration already exists, this should add (or replace) the libraries, rather than starting again.

PR #4568, opened 2026-03-13
feat(migrate): allow incremental and selective migration (Go and Python)

jskeet:
Allows library names to be specified on the command line, after the repo path. Only those libraries are migrated. (If no libraries are specified, all libraries are migrated.)

If an existing configuration is found, only libraries which aren't already present are migrated, and any existing top-level configuration is preserved.

Towards #4566.


tool/cmd/migrate/legacylibrarian.go
```diff
@@ -208,6 +237,41 @@ func fetchGoogleapisWithCommit(ctx context.Context, endpoints *fetch.Endpoints,
...
+		}
+		result = append(result, library)
+	}
+	cfg.Libraries = result
+	return cfg, nil
+}
+
+// mergeLibraries returns a merged slice of libraries, containing all libraries
+// from existingConfig, and libraries in newConfig which don't already exist in
+// newConfig. The order of the libraries in the slice is the libraries in
```
JoeWang1127:
```suggestion
// existingConfig. The order of the libraries in the slice is the libraries in
```

JoeWang1127 (approved):

tool/cmd/migrate/legacylibrarian.go
jskeet: Whoops, thanks. github does bad things if I apply suggestions due to my personal email address - feel free to just add an extra commit yourself.

PR #4619, opened 2026-03-16
feat(internal/legacylibrarian): adapt legacylibrarian release stage for librarian generate

JoeWang1127:
This change modifies Legacylibrarian stage to correctly calculate the next version for changes made by librarian generate.

Introduce a config, `ReleaseOnlyMode`, in state.yaml.

Legacylibrarian stage will create a minor version bump for changed libraries if `ReleaseOnlyMode` sets to `true`, regardless of the commit type. 

Running legacylibrarian generate when `ReleaseOnlyMode` sets to `true` will fail.

Since librarian generate does not produce conventional commit messages, the version analyzer now enforces a minor version bump if the release only mode is set to true.

For #4637

### DIFF
--- a/internal/legacylibrarian/legacylibrarian/release_stage_test.go
+++ b/internal/legacylibrarian/legacylibrarian/release_stage_test.go
@@ -1378,7 +1378,6 @@ func TestRunStageCommand(t *testing.T) {
 		r := &stageRunner{
 			repo:            test.repo,
 			state:           test.state,
-			image:           test.state.Image,
 			librarianConfig: test.config,
 			containerClient: test.client,
 		}


internal/legacylibrarian/legacylibrarian/command.go
```diff
@@ -196,6 +197,7 @@ func newCommandRunner(cfg *legacyconfig.Config) (*commandRunner, error) {
 		state:           state,
 		librarianConfig: librarianConfig,
 		image:           image,
+		migrate:         state.Migrate,
```
jskeet: No need, just read it from state wherever it's currently being read directly.

internal/legacylibrarian/legacylibrarian/commit_version_analyzer_test.go
```diff
@@ -654,6 +654,7 @@ func TestNextVersion(t *testing.T) {
 		name           string
 		commits        []*legacygitrepo.ConventionalCommit
 		currentVersion string
+		migration      bool
```
jskeet: And again.

internal/legacylibrarian/legacylibrarian/release_stage_test.go
```diff
@@ -1303,6 +1407,116 @@ func TestRunStageCommand(t *testing.T) {
...
+func TestProcessLibrary_GoMigration(t *testing.T) {
+	for _, test := range []struct {
+		name            string
+		releaseOnlyMode bool
+		libraryState    *legacyconfig.LibraryState
+		repo            legacygitrepo.Repository
+		want            *legacyconfig.LibraryState
+	}{
+		{
+			name:            "libraries in release only mode do not have change",
```
jskeet: I don't understand (from the name) what this test does - and it sounds very much like the test below. Again, I'd think about "In release-only mode" as a repo state rather than a library state.

internal/legacylibrarian/legacylibrarian/release_stage_test.go
```diff
@@ -1303,6 +1409,116 @@ func TestRunStageCommand(t *testing.T) {
...
+			},
+		},
+	} {
+		state := &legacyconfig.LibrarianState{
+			Libraries: []*legacyconfig.LibraryState{
+				test.libraryState,
+			},
+		}
+		r := &stageRunner{
+			releaseOnlyMode: test.releaseOnlyMode,
```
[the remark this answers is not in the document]
JoeWang1127: Done.

PR #4623, opened 2026-03-17
chore(internal/librarianops): update pr title raised by librarianops

JoeWang1127:
Update pr title raised by librarianops.

Since we only migrating generate in Go SDK, we need to update the generate pr title from `chore` to `feat` so that legacylibrarian can trigger a release.


jskeet:
With this in place, does that mean we don't need #4619?

PR #4627, opened 2026-03-17
chore(rust): remove redundant assignment in createRepoMetadata

jskeet:
Removes an assignment to RepoMetadata.Repo which is already performed in repometadata.FromLibrary.


coryan:
FYI, this introduced a number of unexpected changes to `google-cloud-rust` because the librarian.yaml file there did not have a `repo` field.  The changes happened in:

https://github.com/googleapis/google-cloud-rust/pull/5055

Fixed in:

https://github.com/googleapis/google-cloud-rust/pull/5067

PR #4632, opened 2026-03-17
chore(internal/config): update gapic-showcase dependency

haphungw:
Update `gapic-showcase` source in the test configurations to the latest commit, pulling in the recent fixes for T4 spans and logs.


internal/config/testdata/rust/librarian.yaml
```diff
@@ -1,10 +1,10 @@
-# Copyright 2025 Google LLC
+# Copyright 2026 Google LLC
```
coryan: This should not be needed..

Issue #4453, opened 2026-03-10
nodejs: preserve original copyright year in generated files


julieqiu:
When librarian regenerates a Node.js library, gapic-generator-typescript stamps the current year in copyright headers of all generated files.

This overwrites the original copyright year. For example, a file first generated in 2025 gets its header changed to 2026 on the next regeneration. The copyright year should reflect when the file was first created, not when it was last regenerated.

We should add a post-processing step after generation that restores the original copyright year for files that already existed, and use the copyright-year field in librarian.yaml to store this information.

PR #4693, opened 2026-03-20
fix(librarian/nodejs): preserve original copyright year in generated files

julieqiu:
When librarian regenerates a Node.js library, the TypeScript generator stamps the current year in copyright headers, which overwrites the original copyright year for pre-existing files. This change adds a post-processing step to save
the paths of existing generated files before they are wiped, and restores their original copyright year using `library.CopyrightYear` from `librarian.yaml`.

Fixes https://github.com/googleapis/librarian/issues/4453


julieqiu:
@gemini-cli please add tests

PR #4695, opened 2026-03-20
fix(.github/workflows): replace github mcp server with gh cli

julieqiu:
Remove use of the GitHub MCP and use the pre-installed GitHub CLI directly for use in Gemini workflows.


julieqiu:
@gemini-cli please confirm if you can create an issue with the updated settings in this PR

PR #4696, opened 2026-03-20
refactor(.gemini): extract shared context and support issue creation

julieqiu:
The invoke and plan-execute prompts contained duplicated content, which is moved to a new shared-context.md file and included via @{shared-context.md}.

Both prompts also gain an explicit list of available GitHub MCP tools, guidance on how to create issues, and project-specific conventions for issue titles and commit messages from .agents/skills/.


julieqiu:
@gemini-cli /approve

Issue #4714, opened 2026-03-22
surfer: refactor files for clear separation of concerns


sarahheacock:
This issue tracks the initial structural refactoring of the CLI generator. The primary goal is to make the codebase more readable and establish a clearer separation of concerns for future feature work.
Currently, the generator relies on passing large data structures (`config`, `model`, `service`) through long chains of procedural functions, making code refactor challenging. 

To resolve this, we will:
1. Abstract logic into dedicated objects and files (`CommandGroupBuilder`, `CommandBuilder`, `ArgBuilder`, and `Emitter`) to cleanly encapsulate state and stop passing heavy arguments.
2. Move shared evaluation functions into a distinct `provider` boundary. Long-term, this provider boundary should be improved by moving shared logic into sidekick

These are purely structural shifts to reduce data-clumping; NO functional behavior is being altered yet. This does not cover all refactoring that should be done but is a starting point.

PR #4717, opened 2026-03-22
refactor(internal/surfer/gcloud): encapsulate builder arguments into structs

sarahheacock:
To reduce the overhead of passing redundant parameters like method, overrides, model, and service across multiple helper functions, stateful CommandBuilder and ArgumentBuilder structs are introduced to encapsulate this context.
The core Build methods are refactored to explicitly assign generated components via small, focused receiver methods.

For: #4714


quirogas (approved):
LGTM

PR #4718, opened 2026-03-23
refactor(internal/surfer/gcloud): update argument builder to focus on one argument

### LINKED ISSUE #4525: surfer: Refactor `addFlattenedParams` to return values instead of mutating pointers
Currently, the `addFlattenedParams` function modifies an `*Arguments` pointer passed to it as a parameter in order to append fields during generation. In Go, functions that explicitly return values are generally easier to reason about, test, and maintain than functions that rely on mutating pointer arguments as a side effect. This refactoring will improve the clarity and functional purity of the parameter generation logic.

- Refactor `addFlattenedParams` and change its signature so it returns `[]Param` instead of mutating state.
- Update the callers (like `newArguments`/`newCommand`) to capture the returned slice and append it to the resulting array.
- Consider renaming the function (e.g., `flattenField`, `newFlag`, or `newPositionalParam`) if it helps clarify whether the resulting parameters are flags or positional arguments in the gcloud CLI context.

## Relevant files
- `internal/surfer/gcloud/builder.go`

sarahheacock:
ArgumentBuilder is refactored to exclusively construct a single Argument per api.Field. Its Build method is tightened from a monolithic block into declarative helper methods.

The recursive field extraction logic, argumentsFromField (prev addFlattenedArguments), is moved to CommandBuilder. It now explicitly returns a populated Argument slice rather than mutating a pointer, improving readability and reducing side effects.

For #4714
Fix #4525


quirogas (approved):
🚀

Issue #4721, opened 2026-03-23
sidekick/config: move Sources type to package librarian


julieqiu:
The Sources struct in internal/sidekick/config/source_config.go represents resolved directory paths for source repositories (Googleapis, Showcase, Discovery, ProtobufSrc, Conformance).

Today it’s only used by a subset of language generators (notably rust and dart), but the abstraction itself is not sidekick-specific. It’s just a container for resolved source paths.

Move Sources to a shared package (e.g. internal/librarian) so it reflects its actual role as a general-purpose “resolved sources” type.

PR #4725, opened 2026-03-23
refactor(internal/sources): move Sources types out of sidekick/config

julieqiu:
The Sources and SourceConfig types in internal/sidekick/config represent resolved directory paths for source repositories. However, sources is not a sidekick-specific concept.

While sidekick language generators (e.g. rust, dart) currently use this more heavily than others, the concept itself should be generic and usable across the codebase.

Move both types into a new internal/sources package as a first step toward decoupling and preventing circular dependencies.

Additionally, change Sources and SourceConfig to be passed by reference consistently throughout the codebase.

For https://github.com/googleapis/librarian/issues/4721


quirogas (approved):
LGTM! (surfer)

Issue #4738, opened 2026-03-23
python: revisit SkipReadmeCopy after migration


jskeet:
We're adding functionality to conditionalize the copying of README.rst from the package root to the docs folder.

We should consider whether to just do that unconditionally, once we're prepared to make deliberate changes in the repo.

jskeet:
Consulting @daniel-sanche and @parthea offline.

PR #4739, opened 2026-03-23
feat(python): handle libraries with no docs/README.rst file

jskeet:
This is detected during migration, and added in a new PythonPackage field (SkipReadmeCopy). If the field is set, then at the end of generation we don't copy the file.

After migration is completed, we can potentially remove this. See #4738.


jskeet:
> Confused by the test failing. Converting to a draft while I investigate.

Looks like this was due to CI using an old version of synthtool.

Issue #4786, opened 2026-03-23
rust: IsVeneer() incorrectly classifies libraries with empty apis and non-empty output as veneers


julieqiu:
See https://github.com/googleapis/google-cloud-rust/pull/5117#issue-4124175823

PR #4787, opened 2026-03-23
fix(internal/librarian/rust): check sdk.yaml to determine veneer

julieqiu:
IsVeneer previously treated any library with no APIs and an explicit output path as a veneer. This incorrectly classified nosvc libraries like google-cloud-oslogin-common, which have a known API path in sdk.yaml but no APIs populated yet at the time IsVeneer is called. When treated as a veneer, generateVeneer found no modules and silently produced no output.

IsVeneer now derives the API path from the library name and checks sdk.yaml. If the path is listed, the library is a generated library whose APIs have not been populated yet, not a veneer. Handwritten libraries like google-cloud-auth, whose derived path is not in sdk.yaml, continue to be classified as veneers.

Fixes https://github.com/googleapis/librarian/issues/4786

### DIFF
--- a/internal/librarian/rust/generate.go
+++ b/internal/librarian/rust/generate.go
@@ -25,6 +25,7 @@ import (
 	"github.com/googleapis/librarian/internal/command"
 	"github.com/googleapis/librarian/internal/config"
 	"github.com/googleapis/librarian/internal/repometadata"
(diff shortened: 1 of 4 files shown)


coryan (approved):
Tested and it works. Thanks!

PR #4798, opened 2026-03-24
fix(librarian): do not report coverage until all coverage reports received

### LINKED ISSUE #4761: codecov: status initially fails before updating to passing after coverage upload
There seems to be a delay/race condition in Codecov status reporting.

The Codecov check will initially report as failing, before all tests and coverage uploads have completed successfully. After some time, the status updates to passing without any changes to the commit.

This creates confusing signal in PRs where the check appears to fail, until all the jobs have finished.

jameslynnwu:
There are two approaches to fixing the issue:
- Wait until N reports have been received by code coverage. If the number of coverage reports changes, we will need to update this value to stay in sync.
- Upload all coverage reports in a [final job](https://about.codecov.io/blog/uploading-code-coverage-in-a-separate-job-on-github-actions/#:~:text=You%20can%20upload%20coverage%20reports%20to%20Codecov,a%20dependency%20to%20the%20applicable%20coverage%20jobs.). This eliminates the hardcoded count of coverage reports that needs to stay in sync, but requires rewriting the workflows and may make them more complex.

For now, will wait until the 7 tests that report coverage have finished.

Fixes #4761


codecov.yml
```diff
@@ -11,6 +11,10 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+codecov:
+  notify:
+    # This should be updated if number of test jobs changes
```
julieqiu: Add to the comment:

> This number corresponds to the number of times GitHub Actions in .github/workflows attempt to upload coverage to Codecov. It should be updated if that number changes.

PR #4837, opened 2026-03-25
refactor(internal/librarian/python): use HasRESTNumericEnums

julieqiu:
Python generation now uses sdk.yaml as the source of truth for rest-numeric-enums, as is the case for all other languages.

This change replaces that custom logic with a call to HasRESTNumericEnums, making Python consistent with the other languages. Logic for populating rest numeric enums values using tool/cmd/migrate is removed.


jskeet:
I'll check in the morning whether this introduces any changes during migration. (Even if it does, we can hopefully ignore them.) I know that for transports some APIs were inconsistent between languahes. The same may be true for this flag.

PR #4849, opened 2026-03-26
refactor(internal/config): add LibrarianYAML const for config filename

julieqiu:
Add a config.LibrarianYAML constant and use it everywhere, so that the filename is defined in one place.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -18,6 +18,9 @@ package config
 
 //go:generate go run -tags configdocgen ../../cmd/config_doc_generate.go -input . -output ../../doc/config-schema.md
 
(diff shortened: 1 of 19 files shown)


internal/config/config.go
```diff
@@ -18,6 +18,9 @@ package config
 
 //go:generate go run -tags configdocgen ../../cmd/config_doc_generate.go -input . -output ../../doc/config-schema.md
 
+// LibrarianYAML is the filename for the librarian configuration file.
+const LibrarianYAML = "librarian.yaml"
```
JoeWang1127:
```suggestion
const (
   LibrarianYAML = "librarian.yaml"
)
```

PR #4888, opened 2026-03-28
ci(.github): add a golang integration check

JoeWang1127:
A new GitHub Actions workflow is added to verify Golang integration.

The workflow installs necessary Go tools and the librarian CLI, checks out the google-cloud-go repository, runs `librarian generate`, and executes vet checks.
It also includes a step to automatically create an issue on failure when running on the main branch.

This ensures that changes to the librarian CLI do not break the generation process for the Go client libraries.

### DIFF
--- a/.github/workflows/golang.yaml
+++ b/.github/workflows/golang.yaml
@@ -28,8 +28,8 @@ jobs:
       - name: Install go tools
         run: |
           go install tool
-          # TODO(https://github.com/googleapis/librarian/issues/4894), Group static checks
-          # in golangci-lint
+          # TODO(https://github.com/googleapis/librarian/issues/4894), Use golangci-lint
+          # for static checkers and linters.
           # we need to install golint and staticcheck because
           # the vet check uses this tool.
           go install golang.org/x/lint/golint@latest


.github/workflows/golang.yaml
```diff
@@ -0,0 +1,56 @@
...
+        run: librarian generate --all
+      - name: Run vet check
+        working-directory: google-cloud-go
+        run: ./.github/workflows/vet.sh
+  create-issue-on-failure:
+    needs: [integration]
+    if: ${{ always() && contains(needs.*.result, 'failure') && github.event_name == 'push' && github.ref == 'refs/heads/main' }}
+    uses: ./.github/workflows/create-issue-on-failure.yaml
+    with:
+      language: Golang
```
julieqiu: nit: just "Go"

we use golang in the codebase because "go" is a special keyword in the language

.github/workflows/golang.yaml
```diff
@@ -0,0 +1,52 @@
...
+  integration:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/setup-go@v6
+        with:
+          go-version-file: "go.mod"
+      - name: Install go tools
+        run: |
+          go tool install
```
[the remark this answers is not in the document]
JoeWang1127: Done.

JoeWang1127:
> Add a commit message

Added a message.

.github/workflows/golang.yaml
```diff
@@ -0,0 +1,55 @@
...
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/setup-go@v6
+        with:
+          go-version-file: "go.mod"
+      - name: Install go tools
+        run: |
+          go install tool
+          go install golang.org/x/lint/golint@latest
+          go install honnef.co/go/tools/cmd/staticcheck@latest
```
julieqiu: what's the reason for adding these?

          go install golang.org/x/lint/golint@latest
          go install honnef.co/go/tools/cmd/staticcheck@latest

they should run as part of golangci-lint

julieqiu: file an issue and add a TODO here for context

julieqiu (approved):

.github/workflows/golang.yaml
```diff
@@ -0,0 +1,58 @@
...
+        uses: actions/checkout@v6
+        with:
+          repository: googleapis/google-cloud-go
+          path: google-cloud-go
+      - name: Run librarian generate
+        working-directory: google-cloud-go
+        run: librarian generate --all
+        # The vet.sh script is a comprehensive CI/CD utility in google-cloud-go that enforces code
+        # quality by running static analysis tools. We use this check to validate code generated
+        # by librarian.
```
julieqiu: I assume that we want to get rid of this eventually and standardize on the same static analysis tools across both repos. File a TODO for now and add a comment here

PR #4906, opened 2026-03-30
feat(internal/librarianops): register upgrade command

julieqiu:
The upgradeCommand function was implemented but never added to the Commands list in Run. This registers it so that `librarianops upgrade` works.


jskeet (approved):
I'm happy this change does what it says on the tin - are we happy that the command itself still works appropriately? (I know librarianops is in a bit of an interim state...)

Issue #4881, opened 2026-03-27
rust: starting crates at 0.0.0 not working as expected


coryan:
Onboarding a new library to Rust creates a crate starting a 0.0.0.

I believe the intent was for somebody to take manual action to bump this to 1.0.0, but that is not happening. We just released `visionai-v1` at 0.0.0:

https://crates.io/crates/google-cloud-visionai-v1/versions

And we also release other libraries at 0.0.0 and 0.1.0:

https://crates.io/crates/google-cloud-vectorsearch-v1/versions

Maybe this is a matter of updating the playbooks. Or maybe we need the tools to stop a release for a GAPIC at 0.0.0.... Or maybe we should start the GAPICs at 1.0.0, as sidekick used to do.

noahdietz:
Here is my understanding:

`librarian add` does not populate an initial `version` for any language.

The documented workflow involves running `librarian bump` at some point after `add`/`generate`. In this case, that _did_ happen - `visionai` was added in https://github.com/googleapis/google-cloud-rust/commit/00157125f56394f3ec7456ec1e87bda1420e1424, followed by  a `bump` some time later in https://github.com/googleapis/google-cloud-rust/commit/e3c0870ef6d40eacf6ee050e806445d0d0921b97, but that `bump` did not touch the `visionai` library whenI think it should have. That would've put it at `0.1.0`, which is currently the _general_ default version: https://github.com/googleapis/librarian/blob/9ff1dca0e8aa929cdc705ff44d5e480a2b9e67f6/internal/librarian/bump.go#L38

We can definitely change the initial version used for Rust to `1.0.0` if that's what was done before - #4886.

The more important part here is that `bump` didn't take the `0.0.0` to `0.1.0`.

My initial guess is that it has something to do with the `bump` logic looking for libraries that have changed (by comparing Library.Output to changed file paths) since the last tag (tags only used for releases, like bookmarks). Perhaps a brand new library that didn't appear in the last tag (i.e. all files are "new") is not captured by the "files changed since last tag". That's where I'm starting.

noahdietz:
> bump only has any effect if the files have changed since the last release, and explicitly excludes new files.

The `bump` code called for Rust calls `FilesChangedSince`: https://github.com/googleapis/librarian/blob/a52d7fbc903c5d66fcaffef908544139d4ddd5f3/internal/librarian/bump.go#L408-L414

[...]

PR #4923, opened 2026-03-30
feat: librarian add sets first version for Rust

noahdietz:
`librarian add` now sets the `version` on the `Library` config to the initial, default version for Rust libraries, and provides a mechanism for other languages to also implement `librarian add` "middleware".

For Rust, this means that the first `librarian generate` will utilize this `version` in the first `Cargo.toml` generation, which means the first release will be `1.0.0`, as desired.

Note: This may seem trivial right now, but it could be expanded in the future to do things like reject APIs or API versions that a language doesn't support e.g. Rust/Dart don't generate alpha/beta API versions.

Towards #4881

### DIFF
--- a/internal/librarian/fake.go
+++ b/internal/librarian/fake.go
@@ -100,6 +100,7 @@ func fakeClean(library *config.Library) error {
 	return os.Remove(filepath.Join(library.Output, "README.md"))
 }
 
-func fakeAddLibrary(lib *config.Library, version string) {
+func fakeAdd(lib *config.Library, version string) *config.Library {
 	lib.Version = version
+	return lib
 }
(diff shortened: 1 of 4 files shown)


internal/librarian/fake.go
```diff
@@ -99,3 +99,7 @@ func fakeClean(library *config.Library) error {
 	// This function shouldn't be called if the output directory doesn't exist.
 	return os.Remove(filepath.Join(library.Output, "README.md"))
 }
+
+func fakeAddLibrary(lib *config.Library, version string) {
+	lib.Version = version
```
julieqiu: ditto re: return library

Issue #4664, opened 2026-03-19
nodejs: improve test coverage to 80%


julieqiu:
The internal/librarian/nodejs package currently has 12.72% test coverage, which is below the 80% target.

Improve test coverage for internal/librarian/nodejs/** to at least 80%.

julieqiu:
@gemini-cli /approve

julieqiu:
@gemini-cli you please create the issues for me and parent them to this issue. As of https://github.com/googleapis/librarian/commit/ec03ba51db7f076d1512ea12f2699501bc267239 you should have access to mcp_github_create_issue

julieqiu:
@gemini-cli take a look at https://github.com/googleapis/librarian/commit/ec03ba51db7f076d1512ea12f2699501bc267239 and explain to me why that did not enable mcp_github_create_issue

zhumin8:
coverage recently dropped again: https://screenshot.googleplex.com/8hSmCoW4yANGsut

PR #4977, opened 2026-04-01
feat(tool/cmd/coverage): replace codecov with go tool cover

### LINKED ISSUE #4663: tests(librarian/java): improve test coverage to 80%
The internal/librarian/java package currently has 69.24% test coverage, which is below the 80% target.

Improve test coverage for internal/librarian/java/** to at least 80%.

### LINKED ISSUE #4902: .github/workflows: replace codecov with go tool cover
We currently use Codecov to enforce coverage targets, but it has been slow and unreliable enough that we had to disable it as a presubmit check (https://github.com/googleapis/librarian/issues/4222).

Codecov's numbers also do not match what Go's own toolchain reports. Codecov converts Go's block-level coverage profiles into line-level coverage, and treats "partial" lines (where some blocks on a line are hit and others are not) as uncovered, while Go's `go tool cover` counts any executed statement as covered. 

This creates a gap that makes it hard for contributors to predict CI results locally. Here are the numbers at https://github.com/googleapis/librarian/commit/94cc4426495908e65328b662946ea7c9e88e35fa:

```
  Component            go test    Codecov    Target
  sidekick             90.2%      (none)     80%
  surfer               88.0%      85.48%     80%
  librarian-dart       94.7%      91.30%     80%
  librarian-golang     89.6%      77.55%     75%
  librarian-java       84.5%      76.51%     74%
  librarian-nodejs     73.2%      58.94%     58%
  librarian-python     85.4%      95.42%     80%
  librarian-rust       86.6%      81.45%     80%
```

  We should replace Codecov's coverage gate with a CI step that runs `go test -coverprofile` and checks the percentage from `go tool cover -func` against per-component targets, which is effectively what Codecov does. This lets check coverage locally with a single command, and allows us to re-enable coverage as a presubmit check.

We can then use `go tool cover -html` can generate a local HTML report for inline coverage.

julieqiu:
Replace the Codecov integration with a Go tool that runs tests with -coverprofile and checks the total coverage percentage from "go tool cover -func" against per-component targets.

For https://github.com/googleapis/librarian/issues/4664
Fixes https://github.com/googleapis/librarian/issues/4663
Fixes https://github.com/googleapis/librarian/issues/4902


JoeWang1127 (approved):

suztomo:
I deleted CODECOV_TOKEN in the setting page of this repository.

PR #5024, opened 2026-04-02
feat(librarian/swift): default output directory

coryan:
Librarian needs to compute the default output directory for an API. For Swift, I am selecting the full path of the API, as we do in most languages. The `librarian.yaml` file will include a prefix, such as `generated/`.

### DIFF
--- a/internal/librarian/swift/default_output_test.go
+++ b/internal/librarian/swift/default_output_test.go
@@ -21,7 +21,7 @@ import (
 )
 
 func TestDefaultOutput(t *testing.T) {
-	for _, tt :=  range []struct {
+	for _, tt := range []struct {
 		name   string
 		api    string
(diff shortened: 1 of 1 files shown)


noahdietz (approved):

internal/librarian/swift/default_output_test.go
```diff
@@ -0,0 +1,64 @@
...
+package swift
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestDefaultOutput(t *testing.T) {
+	tests := []struct {
```
julieqiu: nit: use `for _, test := range []struct { ... }{ ... }`

coryan: Done.

PR #5031, opened 2026-04-03
feat(internal/surfer): add descriptor files flags to generate commands

### LINKED ISSUE #4645: surfer: Update sidekick to accept descriptor injection
Sidekick needs to be able to handle descriptors being injected rather than rely on protoc CLI which is not available in all environments

sarahheacock:
The --descriptor-files and --descriptor-files-to-generate flags are added to surfer to allow generating gcloud commands from pre-compiled binary FileDescriptorSet files. This bypasses the need for local source protos in some workflows. Needed for running surfer in google3

Fix #4645

### DIFF
--- a/internal/sidekick/parser/protobuf_test.go
+++ b/internal/sidekick/parser/protobuf_test.go
@@ -17,6 +17,7 @@ package parser
 import (
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


internal/surfer/surfer/surfer_test.go
```diff
@@ -54,30 +72,59 @@ func TestRun(t *testing.T) {
...
+	tmpDir := t.TempDir()
+	descFile := filepath.Join(tmpDir, "field_attributes.desc")
+
+	cmd := exec.CommandContext(t.Context(), "protoc", "-o", descFile, "--include_imports",
+		"-I", scenarioDir,
+		"-I", coreGoogleapisPath,
+		filepath.Join(scenarioDir, "field_attributes.proto"))
+	if err := cmd.Run(); err != nil {
+		t.Fatalf("Failed to generate .desc file: %v", err)
+	}
```
quirogas: Instead of using `protoc`, let's build the descriptors and add them to the test data. That way, we won't have to depend on `protoc`.

coryan: Drive-by, feel free to ignore because I may not have the context.

Why is a dependency on `protoc` so bad? Are you planning to commit the descriptors in binary format? JSON? Proto ASCII?  How do you prevent humans from editing them?

And since there are dozens of tests in `librarian` that depend on `protoc` what do you get from removing it from one test?

Issue #4843, opened 2026-03-25
nodejs: migrate+generate produces diffs for google-cloud-secretmanager


julieqiu:
Running `migrate` followed by `librarian generate` on `google-cloud-secretmanager` produces these diffs:

  **Expected**

```
  D  packages/google-cloud-secretmanager/.OwlBot.yaml (migrate removes OwlBot config)
```

  **Needs google-cloud-node cleanup**

```
  M  packages/google-cloud-secretmanager/samples/generated/v1/snippet_metadata_google.cloud.secretmanager.v1.json (version 0.1.0 → 6.1.1)
  M  packages/google-cloud-secretmanager/samples/generated/v1beta2/snippet_metadata_google.cloud.secretmanager.v1beta2.json (version 0.1.0 →
  6.1.1)
  D  packages/google-cloud-secretmanager/samples/generated/v1/snippet_metadata.google.cloud.secretmanager.v1.json (old dot-pattern superseded
   by underscore-pattern)
  M  packages/google-cloud-secretmanager/src/index.ts (copyright year 2025 → 2026)
```

These diffs exist because google-cloud-node has stale data that will be corrected on the first generate run. 
- The snippet metadata files have a placeholder version (`0.1.0`) that gets updated to the real library version (`6.1.1`).
- The dot-pattern snippet metadata file is a duplicate that existed alongside the underscore-pattern file and gets cleaned up.
- The copyright year in `src/index.ts` was behind by one year.

We should make these changes upstream in google-cloud-node so that when we migrate we do not get these diffs.

  **Needs investigation: proto changes**

```
  D  packages/google-cloud-secretmanager/protos/google/cloud/secrets/v1beta1/resources.proto
  D  packages/google-cloud-secretmanager/protos/google/cloud/secrets/v1beta1/service.proto
  M  packages/google-cloud-secretmanager/protos/protos.d.ts (+522 lines)
  M  packages/google-cloud-secretmanager/protos/protos.js (+1823 lines)
  M  packages/google-cloud-secretmanager/protos/protos.json (+228 lines)
```

  The v1beta1 protos under `google.cloud.secrets` (not `google.cloud.secretmanager`) are deleted because they are not in any API config. See https://github.com/googleapis/librarian/issues/4820 for broader context on proto diff investigation.

  **Needs investigation: code removed**

```
  M  packages/google-cloud-secretmanager/src/v1/secret_manager_service_client.ts (-118 lines)
[...]

PR #5032, opened 2026-04-03
feat(internal/librarian/nodejs): execute librarian.js from repository root

quirogas:
Modify the execution of the custom `librarian.js` post-processing script to run from the repository root instead of the individual package directory.

## Root Cause
Previously, Librarian executed `node librarian.js` with the working directory set to the package directory (e.g., `packages/google-cloud-secretmanager`). However, many existing `librarian.js` scripts were authored with the assumption they would be executed from the repository root, using hardcoded paths like:
`const filePath = 'packages/google-cloud-secretmanager/src/v1/secret_manager_service_client.ts';`

This caused the scripts to fail silently as they looked for a `packages/` directory inside their own package folder.

## Backwards Compatibility
This change is backwards compatible for the following reasons:
1. **Existing Scripts:** Most existing `librarian.js` scripts in the `google-cloud-node` monorepo already use repository-relative paths. This fix allows them to work as intended without modification.
2. **Path Resolution:** By running from the root and passing the absolute or root-relative path to the script itself, Node.js still correctly locates the script, but `process.cwd()` now points to the root, which aligns with the authoring convention for these post-processors.

## Verification
Verified with `google-cloud-secretmanager`:
- **Before Fix:** Custom path templates (`projectPath`, `secretPath`) were removed during generation because `librarian.js` failed to find the target files.
- **After Fix:** `librarian.js` successfully finds and modifies the generated files, restoring the missing handwritten code.

Updates #4843

### DIFF
--- a/internal/librarian/nodejs/generate_test.go
+++ b/internal/librarian/nodejs/generate_test.go
@@ -1036,17 +1036,16 @@ func TestRunPostProcessor_CustomScripts_RootRelativePath(t *testing.T) {
 	testhelper.RequireCommand(t, "compileProtos")
 	testhelper.RequireCommand(t, "node")
 	testhelper.RequireCommand(t, "npx")
-
 	repoRoot := t.TempDir()
 	library := &config.Library{
 		Name: "google-cloud-secretmanager",
 		Keep: []string{"librarian.js"},
 	}
+
 	outDir := filepath.Join(repoRoot, "packages", library.Name)
 	if err := os.MkdirAll(outDir, 0755); err != nil {
 		t.Fatal(err)
 	}
-
 	// This script uses a path relative to the repository root.
 	// This only works if the script is executed from repoRoot.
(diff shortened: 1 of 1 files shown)


internal/librarian/nodejs/generate_test.go
```diff
@@ -523,8 +523,8 @@ func TestRunPostProcessor_CustomScripts(t *testing.T) {
 		t.Error("expected owl-bot-staging to be removed after post-processing")
 	}
 
-	if _, err := os.Stat(filepath.Join(outDir, "librarian-ran.txt")); err != nil {
-		t.Errorf("expected librarian.js to run and create librarian-ran.txt: %v", err)
+	if _, err := os.Stat(filepath.Join(repoRoot, "librarian-ran.txt")); err != nil {
```
julieqiu: add another test case where it fails running from package directory

pearigee (approved):

julieqiu (approved):

PR #5058, opened 2026-04-03
fix(internal/librarian/java): respect distribution override and unify coordinate logic

### LINKED ISSUE #5048: java: client module artifactId special cases should derive from distribution override
The logic that derives module names from library name works common cases, but there are 43 irregular ones that do not follow this pattern.
https://github.com/googleapis/librarian/blob/f2af8417592b22abd1f12dce7ba54c535dfecec5/internal/librarian/java/postprocess.go#L135-L142

When distribution override is present, need to add logic to use artifact id from it instead.

zhumin8:
The module derivation logic is updated to prioritize the artifact ID from distribution_name_override when present. This ensures that restructuring and POM generation use the correct paths.

Also refactored to centralize coordinate management and remove repeating logic. The javaModules struct is  extended to use full coordinate objects for all module types.

Fix #5048

### DIFF
--- a/internal/librarian/java/pom.go
+++ b/internal/librarian/java/pom.go
@@ -180,12 +180,7 @@ func detectIndentation(content string, index int) string {
 // to ensure its dependency list is fully synchronized.
 func collectModules(library *config.Library, libraryDir, monorepoVersion string, metadata *repoMetadata, transports map[string]serviceconfig.Transport) ([]javaModule, error) {
 	var modules []javaModule
-	gapicCoord := deriveGAPICCoordinates(library)
-	parentCoord := coordinates{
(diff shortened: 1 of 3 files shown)


internal/librarian/java/pom.go
```diff
@@ -179,15 +179,14 @@ func detectIndentation(content string, index int) string {
 // module's POM requires a full list of all proto and gRPC dependencies
 // to ensure its dependency list is fully synchronized.
 func collectModules(library *config.Library, libraryDir, monorepoVersion string, metadata *repoMetadata, transports map[string]serviceconfig.Transport) ([]javaModule, error) {
```
JoeWang1127: nit: create a struct for the inputs.

Issue #5079, opened 2026-04-04
golang: run librarian install instead of go install statements


julieqiu:
Instead of these lines:

https://github.com/googleapis/google-cloud-go/blob/01b7e137c0fc0b34b34c724b7e1d4c4cc1d4b845/.github/workflows/generation_check.yaml#L22-L25

We should run `librarian install` before `librarian generate`, so that the tool dependencies are always installed. That way we don't need to worry about version drift.

/cc @JoeWang1127

PR #5102, opened 2026-04-05
fix(internal/librarian/golang): install required tools explicitly

JoeWang1127:
The golang Install function is updated to install specific versions of required tools individually instead of using the Go tool directive.

This approach decouples the generation tools from the project dependencies and avoids the need to run the install command from the repository root.

The golang integration test also updated to include `librarian install`.

For #5079

### DIFF
--- a/.github/workflows/golang.yaml
+++ b/.github/workflows/golang.yaml
@@ -19,6 +19,7 @@ permissions:
 jobs:
   integration:
     runs-on: ubuntu-24.04
+    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
     steps:
       - uses: actions/checkout@v6
       - uses: actions/setup-go@v6


internal/librarian/golang/install.go
```diff
@@ -16,11 +16,25 @@ package golang
...
+	if err != nil {
+		return err
+	}
+	defer os.RemoveAll(dir)
+	if err := os.WriteFile(filepath.Join(dir, "go.mod"), librarian.GoMod, 0644); err != nil {
+		return err
+	}
+	if err := os.WriteFile(filepath.Join(dir, "go.sum"), librarian.GoSum, 0644); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
JoeWang1127: Done.

internal/librarian/golang/install.go
```diff
@@ -16,11 +16,26 @@ package golang
...
+	"fmt"
+	"os"
+	"path/filepath"
 
+	"github.com/googleapis/librarian"
 	"github.com/googleapis/librarian/internal/command"
 )
 
-// Install installs Go tool dependencies defined in the go.mod tool directive.
+// Install installs Go tool dependencies defined in the librarian module's go.mod tool directive.
```
julieqiu: I think we should take a different approach here. Instead of embedding go.mod and reconstructing a module environment at runtime, we should hardcode the tool list explicitly:

```
  var tools = []string{
      "github.com/googleapis/gapic-generator-go/cmd/protoc-gen-go_gapic@v0.57.0",
      "golang.org/x/tools/cmd/goimports@v0.43.0",
      "google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.3.0",
      "google.golang.org/protobuf/cmd/protoc-gen-go@v1.36.11",
  }
```

And then iterate with `go install <tool>@<version>` for each one. This works from any directory without needing a go.mod at all.

We should also remove protoc-gen-go_gapic, protoc-gen-go-grpc, and protoc-gen-go from the go.mod tool directive. It looks like the tool directive is designed for developers working in the repo: https://go.dev/ref/mod#go-mod-file-tool

Tools like golangci-lint, yamlfmt, and goimports belong there because developers actually run go tool <name> locally.

The protoc plugins are runtime dependencies that the compiled binary installs into target repos, so they should be setup independently.

JoeWang1127: Done.

.github/workflows/golang.yaml
```diff
@@ -44,8 +43,9 @@ jobs:
           path: google-cloud-go
       - name: Run librarian generate
         working-directory: google-cloud-go
-        run: librarian generate --all
-        # TODO(https://github.com/googleapis/librarian/issues/4895): Standardize static analysi
+        run: |
+          librarian install && librarian generate --all
```
julieqiu: nit: make `librarian install` its own step so that we have visibility into how long it takes

PR #5121, opened 2026-04-06
refactor(internal/librarian/java): extract logic and struct to derive maven coordinates to file

zhumin8:
Extracts logic and structs around deriving maven coordinates for various modules within a Java library into a separate file.


internal/librarian/java/maven_coordinates.go
JoeWang1127: Use singlar name: https://github.com/googleapis/librarian/blob/main/doc/howwewritego.md?plain=1#L49

JoeWang1127 (approved):

PR #5136, opened 2026-04-06
fix(internal/librarian/java): move run owlbot.py logic to per library instead of per api 

### LINKED ISSUE #5022: java: move run owlbot.py logic to per library instead of per api
run owlbot.py should be run once for the whole library, not per api.

zhumin8:
Run owlbot.py logic should be run once per library to be efficient, not per api. Removed some test setup steps that are no longer needed, and added a test on error cases for the new postProcessLibrary.

Fix #5022

### DIFF
--- a/internal/librarian/java/postprocess_test.go
+++ b/internal/librarian/java/postprocess_test.go
@@ -385,14 +385,8 @@ func TestPostProcessLibrary_ErrorCase(t *testing.T) {
 				metadata: &repoMetadata{NamePretty: "Secret Manager"},
 			}
 			err := postProcessLibrary(t.Context(), params)
-			if test.wantErr == nil {
-				if err != nil {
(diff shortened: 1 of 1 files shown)


internal/librarian/java/postprocess_test.go
```diff
@@ -282,6 +264,148 @@ func TestCopyProtos_ErrorCase(t *testing.T) {
...
+				test.setup(t, outDir)
+			}
+			params := libraryPostProcessParams{
+				cfg:      test.cfg,
+				library:  library,
+				outDir:   outDir,
+				metadata: &repoMetadata{NamePretty: "Secret Manager"},
+			}
+			err := postProcessLibrary(t.Context(), params)
+			if test.wantErr == nil {
```
JoeWang1127: Do you need this line since these tests return error?

PR #5170, opened 2026-04-07
fix(python): only delete client-post-processing directory not scripts

jskeet:
Deletes scripts/client-post-processing and all descendant files, but not scripts/ when cleaning up after post-processing. This allows us to handle google-cloud-bigtable which is generated (or at least should be) and has a scripts/conformance.sh file.

The previous test for a handwritten library is no longer relevant, as the file wouldn't be deleted whether or not the library is fully handwritten.

### DIFF
--- a/internal/librarian/python/generate_test.go
+++ b/internal/librarian/python/generate_test.go
@@ -429,9 +429,9 @@ func TestCleanUpFilesAfterPostProcessing(t *testing.T) {
 			if _, err := os.Stat(filepath.Join(outputDir, "scripts", "client-post-processing")); !os.IsNotExist(err) {
 				t.Errorf("client-post-processing should have been removed")
 			}
-			for _, expectedFile := range test.wantFiles {
-				if _, err := os.Stat(filepath.Join(outputDir, expectedFile)); err != nil {
-					t.Errorf("unable to stat %s which should still exist", expectedFile)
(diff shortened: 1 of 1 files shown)


internal/librarian/python/generate_test.go
```diff
@@ -424,8 +426,13 @@ func TestCleanUpFilesAfterPostProcessing(t *testing.T) {
 			if _, err := os.Stat(filepath.Join(repoRoot, "owl-bot-staging")); !os.IsNotExist(err) {
 				t.Errorf("owl-bot-staging should have been removed")
 			}
-			if _, err := os.Stat(filepath.Join(outputDir, "scripts")); !os.IsNotExist(err) {
-				t.Errorf("scripts should have been removed")
+			if _, err := os.Stat(filepath.Join(outputDir, "scripts", "client-post-processing")); !os.IsNotExist(err) {
```
JoeWang1127:
```suggestion
			if _, err := os.Stat(filepath.Join(outputDir, "scripts", "client-post-processing")); !errors.Is(fs.ErrNotExist) {
```

jskeet: Discussed offline - I've filed https://github.com/googleapis/librarian/issues/5175 and will get to that tomorrow, but merge this as-is to keep it consistent.

PR #5204, opened 2026-04-08
feat(dart): modernize JSON serialization and fix related lints

kevmoo:
- Implemented hybrid toJson generation strategy in annotate.go to handle function calls and null-aware elements efficiently.

- Used _ for fromJson parameter in empty messages to fix avoid_unused_constructor_parameters lint.

- Removed avoid_unused_constructor_parameters and use_null_aware_elements ignores in generated code.

- Added unit tests for HasFields and integration test for empty message fromJson.

### DIFF
--- a/.github/workflows/dart.yaml
+++ b/.github/workflows/dart.yaml
@@ -28,7 +28,7 @@ jobs:
       - name: Install Dart SDK
         uses: dart-lang/setup-dart@e51d8e571e22473a2ddebf0ef8a2123f0ab2c02c
         with:
-          sdk: 3.6.0
+          sdk: 3.9.0
       - name: Display Dart version
         run: dart --version
       - name: Verify git is available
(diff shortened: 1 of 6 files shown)


kevmoo:
@brianquinlan – I *could* split it out into two PRs. But that sounds like a lot of tail for not a lot of ROI.

kevmoo:
@brianquinlan – should we add a test case to validate the output to validate `null`?

natebosch:
> I _could_ split it out into two PRs. But that sounds like a lot of tail for not a lot of ROI.

I haven't looked closely at the code and don't have an opinion on whether this particular PR should be split from a long-term history perspective. In general I'll remind you that the ROI equation for splitting PRs looks different from the perspective of reviewers than authors.

If there are separable changes with independent cost/benefit ratios, there is probably a more than linear improvements for the reviews if they are kept separate.

kevmoo:
Closing out in favor of:

- https://github.com/googleapis/librarian/pull/5715
- https://github.com/googleapis/librarian/pull/5716

Issue #4413, opened 2026-03-08
tool/cmd/migrate: migrate configuration for Node.js


julieqiu:
Update tool/cmd/migrate to migrate configuration for Node.js to librarian.yaml.

See go/sdk-librarian-nodejs for details.

PR #5233, opened 2026-04-08
feat(librarian/nodejs):  add additional_protos support

### LINKED ISSUE #4744: nodejs: generator removes path template helpers from client
The gapic-generator-typescript no longer emits certain path template helpers that exist in the current generated code. For example, projectPath, matchProjectFromProjectName, secretPath, matchProjectFromSecretName, secretVersionPath and their match helpers are removed. The opts.numericEnums setting is also removed. These are breaking changes for users who depend on these convenience methods.

This may be expected if the generator has changed, but we need to verify whether these removals are intentional upstream changes or a configuration issue. If intentional, the migration should happen in a controlled way.

quirogas:
Implement a robust mechanism for including additional proto dependencies during Node.js client generation. It introduces the `additional_protos` configuration field and implements a union logic with common resources.

## Implementation Details
- Added `AdditionalProtos` to the `NodejsPackage` and `NodejsAPI` config structs.
- **Default Behavior**: Librarian automatically includes `google/cloud/common_resources.proto` by default for all Node.js APIs. This provides out-of-the-box support for common path templates without requiring manual configuration for most libraries.
- **Union Logic**: Specifying `additional_protos` (at the package or API level) allows adding extra dependencies. These are appended to the default `common_resources.proto` and deduplicated.
- **Naming Consistency**: Follows the pattern established in the Java generator for naming and YAML tags.

## Verification
- Verified against `google-cloud-secretmanager`, `google-cloud-kms`, and `google-cloud-resourcemanager`.
- Functional parity achieved: `projectPath` and related helpers are correctly restored in the generated clients.
- Unit tests added in `internal/librarian/nodejs/generate_test.go` covering default, hierarchical union, and deduplication scenarios.

--- 
Fix #5208: Provides the necessary infrastructure to include `google/cloud/common_resources.proto`, which contains common path template definitions.

Fix #4744: Directly resolves the problem where the Node.js generator removes path template helpers (like `projectPath`) by ensuring the required resource definitions are available to `protoc`.

For #4404: Updates the Node.js generation logic to support these new configuration fields.

For #4413: The `migrate` tool should be updated to automatically detect `proto_library_with_info` dependencies from `BUILD.bazel` files and populate the `additional_protos` field in `librarian.yaml`.


sofisl (commented):
generally speaking one thing I'm not super clear on in Librarian is how it actually will get rid of Bazel files. But regardless, I think some test you want to set up is something to confirm that those necessary protos specified in the BUILD.bazel are actually passed in correctly. (maybe that's in another PR I assume)?

Not sure if this is more generally helpful, but this is what I mean by those extra necessary protos: [cl/582493526](http://cl/582493526). When I added compileProtos to the generation step (techncially outside of the generator, running in a wrapper script), I had to specify a bunch of these protos to actually get the code to run.

internal/librarian/nodejs/generate.go
```diff
@@ -88,6 +99,31 @@ func generateAPI(ctx context.Context, api *config.API, library *config.Library,
...
+	for _, nodejsAPI := range library.Nodejs.NodejsAPIs {
+		if nodejsAPI.Path != api.Path {
+			continue
+		}
+		// Copy explicit config
+		res.Path = nodejsAPI.Path
+		res.ExtraProtos = nodejsAPI.ExtraProtos
+		if len(res.ExtraProtos) == 0 {
+			res.ExtraProtos = []string{commonProtos}
+		}
```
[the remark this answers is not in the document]
quirogas: I've updated the behavior to overwrite the default `common_resources` if any additional protos are passed, which should prevent the intentional exclusion of the file. This makes the intention explicit and will hopefully avoid unexpected behavior. Let me know what you think :)

internal/config/language.go
```diff
@@ -675,6 +675,18 @@ type NodejsPackage struct {
...
+	NodejsAPIs []*NodejsAPI `yaml:"nodejs_apis,omitempty"`
+
 	// PackageName is the npm package name (e.g., "@google-cloud/access-approval").
 	PackageName string `yaml:"package_name,omitempty"`
 }
+
+// NodejsAPI represents configuration for a single API within a Node.js package.
+type NodejsAPI struct {
+	// ExtraProtos is a list of additional proto files to include in generation.
+	ExtraProtos []string `yaml:"extra_protos,omitempty"`
```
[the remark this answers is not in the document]
quirogas: Thank you for the suggestion. I adopted Java's naming here. It provides a better name (AdditionalProtos) and clearly conveys that these are not dependencies, but rather protos that must also be generated.

doc/config-schema.md
```diff
@@ -273,17 +273,28 @@ This document describes the schema for the librarian.yaml.
 | `rest_documentation` | string | Is the URL for the REST documentation. |
 | `rpc_documentation` | string | Is the URL for the RPC documentation. |
 
+## NodejsAPI Configuration
+
+| Field | Type | Description |
+| :--- | :--- | :--- |
+| `additional_protos` | list of string | Is a list of additional proto files to include in generation. |
+| `path` | string | Is the source path. |
+| `skip_default_additional_protos` | bool | Determines whether to exclude the default additional protos (e.g. common_resources.proto). |
```
sofisl: if additional_protos is included, does this *have* to be true? If so, should we add that inclusion? Or if additional protos is included does this automatically flip to true? Not clear on the behavior.

sofisl: We should also include here why you would want to use this flag, sounds like we want to avoid duplicate protos.


If that's the case, open question for the group: should this linting occur at a higher level (i.e., wherever the new place is to define the global protos) vs. a "cleanup" function at the generator level?

sofisl (approved):

doc/config-schema.md
```diff
@@ -273,17 +273,28 @@ This document describes the schema for the librarian.yaml.
...
+| `additional_protos` | list of string | Is a list of additional proto files to include in generation. This can be overridden at the API level. |
 | `bundle_config` | string | Is the path to a GAPIC bundle config file. |
 | `dependencies` | map[string]string | Maps npm package names to version constraints. |
 | `extra_protoc_parameters` | list of string | Is a list of extra parameters to pass to protoc. |
 | `handwritten_layer` | bool | Indicates the library has a handwritten layer on top of the generated code. |
 | `main_service` | string | Is the name of the main service for libraries with a handwritten layer. |
 | `mixins` | string | Controls mixin behavior (e.g., "none" to disable). |
+| `nodejs_apis` | list of [NodejsAPI](#nodejsapi-configuration) (optional) | Is a list of Node.js-specific API configurations. |
 | `package_name` | string | Is the npm package name (e.g., "@google-cloud/access-approval"). |
+| `skip_default_additional_protos` | bool | Determines whether to exclude the default additional protos (e.g. common_resources.proto). This can be overridden at the API level. |
```
quirogas: TODO: remove this flag and document why we would want or not want to add it.

doc/config-schema.md
```diff
@@ -273,17 +273,28 @@ This document describes the schema for the librarian.yaml.
 | `rest_documentation` | string | Is the URL for the REST documentation. |
 | `rpc_documentation` | string | Is the URL for the RPC documentation. |
 
+## NodejsAPI Configuration
+
+| Field | Type | Description |
+| :--- | :--- | :--- |
+| `additional_protos` | list of string | Is a list of additional proto files to include in generation. |
```
quirogas, line 280: TODO: change the behaviour of adding protos to create a union of the commont_resource_protos

doc/config-schema.md
quirogas: c.c @sofisl

doc/config-schema.md
quirogas: c.c. @sofisl

doc/config-schema.md
quirogas: Done

PR #5234, opened 2026-04-08
test(internal/librarian/golang): add context to errors and improve test coverage

### LINKED ISSUE #4234: golang: improve test coverage for function `generate`
https://github.com/googleapis/librarian/blob/4988face06bc5f6b1aaeb2a9fc04f1cb76438bec/internal/librarian/golang/generate.go#L60

sofisl:
These changes improve error traceability and test coverage for the Go client library generation process. To create these tests, I ran `go tool cover -func=coverage.out | grep "Generate"` to figure out which tests to add. Initially, this had a 64% test coverage. With these new tests, we're at 73.8%.

This PR also adds context for the errors to report so that we can assert specific failures and error messages.

Fixes #4234


sofisl:
We should wait until #5200 is merged.

sofisl:
@JoeWang1127, a few questions/concerns:
1. I didn't pursue full coverage or else we'd either need to restructure a lot of things (like function calls to do dependency injection), or we'd be testing implementation details.
2. I did a best guess of what better test coverage would look like by using untested lines, but if you had something specific you were looking for, please lmk.
3. I added some better error formatting to better test some of the errors. This could also be removed if you want.

quirogas (approved):

internal/librarian/golang/generate_test.go
```diff
@@ -67,6 +67,47 @@ func TestGenerate(t *testing.T) {
...
+					{
+						ClientPackage: "secretmanager",
+						ImportPath:    "secretmanager-multi/apiv1beta2",
+						Path:          "google/cloud/secretmanager/v1beta2",
+					},
+				},
+			},
+		},
+		{
+			Name:          "secretmanager-delete",
```
JoeWang1127: Did you add this test case to verify the path in `DeleteGenerationOutputPaths` is deleted after the generation?

We have this test to verify the same thing: https://github.com/googleapis/librarian/blob/8be5b7f18f7afaa052fac23525492ca905db63fe/internal/librarian/golang/generate_test.go#L314

internal/librarian/golang/generate_test.go
```diff
@@ -67,6 +67,47 @@ func TestGenerate(t *testing.T) {
 				},
 			},
 		},
+		{
+			Name:          "secretmanager-multi",
```
JoeWang1127: Did you add this test case to verify a library with multiple APIs can be generated? If you, you can move this tests case to https://github.com/googleapis/librarian/blob/8be5b7f18f7afaa052fac23525492ca905db63fe/internal/librarian/golang/generate_test.go#L255

The validation logic is more appropriate than this test.

internal/librarian/golang/generate_test.go
```diff
@@ -179,6 +220,38 @@ func TestGenerate_Error(t *testing.T) {
...
+		Output:        filePath,
+		APIs: []*config.API{
+			{
+				Path: "google/cloud/secretmanager/v1",
+			},
+		},
+	}
+
+	gotErr := Generate(t.Context(), library, &sources.Sources{Googleapis: googleapisDir})
+	if gotErr == nil {
```
JoeWang1127: Could you check the error?

JoeWang1127:
> 1. I didn't pursue full coverage or else we'd either need to restructure a lot of things (like function calls to do dependency injection), or we'd be testing implementation details.

Makes sense.

> I did a best guess of what better test coverage would look like by using untested lines, but if you had something specific you were looking for, please lmk.

I think the change LGTM.

> I added some better error formatting to better test some of the errors. This could also be removed if you want.

LGTM.

internal/librarian/golang/generate_test.go
sofisl: actually this test confirms more explicitly that the README was generated/exists (again, just going by lines not covered). What I ended up doing is just combining the two tests, i.e., just adding the assertion that the README exists in this test you pointed out!

internal/librarian/golang/generate_test.go
sofisl: done!

JoeWang1127 (approved):

PR #5247, opened 2026-04-09
fix(internal/librarian/java): update client, parent and BOM pom.xml only when new modules added

### LINKED ISSUE #5244: java: parent and BOM pom.xml are generated with diff in order
BOM and parent pom.xml have managed sections for dependencies and modules, these are always updated now. 
The generate results are correct, but may have different ordering than what's in google-cloud-java.

Change logic to only update these sections if needed (new grpc- proto- modules added), this also avoids unnecessary I/O and slightly improves efficiency.
Alternatively, could update migrate logic to "pre-sort" these sections in a one-time-change. This makes the migrate logic more complicated.

zhumin8:
Only update client, parent and BOM pom.xml if there are new grpc- proto- modules added.

Fix #5244

### DIFF
--- a/internal/librarian/java/pom.go
+++ b/internal/librarian/java/pom.go
@@ -84,10 +84,10 @@ type javaModule struct {
 	template     string
 }
 
-// syncPOMs generates missing proto-*, grpc-*, and client POMs, and surgically updates
-// existing client library POMs to include new dependencies.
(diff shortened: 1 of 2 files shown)


JoeWang1127 (commented):
What are the difference between `isProtoMissing`/`isGRPCMissing` in module and `anyMissingProtoGrpc`?

JoeWang1127 (approved):

Issue #4834, opened 2026-03-25
google-cloud-go: serviceconfig: normalize release level vocabulary across languages


julieqiu:
Go and other languages use different terminology for release levels. Go's generator expects "ga", "alpha", and "beta", while other languages use "stable" and "preview". 

We should normalize release level in google-cloud-go to use a consistent vocabulary (e.g., "stable" and "preview") and have each language's generator map from the normalized value to whatever its tooling expects.

noahdietz:
I'm going to take a look at this one!

PR #5276, opened 2026-04-09
feat: derive release level

### LINKED ISSUE #4860: serviceconfig: default release level is in conflict with new library release version
We default release level to stable in [api.go](https://github.com/googleapis/librarian/blob/cec5d8b73019bdf85d7bd0843a74f84897b641c9/internal/serviceconfig/api.go#L162-L167), but when releasing a new library, it defaults to v0.1.0 in [bump](https://github.com/googleapis/librarian/blob/cec5d8b73019bdf85d7bd0843a74f84897b641c9/internal/librarian/bump.go#L38). This is conflicting.

If we "discover" each new API request and added to sdk.yaml with release-level set to "preview". We should be fine. Otherwise, we are again forced to update sdk.yaml for a new API onboarding.

In Java "release level" is not used in actual code generation, however it is used to generate a badge and a banner in README.md (e.g. [java-chat/README.md](https://github.com/googleapis/google-cloud-java/blob/main/java-chat/README.md)).

Curious if any other languages use "release level" for more than repo-metadata?

noahdietz:
Adds a heuristic for deriving release level of an API path for a library based on the channel in the API path and the version of the library so that we no longer need to manually set the release level in sdk.yaml. Also removes the `release_levels` that were identified as extraneous (i.e. they match the result of applying the heuristic). 

Regenerating Rust with this change only updates the `showcase` testing library to be marked as `preview`, and doing the same for Go produces the [following diffs](https://github.com/noahdietz/google-cloud-go/commit/a887987631a170c5159aa0ae1679ff96d0944406), which all seem reasonable based on either their current released version being `0.x` or the pkg being alpha or beta.

Keeps the ability to manually override as there are some edge cases in Go that we likely need to push towards 1.x and thus `ga` OR just drop the `beta` flag if the module is already 1.x (#5303).

Fixes #4860


internal/librarian/golang/generate.go
```diff
@@ -68,7 +68,7 @@ func Generate(ctx context.Context, library *config.Library, srcs *sources.Source
 		if goAPI == nil {
 			return fmt.Errorf("error finding goAPI associated with API %s: %w", api.Path, errGoAPINotFound)
 		}
-		if err := generateAPI(ctx, goAPI, googleapisDir, tempDir); err != nil {
+		if err := generateAPI(ctx, goAPI, googleapisDir, library.Version, tempDir); err != nil {
```
noahdietz: I think we should just change this to load the release level ahead of time and pass it through this argument rather than pass the version through, which is a bit indirect.

noahdietz: For example
```suggestion
        releaseLevel := sc.ReleaseLevel(config.LanguageGo, library.Version)
		if err := generateAPI(ctx, goAPI, googleapisDir, releaseLevel, tempDir); err != nil {
```

jskeet:
Drive-by comment, based on the Gemini comment: should we have an enum for this, and then a mapping from the enum to language-specific value, rather than sticking to strings?

noahdietz:
> should we have an enum for this, and then a mapping from the enum to language-specific value, rather than sticking to strings?

@jskeet do you mean some constants for these values e.g. `type ReleaseLevel string` `const Preview  ReleaseLevel = "preview"` and `const Stable ReleaseLevel = "stable"` rather than using "magic strings"? Yeah I can do that. Would be cleaner too if i end up precalculating release level and passing it through the call stack - the type being `ReleaseLevel` is more self-describing.

Non-sequitur: I need to double check the sdk.yaml changes, I suspect there wasn't enough context to make an appropriate determination of if the value was correct or not, need to verify.

jskeet:
> > should we have an enum for this, and then a mapping from the enum to language-specific value, rather than sticking to strings?
> 
> @jskeet do you mean some constants for these values e.g. `type ReleaseLevel string` `const Preview ReleaseLevel = "preview"` and `const Stable ReleaseLevel = "stable"` rather than using "magic strings"? Yeah I can do that. Would be cleaner too if i end up precalculating release level and passing it through the call stack - the type being `ReleaseLevel` is more self-describing.

I'd meant an actual non-string enum, which could then be mapped to string on a per language basis. Or maybe we could do the consts as you describe, and just have a separate "map of ReleaseLevel to string" for Go...

internal/serviceconfig/api.go
```diff
@@ -144,20 +144,38 @@ func (api *API) HasRESTNumericEnums(language string) bool {
...
+		if isAlpha {
+			return "alpha"
+		}
+		if isBeta {
+			return "beta"
+		}
+		if level == "stable" {
+			return "ga"
+		}
 	}
```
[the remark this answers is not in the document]
noahdietz: good catch, fixed.

internal/librarian/golang/generate.go
noahdietz: On second thought, this is fine. For Rust this isn't a lot of plumbing and for Go we need to load all of the same information in buildGAPICOpts already, so we don't save anything by pulling it out.

internal/librarian/python/generate_test.go
```diff
@@ -1111,7 +1111,8 @@ func TestCreateRepoMetadata(t *testing.T) {
 		{
 			name: "non-cloud API",
 			library: &config.Library{
-				Name: "google-apps-meet",
+				Name:    "google-apps-meet",
+				Version: "0.4.0",
```
noahdietz, line 1115: This was previously set explicitly to `preview` in sdk.yaml, as in reality it is a pre-1.0 client should be marked as such, but the test didn't include a version so the heuristic fell back on the API path being a stable channel.

noahdietz:
> I'd meant an actual non-string enum, which could then be mapped to string on a per language basis. Or maybe we could do the consts as you describe, and just have a separate "map of ReleaseLevel to string" for Go...

@jskeet I think I see what you are saying, but I'd like to punt on it. I think the language based uniqueness goes away once we get go on the `preview`/`stable` keywords (https://github.com/googleapis/librarian/issues/4834) and the code is cleaner as a result. That shouldn't be hard to do and should be something we do soon as part of Go stabilization.

JoeWang1127 (approved):

jskeet (approved):

PR #5298, opened 2026-04-10
feat(sidekick/rust): inject `ResourceName` into gRPC request extensions in templates

haphungw:
Update the gRPC client templates to inject `ResourceName` into `tonic::Extensions` (or `RequestOptions`) so that the gRPC tracing middleware can extract it and populate `gcp.resource.destination.id` in spans.
This ensures that both unary and streaming RPCs correctly propagate the resource name to the transport layer.


haphungw:
This approach has been replaced by https://github.com/googleapis/google-cloud-rust/pull/5368

PR #5315, opened 2026-04-11
fix(internal/librarian/java): remove hardcoded common_resources.proto

### LINKED ISSUE #5314: java: common proto is added without condition
The common proto (`google/cloud/common_resources.proto`) is added to proto list when running protoc command to generate GAPIC client:
https://github.com/googleapis/librarian/blob/1c5d095e718734786880da9f05ea7cfee1d761e1/internal/librarian/java/generate.go#L288-L306

The common proto should be added only if this proto is in `proto_library_with_info` target in BUILD.bazel.

JoeWang1127:
The hardcoded addition of `google/cloud/common_resources.proto` to all Java APIs is removed from the generation logic. 

Previously, ResolveJavaAPI would automatically add this proto to AdditionalProtos if it was not already present.

Fixes #5314


internal/librarian/java/generate.go
zhumin8: nit: update this comment.

zhumin8 (approved):

internal/librarian/java/generate.go
JoeWang1127: Done.

Issue #5307, opened 2026-04-11
java: extra sample files are generated


zhumin8:
Current config defaults to generate sample when not specified. e.g. observed in bigquerystorage
BUILD.bazel field for sample is optional and only generated on `true`.
Need to verify if missing cases are treated right.

Also, negative config `NoSamples` is hard to process. Change to `Samples *bool`

PR #5346, opened 2026-04-13
refactor(internal/config): replace no_samples with optional samples field

zhumin8:
Replaces the no_samples boolean field with an optional samples pointer-based boolean across the configuration schema and Java generation logic.

The previous `no_samples` field was counter-intuitive. Moving to samples *bool improves config clarity, while allows a default. 

For #5307

### DIFF
--- a/internal/librarian/java/generate.go
+++ b/internal/librarian/java/generate.go
@@ -99,7 +99,7 @@ func generateAPI(ctx context.Context, cfg *config.Config, api *config.API, libra
 		outDir:         outdir,
 		version:        version,
 		googleapisDir:  googleapisDir,
-		includeSamples: !javaAPI.NoSamples,
+		includeSamples: javaAPI.Samples == nil || *javaAPI.Samples,
 	}
 	gapicDir := p.gapicDir()
 	gRPCDir := p.gRPCDir()
--- a/internal/librarian/java/generate_test.go
+++ b/internal/librarian/java/generate_test.go
@@ -256,7 +256,7 @@ func TestResolveJavaAPI(t *testing.T) {
 						{
 							Path:             "google/cloud/secretmanager/v1",
 							AdditionalProtos: []string{"other.proto"},
-							Samples:          boolPtr(false),
+							Samples:          new(false),
 						},
 					},
 				},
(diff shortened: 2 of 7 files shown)


internal/librarian/java/generate.go
```diff
@@ -99,7 +99,7 @@ func generateAPI(ctx context.Context, cfg *config.Config, api *config.API, libra
 		outDir:         outdir,
 		version:        version,
 		googleapisDir:  googleapisDir,
-		includeSamples: !javaAPI.NoSamples,
+		includeSamples: javaAPI.Samples == nil || *javaAPI.Samples,
```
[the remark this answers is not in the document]
zhumin8: Will deal with this in followups, when changing default behavior

internal/librarian/java/generate_test.go
```diff
@@ -265,7 +265,7 @@ func TestResolveJavaAPI(t *testing.T) {
 			want: &config.JavaAPI{
 				Path:             "google/cloud/secretmanager/v1",
 				AdditionalProtos: []string{"other.proto"},
-				NoSamples:        true,
+				Samples:          boolPtr(false),
```
noahdietz: In Go 1.26 you can use `new` to  initialize a pointer with a primitive value `new(true)`

https://go.dev/doc/go1.26#language

noahdietz: So we don't need these boolPtr functions

noahdietz: I think you missed this one `new(false)`
```suggestion
				Samples:          new(false),
```

internal/librarian/java/generate_test.go
```diff
@@ -672,3 +672,7 @@ func TestCollectJavaFiles(t *testing.T) {
 		t.Errorf("mismatch (-want +got):\n%s", diff)
 	}
 }
+
+func boolPtr(b bool) *bool {
+	return &b
+}
```
noahdietz: ditto, not necessary
```suggestion
```

zhumin8: ah! thanks.

noahdietz (approved):

PR #5358, opened 2026-04-13
refactor(internal/librarian/golang): generate in temp directory instead of output directory

### LINKED ISSUE #4777: golang: generate APIs in a temp directory
This still seems to be expecting that we'll be generating within the library's output directory. I thought we were going to use the repo root or a temp directory? (I think it's worth keeping the two pretty separate.)

_Originally posted by @jskeet in https://github.com/googleapis/librarian/pull/4577#discussion_r2976042701_

### LINKED ISSUE #5301: golang: fix failure with cross-linking in the background when creating temp directory
(from @noahdietz when running `librarian generate`): I'm having trouble regenerating things for Go with Librarian at HEAD due to an issue in what seems to be temp directory moves? 
rename /tmp/librarian-gen-165204932/cloud.google.com/go/accessapproval/apiv1/access_approval_client.go /usr/local/google/home/ndietz/golang/google-cloud-go/accessapproval/apiv1/access_approval_client.go: invalid cross-device link

Reverted https://github.com/googleapis/librarian/pull/5302, will need to re-implement with the fix linked below.

sofisl:
Refactors the Go client library generation process to use a temporary directory for protoc operations. This isolates intermediate generation artifacts from the repository, improving build safety and eliminating the need for manual cleanup scripts. Does so within the same filesystem to avoid the error `invalid cross-device` on different os's.

Fixes #4777
Fixes #5301

### DIFF
--- a/internal/librarian/golang/generate_test.go
+++ b/internal/librarian/golang/generate_test.go
@@ -946,7 +946,7 @@ func TestMoveGeneratedFiles(t *testing.T) {
 		t.Run(test.name, func(t *testing.T) {
 			tmpDir := t.TempDir()
 			outDir, apiDir, snippetDir, lib := test.setup(t, tmpDir)
-			err := moveGeneratedFiles(lib, lib.Go.GoAPIs[0], outDir)
+			err := moveGeneratedFiles(lib, lib.Go.GoAPIs[0], outDir, outDir)
 			if err != nil {
 				t.Fatal(err)
 			}
--- a/internal/librarian/golang/generate.go
+++ b/internal/librarian/golang/generate.go
@@ -41,7 +41,7 @@ var (
 )
 
 // Generate generates a Go client library.
-func Generate(ctx context.Context, library *config.Library, srcs *sources.Sources) error {
+func Generate(ctx context.Context, library *config.Library, srcs *sources.Sources) (err error) {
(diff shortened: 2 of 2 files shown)


internal/librarian/golang/generate_test.go
```diff
@@ -946,7 +946,7 @@ func TestMoveGeneratedFiles(t *testing.T) {
 		t.Run(test.name, func(t *testing.T) {
 			tmpDir := t.TempDir()
 			outDir, apiDir, snippetDir, lib := test.setup(t, tmpDir)
-			err := moveGeneratedFiles(lib, lib.Go.GoAPIs[0], outDir)
+			err := moveGeneratedFiles(lib, lib.Go.GoAPIs[0], outDir, outDir)
```
noahdietz, line 950: Would you mind adding a comment here to the effect of "outDir is passed as the srcDir because it contains the temporary generated output and also passed as the destination because the generated code is moved elsewhere within its directory structure"

It's just kind of weird to see `outDir, outDir` :)

noahdietz (approved):

internal/librarian/golang/generate_test.go
sofisl: yes good call!

PR #5368, opened 2026-04-14
refactor(internal/librarian): remove rust_storage language hack

### LINKED ISSUE #4291: rust: remove "rust_storage" language enum
From @julieqiu in https://github.com/googleapis/librarian/pull/4287#discussion_r2872332781_

The rust_storage langauge is a hack used for special handling of the storage client; however, its not really a different language we are generating for.

 Let's clean it up so that the language enum is purely languages.

jameslynnwu:
The rust_storage language was a hack used for special handling of the storage client in the Rust generator. Since it is not a different language, this change removes it from the Language enum and transitions to a template-based detection mechanism (Template == "storage") for specialized generation logic.

Additionally, the redundant Language field in RustModule is removed from the configuration as it always defaulted to LanguageRust. Tests and documentation are updated accordingly.

Fixes https://github.com/googleapis/librarian/issues/4291


noahdietz (approved):

dbolduc (approved):
Thanks!

PR #5374, opened 2026-04-14
fix(internal/surfer): track target vs current for real APIs and decouple unit tests

### LINKED ISSUE #5373: surfer: move test apis into golden directory
Move test apis into golden directory so we can compare against target output

sarahheacock:
Golden file tests are updated to track target versus current generated outputs for real APIs including parallelstore, iam, and seclm. This allows verifying that Librarian output matches the target baseline or identifying intentional deviations. Unit tests in surfer_test.go are decoupled from checked-in production configurations by using dynamically generated temporary files for setup. Blockchainvalidator removed since it is not available in googleapis

Fixes https://github.com/googleapis/librarian/issues/5373


internal/surfer/gcloud/command_writer_test.go
quirogas, line 1: Why are we removing the test for validating the YAML schema?

internal/surfer/gcloud/command_writer_test.go
```diff
@@ -81,60 +78,6 @@ func TestWritePartialFile_Output(t *testing.T) {
...
-				t.Fatal(err)
-			}
-
-			if diff := cmp.Diff(commands, roundTripped); diff != "" {
-				t.Errorf("mismatch (-want +got):\n%s", diff)
-			}
-		})
-	}
-}
-
```
quirogas, line 137: Nit: If we are deleting tests, I think it would be helpful to understand the reasoning. I'm assuming this removal is intended to prioritize golden tests? Will we lose coverage or signaling if we break the schema?

internal/surfer/gcloud/provider/loader.go
```diff
@@ -68,5 +75,22 @@ func ReadGcloudConfig(path string) (*Config, error) {
...
+		if p.Type == "" {
+			return fmt.Errorf("resource_patterns[%d].type is required", i)
+		}
+		if len(p.Patterns) == 0 {
+			return fmt.Errorf("resource_patterns[%d].patterns must not be empty", i)
+		}
+	}
+
+	return nil
+}
```
quirogas, line 96: I have two thoughts here:

1. What if we implemented a Validate() error method directly on the Config struct and its
     nested types?  It would move the validation logic where the data is defined. 
2. should we also validate ServiceName, API.Name, and Selector strings?

quirogas (approved):
LGTM!

internal/surfer/gcloud/command_writer_test.go
sarahheacock: Correct, this was redundant of the golden tests and was a little confusing. It took me a minute to understand that it was looking for missing fields from just the command schema. This was a good starting point but the golden tests are a better fit because they verify that the actual generated output (which implies a correct schema), rather than just checking if we can parse and serialize a specific API file.

internal/surfer/gcloud/command_writer_test.go
sarahheacock: Same answer as below

internal/surfer/gcloud/provider/loader.go
sarahheacock: I think those are good ideas. I think you're right we'll need ServiceName, API.Name, selector. However, given the slight direction change of the project where we might be mass onboarding, I didn't want to add any premature validation for the time being

Issue #5380, opened 2026-04-14
java: generate diff for aiplatform


zhumin8:
resolve generate failures and issues/diffs with aiplatform,

1. migrate did not include additional protos for it in librarian.yaml
2. missing files in generation result.
3. generation diffs: manual files 
      - java-aiplatform/google-cloud-aiplatform/src/main/java/com/google/cloud/aiplatform/util/ValueConverter.java
      - java-aiplatform/google-cloud-aiplatform/src/test/java/com/google/cloud/aiplatform/util/ValueConverterTest.java
4. removed test files: e.g. java-aiplatform/google-cloud-aiplatform/src/test/java/com/google/cloud/aiplatform/util/ValueConverterTest.java 
5. extra proto files

zhumin8:
Here are remaining diffs:
```
java-aiplatform/google-cloud-aiplatform/src/main/java/com/google/cloud/aiplatform/util/ValueConverter.java
java-aiplatform/google-cloud-aiplatform/src/main/resources/META-INF/native-image/com.google.cloud.aiplatform.v1beta1/reflect-config.json
java-aiplatform/google-cloud-aiplatform/src/test/java/com/google/cloud/aiplatform/util/ValueConverterTest.java
java-aiplatform/google-cloud-aiplatform/src/test/java/com/google/cloud/location/MockLocations.java
java-aiplatform/google-cloud-aiplatform/src/test/java/com/google/cloud/location/MockLocationsImpl.java
java-aiplatform/google-cloud-aiplatform/src/test/java/com/google/iam/v1/MockIAMPolicy.java
java-aiplatform/google-cloud-aiplatform/src/test/java/com/google/iam/v1/MockIAMPolicyImpl.java
java-aiplatform/proto-google-cloud-aiplatform-v1beta1/src/main/proto/google/cloud/aiplatform/v1beta1/schema/annotation_payload.proto
java-aiplatform/proto-google-cloud-aiplatform-v1beta1/src/main/proto/google/cloud/aiplatform/v1beta1/schema/annotation_spec_color.proto
java-aiplatform/proto-google-cloud-aiplatform-v1beta1/src/main/proto/google/cloud/aiplatform/v1beta1/schema/data_item_payload.proto
java-aiplatform/proto-google-cloud-aiplatform-v1beta1/src/main/proto/google/cloud/aiplatform/v1beta1/schema/dataset_metadata.proto
[...]

PR #5381, opened 2026-04-14
fix(tool/cmd/migrate): handle is variable when parsing for additional protos

zhumin8:
Handle more complex case whre deps is variable when parsing for additional protos for Java migration.

 The deps attribute of the proto_library_with_info rule uses a list addition (+) with the variable _PROTO_SUBPACKAGE_DEPS (see [source](https://github.com/googleapis/googleapis/blob/fc968708d27f9aeb9f1c50a3e32d53810df843fd/google/cloud/aiplatform/v1/BUILD.bazel#L178-L186)).  Because of this structure, the previous implementation using AttrStrings("deps") failed to see any of the dependencies because it only handles simple, literal string lists. The updated implementation now walks the entire expression and collects all literal strings it finds.

For #5380

### DIFF
--- a/tool/cmd/migrate/java.go
+++ b/tool/cmd/migrate/java.go
@@ -22,6 +22,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/bazelbuild/buildtools/build"
 	"github.com/googleapis/librarian/internal/config"
 	"github.com/googleapis/librarian/internal/librarian"
 	"github.com/googleapis/librarian/internal/librarian/java"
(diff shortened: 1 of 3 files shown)


tool/cmd/migrate/java.go
```diff
@@ -635,3 +638,14 @@ func getLineIndent(line string) string {
 	trimmed := strings.TrimLeft(line, " \t")
 	return line[:len(line)-len(trimmed)]
 }
+
+// extractStrings returns all string literals found within a Bazel expression.
```
sofisl, line 642: perhaps unrelated, but what about using this go package for parsing BUILD files? https://pkg.go.dev/github.com/bazelbuild/buildtools/build (I guess more so for my own learning)

sofisl (approved):

tool/cmd/migrate/java.go
zhumin8: We do use this package, if you look at https://github.com/zhumin8/librarian/blob/95dfd89b6640550e5fd1afd0fdf6e2304b1a194c/tool/cmd/migrate/java.go#L57
In this particular case, I did not found fit function (also did not look deeper because it's only a few lines of code + migrate code is short term). `build.Strings(attr)` is too strict for this case.

PR #5382, opened 2026-04-14
chore(tool/cmd/migrate): add flag for Java insert markers

zhumin8:
Puts Java insert markers behind flag. Now default run `migrate .` only creates librarian.yaml. 
To also insert markers to pom.xml, use `migrate -insert-markers . `

This is a convenience feature for development, to make our life easier when testing migrate repeatively.


tool/cmd/migrate/java.go
```diff
@@ -146,7 +146,7 @@ type GenerationConfig struct {
 	Libraries           []LibraryConfig `yaml:"libraries"`
 }
 
-func runJavaMigration(ctx context.Context, repoPath string) error {
+func runJavaMigration(ctx context.Context, repoPath string, insert bool) error {
```
jameslynnwu, line 149: can we rename insert to something specific to inserting markers?

`shouldInsertMarkers bool`?

jameslynnwu (approved):

PR #5389, opened 2026-04-14
fix(internal/librarian/java): recursively gather protos and sort

### LINKED ISSUE #4198: java: determine if sort needed for proto files
In java hermetic_build generate, it sorts proto files binary order with `LC_COLLATE=C sort`. This is likely to ensure gapic_metadata.json has a deterministic and correct service order.
Code for ref: https://github.com/googleapis/sdk-platform-java/blob/d9b39ed775795c44b1fd778d767f06aa706f6ee9/hermetic_build/library_generation/generate_library.sh#L123

when comparing diff for migration, find concrete case if this sort is needed (likely yes) and implement similar logic to librarian code.

zhumin8:
Recursively gather protos for generate, to include protos in nested dirs, sort to maintain deterministic order.

For #5380
Fix #4198

### DIFF
--- a/internal/librarian/java/generate.go
+++ b/internal/librarian/java/generate.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"sort"
 	"strings"
 
 	"github.com/googleapis/librarian/internal/command"
@@ -29,6 +30,12 @@ import (
 	"github.com/googleapis/librarian/internal/serviceconfig"
 	"github.com/googleapis/librarian/internal/sources"
 )
+
+var nonRecursivePaths = map[string]bool{
+	"google/api":   true,
+	"google/cloud": true,
+	"google/rpc":   true,
+}
 
 var (
 	errExtractVersion    = errors.New("failed to extract version")
@@ -111,11 +118,7 @@ func generateAPI(ctx context.Context, cfg *config.Config, api *config.API, libra
 	}
 
 	apiDir := filepath.Join(googleapisDir, api.Path)
-	// TODO(https://github.com/googleapis/librarian/issues/4198):
-	// Consider recursive gathering and explicit sorting
-	// of proto files to match the behavior of the hermetic build, ensuring
-	// a deterministic order in the generated gapic_metadata.json.
-	apiProtos, err := filepath.Glob(apiDir + "/*.proto")
+	apiProtos, err := gatherProtos(apiDir, api.Path)
 	if err != nil {
 		return fmt.Errorf("failed to find protos: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


internal/librarian/java/generate.go
```diff
@@ -305,3 +308,37 @@ func findBOMVersion(cfg *config.Config) (string, error) {
...
+		if err != nil {
+			return err
+		}
+		if d.IsDir() {
+			if !recursive && path != root {
+				return filepath.SkipDir
+			}
+			return nil
+		}
+		if d.Type().IsRegular() && filepath.Ext(path) == ".proto" {
```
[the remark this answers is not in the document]
zhumin8: I don't think we use symlinks for proto files.

internal/librarian/java/generate_test.go
```diff
@@ -749,3 +749,76 @@ func TestCollectJavaFiles(t *testing.T) {
...
+			relPath: "google/cloud/location",
+			want: []string{
+				filepath.Join(tmpDir, "google/cloud/location/locations.proto"),
+				filepath.Join(tmpDir, "google/cloud/location/sub/sub.proto"),
+			},
+		},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			root := filepath.Join(tmpDir, test.relPath)
+			if test.relPath == "google/cloud/aiplatform/v1" {
```
JoeWang1127: How about create some proto files within `google/cloud/aiplatform/v1` so that you don't need this if condition.

internal/librarian/java/generate.go
```diff
@@ -305,3 +309,37 @@ func findBOMVersion(cfg *config.Config) (string, error) {
 	}
 	return "", errBOMVersionMissing
 }
+
+// gatherProtos returns a sorted list of proto files in the given root directory,
+// ensuring that subpackage protos (e.g., in a "schema" directory) are included
```
JoeWang1127, line 309: Why do you need to mention `schema` in the comment?

internal/librarian/java/generate.go
```diff
@@ -305,3 +309,37 @@ func findBOMVersion(cfg *config.Config) (string, error) {
...
+// in the generation.
+//
+// recursion is disabled for certain base paths (google/api, google/cloud, google/rpc).
```
JoeWang1127:
```suggestion
// recursion is disabled for certain base paths in nonRecursivePaths.
```

quirogas (approved):

internal/librarian/java/generate_test.go
zhumin8: done

internal/librarian/java/generate.go
zhumin8: as example.

JoeWang1127 (approved):

Issue #5385, opened 2026-04-14
java: generate unversioned API


JoeWang1127:
The error message:
```
2026/04/14 15:25:08 librarian: generate library "gsuite-addons" (java): failed to generate api "google/apps/script/type": google/apps/script/type: failed to extract version
```

PR #5391, opened 2026-04-15
feat(internal/librarian/java): support non-versioned APIs

JoeWang1127:
The Java generator is updated to support APIs that do not follow standard versioning patterns (e.g., v1, v2).
If a version cannot be extracted from the API path, the generator now defaults to using the base name of the path as the version.

Additionally, the logic for determining when to include gRPC modules in the generated POM files is refined.
gRPC modules are now skipped if the API is configured as `proto-only` or if the transport is restricted to `rest`.

Comprehensive unit tests are added to verify these changes and ensure correct module collection under various `transport` and `proto-only` configurations.

For #5385


internal/librarian/java/generate.go
```diff
@@ -88,7 +87,7 @@ func Generate(ctx context.Context, cfg *config.Config, library *config.Library,
 func generateAPI(ctx context.Context, cfg *config.Config, api *config.API, library *config.Library, googleapisDir, outdir string, metadata *repoMetadata, apiCfg *serviceconfig.API) error {
 	version := serviceconfig.ExtractVersion(api.Path)
 	if version == "" {
-		return fmt.Errorf("%s: %w", api.Path, errExtractVersion)
+		version = filepath.Base(api.Path)
```
zhumin8: nit: do we want to rename it to something else?

zhumin8 (approved):

internal/librarian/java/generate.go
JoeWang1127: Makes sense.

Since we use the base directory name, updated the name to `apiBase` and simplify the logic.

PR #5392, opened 2026-04-15
feat(internal/librarian/java): support config to exclude protos

zhumin8:
Supports a new JavaAPI.ExcludedProtos config to exclude listed proto files from generation. There are a few cases where proto needs to be excluded, they are previously hardcoded with hermetic build in [here](https://github.com/googleapis/google-cloud-java/blob/5208fc94f69790dbe2ba65ab2685ad9674314b6c/sdk-platform-java/hermetic_build/library_generation/generate_library.sh#L141).

For #5380

### DIFF
--- a/internal/librarian/java/generate.go
+++ b/internal/librarian/java/generate.go
@@ -345,12 +345,12 @@ func filterProtos(fullPaths []string, relExcludes []string, root string) []strin
 	if len(relExcludes) == 0 {
 		return fullPaths
 	}
-	excludedSet := make(map[string]bool)
+	excludedSet := make(map[string]bool, len(relExcludes))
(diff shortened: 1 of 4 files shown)


internal/librarian/java/generate.go
```diff
@@ -338,3 +339,23 @@ func gatherProtos(root, relPath string) ([]string, error) {
 	sort.Strings(protos)
 	return protos, nil
 }
+
+// filterProtos removes entries from fullPaths that match root + relPath in relExcludes.
```
suztomo: `removes` implies this function modifies the list in the argument. Rather this `returns` a filtered list.

suztomo (approved):

PR #5410, opened 2026-04-15
chore(internal/serviceconfig): update release level for java-workstations

JoeWang1127:
Update the release level to `stable` for java-workstations.

The API is v1 but the library version is <1.0.0 so we need to override the default value.


noahdietz (approved):

noahdietz:
> The API is v1 but the library version is <1.0.0 so we need to override the default value.

That's unfortunate :)

PR #5411, opened 2026-04-15
fix(internal/librarian/golang): add title override for README generation

### LINKED ISSUE #5035: golang: README.md changed after onboarding `maps/geocode/v4`
The title in the maps README.md changed after onboarding `maps/geocode/v4`: https://github.com/googleapis/google-cloud-go/pull/14305/changes#diff-51ae4d5709430de920883fdb84dc88c2811e31e6f474b321552819775699af10L5

The reason for this change is this API has the highest major version so it's title is used to generate the README.md.

jameslynnwu:
The title in the maps README.md changed after onboarding maps/geocode/v4 because the APIs are sorted with the highest version first, and the first API's title is used for the README. This is problematic for libraries like Maps that do not have a single root API and instead group multiple independent services.

To address this, a language-neutral title_override field is added to the top-level Library configuration in librarian.yaml.

The Go README generator is refactored to run outside the API loop and handle the fallback to the first API's title internally if title_override is not set.

Fixes https://github.com/googleapis/librarian/issues/5035


internal/librarian/golang/generate.go
```diff
@@ -304,8 +304,21 @@ func generateREADME(library *config.Library, api *serviceconfig.API, moduleRoot
 	if err != nil {
 		return err
 	}
+	title := library.TitleOverride
+	if title == "" {
+		if len(library.APIs) == 0 {
+			f.Close()
+			return fmt.Errorf("no APIs in library %s", library.Name)
+		}
+		api, err := serviceconfig.Find(googleapisDir, library.APIs[0].Path, config.LanguageGo)
```
jameslynnwu: I think this refactor is cleaner than the `if i != 0` check in Generate that only generated a readme for the first API.

internal/librarian/golang/generate_test.go
```diff
@@ -500,6 +496,35 @@ func TestGenerateREADME(t *testing.T) {
...
+	s := string(content)
+	if !strings.Contains(s, "Custom Title") {
+		t.Errorf("want overridden title in README, got:\n%s", s)
+	}
+	if strings.Contains(s, "Secret Manager API") {
+		t.Errorf("did not want original title in README, got:\n%s", s)
+	}
+}
+
 func TestGenerateREADME_Skipped(t *testing.T) {
```
JoeWang1127, line 528: Could you update this test to a table-driven test and add a test case to verify a library without title will skip generating README?

jameslynnwu: fixed

internal/librarian/golang/generate_test.go
```diff
@@ -500,29 +496,74 @@ func TestGenerateREADME(t *testing.T) {
...
+			},
+			fallbackTitle: "Secret Manager API",
+		},
+		{
+			name: "skipped because no title",
+			library: &config.Library{
+				Name: "secretmanager",
+				APIs: []*config.API{{Path: "google/cloud/secretmanager/v1"}},
+			},
+			fallbackTitle: "",
```
JoeWang1127, line 549: nit: remove this line.

JoeWang1127 (approved):

Issue #4559, opened 2026-03-13
java: refactor clean logic


zhumin8:
Should re-evaluate clean logic implemented, and consider reference what Python and Go doing and only clean "generated files". 
Complexity is, unlike python and go, we probably need to search for one of to identify generated files.
`// AUTO-GENERATED DOCUMENTATION AND CLASS.` 
 `// Generated by the protocol buffer compiler.  DO NOT EDIT!`
`@io.grpc.stub.annotations.GrpcGenerated` .

zhumin8:
Will update clean to marker based for gapic modules, it will only wipe out .java files with generated markers + 2 json files that are generated. 
Benefit of this approach:
- Avoid need of extensive keep configs in librarian.yaml. A lot of these are not cleaned today in hermetic build, and not obvious in config yamls. 
- Avoid need of extra step for developers working on these veneer libraries, when they need to add an extra file.

Cost: time difference of scanning files before clean up is neglect-able to running time. Verified with running aiplatform.

For now, I am planning applying marker based clean for gapic modules. For grpc- and proto- module and samples/snippets/generated, will delete all files according to path. These are far less likely to mix  with handwritten files. I spot checked with aiplatform, datastore, logging, speech, bigquerystorage.

PR #5418, opened 2026-04-15
fix(internal/librarian/java): clean to use derived module names

### LINKED ISSUE #5417: java: clean logic should use derived module names
Clean logic should use derived module names from internal/librarian/java/maven_coordinate.go, instead of library names and hardcoded guess for proto- and grpc- modules.

zhumin8:
Change clean to derive clean patterns with `DeriveAPICoordinates` so that it uses same source as generate. Also changed clean scope to /src dirs under each module, to be consistent with that owlbot.yaml does today.

Fix #5417

### DIFF
--- a/internal/librarian/java/clean.go
+++ b/internal/librarian/java/clean.go
@@ -51,21 +51,21 @@ func Clean(library *config.Library) error {
 }
 
 func cleanPatterns(library *config.Library) map[string]bool {
-	lc := DeriveLibraryCoordinates(library)
+	libraryCoordinates := DeriveLibraryCoordinates(library)
 	patterns := map[string]bool{
-		filepath.Join(lc.GAPIC.ArtifactID, "src"):         true,
-		filepath.Join("samples", "snippets", "generated"): true,
-		".repo-metadata.json":                             true,
+		filepath.Join(libraryCoordinates.GAPIC.ArtifactID, "src"): true,
+		filepath.Join("samples", "snippets", "generated"):         true,
+		".repo-metadata.json": true,
 	}
 	for _, api := range library.APIs {
 		javaAPI := ResolveJavaAPI(library, api)
 		version := filepath.Base(api.Path)
-		ac := DeriveAPICoordinates(lc, version, javaAPI)
-		if ac.Proto.ArtifactID != "" {
-			patterns[filepath.Join(ac.Proto.ArtifactID, "src")] = true
+		apiCoordinates := DeriveAPICoordinates(libraryCoordinates, version, javaAPI)
+		if apiCoordinates.Proto.ArtifactID != "" {
+			patterns[filepath.Join(apiCoordinates.Proto.ArtifactID, "src")] = true
 		}
-		if ac.GRPC.ArtifactID != "" {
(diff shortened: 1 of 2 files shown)


internal/librarian/java/clean.go
```diff
@@ -58,6 +50,27 @@ func Clean(library *config.Library) error {
...
+		ac := DeriveAPICoordinates(lc, version, javaAPI)
+		if ac.Proto.ArtifactID != "" {
+			patterns[filepath.Join(ac.Proto.ArtifactID, "src")] = true
+		}
+		if ac.GRPC.ArtifactID != "" {
+			patterns[filepath.Join(ac.GRPC.ArtifactID, "src")] = true
+		}
+	}
+	return patterns
+}
```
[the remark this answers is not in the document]
zhumin8: This is to keep consistent with existing from owlbot.yaml, see template [here](https://github.com/googleapis/google-cloud-java/blob/main/sdk-platform-java/hermetic_build/library_generation/templates/owlbot.yaml.monorepo.j2).
However, this **may** change (https://github.com/googleapis/librarian/issues/4559). I want to get this update in and test out.

internal/librarian/java/clean.go
```diff
@@ -58,6 +50,27 @@ func Clean(library *config.Library) error {
 	return nil
 }
 
+func cleanPatterns(library *config.Library) map[string]bool {
+	lc := DeriveLibraryCoordinates(library)
```
JoeWang1127:
```suggestion
	libCoordinates := DeriveLibraryCoordinates(library)
```

internal/librarian/java/clean.go
```diff
@@ -58,6 +50,27 @@ func Clean(library *config.Library) error {
...
+	patterns := map[string]bool{
+		filepath.Join(lc.GAPIC.ArtifactID, "src"):         true,
+		filepath.Join("samples", "snippets", "generated"): true,
+		".repo-metadata.json":                             true,
+	}
+	for _, api := range library.APIs {
+		javaAPI := ResolveJavaAPI(library, api)
+		version := filepath.Base(api.Path)
+		ac := DeriveAPICoordinates(lc, version, javaAPI)
```
JoeWang1127:
```suggestion
		apiCoordinates := DeriveAPICoordinates(lc, version, javaAPI)
```

internal/librarian/java/clean.go
```diff
@@ -58,6 +50,27 @@ func Clean(library *config.Library) error {
 	return nil
 }
 
+func cleanPatterns(library *config.Library) map[string]bool {
+	lc := DeriveLibraryCoordinates(library)
+	patterns := map[string]bool{
+		filepath.Join(lc.GAPIC.ArtifactID, "src"):         true,
```
JoeWang1127: Do we need to add this pattern if the api is proto-only?

Can we move this line to the for loop in line 60?

zhumin8: Good point. Created a followup https://github.com/googleapis/librarian/issues/5430 to check if logic is compatible with proto only case.

JoeWang1127 (approved):

Issue #5075, opened 2026-04-04
librarian: change update command to use hierarchical dot notation


julieqiu:
Today librarian update uses positional arguments:

```
librarian update googleapis discovery
```

This does not scale well as we add more update targets, such as for updating the librarian version (https://github.com/googleapis/librarian/issues/3756) and tool versions (https://github.com/googleapis/librarian/issues/4848).

We should move librarian update to use hierarchical dot notation, so the UX would be:

```
librarian update sources.googleapis
librarian update version
```

This also aligns well with the proposed config command: https://github.com/googleapis/librarian/issues/5074

JoeWang1127:
I agree with this proposal.

noahdietz:
Let's do it

Issue #5074, opened 2026-04-04
librarian: add config command


julieqiu:
Today we read the librarian version from librarian.yaml using sed in multiple language repos:

https://github.com/googleapis/google-cloud-rust/blob/c1668796234f030c09163985882f2fa7bc8bbd9d/.gcb/scripts/regenerate.sh#L38

https://github.com/googleapis/google-cloud-go/blob/01b7e137c0fc0b34b34c724b7e1d4c4cc1d4b845/.github/workflows/regenerate.sh#L25

Additionally, the only way to update values in librarian.yaml today is to open the file and edit it manually.

We should introduce a config command to make reading and writing values easier. The UX would be:

```
librarian config get version
librarian config set version vX.Y.Z
```

This should use hierarchical dot notation so it can scale to the full librarian.yaml structure. For example:

```
librarian config get sources.googleapis
librarian config set sources.googleapis.commit <commit>
librarian config set sources.googleapis.sha256 <sha256>
```

sofisl:
In talking with Julie, she recommends the following steps:

```

1. Send a PR for just getting and setting the version field, which you already have in that PR. 
2. Once that PR is in, update the language repositories to use that instead of the sed commands
3. Implement https://github.com/googleapis/librarian/issues/5075 by changing the librarian update syntax to use hierarchical dot notation (without changing the underlying functionality yet)
4. Add the functionality to librarian update that actually sets the version using the new syntax (https://github.com/googleapis/librarian/issues/3756)
5. Go back to implementing get / set for sources
```

PR #5422, opened 2026-04-16
feat(internal/librarian): add generic getter/setter for config values

sofisl:
Introduces `yaml.GetConfigValue` and `yaml.SetConfigValue` to parse dot-separated paths, will be used in `update` and in the future, `config`.
For #5074
For #5075


JoeWang1127:
> For https://github.com/googleapis/librarian/issues/5074 and https://github.com/googleapis/librarian/issues/5075

For #5074
For #5075

internal/yaml/dotnotation.go
```diff
@@ -0,0 +1,112 @@
+// Copyright 2025 Google LLC
```
JoeWang1127: nit: update copyright year to 2026

internal/yaml/dotnotation.go
```diff
@@ -0,0 +1,112 @@
...
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+// Get parses a dot-separated path within a map representation of the configuration
+// and returns the value. Returns an error if the key is not found.
+func Get(m map[string]any, path string) (any, error) {
```
JoeWang1127: Could you come up with a more descriptive name, `Get` is too generic: https://github.com/googleapis/librarian/blob/7bf0e08615d35f0c412274b6f420cca5951eea46/doc/howwewritego.md#naming-and-spelling

internal/yaml/dotnotation.go
```diff
@@ -0,0 +1,112 @@
...
+	for _, p := range parts {
+		if p == "" {
+			continue
+		}
+		switch c := current.(type) {
+		case map[string]any:
+			if val, exists := c[p]; exists {
+				current = val
+			} else {
+				return nil, fmt.Errorf("key not found: %s", p)
```
JoeWang1127: Create an error for better testing.

internal/yaml/dotnotation_test.go
```diff
@@ -0,0 +1,137 @@
+// Copyright 2025 Google LLC
```
JoeWang1127: nit: update copyright year to 2026

internal/yaml/dotnotation_test.go
```diff
@@ -0,0 +1,137 @@
...
+		},
+		{
+			path:    "libraries.0.name",
+			want:    "library-a",
+			wantErr: false,
+		},
+	} {
+		t.Run(tt.path, func(t *testing.T) {
+			got, err := Get(m, tt.path)
+			if (err != nil) != tt.wantErr {
```
JoeWang1127: Separate success cases and error cases into different tests: https://github.com/googleapis/librarian/blob/7bf0e08615d35f0c412274b6f420cca5951eea46/doc/howwewritego.md#separate-error-tests

internal/yaml/dotnotation_test.go
```diff
@@ -0,0 +1,137 @@
...
+				t.Fatalf("Get(%q) error = %v, wantErr %v", tt.path, err, tt.wantErr)
+			}
+
+			if !tt.wantErr {
```
JoeWang1127: After returning an error variable, you can check the error here.

internal/yaml/dotnotation_test.go
```diff
@@ -0,0 +1,137 @@
...
+				if diff := cmp.Diff(tt.want, got); diff != "" {
+					t.Errorf("mismatch (-want +got):\n%s", diff)
+				}
+			}
+		})
+	}
+}
+
+func TestSet(t *testing.T) {
+
```
JoeWang1127: nit: avoid empty lines.

internal/yaml/dotnotation_test.go
```diff
@@ -0,0 +1,137 @@
...
+	for _, tt := range []struct {
```
JoeWang1127:
```suggestion
	for _, test := range []struct {
```

internal/yaml/dotnotation_test.go
```diff
@@ -0,0 +1,137 @@
...
+				"commit": "abcd123",
+			},
+		},
+		"version": "v1.0.0",
+		"libraries": []any{
+			map[string]any{"name": "library-a"},
+		},
+	}
+
+	for _, tt := range []struct {
```
JoeWang1127:
```suggestion
	for _, test := range []struct {
```

internal/yaml/dotnotation_test.go
```diff
@@ -0,0 +1,137 @@
...
+		t.Run(tt.path, func(t *testing.T) {
+			m := map[string]any{
+				"version": "v1.0.0",
+				"libraries": []any{
+					map[string]any{"name": "library-a"},
+				},
+			}
+			updated, err := Set(m, tt.path, tt.value)
+			if err != nil {
+				t.Fatalf("Set(%q, %v) error = %v", tt.path, tt.value, err)
```
JoeWang1127:
```suggestion
				t.Fatal(err)
```

https://github.com/googleapis/librarian/blob/7bf0e08615d35f0c412274b6f420cca5951eea46/doc/howwewritego.md#use-tfatal-or-terror-for-simple-error-handling

internal/yaml/dotnotation.go
julieqiu: Putting Get/Set in package yaml implies that this has to do with parsing a YAML file, but it seems like we are just looking things up from a map. I would leave the functions in internal/librarian, since it's an implementation detail of the librarian update command

internal/yaml/dotnotation_test.go
```diff
@@ -0,0 +1,137 @@
...
+		"version": "v1.0.0",
+		"libraries": []any{
+			map[string]any{"name": "library-a"},
+		},
+	}
+
+	for _, tt := range []struct {
+		path    string
+		want    any
+		wantErr bool
```
julieqiu: split error cases: https://github.com/googleapis/librarian/blob/main/doc/howwewritego.md#separate-error-tests

internal/yaml/dotnotation.go
```diff
@@ -0,0 +1,112 @@
...
+	"strings"
+)
+
+// Get parses a dot-separated path within a map representation of the configuration
+// and returns the value. Returns an error if the key is not found.
+func Get(m map[string]any, path string) (any, error) {
+	parts := strings.Split(path, ".")
+	var current any = m
+
+	for _, p := range parts {
```
julieqiu: Since we only need this for version and sources.googleapis.commit right now, I'd keep it simple and just switch on those strings. If we need to expand this to more fields, reflection over `*config.Config` (driven by the existing yaml:"..." tags) is probably the right way to go.

Either way, I don't think map[string]any is the right approach, because it doesn't enforce the existing schema. For example, a user could do:

librarian config set sources.googleapis.commit false (wrong value type)
librarian config set surces.googleapis.commit (misspelled field name)

None of these would be enforced

internal/yaml/dotnotation.go
sofisl: yes you're right, makes sense!

internal/librarian/dotnotation.go
```diff
@@ -0,0 +1,66 @@
...
+	// ErrUnsupportedPath is returned when a dot-notation path is not supported.
+	ErrUnsupportedPath = errors.New("unsupported config path")
+
+	// ErrSourceNotConfigured is returned when attempting to access a source that is not configured.
+	ErrSourceNotConfigured = errors.New("source not configured")
+)
+
+// SetConfigValue sets a value at a specific path within the configuration.
+// It only supports a limited set of paths for now.
+func SetConfigValue(currentConfig *config.Config, path string, value string) error {
```
julieqiu: This doesn't need to be exported

internal/librarian/cfg_value_utility.go
```diff
@@ -0,0 +1,66 @@
...
+	ErrSourceNotConfigured = errors.New("source not configured")
+)
+
+// SetConfigValue sets a value at a specific path within the configuration.
+// It only supports a limited set of paths for now.
+func SetConfigValue(currentConfig *config.Config, path string, value string) error {
+	switch path {
+	case "version":
+		currentConfig.Version = value
+	case "sources.googleapis.commit":
```
julieqiu: When setting the commit, we should also compute sources.googleapis.sha256, since that will change

internal/librarian/cfg_value_utility.go
```diff
@@ -0,0 +1,66 @@
...
+package librarian
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/googleapis/librarian/internal/config"
+)
+
+var (
```
julieqiu, line 26: These don't need to be exported

internal/librarian/dotnotation.go
sofisl: TIL!

internal/librarian/cfg_value_utility.go
sofisl: TIL!

internal/librarian/dotnotation.go
```diff
@@ -0,0 +1,81 @@
...
+	case "version":
+		currentConfig.Version = value
+	case "sources.googleapis.commit":
+		if currentConfig.Sources == nil {
+			currentConfig.Sources = &config.Sources{}
+		}
+		if currentConfig.Sources.Googleapis == nil {
+			currentConfig.Sources.Googleapis = &config.Source{}
+		}
+		currentConfig.Sources.Googleapis.Commit = value
```
julieqiu: if `fetch.LatestCommitAndChecksum` fails, currentConfig shouldn't be mutated

also add a test case for this

internal/librarian/dotnotation.go
```diff
@@ -0,0 +1,81 @@
...
+	// errUnsupportedPath is returned when a dot-notation path is not supported.
+	errUnsupportedPath = errors.New("unsupported config path")
+
+	// errSourceNotConfigured is returned when attempting to access a source that is not configured.
+	errSourceNotConfigured = errors.New("source not configured")
+)
+
+// setConfigValue sets a value at a specific path within the configuration.
+// It only supports a limited set of paths for now.
+func setConfigValue(currentConfig *config.Config, path string, value string) error {
```
julieqiu: rename `currentConfig` to `cfg` for consistency in other parts of the codebase, and return the mutated config, so that it is obvious the config changed in this function

internal/librarian/dotnotation.go
```diff
@@ -0,0 +1,81 @@
...
+		endpoints := &fetch.Endpoints{
+			API:      githubAPI,
+			Download: githubDownload,
+		}
+		repo := &fetch.RepoRef{
+			Org:    "googleapis",
+			Name:   "googleapis",
+			Branch: value,
+		}
+		_, sha256, err := fetch.LatestCommitAndChecksum(endpoints, repo)
```
julieqiu: return the sha256 corresponding to the commit, not the latest one

internal/librarian/cfg_value_utility.go
```diff
@@ -0,0 +1,81 @@
+// Copyright 2026 Google LLC
```
julieqiu, line 1: nit: dotnotation.go is a weird name for this file, since it doesn't match the contents after the refactor

internal/librarian/dotnotation.go
```diff
@@ -0,0 +1,81 @@
...
+		}
+		if currentConfig.Sources.Googleapis == nil {
+			currentConfig.Sources.Googleapis = &config.Source{}
+		}
+		currentConfig.Sources.Googleapis.Commit = value
+		endpoints := &fetch.Endpoints{
+			API:      githubAPI,
+			Download: githubDownload,
+		}
+		repo := &fetch.RepoRef{
```
julieqiu: Reuse this: https://github.com/googleapis/librarian/blob/d80707b00074c891ffd240324b3096beb87e8000/internal/librarian/update.go#L34

internal/librarian/dotnotation.go
```diff
@@ -0,0 +1,81 @@
...
+			Org:    "googleapis",
+			Name:   "googleapis",
+			Branch: value,
+		}
+		_, sha256, err := fetch.LatestCommitAndChecksum(endpoints, repo)
+		if err != nil {
+			return fmt.Errorf("failed to fetch checksum for commit %s: %w", value, err)
```
julieqiu: nit: value could be a ref here instead of a commit, just use "fetching checksum for %s: %w"

julieqiu (changes requested):

internal/fetch/fetch.go
```diff
@@ -145,6 +145,12 @@ func LatestCommitAndChecksum(endpoints *Endpoints, repo *RepoRef) (commit, sha25
 	return commit, sha256, nil
 }
 
+// CommitChecksum fetches the SHA256 of the tarball for a specific commit.
+func CommitChecksum(endpoints *Endpoints, repo *RepoRef, sha string) (string, error) {
```
julieqiu: just pass in endpoints.Download, since that is the only field that is used

internal/librarian/cfg_value_utility.go
```diff
@@ -0,0 +1,79 @@
...
+	case "version":
+		cfg.Version = value
+	case "sources.googleapis.commit":
+		endpoints := &fetch.Endpoints{
+			API:      githubAPI,
+			Download: githubDownload,
+		}
+		repo := sourceRepos["googleapis"]
+		repo.Branch = value
+		sha256, err := fetch.CommitChecksum(endpoints, &repo, value)
```
julieqiu: it looks like you are passing in repo, so I don't think you need to pass in value as well, since `value` is the same as `repo.Branch`

internal/fetch/fetch.go
julieqiu: nit: just name this Checksum, since that's the only checksum of the commit is the only checksum that we could be fetching

julieqiu: I don't think `sha` is the right argument name, since the `sha256` is what this function should be returning

internal/librarian/cfg_value_utility.go
```diff
@@ -0,0 +1,79 @@
...
+	errSourceNotConfigured = errors.New("source not configured")
+)
+
+// setConfigValue sets a value at a specific path within the configuration.
+// It only supports a limited set of paths for now.
+func setConfigValue(cfg *config.Config, path string, value string) (*config.Config, error) {
+	switch path {
+	case "version":
+		cfg.Version = value
+	case "sources.googleapis.commit":
```
julieqiu: We want to support setting sources for any Source, so this should include all of these: https://github.com/googleapis/librarian/blob/1a6932227f882279cd862c5214460ee2d5c6b49a/internal/config/config.go#L91

Feel free to add a TODO if you prefer to do that as a follow up

internal/librarian/cfg_value_utility_test.go
```diff
@@ -0,0 +1,265 @@
...
+			_, err := getConfigValue(currentConfig, test.path)
+			if !errors.Is(err, test.wantErr) {
+				t.Errorf("getConfigValue(%q) error = %v, wantErr %v", test.path, err, test.wantErr)
+			}
+		})
+	}
+}
+
+func TestSetConfigValue(t *testing.T) {
+	// Start a mock server
```
julieqiu: nit: you don't need this comment, since it is clear from the code below

internal/librarian/cfg_value_utility_test.go
```diff
@@ -0,0 +1,265 @@
...
+
+	// Override endpoints
+	oldAPI := githubAPI
+	oldDownload := githubDownload
+	githubAPI = ts.URL
+	githubDownload = ts.URL
+	defer func() {
+		githubAPI = oldAPI
+		githubDownload = oldDownload
+	}()
```
julieqiu, line 135: Move this block inside t.Run, so that you don't need to modify global variables

internal/librarian/cfg_value_utility_test.go
```diff
@@ -0,0 +1,265 @@
...
+		t.Run(test.path, func(t *testing.T) {
+			currentConfig := &config.Config{
+				Version: "v1.0.0",
+				Sources: &config.Sources{
+					Googleapis: &config.Source{
+						Commit: "abcd123",
+					},
+				},
+			}
+			gotCfg, err := setConfigValue(currentConfig, test.path, test.value)
```
julieqiu: nit: just use `got`, since you are only comparing one value

internal/librarian/cfg_value_utility_test.go
```diff
@@ -0,0 +1,265 @@
...
+					Googleapis: &config.Source{
+						Commit: "xyz789",
+						SHA256: dummySHA,
+					},
+				},
+			},
+		},
+	} {
+		t.Run(test.path, func(t *testing.T) {
+			currentConfig := &config.Config{
```
julieqiu: nit: use `cfg` since there is only one config

internal/librarian/cfg_value_utility_test.go
```diff
@@ -0,0 +1,265 @@
...
+
+			if diff := cmp.Diff(test.want, gotCfg); diff != "" {
+				t.Errorf("mismatch (-want +got):\n%s", diff)
+			}
+		})
+	}
+}
+
+func TestSetConfigValue_Error(t *testing.T) {
+	currentConfig := &config.Config{
```
julieqiu: Move this inside t.Run, since it can be modified every time `setConfigValue` is called

internal/librarian/cfg_value_utility_test.go
```diff
@@ -0,0 +1,265 @@
...
+			wantErr: errUnsupportedPath,
+		},
+		{
+			name:    "unsupported path",
+			path:    "unknown.field",
+			wantErr: errUnsupportedPath,
+		},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			gotCfg, err := setConfigValue(currentConfig, test.path, "some-value")
```
julieqiu: nit: just use got, since you are only comparing one value

internal/librarian/cfg_value_utility_test.go
```diff
@@ -0,0 +1,265 @@
...
+		if gotCfg != nil {
+			t.Errorf("expected nil config on failure, got %v", gotCfg)
+		}
+
+		if diff := cmp.Diff(originalConfig, currentConfig); diff != "" {
+			t.Errorf("currentConfig was mutated on failure (-want +got):\n%s", diff)
+		}
+	})
+
+	t.Run("nil sources remains nil", func(t *testing.T) {
```
julieqiu: split this into a separate test, since these two t.Run functions aren't related to each other

julieqiu (changes requested):

internal/librarian/cfg_value_utility.go
sofisl: I implemented it, but I wasn't sure what to do in the other cases (I have the behavior we want for `commit`, but not `branch`, `dir`, etc. LMK if this looks wrong to you.

internal/fetch/fetch.go
```diff
@@ -137,14 +137,19 @@ func LatestCommitAndChecksum(endpoints *Endpoints, repo *RepoRef) (commit, sha25
...
-	sha256, err = urlSha256(tarballURL)
+	sha256, err = Checksum(endpoints.Download, repo, commit)
 	if err != nil {
 		return "", "", err
 	}
 	return commit, sha256, nil
 }
 
+// Checksum fetches the SHA256 of the tarball for a specific commit.
+func Checksum(githubDownload string, repo *RepoRef, sha256 string) (string, error) {
```
julieqiu, line 148: add a test

internal/librarian/cfg_value_utility.go
julieqiu: This doesn't look quite right. It seems like you can only set sources.googleapis.commit.

See https://github.com/googleapis/google-cloud-rust/blob/c290aedec4413a575db3a0b38df86f622466b2a7/librarian.yaml#L17-L33 for an example of what these sources look like.

julieqiu (changes requested):

internal/librarian/cfg_value_utility.go
sofisl: ok, I changed a bit and added this .yaml file as a test case.

sofisl:
Closing in favor of https://github.com/googleapis/librarian/pull/5500.

see updated steps in #5074

PR #5423, opened 2026-04-16
feat(internal/librarian/config): introduce config command for independent settings

### LINKED ISSUE #5074: librarian: add config command

sofisl:
Adds `librarian config get` and `librarian config set` subcommands.
Allows dynamic read/write modifications directly on the config file.

Fixes #5074


sofisl:
Review #5422  first

internal/yaml/dotnotation.go
```diff
@@ -0,0 +1,79 @@
...
+				current = val
+			} else {
+				return nil, fmt.Errorf("key not found: %s", p)
+			}
+		} else {
+			return nil, fmt.Errorf("key not found: %s", p)
+		}
+	}
+	return current, nil
+}
```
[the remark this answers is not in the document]
sofisl: this should be done in #5422, will address there

internal/librarian/config_command.go
```diff
@@ -0,0 +1,100 @@
+// Copyright 2025 Google LLC
```
julieqiu, line 1: nit: just use config.go for the file name, since all of the other filenames are the command name

internal/librarian/config_command.go
```diff
@@ -0,0 +1,100 @@
...
+					if path == "" {
+						return fmt.Errorf("path is required")
+					}
+
+					m, err := yaml.Read[map[string]any](config.LibrarianYAML)
+					if err != nil {
+						return err
+					}
+
+					val, err := yaml.Get(*m, path)
```
julieqiu, line 48: See comments on https://github.com/googleapis/librarian/pull/5422

Issue #5163, opened 2026-04-07
python: improve API onboarding with "librarian add"


jskeet:
For a new library: 

- Set the version to 0.0.0 automatically (although eventually we'll need a better solution so we can make the first release 1.0.0 for GA libraries)
- Create a "python" section and set "default_version" to the last element of the API path
- Potentially modify legacylibrarian's state.yaml file as well

For a new API in an existing library:

- Detect that it *is* in an existing library
- Add an new API in the APIs section

jskeet:
Current plan around python-gapic-namespace, python-gapic-name, and warehouse-package-name. These will be in separate PRs, but will need Librarian releases (or pseudo-versions) so we can anchor on the behavioral changes

- Locally modify `tidy` to fill opt_args_by_api for *all* APIs based on current GAPIC generator default behavior, and commit that change to librarian.yaml. (Checking that regenerating everything causes no diffs.)
  - This is done in https://github.com/googleapis/google-cloud-python/pull/16762 which *does* have diffs, but they're *only* corrections to the package name in comments etc. The actual code hasn't changed.
- Modify `generate` to always provide these options, inferring them in a new way (see below) if they're not present.
- Modify `tidy` to remove values which match the inferred ones. (This can be used for transport as well.)
- Modify `add` to validate that the inferred namespace/name conform with expectations of approved namespaces

**New inference**

- Infer gapic-namespace as "the first two segments of the API path, with the / replaced by dot" (e.g. google.cloud)
- Infer gapic-name as "the rest of the API path, without any trailing version segment, with / replaced by _" (e.g. datacatalog_lineage_configmanagement)

jskeet:
Remaining tasks related to this:

- Update internal documentation for Python team
- Consider nested packages - captured in #5669, so we can close this once the above task is done

jskeet:
Internal documentation has been updated.

PR #5424, opened 2026-04-16
feat(internal/librarian/python): populate attributes when adding new library

jskeet:
Set Library.Version and (where appropriate)
Library.PythonPackage.DefaultVersion.

More work will be potentially be required, setting GAPIC options, which can be addressed separately after more research.

Towards #5163

### DIFF
--- a/internal/librarian/python/add.go
+++ b/internal/librarian/python/add.go
@@ -25,7 +25,7 @@ import (
 // This is set on the initial `librarian add` for a new API.
 const defaultVersion = "0.0.0"
 
-var errNewLibraryMustHaveOneAPI = errors.New("a newly added library must have exactly one API")
+var errNewLibraryMustHaveOneAPI = errors.New("a newly added library (in Python) must have exactly one API so that the default version can be populated")
 
 // Add initializes a new Python library with default values.
 func Add(lib *config.Library) (*config.Library, error) {
--- a/internal/librarian/python/add.go
+++ b/internal/librarian/python/add.go
@@ -1,0 +1,43 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


internal/librarian/python/add.go
```diff
@@ -0,0 +1,45 @@
...
+	lib.Version = defaultVersion
+	if len(lib.APIs) != 1 {
+		return nil, errNewLibraryMustHaveOneAPI
+	}
+	defaultVersion := serviceconfig.ExtractVersion(lib.APIs[0].Path)
+	if defaultVersion != "" {
+		lib.Python = &config.PythonPackage{
+			DefaultVersion: defaultVersion,
+		}
+	}
```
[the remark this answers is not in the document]
jskeet: Fixed.

internal/librarian/python/add.go
```diff
@@ -0,0 +1,42 @@
...
+// defaultVersion is the first version used for a new library.
+// This is set on the initial `librarian add` for a new API.
+const defaultVersion = "0.0.0"
+
+var errNewLibraryMustHaveOneAPI = errors.New("a newly added library must have exactly one API")
+
+// Add initializes a new Python library with default values.
+func Add(lib *config.Library) (*config.Library, error) {
+	lib.Version = defaultVersion
+	if len(lib.APIs) != 1 {
```
zhumin8, line 33: I suppose this should be a requirement for all languages? If so, can we do this check in internal/librarian/add.go instead?

jskeet: No, I suspect most languages can add multiple API paths for a new library at the same time. But Python needs to pick which will be the "default version" from the first API path added.

internal/librarian/python/add.go
```diff
@@ -0,0 +1,42 @@
...
+
+	"github.com/googleapis/librarian/internal/config"
+	"github.com/googleapis/librarian/internal/serviceconfig"
+)
+
+// defaultVersion is the first version used for a new library.
+// This is set on the initial `librarian add` for a new API.
+const defaultVersion = "0.0.0"
+
+var errNewLibraryMustHaveOneAPI = errors.New("a newly added library must have exactly one API")
```
zhumin8: nit: it might be useful to add explanation (that this is due to default version) to the error message.

zhumin8 (approved):

internal/librarian/python/add.go
jskeet: Done.

PR #5437, opened 2026-04-16
fix(internal/librarian/rust): update version bump to handle renamed dependencies

### LINKED ISSUE #5365: rust: bump versions can miss `wkt`
`wkt` is the one crate that still has two names in our top level Cargo manifest. (https://github.com/googleapis/google-cloud-rust/issues/4598)

When we try to bump its version using sidekick, it will only bump the full name of `google_cloud_wkt`. This can lead to minimal version failures.

---

A quick hack (if we don't decide to fix the root of the issue) would be to have this

https://github.com/googleapis/librarian/blob/28721cc6a0af2390473a017e53067523498a93b3/internal/librarian/rust/update_manifest.go#L76

... also pass for `"wkt"` if the crate name is `"google-cloud-wkt"`

jameslynnwu:
The version bump logic for Rust libraries is updated to correctly identify and update dependencies that have been renamed in the workspace manifest using the package field.

Previously, the logic only checked if the line started with the crate name, missing cases where the crate was aliased (like wkt for google-cloud-wkt). The new approach uses regular expressions to check for both direct names and the package field, while also ensuring we do not match commented-out lines.

Fixes https://github.com/googleapis/librarian/issues/5365


internal/librarian/rust/update_manifest_test.go
```diff
@@ -170,6 +170,24 @@ func TestUpdateWorkspaceVersion(t *testing.T) {
 			crateName: "test-crate",
 			want:      "[workspace.dependencies]\nother-crate = { version = \"1.0.0\", path = \"src/other-crate\" }\n",
 		},
+		{
```
sofisl, line 173: nit: with regex I add a test for "not-want"

sofisl (approved):

internal/librarian/rust/update_manifest.go
```diff
@@ -75,13 +75,10 @@ func updateWorkspaceVersion(path, crateName string, newVersion semver.Version) e
 	}
 	lines := strings.Split(string(contents), "\n")
 	updated := false
+	crateRegex := regexp.MustCompile(fmt.Sprintf(`^\s*%s\s*=`, regexp.QuoteMeta(crateName)))
```
julieqiu: Use string matching with strings.HasPrefix instead of regex, since it is more transparent and easier to understand that regex

julieqiu (dismissed):

internal/librarian/rust/update_manifest.go
jameslynnwu: I think pulling the existing crate check and new package check into functions makes things more clear. 

Did you also want to convert the isPackage check to prefix matching?

internal/librarian/rust/update_manifest_test.go
jameslynnwu: lmk if this is what you were thinking?

internal/librarian/rust/update_manifest.go
```diff
@@ -92,6 +90,23 @@ func updateWorkspaceVersion(path, crateName string, newVersion semver.Version) e
...
+	if !strings.HasPrefix(trimmed, crateName) {
+		return false
+	}
+	after := strings.TrimSpace(trimmed[len(crateName):])
+	return strings.HasPrefix(after, "=")
+}
+
+// isRenamedDependency checks if the line describes a renamed dependency that uses the crate.
+// Example: custom_name = { package = "crateName", ... }.
+func isRenamedDependency(line, crateName string) bool {
```
julieqiu, line 103: can this just be:

```
func isRenamed(line, crate string) bool {
	pattern := `package = "` + crate + `"`
	return strings.Contains(line, pattern)
}
```

jameslynnwu: Yes, provided that we don't need to worry about commented out lines or whitespace differences

internal/librarian/rust/update_manifest_test.go
```diff
@@ -170,6 +170,36 @@ func TestUpdateWorkspaceVersion(t *testing.T) {
...
+		},
+		{
+			name:      "inline comment ignored",
+			content:   "[workspace.dependencies]\nother-crate = { version = \"1.0.0\" } # package = \"google-cloud-wkt\"\n",
+			crateName: "google-cloud-wkt",
+			want:      "[workspace.dependencies]\nother-crate = { version = \"1.0.0\" } # package = \"google-cloud-wkt\"\n",
+		},
+		{
+			name:      "prefix crate name no-op",
+			content:   "[workspace.dependencies]\nwkt-types = { version = \"1.0.0\" }\n",
```
dbolduc, line 187: https://en.wikipedia.org/wiki/RAS_syndrome

😉

internal/librarian/rust/update_manifest.go
```diff
@@ -92,6 +88,23 @@ func updateWorkspaceVersion(path, crateName string, newVersion semver.Version) e
...
```
dbolduc, line 103: Smart, this is way more robust than what I was thinking of.

dbolduc (approved):
LGTM, thanks!

internal/librarian/rust/update_manifest_test.go
```diff
@@ -170,6 +170,36 @@ func TestUpdateWorkspaceVersion(t *testing.T) {
...
+			content:   "[workspace.dependencies]\nwkt = { package = \"google-cloud-wkt\", version = \"1.0.0\" }\n",
+			crateName: "google-cloud-wkt",
+			want:      "[workspace.dependencies]\nwkt = { package = \"google-cloud-wkt\", version = \"2.0.0\" }\n",
+		},
+		{
+			name:      "inline comment ignored",
+			content:   "[workspace.dependencies]\nother-crate = { version = \"1.0.0\" } # package = \"google-cloud-wkt\"\n",
+			crateName: "google-cloud-wkt",
+			want:      "[workspace.dependencies]\nother-crate = { version = \"1.0.0\" } # package = \"google-cloud-wkt\"\n",
+		},
```
dbolduc: I don't think we need to support this case. If we did, we should be parsing the toml instead of manipulating strings.

This is also perfectly valid TOML

```toml
[workspace.dependencies.wkt]
default-features = false
version = "1.3.0"
path = "src/wkt"
package = "google-cloud-wkt"
```

... but I don't think we need to support it in this code. We can assume we write our manifest `Cargo.toml` in a certain way.

sofisl (approved):

PR #5438, opened 2026-04-16
fix(internal/librarian/java): clean generated files from gapic module by marker

zhumin8:
Supports cleaning generated files from gapic modules by scanning files and look for one these markers: `// AUTO-GENERATED DOCUMENTATION AND CLASS.`, `@Generated(\"by gapic-generator-java\")`.

For #4559


internal/librarian/java/clean.go
```diff
@@ -125,3 +152,39 @@ func cleanPath(targetPath, root string, keepSet map[string]bool) error {
...
+		}
+	}()
+
+	scanner := bufio.NewScanner(f)
+	for scanner.Scan() {
+		line := scanner.Text()
+		if strings.Contains(line, autoGeneratedMarker) || strings.Contains(line, autoGeneratedAnnotation) {
+			return true, nil
+		}
+	}
```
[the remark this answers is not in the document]
zhumin8: Not choosing read all and a limit line because this marker is after imports, and imports can be long. The current record is about 350 lines in, but it's no guarantee. Instead, updated to use a safer method.

internal/librarian/java/clean.go
```diff
@@ -125,3 +153,45 @@ func cleanPath(targetPath, root string, keepSet map[string]bool) error {
...
+		return false, nil
+	}
+	return hasMarker(path)
+}
+
+// hasMarker checks if the file at path contains the auto-generated marker.
+// It scans the file line by line using [bufio.Reader] and returns true as soon
+// as the marker is found, avoiding reading the entire file.
+func hasMarker(path string) (found bool, err error) {
+	f, err := os.Open(path)
```
JoeWang1127, line 173: I wonder whether this approach have performance downgrade because it needs to open and read every file in the GAPIC directory?

Do you find the "keep files" approach not good enough?

zhumin8: In terms of performance, I tested with aiplatform which should be one of the largest, before and after the change, run time both ranged from 1m20s t0 2m20s on my cloudtop. According to gemini's analysis based on library size of aiplatform (number of files, size of files), the time difference should be less than a second.

With the  "keep files" approach:
- We need to explicitly list out each file/folder to keep, these might be +100 lines long for some libraries. e.g. [bigquerystorage](https://github.com/googleapis/google-cloud-java/blob/main/java-bigquerystorage/.OwlBot-hermetic.yaml) got 70+ lines for preserve. Note that "keep" config does not support globs now, so this list explodes about X5 counting the API versions.
- For developers working on features of bigquerystorage, if a new file is added/deleted, they need to know the extra step to change librarian.yaml. It is not a good experience for who works on handwritten features.  

A bit more context here https://github.com/googleapis/librarian/issues/4559#issuecomment-4263083342.


Alternatively, we can try to derive the full output path like `google-cloud-bigquerystorage/src/main/java/com/google/cloud/bigquery/storage/v1` (we currently clean `google-cloud-bigquerystorage/src` before this change). Benefit of this is should work for most cases. But counter arguments: the 2 points above still stands for bigquerystorage; Logic that works out path `src/main/java/com/google/cloud/bigquery/storage/v1` is inside gapic-generator-java. Although this structure should not expect to change, I am a bit hesitant to derive this again in librarian.

What do you think?

zhumin8: bigtable is another example that have mixed handwritten files and generated one on path.
https://github.com/googleapis/java-bigtable/tree/main/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2
Today, they are simply not cleaned up by [.OwlBot-hermetic.yaml](https://github.com/googleapis/java-bigtable/blob/main/.github/.OwlBot-hermetic.yaml)

JoeWang1127: > bigtable is another example that have mixed handwritten files and generated one on path.
https://github.com/googleapis/java-bigtable/tree/main/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2

Could you point me a generated file in this directory? I think bigtable (not the admin) disabled the gapic code, the library generation didn't have gapic updates, e.g., in https://github.com/googleapis/java-bigtable/pull/2845

zhumin8: Oh. good to know. I did not look that far. But [generation_config.yaml](https://github.com/googleapis/java-bigtable/blob/0aecd6aa7e27a71dcfddb3255c4663000839f05c/generation_config.yaml#L37) still has `google/bigtable/v2` listed. Do you happen to know how is it disabled?

Let me bring this discussion to a doc, we can discuss further.

JoeWang1127: On a 2nd look, the PR did update GAPIC file: https://github.com/googleapis/java-bigtable/pull/2845/changes#diff-2122ff8e0eb2080906dbcdc40c3e76a9f5a7db14218ade216b578c19ac452848

I think the marker of generated by gapic-generator-java is removed from owlbot.py:
https://github.com/googleapis/java-bigtable/blob/main/owlbot.py#L33

zhumin8: Interesting, `make_internal_only` is adding the `@Generated("by gapic-generator")` annotation. But TODO linked is closed, maybe this is outdated? We'll know when we migrate bigtable.
I can also alter to rely on the generated comment only if needed.

Here is the [doc](https://docs.google.com/document/d/1gLOO5WdI3l-zOh-F4FGz2AAUhYZ1Ove3FSVQoxJ_hDA/edit?resourcekey=0-_cqHVvSD_fDt6XnYTfr-kg&tab=t.0) for more discussions.

zhumin8:
Closing this temporarily as there are discussions going on in [doc](https://docs.google.com/document/d/1gLOO5WdI3l-zOh-F4FGz2AAUhYZ1Ove3FSVQoxJ_hDA/edit?resourcekey=0-_cqHVvSD_fDt6XnYTfr-kg&tab=t.0). Will re-open if we feel this is the right approach.

PR #5494, opened 2026-04-20
feat(sidekick/swift): generate query parameters

coryan:
Compute the query parameters for the default binding and use them in the generated code. We have existing code in sidekick to compute the query parameters, so this is not that difficult.

Fixes #5220

### DIFF
--- a/internal/sidekick/swift/annotate_method.go
+++ b/internal/sidekick/swift/annotate_method.go
@@ -32,7 +32,8 @@ type methodAnnotations struct {
 
 // HasQueryParams returns true if the method's default binding has query parameters
 //
-// The mustache templates use this to avoid
+// The mustache templates use this to (1) use a `var query` vs. `let query` for the collection of
+// query parameters, and (2) generate the query parameter encoder only once, and only if needed.
 func (ann *methodAnnotations) HasQueryParams() bool {
 	return len(ann.QueryParams) != 0
 }
--- a/internal/sidekick/swift/annotate_method.go
+++ b/internal/sidekick/swift/annotate_method.go
@@ -16,6 +16,7 @@ package swift
 
 import (
 	"github.com/googleapis/librarian/internal/sidekick/api"
+	"github.com/googleapis/librarian/internal/sidekick/language"
 )
 
 type methodAnnotations struct {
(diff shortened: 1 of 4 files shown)


internal/sidekick/swift/annotate_method.go
```diff
@@ -26,6 +27,14 @@ type methodAnnotations struct {
 	HasBody        bool
 	IsBodyWildcard bool
 	BodyField      string
+	QueryParams    []*api.Field
+}
+
+// HasQueryParams returns true if the method's default binding has query parameters
+//
+// The mustache templates use this to avoid
```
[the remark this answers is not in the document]
coryan: Fixed.

internal/sidekick/swift/annotate_method.go
```diff
@@ -26,6 +27,15 @@ type methodAnnotations struct {
...
 	IsBodyWildcard bool
 	BodyField      string
+	QueryParams    []*api.Field
+}
+
+// HasQueryParams returns true if the method's default binding has query parameters
+//
+// The mustache templates use this to (1) use a `var query` vs. `let query` for the collection of
+// query parameters, and (2) generate the query parameter encoder only once, and only if needed.
+func (ann *methodAnnotations) HasQueryParams() bool {
```
julieqiu, line 37: nit: I would personally just make this a function since it only uses one field

internal/sidekick/swift/annotate_service_test.go
```diff
@@ -70,24 +70,48 @@ func TestAnnotateService(t *testing.T) {
...
 			},
 			{
-				Name:     "NilPathInfoMethod",
-				PathInfo: nil,
+				Name:         "NilPathInfoMethod",
+				InputTypeID:  inputType.ID,
+				InputType:    inputType,
+				OutputTypeID: outputType.ID,
+				OutputType:   outputType,
+				PathInfo:     nil,
```
julieqiu: you don't need to list this since it is nil

julieqiu (approved):

internal/sidekick/swift/annotate_service_test.go
coryan: Done.

internal/sidekick/swift/annotate_method.go
coryan: You mean `HasQueryParams(ann *methodAnnotations)` ? I don't think that works with mustache, I tried. And the mustache library we use only documents method receivers:

https://github.com/cbroglie/mustache#a-note-about-method-receivers

Issue #3756, opened 2026-01-29
librarian: add update version command


julieqiu:
librarian should be able to update itself to the latest version. 

The version in librarian.yaml should be updated accordingly.

PR #5500, opened 2026-04-20
feat(internal/librarian): add get and set for version field

sofisl:
This prepares for updating language repositories to use the new syntax instead of sed.

Step #1 in https://github.com/googleapis/librarian/issues/5074#issuecomment-4284406066

For #3756

### DIFF
--- a/internal/librarian/cfg_value_utility_test.go
+++ b/internal/librarian/cfg_value_utility_test.go
@@ -1,0 +1,130 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package librarian
+
+import (
+	"errors"
+	"testing"
+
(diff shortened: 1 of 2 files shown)


JoeWang1127:
> For https://github.com/googleapis/librarian/issues/3756

Could you modify this to be the last line of description?

internal/librarian/cfg_value_utility_test.go
```diff
@@ -0,0 +1,130 @@
...
+			path: "version",
+			want: "v1.0.0",
+		},
+	} {
+		t.Run(test.path, func(t *testing.T) {
+			got, err := getConfigValue(currentConfig, test.path)
+			if err != nil {
+				t.Fatal(err)
+			}
+			if got != test.want {
```
JoeWang1127:
```suggestion
if diff := cmp.Diff(want, got); diff != "" {
	t.Errorf("mismatch (-want +got):\n%s", diff)
}
```
https://github.com/googleapis/librarian/blob/7bf0e08615d35f0c412274b6f420cca5951eea46/doc/howwewritego.md#use-cmpdiff-for-comparisons

internal/librarian/cfg_value_utility_test.go
```diff
@@ -0,0 +1,130 @@
...
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			cfg := &config.Config{
+				Version: "v1.0.0",
+			}
+			got, err := setConfigValue(cfg, test.path, "some-value")
+			if !errors.Is(err, test.wantErr) {
+				t.Errorf("setConfigValue(%q) error = %v, wantErr %v", test.path, err, test.wantErr)
+			}
+			if got != nil {
```
JoeWang1127: I think you don't need to check the returned config since `err != nil`.

JoeWang1127 (approved):

PR #5503, opened 2026-04-20
feat(dart): specify base API URL

brianquinlan:


internal/sidekick/dart/templates/lib/service.mustache
```diff
@@ -27,8 +28,13 @@ final class {{Codec.Name}} {
   /// The provided [http.Client] must be configured to provide whatever
   /// authentication is required by `{{Codec.Name}}`. You can do that using
   /// [`package:googleapis_auth`](https://pub.dev/packages/googleapis_auth).
-  {{Codec.Name}}({required http.Client client})
-      : _client = ServiceClient(client: client);
+  ///
+  /// If [endPoint] is provided then its `host` and `scheme` are used for all
```
natebosch: As written, this uses more than `host` and `scheme`. It also uses the `query`, and `fragment`. If we want to be pedantic (and I think we should) we also use more than the `host` - it would be more correct I think to say we use it's `authority` which includes the `host`, `port`, and `userInfo`.

If we want to match the comment we either need to `replace` everything else with `null`, or use the `Uri` constructor and explicitly pass the fields we want to preserve.

If we want the comment to match the implementation I would say something like "If [endPoint] is provided it is used as the base Uri for API requests, with the path replaced. Typically this is constructed with an appropriate `host` and `scheme` for the server."

natebosch: Oh, two of the uses do replace the `query`, but I think one doesn't..

natebosch (approved):

internal/sidekick/dart/templates/lib/service.mustache
brianquinlan: I use the `Uri` constructor and pass the fields that I want to reserve.

Issue #5256, opened 2026-04-09
nodejs: resolve migrate+generate diffs for google-cloud-compute


julieqiu:
Once we finish https://github.com/googleapis/librarian/issues/4843, we should repeat these steps and resolve diffs for google-cloud-compute.

```
  cd /path/to/googleapis/librarian
  go install ./cmd/librarian && go install ./tool/cmd/migrate
  cd /path/to/googleapis/google-cloud-node
  git fetch upstream && git reset --hard upstream/main
  migrate .
  librarian generate google-cloud-compute
  git diff --stat
```

quirogas:
As of this comment, the current diff from running `migrate` and `generate` for `google-cloud-compute` is massive.

### Diff Summary (after running migrate+generate)
```console
1135 files changed, 150859 insertions(+), 323603 deletions(-)
```

### Categorization of Changes

#### 1. Expected Deletions (Cleanup)
- `packages/google-cloud-compute/.OwlBot.yaml`
- `packages/google-cloud-compute/.readme-partials.yaml`

#### 2. Deprecated / Removed Protos & Code
- `packages/google-cloud-compute/protos/google/cloud/compute/v1small/compute_small.proto`
- `packages/google-cloud-compute/src/v1small/`
- `packages/google-cloud-compute/test/gapic_addresses_v1small.ts`

**Actionable Item:** `v1small` should be removed from `librarian.yaml` because it is marked for "testing and prototyping" and its name violates Node.js versioning conventions, which causes generate to fail. @pearigee WDYT?

#### 3. Structural Regressions (Missing Helpers)
Helper methods like `projectPath` are missing from all client classes (e.g., `AddressesClient`).
**Actionable Item:** Investigate why the helpers are missing. Look at [#4744 commnet](https://github.com/googleapis/librarian/issues/4744#issuecomment-4227530684) for guidance. 

#### 4. Massive Code Drift
The majority of the 400k+ line diff resides in:
  - `packages/google-cloud-compute/protos/protos.{js,d.ts,json}`
  - `packages/google-cloud-compute/src/v1/*.ts`
  - `packages/google-cloud-compute/test/*.ts`

Unlike standard GAPICs (like Secret Manager), Compute is a "Diregapic" (http://go/cloudsdk-compute). The generated code shows significant differences in transport handling (REST vs gRPC) and auth initialization.
**Actionable Item:** Verify if these changes are expected improvements in the modern generator or regressions in feature parity.

[...]

PR #5505, opened 2026-04-21
feat(nodejs): add native diregapic support

quirogas:
Adds native support for the `diregapic` flag in Node.js client library generation. 

`librarian generate` for Node.js lacks the ability to pass the `--diregapic` flag to `gapic-generator-typescript`, which results in a massive diff when generating  `google-cloud-compute`.  This PR introduces changes to both pass the flag to the generator, and detect the flag when running the migration. 

## Changes
* **Config:** Added `DIREGAPIC` boolean to the `NodejsAPI` configuration struct (`internal/config/language.go`).
* **Generation:** Updated `resolveNodejsAPI` and `buildGeneratorArgs` in `internal/librarian/nodejs/generate.go` to parse the `DIREGAPIC` field and append the `--diregapic` flag to the TypeScript generator CLI arguments.
* **Migrate Tool:** Enhanced `parseBazelNodejsInfo` and `buildNodejsLibraries` in `tool/cmd/migrate/nodejs.go` to automatically detect the `diregapic = True` attribute in `nodejs_gapic_library` Bazel rules and populate the resulting `librarian.yaml`.

For #5256


internal/config/config_test.go
```diff
@@ -434,3 +434,37 @@ libraries:
 		t.Errorf("mismatch (-want +got):\n%s", diff)
 	}
 }
+
+func TestNodejsConfig_DIREGAPIC(t *testing.T) {
```
jameslynnwu: I see that TestSwiftConfig_IncludeList has a similar style case, but I wonder if this is worth testing? Effectively, we're just testing that we have set the yaml annotations correctly?

tool/cmd/migrate/nodejs.go
```diff
@@ -168,9 +169,17 @@ func buildNodejsLibraries(repoPath, googleapisDir string) ([]*config.Library, er
 			if err != nil {
 				return nil, fmt.Errorf("parsing API paths for %s: %w", libraryName, err)
 			}
+			for _, api := range apis {
```
jameslynnwu: can we restructure to something like:
```
var nodejsAPIs []*config.NodejsAPI
for _, api := range apis {
  append(...
}
library.Nodejs.NodejsAPIs = nodejsAPIs
```
Maybe also extract into a separate buildNodejsLibraryAPIs (and buildNodejsLibrary) function

Lastly, is ensureNodejsPackage necessary? Can we just set it library.Nodejs on line 156?

internal/config/config_test.go
quirogas: Yes, we are. This was intended to maintain high test coverage, but you are right, it is redundant.

tool/cmd/migrate/nodejs.go
quirogas: >can we restructure to something like:
>```
>var nodejsAPIs []*config.NodejsAPI
>for _, api := range apis {
>  append(...
>}
>library.Nodejs.NodejsAPIs = nodejsAPIs
>```
>Maybe also extract into a separate buildNodejsLibraryAPIs (and buildNodejsLibrary) function

done

> Lastly, is ensureNodejsPackage necessary? Can we just set it library.Nodejs on line 156?

Done. I believe the original intent was to perform some lazy dependency gathering; if this becomes a problem, we can revert it back.

tool/cmd/migrate/nodejs.go
```diff
@@ -143,99 +144,118 @@ func buildNodejsLibraries(repoPath, googleapisDir string) ([]*config.Library, er
...
+		library, err := buildNodejsLibrary(googleapisDir, packagesDir, entry.Name())
 		if err != nil {
-			return nil, fmt.Errorf("reading package.json for %s: %w", libraryName, err)
+			return nil, err
 		}
-
-		library := &config.Library{
-			Name:    libraryName,
-			Version: pkgJSON.Version,
+		if library != nil {
```
jameslynnwu: if there's no error, then library should exist. no need to check both.

jameslynnwu (approved):

PR #5530, opened 2026-04-22
feat(sidekick/rust): Format resource names in RPC samples.

amanda-tarafa:
RPC rely on resource names to identify the resource they operate on. If we know the format of the resource name, we show it on RPC samples.

### DIFF
--- a/internal/sidekick/api/xref.go
+++ b/internal/sidekick/api/xref.go
@@ -105,6 +105,9 @@ func enrichSamples(model *API) {
 				o.ExampleField = slices.MaxFunc(o.Fields, sortOneOfFieldForExamples)
 			}
 		}
+		for _, f := range m.Fields {
+			enrichWithResourceNamePattern(f, m, model)
+		}
 	}
 
 	for m := range model.AllMethods() {
@@ -381,6 +384,69 @@ func enrichMethodSamples(m *Method) {
 	m.IsAIPStandard = m.SampleInfo != nil
 }
 
+func enrichWithResourceNamePattern(f *Field, m *Message, model *API) {
+	var resource *Resource
+	truncateChild := false
+
+	if f.ResourceReference != nil {
+		// The field is a resource reference.
+		if res := model.Resource(f.ResourceReference.Type); res != nil {
+			resource = res
+		} else if res := model.Resource(f.ResourceReference.ChildType); res != nil {
+			resource = res
+			truncateChild = true
+		}
+	} else if f.Name == StandardFieldNameForResourceRef {
+		// The field is the `name` field.
+		resource = m.Resource
+	}
+
+	if resource != nil && len(resource.Patterns) > 0 {
+		f.ResourceNamePattern = toResourceNamePattern(resource.Patterns[0], truncateChild)
+	}
+}
+
+// toResourceNamePattern converts a ResourcePattern into a ResourceNamePattern.
+func toResourceNamePattern(pattern ResourcePattern, skipLast bool) *ResourceNamePattern {
+	var segments []ResourceNameSegment
+	for i := 0; i < len(pattern); i++ {
+		s := pattern[i]
+		if s.Literal != nil && strings.HasPrefix(*s.Literal, "//") {
+			continue
+		}
+		seg := ResourceNameSegment{}
+		if s.Literal != nil {
+			seg.Literal = *s.Literal
+		}
+		if s.Variable != nil && len(s.Variable.FieldPath) > 0 {
+			// The parser used for parsing resource name patterns is the same used for
+			// parsing HTTP rules. Resource patterns do not have "field paths",
+			// instead they only have "placeholders" for each of the resource IDs of
+			// the resources that are part of the resource name hierarchy.
+			// These placeholders end up on the first, and only, field path
+			// that results when parsing a segment of a resource name pattern.
+			seg.Variable = s.Variable.FieldPath[0]
+		}
+
+		// Try to combine with next segment if this is a literal and next is variable
+		if s.Literal != nil && i+1 < len(pattern) && pattern[i+1].Variable != nil {
+			if len(pattern[i+1].Variable.FieldPath) > 0 {
+				seg.Variable = pattern[i+1].Variable.FieldPath[0]
+			}
+			i++
+		}
+
+		// Clean up leading and trailing slashes
+		seg.Literal = strings.TrimSuffix(strings.TrimPrefix(seg.Literal, "/"), "/")
+
+		segments = append(segments, seg)
+	}
+	if skipLast && len(segments) > 0 {
+		segments = segments[:len(segments)-1]
+	}
+	return &ResourceNamePattern{Segments: segments}
+}
+
 func aipStandardGetInfo(m *Method) *SampleInfo {
 	if !m.IsSimple || m.InputType == nil || m.ReturnsEmpty {
 		return nil
@@ -410,7 +476,8 @@ func aipStandardGetInfo(m *Method) *SampleInfo {
 	}
 
 	return &SampleInfo{
-		ResourceNameField: resourceField,
+		ResourceNameField:     resourceField,
+		IsRequestResourceName: true,
 	}
 }
 
@@ -434,7 +501,8 @@ func aipStandardDeleteInfo(m *Method) *SampleInfo {
 	}
 
 	return &SampleInfo{
-		ResourceNameField: resourceField,
+		ResourceNameField:     resourceField,
+		IsRequestResourceName: true,
 	}
 }
 
@@ -458,7 +526,8 @@ func aipStandardUndeleteInfo(m *Method) *SampleInfo {
 	}
 
 	return &SampleInfo{
-		ResourceNameField: resourceField,
+		ResourceNameField:     resourceField,
+		IsRequestResourceName: true,
 	}
 }
 
@@ -502,8 +571,9 @@ func aipStandardCreateInfo(m *Method) *SampleInfo {
 	resourceIDField := findResourceIDField(m.InputType, maybeSingular)
 
 	info := &SampleInfo{
-		ResourceNameField: parentField,
-		MessageField:      resourceField,
+		ResourceNameField:     parentField,
+		IsRequestResourceName: true,
+		MessageField:          resourceField,
 	}
 	if resourceIDField != nil {
 		info.ResourceIDField = resourceIDField
@@ -548,9 +618,21 @@ func aipStandardUpdateInfo(m *Method) *SampleInfo {
 		}
 	}
 
+	var resourceNameField *Field
+	if resourceField.MessageType != nil {
+		for _, f := range resourceField.MessageType.Fields {
+			if f.Name == StandardFieldNameForResourceRef {
+				resourceNameField = f
+				break
+			}
+		}
+	}
+
 	return &SampleInfo{
-		MessageField:    resourceField,
-		UpdateMaskField: updateMaskField,
+		ResourceNameField:     resourceNameField,
+		IsMessageResourceName: true,
+		MessageField:          resourceField,
+		UpdateMaskField:       updateMaskField,
 	}
 }
 
@@ -585,7 +667,8 @@ func aipStandardListInfo(m *Method) *SampleInfo {
 	}
 
 	return &SampleInfo{
-		ResourceNameField: parentField,
+		ResourceNameField:     parentField,
+		IsRequestResourceName: true,
 	}
 }
 
--- a/internal/sidekick/api/xref_test.go
+++ b/internal/sidekick/api/xref_test.go
@@ -419,6 +419,207 @@ func TestEnrichSamplesOneOfExampleField(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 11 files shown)


internal/sidekick/api/xref.go
```diff
@@ -89,10 +89,84 @@ func CrossReference(model *API) error {
...
+						pattern := ToResourceNamePattern(res.Patterns[0])
+						if len(pattern.Segments) > 1 {
+							pattern.Segments = pattern.Segments[:len(pattern.Segments)-1]
+						}
+						f.ResourceNamePattern = pattern
+					}
+				}
+			}
+		}
+	}
```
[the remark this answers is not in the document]
amanda-tarafa: I've significantly refactored this code.

amanda-tarafa:
This is ready for review. Generation results can be seen in https://github.com/googleapis/google-cloud-rust/pull/5504.

internal/sidekick/api/xref.go
```diff
@@ -381,6 +384,65 @@ func enrichMethodSamples(m *Method) {
...
 }
 
+func enrichWithResourceNamePattern(f *Field, m *Message, model *API) {
+
+	var resource *Resource
+	truncateChild := false
+
+	// The field is a resource reference.
+	if f.ResourceReference != nil {
+		if res, ok := model.State.ResourceByType[f.ResourceReference.Type]; ok {
```
coryan: I think @ianthehat is making changes to hide `.ResourceByType` and `.State`.  Maybe both of you should talk to avoid a last minute conflict / broken build.

internal/sidekick/api/xref.go
```diff
@@ -381,6 +384,65 @@ func enrichMethodSamples(m *Method) {
 	m.IsAIPStandard = m.SampleInfo != nil
 }
 
+func enrichWithResourceNamePattern(f *Field, m *Message, model *API) {
+
```
coryan: nit: blank line?

coryan (commented):
I won't be able to complete the review, just a heads up...

internal/sidekick/api/xref.go
amanda-tarafa: Removed, thanks.

internal/sidekick/api/xref.go
amanda-tarafa: Refactored to use all the accessors instead. I think I didn't miss any.

internal/sidekick/api/xref_test.go
```diff
@@ -419,6 +419,214 @@ func TestEnrichSamplesOneOfExampleField(t *testing.T) {
...
+		}
+		want := ResourceNameSegment{Literal: "resources", Variable: "resource"}
+		if field.ResourceNamePattern.Segments[0] != want {
+			t.Errorf("mismatch: got %+v, want %+v", field.ResourceNamePattern.Segments[0], want)
+		}
+	})
+}
+
+func TestToResourceNamePattern(t *testing.T) {
+	tests := []struct {
```
coryan: I have been taught to follow the [guidelines](https://github.com/googleapis/librarian/blob/main/doc/howwewritego.md) which require this to be:


```suggestion
	for _, test := range []struct {
```

internal/sidekick/api/xref_test.go
```diff
@@ -419,6 +419,214 @@ func TestEnrichSamplesOneOfExampleField(t *testing.T) {
...
+				},
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := toResourceNamePattern(tt.pattern, tt.skipLast)
+			if diff := cmp.Diff(tt.want, got); diff != "" {
+				t.Errorf("ToResourceNamePattern() mismatch (-want +got):\n%s", diff)
```
coryan: The guidelines also have opinions on the format of this error message.

internal/sidekick/api/xref_test.go
```diff
@@ -419,6 +419,214 @@ func TestEnrichSamplesOneOfExampleField(t *testing.T) {
...
+			t.Fatal("ResourceNamePattern should not be nil")
+		}
+		if len(field.ResourceNamePattern.Segments) != 1 {
+			t.Fatalf("expected 1 segment, got %d", len(field.ResourceNamePattern.Segments))
+		}
+		want := ResourceNameSegment{Literal: "resources", Variable: "resource"}
+		if field.ResourceNamePattern.Segments[0] != want {
+			t.Errorf("mismatch: got %+v, want %+v", field.ResourceNamePattern.Segments[0], want)
+		}
+	})
```
coryan: I think you can (and should) use `cmp.Diff()` in this test

internal/sidekick/api/xref.go
```diff
@@ -381,6 +384,64 @@ func enrichMethodSamples(m *Method) {
...
+		if s.Literal != nil && i+1 < len(pattern) && pattern[i+1].Variable != nil {
+			if len(pattern[i+1].Variable.FieldPath) > 0 {
+				seg.Variable = pattern[i+1].Variable.FieldPath[0]
+			}
+			i++
+		}
+
+		// Clean up leading and trailing slashes
+		seg.Literal = strings.TrimPrefix(seg.Literal, "/")
+		seg.Literal = strings.TrimSuffix(seg.Literal, "/")
```
coryan:
```suggestion
		seg.Literal = strings.TrimSuffix(strings.TrimPrefix(seg.Literal, "/"), "/")
```

internal/sidekick/api/xref.go
```diff
@@ -381,6 +384,64 @@ func enrichMethodSamples(m *Method) {
...
+		s := pattern[i]
+		if s.Literal != nil && strings.HasPrefix(*s.Literal, "//") {
+			continue
+		}
+		seg := ResourceNameSegment{}
+		if s.Literal != nil {
+			seg.Literal = *s.Literal
+		}
+		if s.Variable != nil && len(s.Variable.FieldPath) > 0 {
+			seg.Variable = s.Variable.FieldPath[0]
```
coryan, line 428: Why just the first element in the FieldPath? Can you leave a comment?

internal/sidekick/rust/annotate.go
```diff
@@ -1391,6 +1436,35 @@ func (c *codec) annotateField(field *api.Field, message *api.Message, model *api
...
+				formatString.WriteString("/")
+			}
+			if s.Variable != "" {
+				formatString.WriteString("{}/")
+				// Convert variable to snake_case and add _id suffix if needed
+				arg := strcase.ToSnake(s.Variable)
+				if !strings.HasSuffix(arg, "_id") && !strings.HasSuffix(arg, "_name") {
+					arg += "_id"
+				}
+				fr.FormatArgs = append(fr.FormatArgs, arg)
```
coryan: I don't think you need the `FormatArgs` and the generated code can look better if you do this:

```suggestion
				// Convert variable to snake_case and add _id suffix if needed
				arg := strcase.ToSnake(s.Variable)
				if !strings.HasSuffix(arg, "_id") && !strings.HasSuffix(arg, "_name") {
					arg += "_id"
				}
				fmt.Fprintf(&formatString, "{%s}/", arg)
```

internal/sidekick/rust/annotate.go
```diff
@@ -1391,6 +1436,35 @@ func (c *codec) annotateField(field *api.Field, message *api.Message, model *api
...
+		for _, s := range field.ResourceNamePattern.Segments {
+			if s.Literal != "" {
+				hasLiterals = true
+				formatString.WriteString(s.Literal)
+				formatString.WriteString("/")
+			}
+			if s.Variable != "" {
+				formatString.WriteString("{}/")
+				// Convert variable to snake_case and add _id suffix if needed
+				arg := strcase.ToSnake(s.Variable)
```
coryan: Hmmm... This is only safe because of the `_id/_name` suffix, otherwise you could land on a keyword.  Please use `toSnakeNoMangling()` so we can keep all the name mapping in one place.

internal/sidekick/rust/annotate.go
```diff
@@ -1391,6 +1436,35 @@ func (c *codec) annotateField(field *api.Field, message *api.Message, model *api
...
+				if !strings.HasSuffix(arg, "_id") && !strings.HasSuffix(arg, "_name") {
+					arg += "_id"
+				}
+				fr.FormatArgs = append(fr.FormatArgs, arg)
+			}
+		}
+		if hasLiterals && len(fr.FormatArgs) > 0 {
+			fr.FormatString = formatString.String()
+			// Remove the last slash
+			fr.FormatString = fr.FormatString[:len(fr.FormatString)-1]
```
coryan: Consider putting array and then use `strings.Join(array, "/")` instead of adding the slashes manually.

coryan (changes requested):
Looking good, I think we can cleanup the output a little bit more using `format!("foo/{bar_id}").

internal/sidekick/api/xref.go
amanda-tarafa: Yes, better, done.

internal/sidekick/api/xref_test.go
amanda-tarafa: Yes, done.

internal/sidekick/api/xref_test.go
amanda-tarafa: Done here as well.

internal/sidekick/api/xref_test.go
amanda-tarafa: And done here. Sorry about all of these, this file is older than the guidelines and I was copying around some for these tests.

internal/sidekick/rust/annotate.go
amanda-tarafa: Thanks, TIL `toSnakeNoMangling()`

internal/sidekick/rust/annotate.go
amanda-tarafa: I like including the names of the arguments straight in the format here so that the output is cleaner. 

But I'll keep around the `FormatArgs` because that's what we use later on to annotate the sample with the parameters it has to include in the sample. Otherwise, we'll have to do something like this again when annotating the sample.

Note that it goes like this:

- We first annotate each field that is a resource name, with the format and format args.
- We then annotate each sample with the parameters it has to show, it will match the resource name format args if the sample has a resource name with a known format, but otherwise it will be different, see `annotateSampleInfo`. This takes a lot of complexity away from the templates.
- If we don't keep the args around, we could still recalculate them in `annoteteSampleInfo` but that seems like an unnecessary duplication, since we are already doing that here regardless.

internal/sidekick/rust/annotate.go
amanda-tarafa: Done.

internal/sidekick/api/xref.go
amanda-tarafa: Done, it took me a little to figure this out myself so I should have added the comment earlier.

amanda-tarafa (commented):
All comments addressed, pushing shortly both here and in https://github.com/googleapis/google-cloud-rust/pull/5504. Will flag when green.

amanda-tarafa:
This is ready for re-review alongside https://github.com/googleapis/google-cloud-rust/pull/5504.

internal/sidekick/api/xref.go
```diff
@@ -381,6 +384,69 @@ func enrichMethodSamples(m *Method) {
...
+	// The field is a resource reference.
+	if f.ResourceReference != nil {
+		if res := model.Resource(f.ResourceReference.Type); res != nil {
+			resource = res
+		} else if res := model.Resource(f.ResourceReference.ChildType); res != nil {
+			resource = res
+			truncateChild = true
+		}
+		// The field is the `name` field.
+	} else if f.Name == StandardFieldNameForResourceRef {
```
coryan:
```suggestion
	} else if f.Name == StandardFieldNameForResourceRef {
		// The field is the `name` field.
```

internal/sidekick/rust/annotate.go
```diff
@@ -1391,6 +1436,33 @@ func (c *codec) annotateField(field *api.Field, message *api.Message, model *api
...
+				if !strings.HasSuffix(arg, "_id") && !strings.HasSuffix(arg, "_name") {
+					arg += "_id"
+				}
+				formatString = append(formatString, "{"+arg+"}")
+				fr.FormatArgs = append(fr.FormatArgs, arg)
+			}
+		}
+		if hasLiterals && len(fr.FormatArgs) > 0 {
+			// Remove the last slash
+			fr.FormatString = strings.Join(formatString, "/")
```
coryan, line 1460:
```suggestion
		if hasLiterals && len(fr.FormatArgs) > 0 {
			fr.FormatString = strings.Join(formatString, "/")
```

coryan (approved):

internal/sidekick/api/xref.go
amanda-tarafa: Done.

internal/sidekick/rust/annotate.go
amanda-tarafa: Duh myself, thanks.

amanda-tarafa (commented):
Latest comments addressed, will merge on green. Thanks!

Issue #5181, opened 2026-04-07
sidekick: generate Swift version of `google/iam/v1`


coryan:
This API is a dependency of `google/cloud/secretmanager/v1`, so we need to generate it before we can complete the generation of `secretmanager-v1`.

coryan:
Fixed.

PR #5532, opened 2026-04-22
feat(sidekick/swift): skip self-dependencies

coryan:
For Swift, we want to declare a lot of common libraries in the default section. When generating a library that is in that section, we don't want to add it as a dependency of itself. This PR implements that feature.

Towards #5181

### DIFF
--- a/internal/sidekick/swift/annotate_model.go
+++ b/internal/sidekick/swift/annotate_model.go
@@ -81,7 +81,7 @@ func (c *codec) annotateModel() error {
 	var serviceImports []string
 	var messageImports []string
 	for _, p := range c.Dependencies {
-		if p.ApiPackage == c.Model.PackageName {
+		if p.ApiPackage == c.Model.PackageName || p.Name == c.PackageName {
 			continue
 		}
 		if p.RequiredByServices && len(c.Model.Services) != 0 {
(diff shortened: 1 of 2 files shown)


internal/sidekick/swift/annotate_model.go
```diff
@@ -81,6 +81,9 @@ func (c *codec) annotateModel() error {
 	var serviceImports []string
 	var messageImports []string
 	for _, p := range c.Dependencies {
+		if p.ApiPackage == c.Model.PackageName {
+			continue
+		}
```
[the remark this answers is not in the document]
coryan: Done.

internal/sidekick/swift/annotate_model.go
```diff
@@ -81,6 +81,9 @@ func (c *codec) annotateModel() error {
 	var serviceImports []string
 	var messageImports []string
 	for _, p := range c.Dependencies {
+		if p.ApiPackage == c.Model.PackageName || p.Name == c.PackageName {
```
AlexGhiondea, line 84: Do we care about case insensitive checks here?

coryan: no.

AlexGhiondea (approved):

PR #5534, opened 2026-04-23
feat(internal/librarian): add config command for version

### LINKED ISSUE #5074: librarian: add config command

sofisl:
This change implements the new `config` command in the `librarian` tool to read and write configuration values using dot notation. This is a very specific implementation of `config` specifically; a separate PR will be written for the full librarian.yaml structure.

Fixes #5074


cmd/librarian/doc.go
```diff
@@ -25,6 +25,61 @@ Usage:
...
 
+# config
+
+NAME:
+
+	librarian config - read and write librarian.yaml configuration
+
+USAGE:
+
+	librarian config [get|set] [path] [value]
```
jameslynnwu, line 39: With new commands, let's add it to https://github.com/googleapis/librarian/blob/main/internal/command/usage_test.go

Its a hacky test, since it requires all new commands to be added in and also relies on hardcoded expectations.

jameslynnwu (approved):

internal/librarian/config_command.go
```diff
@@ -0,0 +1,91 @@
...
+					if path == "" {
+						return fmt.Errorf("path is required")
+					}
+					if value == "" {
+						return fmt.Errorf("value is required")
+					}
+
+					cfg, err := yaml.Read[config.Config](config.LibrarianYAML)
+					if err != nil {
+						if !errors.Is(err, fs.ErrNotExist) {
```
julieqiu: what's the reason for checking for this error?

internal/librarian/config_command.go
```diff
@@ -0,0 +1,91 @@
...
+							return err
+						}
+						cfg = &config.Config{}
```
julieqiu: what's the reason for setting cfg here?

internal/librarian/config_command.go
```diff
@@ -0,0 +1,91 @@
...
+					}
+
+					updated, err := setConfigValue(cfg, path, value)
+					if err != nil {
+						return err
+					}
+
```
julieqiu: you don't need all of these newlines

see https://github.com/googleapis/librarian/blob/fa79d1c1732df4d5d932b64f49fd909d43581ce6/.gemini/styleguide.md?plain=1#L21 (the gemini bot doesn't seem to pick this up well)

internal/librarian/config_command_test.go
```diff
@@ -0,0 +1,222 @@
...
+			args:       []string{"librarian", "config", "get", "version"},
+			configYAML: "version: 1.2.3\n",
+			wantOutput: "1.2.3\n",
+		},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			tempDir := t.TempDir()
+			t.Chdir(tempDir)
+
+			if err := os.WriteFile("librarian.yaml", []byte(test.configYAML), 0644); err != nil {
```
julieqiu: use yaml.Write

internal/librarian/config_command.go
```diff
@@ -0,0 +1,91 @@
...
+					path := cmd.Args().Get(0)
+					value := cmd.Args().Get(1)
+					if path == "" {
+						return fmt.Errorf("path is required")
+					}
+					if value == "" {
+						return fmt.Errorf("value is required")
+					}
+
+					cfg, err := yaml.Read[config.Config](config.LibrarianYAML)
```
julieqiu: Move this code into a function. That way you can just test the function, instead of having to run `librarian config` in the tests

internal/librarian/config_command.go
```diff
@@ -0,0 +1,91 @@
...
+				Name:      "get",
+				Usage:     "get a configuration value",
+				UsageText: "librarian config get [path]",
+				Action: func(ctx context.Context, cmd *cli.Command) error {
+					path := cmd.Args().First()
+					if path == "" {
+						return fmt.Errorf("path is required")
+					}
+
+					cfg, err := yaml.Read[config.Config](config.LibrarianYAML)
```
julieqiu: Move this code into a function. That way you can just test the function, instead of having to run librarian config in the tests

internal/librarian/config_command.go
```diff
@@ -0,0 +1,91 @@
...
+					if err != nil {
+						return err
+					}
+
+					val, err := getConfigValue(cfg, path)
+					if err != nil {
+						return err
+					}
+
+					fmt.Println(val)
```
julieqiu: as part of moving this code into a function, pass in `cmd.Root().Writer` as the writer, and call this instead:

```
_, err = fmt.Fprintln(w, val)
```

where `w = io.Writer`
https://pkg.go.dev/io#Writer

that way you can use a different writer in tests

internal/librarian/config_command_test.go
```diff
@@ -0,0 +1,222 @@
...
+	"github.com/google/go-cmp/cmp"
+)
+
+// TestConfigCommand_Get tests that the config get command successfully reads a value from librarian.yaml.
+func TestConfigCommand_Get(t *testing.T) {
+	for _, test := range []struct {
+		name       string
+		args       []string
+		configYAML string
+		wantOutput string
```
julieqiu: nit: juse use `want` since you're only checking one value

internal/librarian/config_command_test.go
```diff
@@ -0,0 +1,222 @@
...
+				t.Fatal(err)
+			}
+
+			err := Run(t.Context(), test.args...)
+
+			if err == nil {
+				t.Fatal("expected error; got nil")
+			}
+
+			if !strings.Contains(err.Error(), test.wantErrStr) {
```
julieqiu: use named errors: https://github.com/googleapis/librarian/blob/main/doc/howwewritego.md#verify-with-errorsis-and-errorsas

internal/librarian/config_command.go
sofisl: This one was a gemini recommendation? https://github.com/googleapis/librarian/pull/5534#discussion_r3127613119.

internal/librarian/config_command.go
sofisl: If line 73 fails, then cfg would be nil. IIUC, that would cause go to panic if you try to set it to something else, so you need it to point to an empty struct? Not sure though, still learning go!

zhumin8:
extra nit: extra "/" in title

internal/librarian/config_command_test.go
sofisl: just out of curiosity, why does `wantOutput` not work?

noahdietz: I was curious about this myself as it is something I've taken for granted. I couldn't find style guide ref explicitly stating "use got/want" (but I did find a [relevant issue](https://github.com/golang/go/issues/44179)). That said, there are a few examples of using/guidelines relating to got/want

https://go.dev/wiki/TestComments#got-before-want
https://go.dev/wiki/TestComments#identify-the-function

Why `want` over `wantOutput`, I think it's related to Go's preference for brief field names, and the convention that `got` and `want` in tests represent the "output" of the code under test. Combined, `wantOutput` is overly verbose compared to `want` when there is no need to differentiate it from another test output.

internal/librarian/config_command.go
```diff
@@ -0,0 +1,100 @@
+// Copyright 2025 Google LLC
```
julieqiu: 2026

internal/librarian/config_command.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	if value == "" {
+		return errValueRequired
+	}
+	cfg, err := yaml.Read[config.Config](config.LibrarianYAML)
+	if err != nil {
+		if !errors.Is(err, fs.ErrNotExist) {
+			return err
+		}
+		cfg = &config.Config{}
```
julieqiu: what's the rationale of the error handling here? why not just return an error in all cases?

internal/librarian/config_command.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	updated, err := setConfigValue(cfg, path, value)
+	if err != nil {
+		return err
+	}
+
```
julieqiu: nit: remove unnecessary line break

internal/librarian/config_command_test.go
```diff
@@ -0,0 +1,220 @@
...
+
+			if !errors.Is(err, test.wantErr) {
+				t.Fatalf("got error %v, want %v", err, test.wantErr)
+			}
+		})
+	}
+}
+
+// TestConfigCommand_Set tests that the config set command successfully updates a value in librarian.yaml.
+func TestConfigCommand_Set(t *testing.T) {
```
julieqiu: nit: use TestRunConfigSet, since the name of the function you are testing is runConfigSet

internal/librarian/config_command_test.go
```diff
@@ -0,0 +1,220 @@
...
+		t.Run(test.name, func(t *testing.T) {
+			tempDir := t.TempDir()
+			t.Chdir(tempDir)
+
+			if err := os.WriteFile("librarian.yaml", []byte(test.configYAML), 0644); err != nil {
+				t.Fatal(err)
+			}
+
+			err := runConfigSet(test.path, test.value)
+
```
julieqiu: remove unnecessary line breaks

internal/librarian/config_command_test.go
```diff
@@ -0,0 +1,220 @@
...
```
julieqiu: nit: remove unnecessary line breaks

internal/librarian/config_command_test.go
```diff
@@ -0,0 +1,220 @@
...
+func TestConfigCommand_Set_ReadError(t *testing.T) {
+	tempDir := t.TempDir()
+	t.Chdir(tempDir)
+
+	if err := os.WriteFile("librarian.yaml", []byte("version: 1.2.3\n"), 0000); err != nil {
+		t.Fatal(err)
+	}
+
+	err := runConfigSet("version", "1.2.4")
+
```
julieqiu: nit: remove unnecessary line breaks

internal/librarian/config_command_test.go
```diff
@@ -0,0 +1,220 @@
...
+			if !strings.HasSuffix(string(gotYAML), test.wantYAML) {
+				t.Errorf("got YAML =\n%s\nwant ending with %q", string(gotYAML), test.wantYAML)
+			}
+		})
+	}
+}
+
+// TestConfigCommand_Set_Error tests that the config set command returns an error when the path or value is missing.
+func TestConfigCommand_Set_Error(t *testing.T) {
```
julieqiu: s/TestConfigCommand_Set_ReadError/TestRunConfigCommand_Error

internal/librarian/config_command_test.go
```diff
@@ -0,0 +1,220 @@
...
+
+			if !errors.Is(err, test.wantErr) {
+				t.Fatalf("got error %v, want %v", err, test.wantErr)
+			}
+		})
+	}
+}
+
+// TestConfigCommand_Set_ReadError tests that the config set command returns an error when librarian.yaml is unreadable.
+func TestConfigCommand_Set_ReadError(t *testing.T) {
```
julieqiu: this test doesn't seem super useful to me - under what conditions are you worried about a permission error?

internal/librarian/config_command.go
julieqiu: nit: name the file internal/librarian/config.go to match with the conventions used for other files

julieqiu (changes requested):

internal/librarian/config_command_test.go
```diff
@@ -0,0 +1,220 @@
...
+			tempDir := t.TempDir()
+			t.Chdir(tempDir)
+
+			if err := os.WriteFile("librarian.yaml", []byte(test.configYAML), 0644); err != nil {
+				t.Fatal(err)
+			}
+
+			var buf bytes.Buffer
+			err := runConfigGet(&buf, test.path)
+
```
julieqiu: remove unnecessary line breaks

internal/librarian/config_command_test.go
```diff
@@ -0,0 +1,220 @@
...
```
julieqiu: remove unnecessary line breaks

internal/librarian/config_command_test.go
```diff
@@ -0,0 +1,220 @@
...
+			if diff := cmp.Diff(test.want, buf.String()); diff != "" {
+				t.Errorf("mismatch (-want +got):\n%s", diff)
+			}
+		})
+	}
+}
+
+// TestConfigCommand_Get_Error tests that the config get command returns an error when the path is missing or the key is not found.
+func TestConfigCommand_Get_Error(t *testing.T) {
```
julieqiu: s/TestConfigCommand_Get_Error/TestRunConfigGet_Error

internal/librarian/config_command_test.go
```diff
@@ -0,0 +1,220 @@
...
+	"errors"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+// TestConfigCommand_Get tests that the config get command successfully reads a value from librarian.yaml.
+func TestConfigCommand_Get(t *testing.T) {
```
julieqiu: s/TestConfigCommand_Get/TestRunConfigGet

internal/librarian/config_command_test.go
sofisl: thanks @noahdietz!

internal/librarian/config_command.go
sofisl: I don't think I have a good handle on which line breaks are necessary. Is it just between functions that we have newlines, otherwise no newlines?

internal/librarian/config_command.go
sofisl: I'm kind of assuming here - If the file doesn't exist yet, we can initialize an empty Config so that they can run `librarian config set <key value>` directly without having to manually create an empty file first. For any other error (like permission denied), we do still return the error. Or do we always want to make sure the file is manually created first?

julieqiu: we should make sure the file is created first - if it doesn't exist just return an error. `librarian config set` shouldn't create a librarian.yaml for you

internal/librarian/config_command.go
julieqiu: Think about it like the way you would use blank lines as paragraphs in an essay.

A paragraph is a group of related ideas, and you start a new one when the topic or intent shifts. You wouldn't want to put a line break after every sentence. 

It's the same idea with code. Blank lines should be used to separated logical chunks, not just functions.

internal/librarian/config_command.go
sofisl: got it! Thanks!

internal/librarian/config_command_test.go
sofisl: given our conversation above, I rewrote it to just test if there's a missing file.

internal/librarian/config_command.go
sofisl: Thanks!

internal/librarian/config.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+}
+
+func runConfigGet(w io.Writer, path string) error {
+	if path == "" {
+		return errPathRequired
+	}
+	cfg, err := yaml.Read[config.Config](config.LibrarianYAML)
+	if err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
```
julieqiu: I don't think I see the value of returning a named error here - I would just return the err in all cases

internal/librarian/config.go
```diff
@@ -0,0 +1,100 @@
...
+func runConfigSet(path, value string) error {
+	if path == "" {
+		return errPathRequired
+	}
+	if value == "" {
+		return errValueRequired
+	}
+	cfg, err := yaml.Read[config.Config](config.LibrarianYAML)
+	if err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
```
julieqiu: ditto: re just return an err in all cases

julieqiu (approved):

PR #5546, opened 2026-04-23
fix(internal/librarian/python): remove hard-coded gRPC service config exception

### LINKED ISSUE #4175: python: remove package-specific overrides for .repo-metadata.json
In order to minimize the diff for Python, `PythonPackage` has fields to override the values in `.repo-metadata.json`. These are in `PythonPackage` rather than `Library` as we hope they'll be temporary. Once the migration has been completed, we can remove them, check the diff, and either adjust sdk.yaml or live with the changes.

Note that although the Librarian code only changes how .repo-metadata.json is written, this then affects documentation - which is why it's safer to minimize the diff first and then check for changes after migration.

jskeet:
The Compute packages in Python previously ignored the gRPC service config, to match the original BUILD.bazel config and produce a minimal diff for migration. This exception is removed, so that the gRPC service config is used in Python as it is in other languages.

Fixes #4175

### DIFF
--- a/internal/librarian/python/generate.go
+++ b/internal/librarian/python/generate.go
@@ -325,11 +325,6 @@ func createProtocOptions(api *config.API, library *config.Library, googleapisDir
 	if err != nil {
 		return nil, err
 	}
-	// TODO(https://github.com/googleapis/librarian/issues/3827): remove this
-	// hardcoding once we can use the gRPC service config for Compute.
-	if strings.HasPrefix(library.Name, "google-cloud-compute") {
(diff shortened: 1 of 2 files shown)


jskeet:
Note: deliberately not setting to automerge, as we need confirmation from the Python team first. But https://github.com/googleapis/google-cloud-python/pull/16751 shows that the result passes tests.

jskeet:
We've merged the PR in the Python repo now, so this is ready to merge too.

codyoss (approved):

PR #5548, opened 2026-04-23
refactor(internal/serviceconfig): centralize transport logic for repometadata

### LINKED ISSUE #5547: java: storage generate diff in repo-metadata
Generated .repo-metadata.json for storage has `  "transport": "rest",` when expected `  "transport": "http",`.

Logic for TransportOverride in repo-metadata should be centralized in RepoMetadataTransport.
The transport_override config was introduced in https://github.com/googleapis/librarian/pull/5483.

zhumin8:
RepoMetadataTransport now accepts a *config.Library argument to internalize transport override logic. This ensures that transport values, including overrides, undergo the same platform-specific transformations (such as mapping "rest" to "http" for Java) in a single location.

Fix #5547

### DIFF
--- a/internal/librarian/java/repometadata.go
+++ b/internal/librarian/java/repometadata.go
@@ -148,9 +148,6 @@ func deriveRepoMetadata(cfg *config.Config, library *config.Library, googleapisD
 		metadata.RecommendedPackage = library.Java.RecommendedPackage
 		metadata.RestDocumentation = library.Java.RestDocumentation
 		metadata.RpcDocumentation = library.Java.RpcDocumentation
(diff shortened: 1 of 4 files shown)


jskeet (approved):
Any language-specific code in language-agnostic packages feels a bit odd, but this is already a function that behaves in a language-specific way... and we're planning on fixing the discrepancy in the longer term anyway, IIRC.

zhumin8:
> Any language-specific code in language-agnostic packages feels a bit odd
I agree. This is more of a temporarily solution until we figure out. If there are other onboarding languages that needs similar manipulation, centralizing it makes it easier for future discovery and changes.

PR #5550, opened 2026-04-23
refactor(sidekick): use API helper methods for State access

ianthehat:
Eventually direct access to the maps will become impossible, this catches most (but not all) of the read accesses.

### DIFF
--- a/internal/sidekick/parser/auto_populated.go
+++ b/internal/sidekick/parser/auto_populated.go
@@ -32,12 +32,12 @@ func updateAutoPopulatedFields(serviceConfig *serviceconfig.Service, model *api.
 	}
 	for _, m := range serviceConfig.GetPublishing().GetMethodSettings() {
 		selector := m.GetSelector()
-		method, ok := model.State.MethodByID["."+selector]
-		if !ok {
+		method := model.Method("." + selector)
+		if method == nil {
 			continue
 		}
-		message, ok := model.State.MessageByID[method.InputTypeID]
-		if !ok {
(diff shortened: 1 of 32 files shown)


internal/sidekick/parser/auto_populated.go
```diff
@@ -32,12 +32,12 @@ func updateAutoPopulatedFields(serviceConfig *serviceconfig.Service, model *api.
 	}
 	for _, m := range serviceConfig.GetPublishing().GetMethodSettings() {
 		selector := m.GetSelector()
-		method, ok := model.State.MethodByID["."+selector]
-		if !ok {
+		method := model.Method("." + selector)
```
coryan, line 35: TIL, what `go fmt` treats  the expression differently inside square vs. round brackets

internal/sidekick/swift/lookup.go
coryan: Why not this one too?

coryan (approved):
Fix the new one in Swift and then merge.

internal/sidekick/swift/lookup.go
ianthehat: Been rebasing this lump for a while, maybe it arrived since, or got dropped in one of the conflict resolutions, will fix.

internal/sidekick/parser/auto_populated.go
ianthehat: Yeah, the rules are weird, the same operator can make different spacing decisions even within a single statement.

coryan (approved):

PR #5555, opened 2026-04-23
fix(tool/cmd/migrate): add hardcoded transport override for Java

### LINKED ISSUE #5554: java: java-alloydb-connectors repo-metadata transport diff
Using default transport grpc+rest, generated repo-metadata has ` "transport": "both",`, expecting ` "transport": "grpc",`.

Note `java-alloydb-connectors` is a proto only library. So the transport setting does not actually affects generated code other than .repo-metadata.json and README.

Here is current config for this library:
```
  - name: alloydb-connectors
    version: 0.67.0-SNAPSHOT
    apis:
      - path: google/cloud/alloydb/connectors/v1
      - path: google/cloud/alloydb/connectors/v1beta
      - path: google/cloud/alloydb/connectors/v1alpha
    java:
      api_id_override: connectors.googleapis.com
      api_description_override: AlloyDB is a fully-managed, PostgreSQL-compatible database for demanding transactional workloads. It provides enterprise-grade performance and availability while maintaining 100% compatibility with open-source PostgreSQL.
      name_pretty_override: AlloyDB connectors
      java_apis:
        - path: google/cloud/alloydb/connectors/v1
          proto_only: true
          samples: false
        - path: google/cloud/alloydb/connectors/v1alpha
          proto_only: true
          samples: false
        - path: google/cloud/alloydb/connectors/v1beta
          proto_only: true
          samples: false
      product_documentation_override: https://cloud.google.com/alloydb/docs
      rest_documentation: https://cloud.google.com/alloydb/docs/reference/rest
```

zhumin8:
Adds a hardcoded transport override for `alloydb-connectors`. This is used only for .repo-metadata.json and documentation to minimize generation diff, no affect on Java code generation because this is an proto-only library. 
More context in #5554.

Fix #5554

### DIFF
--- a/tool/cmd/migrate/java_module.go
+++ b/tool/cmd/migrate/java_module.go
@@ -110,6 +110,10 @@ var (
 			gapicArtifactID: "google-cloud-storage-control",
 		},
 	}
+
+	javaTransportOverrides = map[string]string{
+		"alloydb-connectors": "grpc",
+	}
 )
 
 type javaArtifactOverrides struct {
--- a/tool/cmd/migrate/java_module.go
+++ b/tool/cmd/migrate/java_module.go
@@ -113,7 +113,7 @@ var (
 
 	javaTransportOverrides = map[string]string{
 		//This is added here instead of sdk.yaml change because this is
-		//a proto-only library and transport affect on Java code generation.
+		//a proto-only library and transport does not affect Java code generated.
 		"alloydb-connectors": "grpc",
 	}
 )
--- a/tool/cmd/migrate/java_module.go
+++ b/tool/cmd/migrate/java_module.go
@@ -112,6 +112,8 @@ var (
 	}
 
 	javaTransportOverrides = map[string]string{
+		//This is added here instead of sdk.yaml change because this is
+		//a proto-only library and transport affect on Java code generation.
 		"alloydb-connectors": "grpc",
 	}
 )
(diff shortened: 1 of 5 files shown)


tool/cmd/migrate/java_module.go
```diff
@@ -110,6 +110,10 @@ var (
 			gapicArtifactID: "google-cloud-storage-control",
 		},
 	}
+
```
noahdietz, line 113: Nit: Can you add a comment similar to the one in the PR description as to what this is for and why its OK?

tool/cmd/migrate/java_module.go
```diff
@@ -110,6 +110,10 @@ var (
...
+	javaTransportOverrides = map[string]string{
```
noahdietz, line 114: Is this something we will need to fix/cleanup somehow in the future? Not sure if it is worth filing a TODO for or not, I'll let you be the judge.

zhumin8: Good point. Added a issue link to the config (internal/config/language.go).

noahdietz:
Looks like the TODO snuck its way into the comments that go-generate consumes

noahdietz:
> Looks like the TODO snuck its way into the comments that go-generate consumes

Consider adding it after the property so it doesn't show up in the generated documentation? I'm not sure how to make it not appear there :)

zhumin8:
updated the doc. Not terrible to have TODO there, we already got a bunch.

noahdietz (approved):

PR #5553, opened 2026-04-23
feat(dart): callbacks in fakes

brianquinlan:
Part of https://github.com/googleapis/google-cloud-dart/issues/246

Change the generated fakes so that they can be used either by subclassing or by providing a callback. For example:

```dart
/// Testing fake for [Echo].
base class FakeEcho implements Echo {
  final Future<EchoResponse> Function(EchoRequest request)? _echo;
  @override
  Uri get _endPoint => throw UnsupportedError('_endPoint');
  @override
  ServiceClient get _client => throw UnsupportedError('_client');

  bool isClosed = false;

  FakeEcho({
    Future<EchoResponse> Function(EchoRequest request)? echo,
  }) : _echo = echo;

  /// This method simply echoes the request. This method showcases unary RPCs.
  ///
  /// Throws a [http.ClientException] if there were problems communicating with
  /// the API service. Throws a [ServiceException] if the API method failed for
  /// any reason.
  @override
  Future<EchoResponse> echo(EchoRequest request) async {
    if (isClosed) {
      throw StateError('Service is closed');
    }

    if (_echo case final echo?) {
      return echo(request);
    }
    throw UnsupportedError('echo');
  }
}
```

There are tests for such a service [here](https://github.com/googleapis/google-cloud-dart/pull/248/changes#diff-0c1c3a8f36842aad8975f9e9c73839007ce62c615c6217a41955911e6b86e9b9).

PR #5572, opened 2026-04-24
chore: remove accidentally-included binary from root directory

jskeet:
Remove a binary accidentally added in #5553, and add appropriate .gitignore lines to avoid it happening again.


brianquinlan (approved):
Sorry about that and thanks for fixing!

Can't you just use the bare names (e.g. `legacyautomation`) so that they won't be included regardless of directory level?

jskeet:
> Can't you just use the bare names (e.g. legacyautomation) so that they won't be included regardless of directory level?

I would *expect* that to stop the directories themselves (internal/librarian etc) being included, which would be a Bad Thing.

Issue #5465, opened 2026-04-18
python: remove issue_tracker_override


julieqiu:
https://github.com/googleapis/google-cloud-python/blob/fa20a740b15accf2b1ae18a9ac20b75f006dbcad/librarian.yaml currently contains a large number of issue_tracker_override fields.

Since google-cloud-python is the only supported repository, we do not need this override. In almost all (if not all) cases, they are set to https://github.com/googleapis/google-cloud-python/issues.

jskeet:
One problem that's come up with this in Python is with Cloud DNS. It's a handwritten library, so there's nowhere in sdk.yaml to put the (genuinely language agnostic) link to the issue tracker, but we *do* currently use it in the top-level readme, and that saves our engineers time by avoiding (some) people filing bugs directly in the Python repo.

Some options:

- Rename `issue_tracker_override` to `issue_tracker`, and only allow it for handwritten libraries - downside is that we could end up with information duplication across languages. I don't know how many other languages *have* a Cloud DNS client, but if this occurs for other APIs, it could be relevant across languages.
- Have some sort of "notional" API path for Cloud DNS. (I don't generally like data lying about what it represents though.)

jskeet:
There are currently four overrides:

- google-cloud-dns: keep, as it's a handwritten library
- google-cloud-runtimeconfig: keep, as it's a handwritten library
- google-cloud-compute-v1beta: now in sdk.yaml; will remove when we've upgraded librarian
- google-cloud-storage: now in sdk.yaml; will remove when we've upgraded librarian

Once we're down to just the first two, I'll modify the field comment in PythonPackage to explain why it still exists, and close this issue as "done as far as we can reasonably do".

PR #5573, opened 2026-04-24
chore(internal/serviceconfig): add documentation_uri values to sdk.yaml

jskeet:
Documentation URIs are added to sdk.yaml, based on Python overrides in the google-cloud-python librarian.yaml, as they're language-agnostic values. Two of these are just generic cloud.google.com URIs, where there is nothing more appropriate to specify.

Towards #5465

### DIFF
--- a/internal/serviceconfig/sdk.yaml
+++ b/internal/serviceconfig/sdk.yaml
@@ -955,6 +955,7 @@
   release_level:
     java: preview
 - path: google/cloud/common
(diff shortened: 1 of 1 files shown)


zhumin8 (approved):
Note to self: java has "product_documentation_override" today in migrated [librarian.yaml](https://paste.googleplex.com/6109015173103616#l=1458). Will move and unify after migration filed https://github.com/googleapis/librarian/issues/5585

PR #5574, opened 2026-04-24
feat(internal/librarian/python): derive Python GAPIC generator options

jskeet:
The GAPIC generator options for package-warehouse-name, python-gapic-namespace and python-gapic-name are *always* specified to the generator. They are derived if they're not specified explicitly in librarian.yaml. This makes the inference algorithm in the GAPIC generator irrelevant.

This change is a no-op for the current librarian.yaml for Python, as we explicitly specify *all* options. After updating to a version of librarian.yaml that includes this code, we can then drop options that already have the same value that would be inferred.

Towards #5163

### DIFF
--- a/internal/librarian/python/clean.go
+++ b/internal/librarian/python/clean.go
@@ -24,6 +24,7 @@ import (
 	"strings"
 
 	"github.com/googleapis/librarian/internal/config"
+	"github.com/googleapis/librarian/internal/serviceconfig"
 )
 
 // gapicGenerationInfo contains useful information about the expected directory
@@ -46,7 +47,6 @@ type gapicGenerationInfo struct {
 const neutralSourcePlaceholder = "{neutral-source}"
 
 var (
-	errBadAPIPath               = errors.New("invalid API path")
 	errNoCommonGAPICFilesConfig = errors.New("when cleaning a GAPIC package, a config with common GAPIC paths must be provided")
 	// versionedGAPICRelativePathsToClean is the set of paths to remove for each
 	// API-versioned GAPIC output directory, relative to that directory.
@@ -129,10 +129,7 @@ func cleanProtoOnly(api *config.API, lib *config.Library) error {
 // be completely deleted. Likewise the documentation directory (e.g.
 // "docs/run_v2") will be completely deleted.
 func cleanGAPIC(api *config.API, lib *config.Library) error {
-	generationInfo, err := deriveGAPICGenerationInfo(api, lib)
-	if err != nil {
-		return err
-	}
+	generationInfo := deriveGAPICGenerationInfo(api, lib)
 	// Unusual, but it does happen, e.g. google/shopping/type and
 	// google/apps/script/type/{xyz}. We'll delete files as "common" GAPIC
 	// files instead.
@@ -152,10 +149,7 @@ func cleanGAPIC(api *config.API, lib *config.Library) error {
 // cleanGAPICCommon cleans the common output created for packages containing
 // any GAPIC libraries.
 func cleanGAPICCommon(lib *config.Library) error {
-	apiInfo, err := deriveGAPICGenerationInfo(lib.APIs[0], lib)
-	if err != nil {
-		return err
-	}
+	apiInfo := deriveGAPICGenerationInfo(lib.APIs[0], lib)
 	if lib.Python == nil {
 		return errNoCommonGAPICFilesConfig
 	}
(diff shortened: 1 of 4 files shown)


internal/librarian/python/generate.go
```diff
@@ -470,3 +480,44 @@ func DefaultLibraryName(api string) string {
...
+// deriveGAPICNamespace derives the value to pass as python-gapic-namespace when
+// it's not specified explicitly. This is the first two components of the API
+// path, dot-separated.
+func deriveGAPICNamespace(path string) string {
+	parts := strings.Split(path, "/")
+	if len(parts) < 2 {
+		return path
+	}
+	return parts[0] + "." + parts[1]
+}
```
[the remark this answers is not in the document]
jskeet: Done.

internal/librarian/python/generate.go
```diff
@@ -470,3 +480,44 @@ func DefaultLibraryName(api string) string {
...
+	version := serviceconfig.ExtractVersion(path)
+	if version != "" {
+		path = strings.TrimSuffix(path, version)
+	}
+	derivedNamespace := deriveGAPICNamespace(path)
+	path = strings.TrimPrefix(path, strings.ReplaceAll(derivedNamespace, ".", "/"))
+	path = strings.TrimSuffix(path, version)
+	path = strings.Trim(path, "/")
+	return strings.ReplaceAll(path, "/", "_")
+}
```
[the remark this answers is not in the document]
jskeet: Will defer to @noahdietz on this.

internal/librarian/python/clean.go
```diff
@@ -208,59 +202,27 @@ func deleteUnlessKept(lib *config.Library, path string) error {
...
 	}
-	var rootDir string
-	if namespace == "" {
-		rootDir = strings.Join(splitPath[:len(splitPath)-1], "/")
-	} else {
-		rootDir = strings.ReplaceAll(namespace, ".", "/")
-	}
-	if gapicName == "" {
-		gapicName = splitPath[len(splitPath)-1]
+	name, gotName := findOption(options, gapicNameOption)
```
noahdietz: Nit: I'd rename `gotName` to `ok` or something less similar to the testing pattern of naming the "actual" value with `got` :)

`ok` is relatively common identifier for the "did it work?" signal of a `(type, bool)` return signature, I think it reads fine. Alternatively, `found` might be more explicit. Either way.

internal/librarian/python/clean.go
```diff
@@ -208,59 +202,27 @@ func deleteUnlessKept(lib *config.Library, path string) error {
...
 	}
-	namespace := findOptArg(api, lib.Python, "python-gapic-namespace")
-	gapicName := findOptArg(api, lib.Python, "python-gapic-name")
-
-	lastElement := splitPath[len(splitPath)-1]
-	version := ""
-	if strings.HasPrefix(lastElement, "v") {
-		version = lastElement
-		splitPath = splitPath[:len(splitPath)-1]
+	namespace, gotNamespace := findOption(options, gapicNamespaceOption)
```
noahdietz: Same here s/`gotName`/`ok`/

internal/librarian/python/generate.go
```diff
@@ -304,15 +308,21 @@ func createProtocOptions(api *config.API, library *config.Library, googleapisDir
...
-	addTransport := true
-	for _, opt := range opts {
-		if strings.HasPrefix(opt, "transport=") {
-			addTransport = false
-		}
-	}
 	// Add transport option, if we haven't already got it.
-	if addTransport {
-		opts = append(opts, fmt.Sprintf("transport=%s", transport))
+	if _, gotOption := findOption(opts, transportOption); !gotOption {
```
noahdietz: ditto re: s/`gotOption`/`ok`/

internal/librarian/python/generate.go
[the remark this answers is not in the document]
noahdietz: I think the naming is fine - we aren't getting a property of a struct or type. If it were a property we were accessing or deriving, I'd say we should go with something like `gapicName` (unexported).

I'm not sure about the "previous implementation" comment - do we have a test covering the case it mentions to validate the desired behavior or not? Something like `google/spanner/v1` I think?

internal/librarian/python/clean.go
jskeet: Done (although we're then redeclaring it, of course...)

internal/librarian/python/clean.go
jskeet: Done.

internal/librarian/python/generate.go
jskeet: Done.

jskeet: Done.

internal/librarian/python/generate.go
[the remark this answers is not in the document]
jskeet: I'm not sure what Gemini is talking about now either, but we've got a test for grafeas/v1

internal/librarian/python/generate_test.go
```diff
@@ -1328,6 +1352,111 @@ func TestCreateRepoMetadata_Error(t *testing.T) {
...
+		})
+	}
+}
+
+func TestFindOption(t *testing.T) {
+	for _, test := range []struct {
+		name      string
+		options   []string
+		wantValue string
+		wantFound bool
```
jskeet, line 1422: I've renamed these to wantOk and gotOk as well.

noahdietz (approved):

PR #5588, opened 2026-04-24
feat(internal/librarian/java): add skip_proto_class_generation config to skip proto generation only

### LINKED ISSUE #5380: java: generate diff for aiplatform

zhumin8:
Adds a new `skip_proto_class_generation` config in Java to specify protos that needs to be skipped in proto generation but included in grpc- and gapic- modules generations.  The equivalent logic is achieved in hermetic build with these proto path hardcoded in generate_library.sh [section 3](https://github.com/googleapis/google-cloud-java/blob/3deec8a0b4cb24a88868b0e830f83e5c6725e9bc/sdk-platform-java/hermetic_build/library_generation/generate_library.sh#L225-L266).

This is added for special case of `aiplatform` maintain same behavior with hermetic build. After migration, should re-evaluate if moving these to `excluded_protos` instead is enough (https://github.com/googleapis/google-cloud-java/issues/12817).

Fix #5380


internal/librarian/java/generate_test.go
```diff
@@ -636,6 +636,105 @@ func TestGenerate_Logic(t *testing.T) {
 	}
 }
 
+func TestGenerate_ProtoExclusion(t *testing.T) {
+	var capturedArgs [][]string
+	oldRunProtoc := runProtoc
```
julieqiu: just call protoc directly instead of mocking

internal/librarian/java/generate_test.go
```diff
@@ -636,6 +636,105 @@ func TestGenerate_Logic(t *testing.T) {
...
+			break
+		}
+	}
+	if !found {
+		t.Errorf("Step 2 (gRPC) should include resources.proto, but it was not found in args")
+	}
+	// Verify Step 3 (GAPIC) includes resources.proto
+	gapicArgs := capturedArgs[2]
+	found = false
+	for _, arg := range gapicArgs {
```
julieqiu: you do this 3 times in this function. use a helper function like:

```
containsArg := func(args []string, target string) bool {
	for _, arg := range args {
		if strings.Contains(arg, target) {
			return true
		}
	}
	return false
}

if containsArg(capturedArgs[0], "resources.proto") {
	t.Errorf("Step 1 (proto) should exclude resources.proto")
}
if !containsArg(capturedArgs[1], "resources.proto") {
	t.Errorf("Step 2 (gRPC) should include resources.proto")
}
// ...
```

julieqiu (approved):

PR #5593, opened 2026-04-24
fix(internal/librarian/java): derive additional protos from OmitCommonResources and AdditionalProtos

zhumin8:
Update generate logic to respect OmitCommonResources config. Also update validate logic to return err with librarian tidy when both OmitCommonResources is true and AdditionalProtos includes "google/cloud/common_resources.proto".
This is a follow-up change to https://github.com/googleapis/librarian/pull/5592.

Fix #5225

### DIFF
--- a/internal/librarian/java/defaults.go
+++ b/internal/librarian/java/defaults.go
@@ -61,18 +61,35 @@ var (
 	// ErrInvalidDistributionName is returned when a distribution name override
 	// is incorrectly formatted.
 	ErrInvalidDistributionName = fmt.Errorf("invalid distribution name override")
+
+	// ErrOmitCommonResourcesConflict is returned when OmitCommonResources is true
+	// but common_resources.proto is also explicitly listed in AdditionalProtos.
+	ErrOmitCommonResourcesConflict = fmt.Errorf("conflict: OmitCommonResources is true but google/cloud/common_resources.proto is explicitly listed in AdditionalProtos")
 )
 
 // Validate checks that the Java-specific configuration for a library is
 // correctly formatted. It ensures that the distribution name override
-// contains exactly two parts separated by a colon.
+// contains exactly two parts separated by a colon, and that there are no
+// conflicts in common resources configuration.
 func Validate(library *config.Library) error {
-	if library.Java == nil || library.Java.DistributionNameOverride == "" {
+	if library.Java == nil {
 		return nil
 	}
-	parts := strings.Split(library.Java.DistributionNameOverride, ":")
-	if len(parts) != 2 || parts[0] == "" || parts[1] == "" {
-		return fmt.Errorf("%w: %s: want \"groupId:artifactId\", got %q", ErrInvalidDistributionName, library.Name, library.Java.DistributionNameOverride)
+	if library.Java.DistributionNameOverride != "" {
+		parts := strings.Split(library.Java.DistributionNameOverride, ":")
+		if len(parts) != 2 || parts[0] == "" || parts[1] == "" {
+			return fmt.Errorf("%w: %s: want \"groupId:artifactId\", got %q", ErrInvalidDistributionName, library.Name, library.Java.DistributionNameOverride)
+		}
+	}
+	for _, javaAPI := range library.Java.JavaAPIs {
+		if !javaAPI.OmitCommonResources {
+			continue
+		}
+		for _, p := range javaAPI.AdditionalProtos {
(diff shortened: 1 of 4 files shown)


internal/librarian/java/generate_test.go
```diff
@@ -832,3 +832,62 @@ func TestFilterProtos(t *testing.T) {
...
+			name: "omitted via flag with other additional protos",
+			javaAPI: &config.JavaAPI{
+				OmitCommonResources: true,
+				AdditionalProtos:    []string{"other.proto"},
+			},
+			want: []string{
+				filepath.Join(googleapisDir, "other.proto"),
+			},
+		},
+	} {
```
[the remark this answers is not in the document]
zhumin8: This is really should be an invalid config. Updated librarian tidy to return err on validatation step.

internal/librarian/java/generate.go
```diff
@@ -156,6 +155,23 @@ func generateAPI(ctx context.Context, cfg *config.Config, api *config.API, libra
...
+		additionalProtos = append(additionalProtos, filepath.Join(googleapisDir, commonResourcesProto))
+	}
+	for _, p := range javaAPI.AdditionalProtos {
+		if p == commonResourcesProto {
+			continue
+		}
+		additionalProtos = append(additionalProtos, filepath.Join(googleapisDir, filepath.FromSlash(p)))
+	}
+	return additionalProtos
+}
```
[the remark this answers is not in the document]
zhumin8: same as below

internal/librarian/java/defaults.go
```diff
@@ -61,18 +61,35 @@ var (
...
+	// but common_resources.proto is also explicitly listed in AdditionalProtos.
+	ErrOmitCommonResourcesConflict = fmt.Errorf("conflict: OmitCommonResources is true but google/cloud/common_resources.proto is explicitly listed in AdditionalProtos")
 )
 
 // Validate checks that the Java-specific configuration for a library is
 // correctly formatted. It ensures that the distribution name override
-// contains exactly two parts separated by a colon.
+// contains exactly two parts separated by a colon, and that there are no
+// conflicts in common resources configuration.
 func Validate(library *config.Library) error {
```
julieqiu, line 74: where is this being called?

internal/librarian/java/defaults.go
```diff
@@ -61,18 +61,35 @@ var (
...
-	if library.Java == nil || library.Java.DistributionNameOverride == "" {
+	if library.Java == nil {
 		return nil
 	}
-	parts := strings.Split(library.Java.DistributionNameOverride, ":")
-	if len(parts) != 2 || parts[0] == "" || parts[1] == "" {
-		return fmt.Errorf("%w: %s: want \"groupId:artifactId\", got %q", ErrInvalidDistributionName, library.Name, library.Java.DistributionNameOverride)
+	if library.Java.DistributionNameOverride != "" {
+		parts := strings.Split(library.Java.DistributionNameOverride, ":")
```
julieqiu, line 79: I would use strings.SplitN(library.Java.DistributionNameOverride, ":", 2)

then you are guaranteed two parts

internal/librarian/java/defaults_test.go
```diff
@@ -48,7 +50,9 @@ func TestFill(t *testing.T) {
 			want: &config.Library{
 				Name:   "secretmanager",
 				Output: "custom-output",
-				Java:   &config.JavaModule{},
+				Java: &config.JavaModule{
+					JavaAPIs: nil,
```
julieqiu: you can omit this line since it is nil

julieqiu (approved):

internal/librarian/java/defaults.go
zhumin8: In tidy validateLibraries here: https://github.com/googleapis/librarian/blob/8c74d3df23146b1c0744bcad21c968cb792b305e/internal/librarian/tidy.go#L139

internal/librarian/java/defaults.go
zhumin8: This is the validate logic, and expect to get an error if this override is malformed.
In actual logic that works out the maven coordinates from it, uses  strings.SplitN(x, ":", 2)
https://github.com/googleapis/librarian/blob/8c74d3df23146b1c0744bcad21c968cb792b305e/internal/librarian/java/maven_coordinate.go#L78

PR #5594, opened 2026-04-24
feat(internal/librarian/python): remove redundant GAPIC options from Python configuration

jskeet:
The following options can be derived during generation, so don't need to be specified if the value currently in the config matches the value which would be derived:

- warehouse-package-name
- python-gapic-namespace
- python-gapic-name

Towards https://github.com/googleapis/librarian/issues/5163

### DIFF
--- a/internal/librarian/python/tidy.go
+++ b/internal/librarian/python/tidy.go
@@ -1,0 +1,68 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


internal/librarian/python/tidy.go
```diff
@@ -0,0 +1,68 @@
...
+	return lib
+}
+
+// tidyAPI removes redundant OptArgsByAPI values for a single API.
+func tidyAPI(lib *config.Library, api *config.API) *config.Library {
+	if isProtoOnly(api, lib) || lib.Python == nil {
+		return lib
+	}
+	pythonPackage := lib.Python
+	if pythonPackage.OptArgsByAPI == nil {
```
noahdietz, line 38:
```suggestion
	if len(pythonPackage.OptArgsByAPI) == 0 {
```

`len` check on map/slice is more robust - it captures set-empty and default nil

noahdietz (approved):

internal/librarian/python/tidy.go
jskeet: But in this case, I think we actually *don't* want to capture set-empty. If someone has a manually-update config with an empty set of args, we don't want to return it unchanged - we want to go through the code below, and set it to nil so it'll be removed from the file.

Could add a comment to that effect though.

noahdietz: nvm we want to tidy an empty slice if the key is present, so let it fall through

PR #5632, opened 2026-04-24
refactor(internal/sidekick/gcloud): generalize gcloud command tracks and update API versioning

### LINKED ISSUE #4530: surfer: Implement Multi-Track Surface Generation
Surfer must generate commands for services that span multiple release tracks (e.g., Alpha, Beta, GA). Currently, processing multiple versions of a service sequentially results in files like `__init__.py` being overwritten, losing the track definitions of previously processed versions. We need to implement a mechanism (based on the outcome of the `gcloud.yaml` handling spike) to correctly output multi-track YAML partials (`_alpha.yaml`, `_beta.yaml`) and aggregate all track classes into shared Python initializers.

*This issue aims to address the core requirements outlined in existing issues **#3037** (support multiple release tracks) and **#3289** (allow config override for tracks) by implementing the new GcloudModel Builder/Writer architecture.*

- Update the generation pipeline to support building an aggregated model of tracks before executing file writes.
- Update the Python scaffolding templates to generate standard boilerplate, multi-track `@base.ReleaseTracks` decorators, and dynamic imports for `extensions` and base groups.

## Relevant files
- `internal/surfer/gcloud/builder.go` (or equivalent parsing logic)
- `internal/surfer/gcloud/generate.go`

sarahheacock:
* Simplified surface generation given we only generate one api version at a time
* Fixed api version mapping in the command output
* Enabled multi_service golden and fixed class naming for it to pass
* Updated targets to only output one api version at a time

Fix #4530


internal/sidekick/gcloud/provider/resource.go
```diff
@@ -335,20 +336,42 @@ func GetLiteralSegments(raw []api.PathSegment) []string {
...
+// GetResourceDisplayNames returns the singular and plural names for a resource,
+// prioritizing explicit annotations in the resource definition and falling back
+// to the message name to provide more descriptive names for help text and classes.
+func GetResourceDisplayNames(model *api.API, methodPath []string) (singular string, plural string) {
+	res := GetResourceForPath(model, methodPath)
+	if res == nil {
+		// Fallback if no resource found
+		seg := methodPath[len(methodPath)-1]
+		singular = strings.TrimSuffix(seg, "s")
+		plural = seg
```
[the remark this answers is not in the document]
sarahheacock: Simplified logic from original gen_sfc

internal/sidekick/gcloud/surface_writer.go
```diff
@@ -25,166 +25,103 @@ import (
...
-	name := groupName(b)
-	if name == "" {
+func (w *surfaceWriter) writeGroup(g *CommandGroup, parentPath []string) error {
+	if g == nil {
 		return nil
 	}
 
-	moduleName := strcase.ToSnake(name)
-	groupDir := filepath.Join(outputDir, moduleName)
+	path := append(parentPath, strcase.ToSnake(g.FileName))
```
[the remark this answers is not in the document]
sarahheacock: Updated

internal/sidekick/gcloud/provider/service_info.go
```diff
@@ -21,25 +21,43 @@ import (
...
+	return parts[len(parts)-1]
+}
 
+// APIVersionFromModel extracts the API version from the API model's package name.
+func APIVersionFromModel(model *api.API) string {
+	return apiVersionFromPackage(model.PackageName)
+}
+
+// Tracks infers the release tracks from the API version string.
+// as mandated per AIP-185
```
jameslynnwu: maybe remove this line or delete the period from previous line

internal/sidekick/gcloud/provider/resource.go
```diff
@@ -335,20 +336,35 @@ func GetLiteralSegments(raw []api.PathSegment) []string {
...
 	}
-	return ""
+	return nil
+}
+
+// GetResourceTypeNames returns the singular and plural names for a resource.
+// It uses the resource definition's explicit `singular` and `plural` fields
+// in that order and falls back to the resource type name if neither is present.
+// Values must be explicit so if plural is not provided, we cannot determine it.
+func GetResourceTypeNames(model *api.API, methodPath []string) (singular string, plural string) {
```
jameslynnwu: I think ideally, we have two separate functions for singular and plural rather than one

Is this useful outside of gcloud command generation?

internal/sidekick/gcloud/provider/resource.go
```diff
@@ -335,20 +336,35 @@ func GetLiteralSegments(raw []api.PathSegment) []string {
...
+	res := GetResourceForPath(model, methodPath)
+	if res == nil {
+		return "", ""
```
jameslynnwu: // Values must be explicit so if plural is not provided, we cannot determine it.

Should this be an error then?

internal/sidekick/gcloud/provider/service_info.go
```diff
@@ -21,25 +21,43 @@ import (
 	"github.com/iancoleman/strcase"
 )
 
-// InferTrackFromPackage infers the release track from the proto package name.
-// as mandated per AIP-185
-// e.g. "google.cloud.parallelstore.v1beta" -> "beta".
-func InferTrackFromPackage(pkg string) string {
+const (
+	// ALPHA release track.
```
jameslynnwu: let's define a type for release tracks

internal/sidekick/gcloud/command.go
```diff
@@ -14,22 +14,21 @@
...
 // CommandGroup defines the metadata for a structural grouping of gcloud commands
 // (typically mapping to an API service or an AIP-122 Resource collection).
 // It is used to generate the Python `__init__.py` registration files.
 type CommandGroup struct {
-	Name     string
-	Path     []string
-	HelpText string
-	Commands map[string]*Command
-	Groups   map[string]*CommandGroup
+	ClassName string
```
jameslynnwu, line 39: how do you set the release track for a command group or a command?

internal/sidekick/gcloud/provider/method.go
```diff
@@ -275,3 +275,11 @@ func GetMethodHelpText(overrides *Config, method *api.Method, model *api.API) He
 		Examples:    examples,
 	}
 }
+
+// APIVersionFromMethod extracts the API version from the method's service package name.
+func APIVersionFromMethod(method *api.Method) string {
+	if method.Service != nil {
+		return apiVersionFromPackage(method.Service.Package)
+	}
+	return ""
```
jameslynnwu: should this be an error?

internal/sidekick/gcloud/provider/method.go
```diff
@@ -275,3 +275,11 @@ func GetMethodHelpText(overrides *Config, method *api.Method, model *api.API) He
 		Examples:    examples,
 	}
 }
+
+// APIVersionFromMethod extracts the API version from the method's service package name.
+func APIVersionFromMethod(method *api.Method) string {
+	if method.Service != nil {
```
jameslynnwu: when is method service nil?

sarahheacock: It shouldn't. But because it's a pointer and due to type safety, we have to do this. That being said, my hope is that the provider (and later sidekick by moving the code over) will be more agressive about requiring certain fields / preventing unnecessary nil checks / runtime errors

internal/sidekick/gcloud/provider/service_info.go
sarahheacock: It was there before but can remove

internal/sidekick/gcloud/provider/resource.go
sarahheacock: Sure, and I imagine it might be helpful but have refrained from editing sidekick until surfer is more stable

internal/sidekick/gcloud/provider/resource.go
sarahheacock: Fair question. Empty string is valid given some method patterns do not have a resource associated with it. Returning an error would just be noise. But would like for there to be a more established pattern here

internal/sidekick/gcloud/provider/service_info.go
sarahheacock: Did not know that was an option! Done!

internal/sidekick/gcloud/command.go
sarahheacock: "type Surface struct" set once at the root. When surfer is executed, it will have the same release tracks for all of the sub commands. This is because only one package (api version) is generated at time. More notes are in the internal bug ticket

internal/sidekick/gcloud/provider/method.go
sarahheacock: I think an error is fair here. Added but we should think more carefully about this

jameslynnwu (approved):

PR #5644, opened 2026-04-26
feat(tool/cmd/docgen): rewrite CLI help

### LINKED ISSUE #5638: librarian: improve package documentation
Improve documentation at https://github.com/googleapis/librarian/blob/main/cmd/librarian/doc.go

This shows up at https://pkg.go.dev/github.com/googleapis/librarian/cmd/librarian

julieqiu:
Rewrite the help text for each librarian subcommand. Each command now has a Description with a short explanation, concrete examples, and references to related commands. Regenerate the cmd/librarian package docs from that help text.

Refactor tool/cmd/docgen to parse `--help` output into structured fields and render a more readable document instead of embedding the raw help text.

Freeze the help text for cmd/legacylibrarian and cmd/legacyautomation, since they use the same tool but expect a different format, and those tools are getting decomissioned. 

Update the documentation cmd/librarianops with docgen as well.

Fixes https://github.com/googleapis/librarian/issues/5638

### DIFF
--- a/tool/cmd/docgen/main.go
+++ b/tool/cmd/docgen/main.go
@@ -28,6 +28,8 @@ import (
 	"regexp"
 	"strings"
 	"text/template"
+	"unicode"
+	"unicode/utf8"
 )
 
 const (
@@ -390,7 +392,8 @@ func sentenceCase(s string) string {
 	if s == "" {
 		return s
 	}
-	return strings.ToUpper(s[:1]) + s[1:]
+	r, size := utf8.DecodeRuneInString(s)
+	return string(unicode.ToUpper(r)) + s[size:]
 }
 
 func getCommandHelpText(command string) (string, error) {
--- a/tool/cmd/docgen/main.go
+++ b/tool/cmd/docgen/main.go
@@ -25,39 +25,34 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"text/template"
 )
 
 const (
-	librarianDesc = `Librarian CLI runs local workflow that 
-	adds, generates, updates and publishes client libraries.
+	librarianDesc = `Librarian manages Google Cloud client libraries. It runs a local workflow
+that onboards new APIs, generates client code, bumps versions, publishes
+releases, and tags release commits. Language-specific work, such as code
+generation, building, and testing, is delegated to per-language tooling.
+
+All behavior is driven by librarian.yaml at the root of the repository,
+whose schema is documented at
+https://github.com/googleapis/librarian/blob/main/doc/config-schema.md.
 
 Usage:
 
 	librarian <command> [arguments]
+
+Global flags:
+
+	--verbose, -v    enable verbose logging
 `
 	librarianopsDesc = `Librarianops orchestrates librarian operations across multiple repositories.
 
 Usage:
 
 	librarianops <command> [arguments]
-`
-	legacyLibrarianDesc = `Librarian manages Google API client libraries by automating onboarding,
-regeneration, and release. It runs language-agnostic workflows while
-delegating language-specific tasks—such as code generation, building, and
-testing—to Docker images.
-
-Usage:
-
-	librarian <command> [arguments]
-`
-	automationDesc = `Automation provides logic to trigger Cloud Build jobs that run Librarian commands for
-any repository listed in internal/legacylibrarian/legacyautomation/prod/repositories.yaml.
-
-Usage:
-
-	automation <command> [arguments]
 `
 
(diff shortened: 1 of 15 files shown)


tool/cmd/docgen/main.go
```diff
@@ -222,22 +229,173 @@ func buildCommandDocs(parentCommand string) ([]CommandDoc, error) {
...
+// GLOBAL OPTIONS are dropped (they are documented once at the package level).
+// The "--help, -h" flag is filtered out of OPTIONS.
+func parseHelp(help string) CommandDoc {
+	sections := splitSections(help)
+
+	var doc CommandDoc
+	if name, ok := sections["NAME"]; ok {
+		name = strings.TrimSpace(name)
+		if i := strings.Index(name, " - "); i >= 0 {
+			doc.Summary = strings.TrimSpace(name[i+3:])
```
sofisl, line 275: these two lines here seem to expect a very specific format. Should we provide a helpful error if `name` has no dashes, or if i+3 doesn't exist or somehow otherwise malformed?

tool/cmd/docgen/main.go
```diff
@@ -222,22 +229,173 @@ func buildCommandDocs(parentCommand string) ([]CommandDoc, error) {
...
+	if o, ok := sections["OPTIONS"]; ok {
+		doc.Flags = filterFlags(o)
+	}
+	return doc
+}
+
+// splitSections splits urfave/cli help text into sections keyed by their
+// uppercase header (NAME, USAGE, DESCRIPTION, OPTIONS, GLOBAL OPTIONS).
+func splitSections(help string) map[string]string {
+	matches := sectionRE.FindAllStringIndex(help, -1)
```
sofisl, line 296: should we make sure to normalize string case before checking matching? (toUpper?)

tool/cmd/docgen/main_test.go
```diff
@@ -0,0 +1,427 @@
...
+			want: map[string]string{},
+		},
+		{
+			name: "single section",
+			help: `USAGE:
+   librarian generate [options]
+`,
+			want: map[string]string{
+				"USAGE": "\n   librarian generate [options]\n",
+			},
```
sofisl, line 167: for completeness, should we try to nonconsecutive commands? (what if USAGE and OPTIONS are the only ones present?)

tool/cmd/docgen/main_test.go
```diff
@@ -0,0 +1,427 @@
...
+
+`,
+			want: []string{"generate"},
+		},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			got, err := extractCommandNames([]byte(test.helpText))
+			if test.wantErr {
+				if err == nil {
+					t.Fatal("expected error, got nil")
```
sofisl, line 428: just for my learning - this seems contrary to https://github.com/googleapis/librarian/blob/main/doc/howwewritego.md#use-tfatal-or-terror-for-simple-error-handling?

tool/cmd/docgen/main.go
```diff
@@ -222,22 +229,173 @@ func buildCommandDocs(parentCommand string) ([]CommandDoc, error) {
...
+	if min <= 0 {
+		return strings.Join(lines, "\n")
+	}
+	out := make([]string, len(lines))
+	for i, l := range lines {
+		if len(l) >= min {
+			out[i] = l[min:]
+		} else {
+			out[i] = l
+		}
```
sofisl, line 361: I think you could remove the if/else IIUC:

if len(l) < min {
        out[i] = l
        continue
    }
    out[i] = l[min:]
}

tool/cmd/docgen/main_test.go
sofisl: actually, i see this is a pattern now.

tool/cmd/docgen/main.go
julieqiu: urfave/cli always emits NAME:, USAGE:, etc. in uppercase (see https://pkg.go.dev/github.com/urfave/cli/v3), so a lower case format (i.e. Names) would mask format drift rather than help catch it

tool/cmd/docgen/main.go
julieqiu: added a comment on the format

if tagline doesn't exist, it is just an empty string

tool/cmd/docgen/main.go
julieqiu: fixed

tool/cmd/docgen/main_test.go
julieqiu: added a test

sofisl (approved):

Issue #4650, opened 2026-03-18
google-cloud-java: unify submodule versions


zhumin8:
Proposed previously in: [Java Librarian Onboarding Design](https://docs.google.com/document/d/1VuwQLH8KcPgbnbcFj-a9EBDz2FOoRsEbqtZFSAnN3us/edit?tab=t.0#heading=h.il6d2s21eeoi) and [Java client library artifacts with multiple major versions: java-workflows](https://docs.google.com/document/d/1y8MbPF1hwGliqXmkW42SmhWcU9IgBi4w4MVSIcYltpE/edit?resourcekey=0-ZcFkzXcJp6X22PWYF-TxkA&tab=t.0#heading=h.842gh0jp01v) and approved by @blakeli0.

Was previously deprioritized as librarian re-desgin. This is needed before Java migrate to librarian as a one time change.

zhumin8:
note: needs a snapshot version open up before doing this.

analyse on libraries that needs a version bump: 
https://paste.googleplex.com/6330588039610368
there are 8 google-cloud- modules that should be bypassed

Used [this script](https://paste.googleplex.com/5078192487530496) to test run, a total of 51 libraries needed to be changed.

zhumin8:
after https://github.com/googleapis/google-cloud-java/pull/13220, there are still these top  java-<name>  modules that contain submodules with different versions:

  -  java-bigquery :
      -  google-cloud-bigquery-jdbc  is at  0.11.0-SNAPSHOT  while the main library is at  2.67.0-SNAPSHOT . This is expected as the JDBC driver
      follows its own independent versioning scheme.
  -  java-bigtable :
      -  google-cloud-bigtable-emulator  and  google-cloud-bigtable-emulator-core  are at  0.216.0-SNAPSHOT .
      -  google-cloud-bigtable-test-proxy  is at  0.0.1-SNAPSHOT .
      - These are utility/testing tools and are versioned independently from the main client library ( 2.79.0-SNAPSHOT ).
  -  java-storage :
      -  storage-shared-benchmarking  is at  0.0.1-SNAPSHOT . This is a benchmarking utility and I assume is never released (not in versions.txt file).

zhumin8:
For generate migrate purpose:
- java-bigquery/google-cloud-bigquery-jdbc, java-bigtable/google-cloud-bigtable-emulator and java-bigtable//google-cloud-bigtable-emulator-core
[...]

PR #5660, opened 2026-04-27
feat(internal/librarian/java): appends to versions.txt when new modules is added in generate

### LINKED ISSUE #4927: java: add entry to versions.txt if new modules added
If generated new modules, add entries to versions.txt. This is needed during migration when release-please is still in use. 
- proto- and grpc- modules for new API versions, or 
- proto-, grpc- and gapic client modules for new API

zhumin8:
Appends version line to versions.txt file for each new module added in generate. The implememtation takes refactors existing logic for collecting modules to update pom.xml, and implements a `IdentifyMissingModules` to record artifacts that needs to be appended to versions.txt. Then as a monorepo level post generate task, append these to versions.txt which is at the repo root.

Context: versions.txt is a release-please file that tracks the released version and snapshot version for artifacts. Until Java onboards to librarian for releases, we need to update this versions.txt file when new artifacts is added by generate. This happens when new library is introduced or new version API path is added to existing libraries.

Fix #4927


internal/librarian/java/postgenerate.go
```diff
@@ -84,6 +90,29 @@ func PostGenerate(ctx context.Context, repoPath string, cfg *config.Config) erro
 	return nil
 }
 
+func appendVersions(repoPath string, versions []string) (err error) {
+	if len(versions) == 0 {
+		return nil
+	}
+	versionsPath := filepath.Join(repoPath, "versions.txt")
+	f, err := os.OpenFile(versionsPath, os.O_APPEND|os.O_WRONLY, 0644)
```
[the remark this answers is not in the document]
zhumin8: "If versions.txt does not exist, this function will return an error. " This is the expected behavior. versions.txt is already present in google-cloud-java. 

"checking for a trailing newline" good point. Will add check

internal/filesystem/filesystem.go
```diff
@@ -82,3 +82,48 @@ func CopyFile(src, dest string) error {
 func Unzip(ctx context.Context, src, dest string) error {
 	return command.Run(ctx, "unzip", "-q", "-o", src, "-d", dest)
 }
+
+// AppendLines appends the given lines to an existing file, ensuring that it
```
sofisl: Is this function for linting? If so, a few questions (again mostly for me to learn):
1. Is there no go linting tool that adds a newline automatically, if that's what we're trying to account for?
2. Is there a go library that adds a newline?

internal/filesystem/filesystem_test.go
```diff
@@ -290,3 +290,67 @@ func createZip(t *testing.T, path string, files map[string]string, modes map[str
...
+			want:    "content",
+		},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			path := filepath.Join(t.TempDir(), "test.txt")
+			if err := os.WriteFile(path, []byte(test.initial), 0644); err != nil {
+				t.Fatal(err)
+			}
+			if err := AppendLines(path, test.lines); err != nil {
+				t.Fatalf("AppendLines() error = %v", err)
```
julieqiu: use t.Fatal: https://github.com/googleapis/librarian/blob/main/doc/howwewritego.md#use-tfatal-or-terror-for-simple-error-handling

internal/filesystem/filesystem.go
```diff
@@ -82,3 +82,48 @@ func CopyFile(src, dest string) error {
...
+// ends with a newline character before appending. It returns an error if the
+// file does not exist.
+func AppendLines(path string, lines []string) (err error) {
```
julieqiu: move this to internal/librarian/java, since that's the only place where this is called

julieqiu: what about simplifying this function to this:

```
  func AppendLines(path string, lines []string) error {
        if len(lines) == 0 {
                return nil
        }
        existing, err := os.ReadFile(path)
        if err != nil {
                return err
        }

        var buf bytes.Buffer
        buf.Write(existing)
        if len(existing) > 0 && existing[len(existing)-1] != '\n' {
                buf.WriteByte('\n')
        }
        for _, line := range lines {
                buf.WriteString(line)
                buf.WriteByte('\n')
        }
        return os.WriteFile(path, buf.Bytes(), 0644)
  }
```

This will rewrite the whole file each time, but for a small file like versions.txt it seems cleaner

What I don't quite understand is what we need to append the newline at all though?

internal/filesystem/filesystem_test.go
```diff
@@ -290,3 +290,67 @@ func createZip(t *testing.T, path string, files map[string]string, modes map[str
...
+				t.Fatal(err)
+			}
+			if diff := cmp.Diff(test.want, string(got)); diff != "" {
+				t.Errorf("mismatch (-want +got):\n%s", diff)
+			}
+		})
+	}
+}
+
+func TestAppendLines_Error(t *testing.T) {
```
julieqiu: this doesn't seem like a useful test

internal/librarian/java/pom.go
```diff
@@ -45,7 +45,9 @@ const (
 	managedModulesEndMarker        = "<!-- {x-generated-modules-end} -->"
 )
 
-var errTargetDir = errors.New("target directory does not exist")
+var (
```
julieqiu: nit: if you have one error, the previous way of having it on one line is actually more idiomatic

internal/librarian/java/pom.go
```diff
@@ -100,33 +104,42 @@ func syncPOMs(library *config.Library, libraryDir, monorepoVersion string, metad
...
 			}
 		case parentPOMTemplateName:
 			if err := updateParentPOM(pomPath, m.templateData.(bomParentPOMData)); err != nil {
-				return fmt.Errorf("failed to update parent pom.xml %s: %w", m.artifactID, err)
+				return nil, fmt.Errorf("failed to update parent pom.xml %s: %w", m.artifactID, err)
 			}
 		}
 	}
-	return nil
+
```
julieqiu, line 246: nit: remove extra newline

internal/librarian/java/pom.go
```diff
@@ -100,33 +104,42 @@ func syncPOMs(library *config.Library, libraryDir, monorepoVersion string, metad
 		}
 	}
 
+	releasedVersion, err := deriveLastReleasedVersion(library.Version)
```
julieqiu: what if a library has no versions to parse?

internal/librarian/java/postprocess_test.go
```diff
@@ -404,10 +405,27 @@ func TestPostProcessLibrary_ErrorCase(t *testing.T) {
...
+			got, err := postProcessLibrary(t.Context(), params)
 			if !errors.Is(err, test.wantErr) {
 				t.Fatalf("error = %v, wantErr %v", err, test.wantErr)
 			}
+			if err == nil && test.name == "success" {
+				libCoords := DeriveLibraryCoordinates(library)
+				apiCoords := DeriveAPICoordinates(libCoords, "v1", &config.JavaAPI{})
+				released, _ := deriveLastReleasedVersion(library.Version)
+				want := []string{
+					fmt.Sprintf("%s:%s:%s", apiCoords.Proto.ArtifactID, released, library.Version),
```
julieqiu: since this is repeated, use an inline helper like:

formatVersionEntry(artifactID, released, snapshot string) string

internal/filesystem/filesystem.go
zhumin8: 1. No, this function is used to append new entries to versions.txt file. versions.txt exists in google-cloud-java, and although not encouraged, can be edited by human and I believe does not have a check for trailing empty line.
2. I am not aware of. I think we need to read in the file and write.

internal/filesystem/filesystem.go
zhumin8: > why we need to append the newline at all though?

It's more of a safety check. Although not encouraged, this versions.txt can be edited by human and I believe there is no check for trailing empty line in google-cloud-java. Alternatively, we can add a dedicated workflow in google-cloud-java to check for it. But I think it's easier to check when attempting to update.

internal/librarian/java/pom.go
zhumin8: A Java library should always have version.  librarian add sets a default version https://github.com/googleapis/librarian/blob/b92da7279c9783c8ea9fe7d2b7c494e85f11dd10/internal/librarian/java/add.go#L37

internal/librarian/generate.go
```diff
@@ -245,15 +245,18 @@ func generateLibraries(ctx context.Context, cfg *config.Config, libraries []*con
 		}
 		return g.Wait()
 	case config.LanguageJava:
+		var allNewVersions []string
 		for _, library := range libraries {
-			if err := java.Generate(ctx, cfg, library, src); err != nil {
+			newVersions, err := java.Generate(ctx, cfg, library, src)
```
julieqiu: we should try to keep the same API interface across languages for Generate. what about moving the post generate logic inside java.Generate?

internal/librarian/java/generate.go
```diff
@@ -46,52 +46,54 @@ var (
 	errBOMVersionMissing = errors.New("libraries bom version not found in config")
 )
 
-// Generate generates a Java client library.
-func Generate(ctx context.Context, cfg *config.Config, library *config.Library, srcs *sources.Sources) error {
+// Generate generates a Java client library and returns a list of newly created
+// artifact version entries to be added to versions.txt.
```
suztomo: I see we will stop adding multiple lines into version txt for proto- and grpc- artifacts https://github.com/googleapis/librarian/issues/4650.

internal/librarian/java/generate.go
```diff
@@ -46,52 +46,54 @@ var (
 	errBOMVersionMissing = errors.New("libraries bom version not found in config")
 )
 
-// Generate generates a Java client library.
-func Generate(ctx context.Context, cfg *config.Config, library *config.Library, srcs *sources.Sources) error {
+// Generate generates a Java client library and returns a list of newly created
```
suztomo: Can you define the canonical way within Java to detect a library is new or not?

(I believe that the responsibility of detecting the condition is not in `Generate` function.)

internal/librarian/java/generate.go
zhumin8: Nope. https://github.com/googleapis/librarian/issues/4650 is to unify the versions. No changes to versions.txt structure. We can potentially do that, but it does not change that we still need append to versions.txt when new library onboards.

internal/librarian/java/generate.go
zhumin8: @julieqiu your comment https://github.com/googleapis/librarian/pull/5660/changes#r3156502889 is related, I'll bring this discussion together:

- Currently, Generate does post processing steps that generates pom.xml files when it is a new library (that is where new library/module is detected and bubbled up).
- I choose to change Generate to return these version entries and let `java.PostGenerate` write to versions.txt, because versions.txt is located at repo root, and it's natural to update it once for the whole generation. It should be technically possible to move inside java.Generate and workout the parent dir of outdir, but it feels a bit wrong as a per library function updates a repo level file, and may be problematic if we seek to do parallel in the future.
- By the time it gets to `java.PostGenerate`, pom.xml are already added, I don't see a way to identify which library is new. @suztomo any ideas you have in mind?

zhumin8: > we should try to keep the same API interface across languages for Generate. what about moving the post generate logic inside java.Generate?

I agree in general. But in this case, moving versions.txt update into java.Generate feels a bit hacky to me. java.Generate should only emit code within the library's output path. Plus this is a mid-term temporary change, we can remove it when onboarding to release.

> I believe that the responsibility of detecting the condition is not in Generate function

I may be misunderstanding your point, but it sounds to me you want a change in the general design. Currently Generate function generates for both existing library and newly added configs, so detecting "is it a new library" is part of its logic.

zhumin8:
Keeping this open, turning to draft now, I'll try to update this by EOD. Otherwise will close and re-open next week.

Update: open for review again. Refactored to not change java.Generate API interface. Introduced a separate func prior to generate to identify new artifacts.

internal/librarian/java/generate.go
zhumin8: After discussion in chat, I decided to refactor to: 
- identify new modules prior to Generate.
- keep API interface for Generate.
- still do update inside PostGenerate.

@julieqiu @suztomo Can you please take another look?

julieqiu (approved):
This seems kind of complicated, but approving and we can refactor in follow ups.

PR #5686, opened 2026-04-28
fix(internal/config): remove unused source.Branch

julieqiu:
The source.Branch config field is no longer be used.


internal/librarian/update_test.go
```diff
@@ -58,24 +57,19 @@ var (
...
-	discoveryBranch := determineBranch("discovery", conf.Sources.Discovery)
-	conformanceBranch := determineBranch("conformance", conf.Sources.Conformance)
-	protobufBranch := determineBranch("protobuf", conf.Sources.ProtobufSrc)
-	showcaseBranch := determineBranch("showcase", conf.Sources.Showcase)
+	// Update defaults to using the branch configured in [sourceRepos].
+	// We set up the test server handlers accordingly.
+	googleapisBranch := sourceRepos["googleapis"].Branch
+	discoveryBranch := sourceRepos["discovery"].Branch
+	protobufBranch := sourceRepos["protobuf"].Branch
+	showcaseBranch := sourceRepos["showcase"].Branch
```
zhumin8, line 65: I missed the previous changes in https://github.com/googleapis/librarian/pull/4969. Does this mean that release is only supporting release from main branches? Java LTS needs to be able to release from branches other than main.

julieqiu: Will we be generating from different branches of a repository? I assume that it will always be the same branch, just a different commit. In that case, I don't believe this should affect anything

zhumin8: Yes, we need to have ability to generate from LTS branches (rare case, but possible). Java cuts LTS branches, to be able to do a patch fix on it.

julieqiu: resolved over chat that this isn't an issue

zhumin8 (approved):

PR #5688, opened 2026-04-28
feat(gcloud): add GcloudSurface config and map to provider.Config

jameslynnwu:
Add support for GcloudSurface configuration in librarian.yaml.

Help text overrides are now supported in librarian.yaml for gcloud generation, allowing customization of method and field help text. Only MethodRules and FieldRules help text overrides are implemented, even though ServiceRules and MessageRules help text overrides are defined in gcloud.yaml schema, since they are not implemented in librarian yet.


doc/config-schema.md
```diff
@@ -223,6 +224,12 @@ This document describes the schema for the librarian.yaml.
 | `to` | string |  |
 | `wire_name` | string |  |
 
+## GcloudSurface Configuration
+
+| Field | Type | Description |
+| :--- | :--- | :--- |
+| `generate_operations` | bool (optional) | Indicates whether to generate top-level operations commands. |
```
julieqiu: Under what circumstances would we not want to generate opertaions?

internal/config/language.go
```diff
@@ -769,3 +769,9 @@ type NodejsAPI struct {
 	// Path is the source path.
 	Path string `yaml:"path,omitempty"`
 }
+
+// GcloudSurface contains gcloud-specific library configuration.
+type GcloudSurface struct {
+	// GenerateOperations indicates whether to generate top-level operations commands.
+	GenerateOperations *bool `yaml:"generate_operations,omitempty"`
```
julieqiu: rename this to `SkipOperations bool yaml:"skip_operations,omitempty"`, so that we don't need to manage three states. that way instead of:

`if *lib.Gcloud.GenerateOperations != false`

you can just do:

`if !lib.Gcloud.SkipOperations`

internal/config/config_test.go
```diff
@@ -434,3 +434,34 @@ libraries:
 		t.Errorf("mismatch (-want +got):\n%s", diff)
 	}
 }
+
+func TestGcloudConfig_GenerateOperations(t *testing.T) {
```
julieqiu: I don't think you need to add a test for this, since we are just marshaling and unmarshaling a YAML file

doc/config-schema.md
sarahheacock: Looked at old notes. The generate_operations was added after the fact because code was not written defensively enough and there was a runtime error for services that did not have operations in their service yaml. 

~Double checked and there are no APIs that set generate_operations to false and have operations specified in the service yaml. It's safe to remove now or later. But would likely remove the added skip logic~

Misread the output. iam service only has the get operation defined in the service config. However, as is, iam is not generating operations commands for methods that are not available in the service config like gen_sfc was. So it is safe to remove / can always re-add

internal/config/config_test.go
jameslynnwu: Deleted this test

doc/config-schema.md
```diff
@@ -124,6 +124,7 @@ This document describes the schema for the librarian.yaml.
 | `specification_format` | string | Specifies the API specification format. Valid values are "protobuf" (default) or "discovery". |
 | `dotnet` | [DotnetPackage](#dotnetpackage-configuration) (optional) | Contains .NET-specific library configuration. |
 | `dart` | [DartPackage](#dartpackage-configuration) (optional) | Contains Dart-specific library configuration. |
+| `gcloud` | [GcloudSurface](#gcloudsurface-configuration) (optional) | Contains gcloud-specific library configuration. |
```
julieqiu: use surfer instead of gcloud given the recent renames

internal/config/config.go
```diff
@@ -292,6 +292,9 @@ type Library struct {
 	// Dart contains Dart-specific library configuration.
 	Dart *DartPackage `yaml:"dart,omitempty"`
 
+	// Gcloud contains gcloud-specific library configuration.
+	Gcloud *GcloudSurface `yaml:"gcloud,omitempty"`
```
julieqiu: use surfer instead of gcloud (and below also)

julieqiu (approved):

sarahheacock (approved):

PR #5689, opened 2026-04-28
chore(internal/librarian/golang): remove go fallback tools

### LINKED ISSUE #5436: golang: remove hardcoded fallback tools from install.go
> Left the hardocded tools as a fallback

Once the tools section is introduced to librarian.yaml in google-cloud-go (and other relevant repositories), we should remove the hardcoded fallback tools in internal/librarian/golang/install.go.

Currently, install.go uses a hardcoded list of fallback tools if the configuration doesn't specify any:

```
var fallbackTools = []string{
	"github.com/googleapis/gapic-generator-go/cmd/protoc-gen-go_gapic@v0.58.0",
	"golang.org/x/tools/cmd/goimports@latest",
	"google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.3.0",
	"google.golang.org/protobuf/cmd/protoc-gen-go@v1.36.11",
}
```
This was left as a stopgap to avoid breaking existing repositories during the transition. Removing this fallback will ensure that all tool dependencies are explicitly declared in the configuration file, making the system more maintainable and reproducible.

see https://github.com/googleapis/librarian/pull/5421

sofisl:
The hardcoded fallback tools are removed to avoid maintaining duplicate tool lists. The installation logic now requires tools to be explicitly specified in the configuration. The file librarian.yaml that contained the fallback tools has been deleted.

From https://github.com/googleapis/librarian/pull/5421#issuecomment-4260532511
Fixes #5436


sofisl:
For whoever reviews, it seems that gemini recommends using sentinel errors, but I've found in the past that gemini is a bit too excited to recommend this option (compared to this codebase's usage of sentinel errors). I'll wait to get a review before changing the code.

sofisl:
> For whoever reviews, it seems that gemini recommends using sentinel errors, but I've found in the past that gemini is a bit too excited to recommend this option (compared to this codebase's usage of sentinel errors). I'll wait to get a review before changing the code.

Nevermind, I think I misunderstood gemini's suggestions. Applying to stay in line with missing versions as well.

zhumin8:
Should we also remove internal/librarian/golang/librarian.yaml?

sofisl:
> Should we also remove internal/librarian/golang/librarian.yaml?

Oops, I did but I think I accidentally re-added it when merging with main. Will re-delete! Good catch.

.github/workflows/librarian.yaml
```diff
@@ -48,6 +48,9 @@ jobs:
     steps:
       - uses: actions/checkout@v6
       - uses: ./.github/actions/setup-librarian
+      - name: Fetch librarian.yaml from google-cloud-go
```
noahdietz: Can you clarify why librarian tests have a dependency on the Go SDK repo's librarian.yaml?

sofisl: Yeah good question! Since the tools are now listed in google-cloud-go's librarian.yaml, and we're removing the hardcoded fallback in this PR, the workflow will fail because it can't find the tools. I couldn't find any other example of how to solve this problem in the codebase so this was just the solution that made most sense - lmk if we should follow a different pattern.

.github/workflows/librarian.yaml
```diff
@@ -48,6 +48,9 @@ jobs:
...
+        run: |
+          curl -fsSL -o librarian.yaml https://raw.githubusercontent.com/googleapis/google-cloud-go/b04765d40282a80b9d2aedfdf9d492c9440f9d91/librarian.yaml
       - name: Install tools
         run: |
           go run ./cmd/librarian install go
```
zhumin8: Are these step needed to run generate tests in golang package? Do we need the exact same version of tools from google-cloud-go for these tests? 

We have https://github.com/googleapis/librarian/blob/main/.github/workflows/go.yaml that verifies `librarian generate --all` on google-cloud-go tools

sofisl:
Ok so after discussing with @noahdietz, I have a better picture of the problem! It seems like we shouldn't be taking a dependency on google-cloud-go because it could cause circular and/or incorrect failures on Librarian head (what if the go the generator breaks something, which then breaks `librarian test`?). However this change explicitly creates a dependency on google-cloud-go because it removes the hardcoded fallback we have in the internal go librarian.yaml.

Here are two options as we see it:
1) Keep the librarian.yaml internally and use it as a test mock. It could cause some drift between what go is using, but that would also insulate us from breakages in the dependencies.
2) Close this PR entirely and leave the code as is. As it stands, it will prefer a config over the hardcoded fallback, so we should be OK to keep using as is.

What do people think?

.github/workflows/librarian.yaml
sofisl: good question! Yes we do given the changes in the PR, specifically, to delete the hardcoded tools locally and move them to librarian.yaml. I wrote up a better explanation of the issue [here](https://github.com/googleapis/librarian/pull/5689#issuecomment-4339527212)

noahdietz:
The way I see it is...we had hardcoded Go gen deps in `install.go` and this `librarian-ci.yaml` workflow depends on `install go` being run in order to execute the `internal/librarian/golang/generate_test.go`.

Rather than maintain a hardcoded list in `.go` with fallback logic (polluting our implementation), it would be cleaner/easier to keep this "mock" `internal/librarian/golang/librarian.yaml` as Sofia puts it, and reference it as our source of integration test dependencies in this CI (and in our dev guide if necessary). We add a comment to the file indicating it is just for sourcing integration test deps for Go, and it only really needs to be updated when integrating a new generator feature e.g. supporting a new flag.

zhumin8:
Plus one to keep a `librarian.yaml` for this test. I don't think the drift with that google-cloud-go uses is a stopper here, as the tests run are more unit tests. We have separately [.github/workflows/go.yaml](https://github.com/googleapis/librarian/blob/main/.github/workflows/go.yaml) that tests librarian with google-cloud-go setups.
Maybe put  this "mock" `librarian.yaml`  under a testdata dir.

sofisl:
I'll confirm this in our meeting tomorrow but generally sgtm!

.github/workflows/librarian-ci.yaml
```diff
@@ -48,6 +48,9 @@ jobs:
     steps:
       - uses: actions/checkout@v6
       - uses: ./.github/actions/setup-librarian
+      - name: Use mock librarian.yaml for tests
```
noahdietz: Kind of a nit/pedantic comment: I wouldn't call this a mock - this implies we are faking something, when in reality this is a real librarian.yaml with only tools defined that we use for managing integration test dependencies. Maybe something like:
```suggestion
      - name: Setup librarian.yaml for test tool dependency installation
```

internal/librarian/golang/testdata/librarian.yaml
```diff
@@ -11,15 +11,14 @@
...
-# librarian.
+# This file is used for sourcing integration test dependencies for Go.
+# It only needs to be updated when integrating a new generator feature
+# (e.g., supporting a new flag).
 language: go
 tools:
   go:
     - name: github.com/googleapis/gapic-generator-go/cmd/protoc-gen-go_gapic
-      version: v0.59.0
+      version: v0.58.0
```
noahdietz: Any reason for the downgrade?

internal/librarian/golang/install_test.go
```diff
@@ -15,16 +15,43 @@
...
+			}
+		})
+	}
+}
+
+func TestInstall_Success(t *testing.T) {
 	gobin := t.TempDir()
 	t.Setenv("GOBIN", gobin)
-	if err := Install(t.Context(), nil); err != nil {
+	tools := &config.Tools{
```
noahdietz, line 49: Nit: Maybe add a comment just to say "The versions here do not need to be updated, they are not used for actual tests."

internal/librarian/golang/testdata/librarian.yaml
sofisl: nope that is just an oversight!

internal/librarian/golang/install_test.go
sofisl: hm - but they are used in a test? wrote up a comment - lmk if this satisfies what you were thinking!

noahdietz: Yeah this is exactly right. We could install anything in these tests, and it wouldn't matter, we just want to see the thing(s) installed. The `GOBIN` redirection in this test to tempdir means we aren't actually installing these deps in the system running the test so they are not used.

noahdietz (approved):

PR #5697, opened 2026-04-29
refactor(sidekick): use number of services not number of known services

ianthehat:
This changes to testing the number of services that are part of this API
rather than the number of services whose names are known when deciding if
the API has services.

### DIFF
--- a/internal/sidekick/rust/codec.go
+++ b/internal/sidekick/rust/codec.go
@@ -369,7 +369,7 @@ type packagez struct {
 }
 
 func resolveUsedPackages(model *api.API, extraPackages []*packagez) {
-	hasServices := len(model.State.ServiceByID) > 0
+	hasServices := len(model.Services) > 0
 	hasLROs := false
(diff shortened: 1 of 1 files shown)


internal/sidekick/rust/codec.go
```diff
@@ -369,7 +369,11 @@ type packagez struct {
 }
 
 func resolveUsedPackages(model *api.API, extraPackages []*packagez) {
-	hasServices := len(model.State.ServiceByID) > 0
+	hasServices := false
+	for range model.AllServices() {
+		hasServices = true
+		break
+	}
```
coryan: > Should it be checking the number of direct entries, rather than including the extra external ones included in the name cache?

I think it should be using direct entries, yes. This works for me:

```suggestion
	hasServices := len(model.Services) > 0
```

As it "it generates the same code as before" and the tests pass.

coryan (changes requested):

coryan (approved):

PR #5705, opened 2026-04-29
feat(internal/librarian): introduce `update version` command

### LINKED ISSUE #3756: librarian: add update version command

sofisl:
`Version` is now added as an `update`-able field, following up on the work in #5500. A secondary PR will be opened later to update the full sources argument (towards #5075). Also introduced a small refactor to write to yaml once instead of writing multiple times (one for version, one for sources)

Fixes #3756


internal/librarian/update.go
```diff
@@ -85,40 +90,59 @@ latest API definitions is:
...
-		if err := updateSource(endpoints, repo, source, cfg); err != nil {
-			return err
+	for _, target := range targets {
+		if target == "version" {
+			repo := fetch.RepoRef{Org: "googleapis", Name: "librarian", Branch: config.BranchMain}
+			commit, _, err := fetch.LatestCommitAndChecksum(endpoints, &repo)
+			if err != nil {
+				return nil, err
+			}
+			cfg, err = setConfigValue(cfg, "version", commit)
```
noahdietz: This will always set it to a commit of the librarian repo, when I think we need to be using a tag or a Go pseudo-version.

My gut says that we should actually be exec'ing the same command we run to query the Go registry:

* latest tag: `GOPROXY=direct go list -m -f '{{.Version}}' github.com/googleapis/librarian@latest`
* HEAD@main (pseudo-version): `GOPROXY=direct go list -m -f '{{.Version}}' github.com/googleapis/librarian@main` (only diff is `@main` instead of `@librarian`)

I think this should default to the latest tag, but may need a flag to opt into latest pseudo-version e.g. `librarian update version --at=main` (`--at` defaults to `latest` otherwise). I'm not sure how useful `--at` would be to updating API sources though, so probably only supported when the targeting `version` for update. Flag name totally up for debate.

WDYT?

sofisl: Yeah you're totally right, we should follow what we do manually. my bad!

sofisl: WDYT about the mocking for the tests?

noahdietz: This is tricky...let's bring it up in team chat!

internal/librarian/update.go
```diff
@@ -85,40 +98,59 @@ latest API definitions is:
...
+				return nil, err
+			}
+		} else {
+			// TODO(https://github.com/googleapis/librarian/issues/5075): change the update command to use hierarchical dot notation for sources.
+			if cfg.Sources == nil {
+				return nil, errEmptySources
+			}
+			source := sourcesMap[target]
+			repo := sourceRepos[target]
+			if err := updateSource(endpoints, repo, source); err != nil {
```
julieqiu, line 140: refactor this so that we just call setConfigValue to set the config value, so that that operation is consistent for every value that is set in librarian.yaml

internal/librarian/update.go
```diff
@@ -45,38 +53,43 @@ var (
...
 	librarian update googleapis protobuf
+	librarian update version
 
 A typical librarian workflow for regenerating every library against the
 latest API definitions is:
 
 	librarian update googleapis
 	librarian generate --all`,
-		UsageText: "librarian update <sources...>",
+		UsageText: "librarian update <targets...>",
```
julieqiu: targets is a bit vague, maybe "librarian update <version | source> to be more precise?

internal/librarian/update.go
julieqiu: actually it doesn't look like setConfigValue is doing much, so consider just calling cfg.Version directly

https://github.com/googleapis/librarian/blob/81f070bccb5bac1b1fcd4f39ca15b67bb3353c7f/internal/librarian/cfg_value_utility.go#L30

sofisl: I think for consistency's sake, I would advocate for keeping `setConfigValue`, since we also use it in the config command (#5534) that way we don't drift in how we get/set this version between `update` and `config`.

Also, I was going to do `updateSource` in a separate PR since we had tried to do it here (#5422) but it was too big. Ideally, I was going to leave this part of the code the same and do it in a F/U PR.

internal/librarian/update.go
sofisl: I ended up removing it! I ended up just mocking a fake go binary for the test (https://github.com/googleapis/librarian/pull/5705/changes/ea4ad6043599168448f9adbda4127677a509bb5a). LMK what you think.

noahdietz: Hey this is really nifty!

internal/librarian/update_test.go
```diff
@@ -209,14 +209,25 @@ func TestUpdateCommand(t *testing.T) {
...
+			// accessing the network. This avoids using test-driven function
+			// abstraction in production code solely for easier mocking.
 			before: func(t *testing.T) {
-				old := runGoList
-				runGoList = func(ctx context.Context, env map[string]string, arg ...string) (string, error) {
-					return librarianTestCommit, nil
+				goDir := t.TempDir()
+				goPath := filepath.Join(goDir, "go")
+				script := `#!/bin/bash
+if [[ "$*" == *"list -m -f {{.Version}} github.com/googleapis/librarian@latest"* ]]; then
```
noahdietz: I think we will want to add another test whenever we add a `librarian update version -at=main` (to produce a pseudo-version) in which case we'd want to reuse this script but "templatize" it.

Could we put this script setup in a little test helper function (don't forget to call `t.Helper()` ([link](https://pkg.go.dev/testing#T.Helper)) on the first line!) with two args `target, want string` which for this test case be `"latest", "v1.2.3"`

So the `before` becomes `fakeGoList("latest", "v1.2.3")`

WDYT?

It would also allow us to dedent this script code a bit and declutter the table test cases.

sofisl: yeah good idea! made it into its own helper function.

agreed, didn't love the script ugliness but also didn't want to create an abstraction if it didn't make sense - but makes sense for future-proofing

internal/librarian/update_test.go
```diff
@@ -330,3 +313,20 @@ func updateTestConfig() *config.Config {
 	}
 	return cfg
 }
+
+// fakeGoList returns a function that mocks `go list` execution by creating a
+// fake go binary in a temporary directory and adding it to the front of PATH.
+// It matches arguments containing "list -m -f {{.Version}} github.com/googleapis/librarian@<target>"
+// and returns the specified <want> version.
+func fakeGoList(target, want string) func(*testing.T) {
```
noahdietz, line 321: Make sure this has `t.Helper()`

I don't think the return anonymous function needs it though.

internal/librarian/update_test.go
```diff
@@ -330,3 +313,20 @@ func updateTestConfig() *config.Config {
...
+	return func(t *testing.T) {
+		t.Helper()
+		goDir := t.TempDir()
+		goPath := filepath.Join(goDir, "go")
+		script := fmt.Sprintf("#!/bin/bash\nif [[ \"$*\" == *\"list -m -f {{.Version}} github.com/googleapis/librarian@%s\"* ]]; then\n  echo \"%s\"\n  exit 0\nfi\nexit 1\n", target, want)
```
noahdietz: I think the `echo \"%s\"` can be `echo %q`, which will wrap the string in quotes when printed

internal/librarian/update_test.go
sofisl: oh I see, I think I misunderstood then..sorry this is a new feature for me! thanks for pointing it out!

internal/librarian/update.go
sofisl: Hi @julieqiu! Friendly bump on this conversation - thanks!

noahdietz (approved):
Approving but want @julieqiu to review as original reviewer.

julieqiu (approved):

julieqiu:
> `Version` is now added as an `update`-able field, following up on the work in #5500. A secondary PR will be opened later to update the full sources argument (towards #5075). Also introduced a small refactor to write to yaml once instead of writing multiple times (one for version, one for sources)
> 
> Will need to merge 

This line seems like a stray comment

PR #5707, opened 2026-04-29
fix(internal/librarian): tidy to skip duplicate API path check for hardcoded list in java

### LINKED ISSUE #5706: java: API paths occurs in more than one library failed tidy
tidy fails for java because these API paths appears more than once in librarian.yaml:
duplicate api path: google/iam/v3 (appears 2 times)
duplicate api path: google/iam/v2beta (appears 2 times)
duplicate api path: google/iam/v1 (appears 2 times)
duplicate api path: google/iam/v2 (appears 2 times)

These paths are listed in `iam-policy` and `iam` library

zhumin8:
Add logic in tidy to skip duplicate API path checks for this hardcoded list when language is Java.

These API paths are duplicated in java's librarian.yaml config because java geneates proto and grpc modules from these into [java-iam](https://github.com/googleapis/google-cloud-java/tree/main/java-iam) and gapic modules separately into [java-iam-policy](https://github.com/googleapis/google-cloud-java/tree/main/java-iam-policy). 

Fix #5706

### DIFF
--- a/internal/librarian/tidy.go
+++ b/internal/librarian/tidy.go
@@ -36,6 +36,16 @@ var (
 	errDuplicateLibraryName  = errors.New("duplicate library name")
 	errDuplicateAPIPath      = errors.New("duplicate api path")
 	errNoGoogleapiSourceInfo = errors.New("googleapis source not configured in librarian.yaml")
+
+	// skipPaths lists special API paths that are allowed to appear in multiple
+	// libraries in Java without triggering the duplicate API path error.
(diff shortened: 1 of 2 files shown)


sofisl:
Just adding this as a comment for future tracking. THis might be a cross-language thing because mixins are weird for other languages too and can be potentially duplicated. We can use this comment as the basis for a bug if this shows up in another language in the future.

sofisl (approved):

Issue #3687, opened 2026-01-22
command: pass verbose flag as config instead of global


julieqiu:
command.Verbose is currently a global, which forces tests to manage shared state.

We should pass the verbose flag as a config to avoid global mutation, simplify tests, and make the command behavior easier to reason about.

PR #5708, opened 2026-04-29
feat(publish): stream output with verbose flag

### LINKED ISSUE #5369: rust: publish lacks informative logging
`librarian publish` for Rust does not log progress at all. These commands can run for many hours and there is no signal in the terminal if things are progressing or if they are stuck. 

We should add some minimal logging to `publish` that at least gives a "starting on X library" then "finished X library" type of communication.

noahdietz:
When running `librarian publish -v` for Rust, the output of `cargo` commands executed will be streamed directly to stdout/stderr as opposed to being swallowed. Without this, the command only prints the error message on failure and does not log any progress or notices.

Fixes #5369

Note to #3687


sofisl:
I checked with jetski and apparently this won't publish the command. It recommends adding:
`command.Verbose = verbose` in `func legacyRustPublish` in librarian/internal/librarian/publish.go to also log the command being run.

noahdietz:
> I checked with jetski and apparently this won't publish the command. It recommends adding: `command.Verbose = verbose` in `func legacyRustPublish` in librarian/internal/librarian/publish.go to also log the command being run.

Good catch! Done

sofisl (approved):

PR #5710, opened 2026-04-29
chore(serviceconfig): add common-protos paths to allow Java

zhumin8:
Add API paths used in common-protos to allow Java.

For #5729

### DIFF
--- a/internal/serviceconfig/sdk.yaml
+++ b/internal/serviceconfig/sdk.yaml
@@ -151,6 +151,7 @@
   description: Card interface displayed in a Google Chat message or Google Workspace Add-on.
   documentation_uri: https://developers.google.com/chat
   languages:
+    - java
     - python
   skip_rest_numeric_enums:
     - python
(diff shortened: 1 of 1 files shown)


internal/serviceconfig/sdk.yaml
```diff
@@ -281,6 +282,9 @@
     - python
   release_level:
     java: preview
+- path: google/cloud
+  languages:
+    - java
```
zhumin8, line 287: This is needed for common-protos. It generates common_resources.proto and extended_operations.proto.

see hermetic-build config: https://github.com/googleapis/google-cloud-java/blob/66ded327ed12fb64acbe076d957f882d35ba25e3/generation_config.yaml#L3048

Wonder how Python and Go managed to do without this path?

zhumin8: Ok, the validation looks for  "google/cloud/" prefix.
https://github.com/googleapis/librarian/blob/cdb7cea6dd544a23755f5212cac60d620815254e/internal/serviceconfig/serviceconfig.go#L213-L216

zhumin8: Removing the trailing slash would potentialy include unwanted sibling dir like "google/cloudxyx". (There isn't one today, but not worth the risk.

julieqiu (approved):

PR #5711, opened 2026-04-29
refactor(internal/sidekick/gcloud): fix api_field to match apitools

### LINKED ISSUE #4523: surfer: Investigate & Fix Custom Method `api_field` Stripping
For custom methods (e.g., `import_data`, `export_data`), the generated `api_field` paths in the YAML output does not match autogen's output:
*   **Current Output:** `api_field: sourceParallelstore.path`
*   **Expected Output:** `api_field: exportDataRequest.sourceParallelstore.path`

We need to investigate whether this is a general failure in our programmatic AIP heuristics for custom methods, or a specific design requirement in gcloud's framework that we should handle through overrides in passed through the `gcloud.yaml`.

- Analyze the `parallelstore` custom methods and their corresponding proto definitions.
- Investigate how the value is constructed in autogen and determine if the value can be derived through the use of AIPs.
- Depending on the investigation outcome, either:
  1. Fix the core programmatic mapping logic in Surfer to retain the request wrapper for custom methods.
  2. Implement this strictly as a set of manual configuration overrides in `gcloud.yaml`.

## Relevant files
- `internal/surfer/gcloud/builder.go`
- `internal/surfer/gcloud/command.go`

sarahheacock:
api_field is updated to use correct api field for wildcard body fields.

The casing logic for argument names and API fields is moved from the builder to the writer to better separate domain logic from output formatting.

Fix https://github.com/googleapis/librarian/issues/4523


internal/sidekick/surfer/command_writer.go
```diff
@@ -174,9 +174,28 @@ func mapArgumentsToYAML(args []Argument) []declarative.Argument {
...
+		var apiField string
+		var rmp map[string]string
+
+		if len(a.APIField) > 0 {
+			parts := make([]string, len(a.APIField))
+			for i, p := range a.APIField {
+				parts[i] = strcase.ToLowerCamel(p)
+			}
+			joinedPath := strings.Join(parts, ".")
+
```
jameslynnwu, line 186: rm line

internal/sidekick/surfer/command_writer.go
```diff
@@ -174,9 +174,28 @@ func mapArgumentsToYAML(args []Argument) []declarative.Argument {
...
+			if a.ResourceSpec != nil {
+				rmp = map[string]string{
+					joinedPath: "{__relative_name__}",
```
jameslynnwu, line 189: what's this {__relative_name__} about?

jameslynnwu (approved):

internal/sidekick/surfer/command_writer.go
sarahheacock: Old declarative syntax. It's saying take the uri path of the resource arg and map it to this request field

PR #5712, opened 2026-04-29
refactor(sidekick): a resource merge that does not need state access

ianthehat:
Remove the need to see the APIState class.
Builds a map to track index of matching resource, and then builds the
actual list in place rather than from the map.

### DIFF
--- a/internal/sidekick/parser/protobuf.go
+++ b/internal/sidekick/parser/protobuf.go
@@ -311,14 +311,20 @@ func makeAPIForProtobuf(serviceConfig *serviceconfig.Service, req *pluginpb.Code
 	}
 
 	// Consolidate resources.
-	// Message-level resources (in state.ResourceByType) take precedence over
+	// Message-level resources (in result.AllResources take precedence over
 	// file-level resources (already in result.ResourceDefinitions).
-	allResources := make(map[string]*api.Resource, len(result.ResourceDefinitions)+len(state.ResourceByType))
-	for _, r := range result.ResourceDefinitions {
-		allResources[r.Type] = r
+	seen := map[string]int{}
+	for i, r := range result.ResourceDefinitions {
+		seen[r.Type] = i
 	}
-	maps.Copy(allResources, state.ResourceByType)
-	result.ResourceDefinitions = slices.Collect(maps.Values(allResources))
+	for r := range result.AllResources() {
+		if i, found := seen[r.Type]; found {
+			result.ResourceDefinitions[i] = r
+		} else {
+			seen[r.Type] = len(result.ResourceDefinitions)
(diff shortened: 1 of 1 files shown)


internal/sidekick/parser/protobuf.go
```diff
@@ -313,17 +317,15 @@ func makeAPIForProtobuf(serviceConfig *serviceconfig.Service, req *pluginpb.Code
...
 
-	// Sort to ensure deterministic output.
-	slices.SortFunc(result.ResourceDefinitions, func(a, b *api.Resource) int {
-		return strings.Compare(a.Type, b.Type)
-	})
+	for r := range result.AllResources() {
+		if i, found := slices.BinarySearchFunc(result.ResourceDefinitions, r, compareResources); found {
+			result.ResourceDefinitions[i] = r
+		} else {
+			result.ResourceDefinitions = slices.Insert(result.ResourceDefinitions, i, r)
```
coryan: Isn't this insert a $O(n)$ worst-case and it is called $n$ times so the whole algorithm is $O(n^2)$ ? Using a map + sort is $O(n + n \log(n))$ no?

I may be worrying about nothing, since $n$ is probably small, but when else do I get to use LaTeX expressions in my GitHub comments?

coryan (changes requested):

internal/sidekick/parser/protobuf.go
ianthehat: I originally decided it would be fine, but then I persuaded myself that a different solution was probably not less efficient and would scale higher, so I changed it again...

internal/sidekick/parser/protobuf.go
```diff
@@ -311,14 +311,20 @@ func makeAPIForProtobuf(serviceConfig *serviceconfig.Service, req *pluginpb.Code
 	}
 
 	// Consolidate resources.
-	// Message-level resources (in state.ResourceByType) take precedence over
+	// Message-level resources (in result.AllResources take precedence over
 	// file-level resources (already in result.ResourceDefinitions).
-	allResources := make(map[string]*api.Resource, len(result.ResourceDefinitions)+len(state.ResourceByType))
-	for _, r := range result.ResourceDefinitions {
-		allResources[r.Type] = r
+	seen := map[string]int{}
```
coryan, line 316: TIL: neat trick.

coryan (approved):
Yes, this is more efficient and cleaner and all, but do you get to use LaTeX ?

ianthehat:
Nope, sadly no latex for me.
Maybe we should proposed that go docs can have embedded latex that renders nicely so we can fill our descriptions with prettiness?

PR #5715, opened 2026-04-30
feat(internal/sidekick/dart): use null-aware elements in generated code

kevmoo:
Updates the generator to use null-aware elements ('key': ?value) in map literals for optional fields in toJson and query parameter serialization.

Removes the corresponding ignore from the templates and updates tests to expect the new syntax.


internal/sidekick/dart/annotate.go
```diff
@@ -1081,6 +1087,39 @@ func createToJsonLine(field *api.Field, model *api.API) string {
...
+// createToJsonElement creates a JSON element expression for map literals.
+func createToJsonElement(field *api.Field) string {
+	name := fieldName(field)
+	jsonName := field.JSONName
+
+	switch field.Typez {
+	case api.TypezFloat, api.TypezDouble:
+		return fmt.Sprintf("if (%s case final $1?) '%s': encodeDouble($1)", name, jsonName)
+	case api.TypezBytes:
+		return fmt.Sprintf("if (%s case final $1?) '%s': encodeBytes($1)", name, jsonName)
```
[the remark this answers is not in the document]
kevmoo: Solid idea, but we need to change how `package:google_cloud_protobuf` works!

kevmoo:
@brianquinlan – totally right to run this against the real code. Found a lot more bits!

internal/sidekick/dart/annotate.go
```diff
@@ -197,9 +197,10 @@ type fieldAnnotation struct {
 	// The default value for the string, e.g. "0" for an integer type.
 	DefaultValue string
 	// Whether the default value is constant or not, e.g. "0" is constant but "Uint8List(0)" is not.
-	ConstDefault bool
-	FromJson     string
-	ToJson       string
+	ConstDefault  bool
+	FromJson      string
+	ToJson        string
+	ToJsonElement string
```
brianquinlan: Could you add a comment explaining what the different between `ToJson` and `ToJsonElement` is.

internal/sidekick/dart/annotate.go
```diff
@@ -197,9 +197,10 @@ type fieldAnnotation struct {
 	// The default value for the string, e.g. "0" for an integer type.
 	DefaultValue string
 	// Whether the default value is constant or not, e.g. "0" is constant but "Uint8List(0)" is not.
-	ConstDefault bool
-	FromJson     string
-	ToJson       string
+	ConstDefault  bool
+	FromJson      string
+	ToJson        string
```
brianquinlan: This is no longer used, right?

internal/sidekick/dart/annotate_test.go
```diff
@@ -1951,11 +1959,10 @@ func TestAnnotateField(t *testing.T) {
 				FieldBehaviorRequired: true,
 				DefaultValue:          "",
 				ConstDefault:          true,
+				ToJsonElement:         "",
 			},
 		},
 		{
-			// `google.protobuf.Empty` is a special because, in some cases, it is
```
brianquinlan, line 1993: Why did you remove this?

internal/sidekick/dart/annotate.go
brianquinlan: Or do you want to just change `ToJson` to mean `ToJsonElement`?

internal/sidekick/dart/annotate_test.go
brianquinlan: Should you repurpose this for `ToJsonElement`?

internal/sidekick/dart/annotate.go
kevmoo: Fixed!

internal/sidekick/dart/annotate_test.go
kevmoo: restored. sorry.

kevmoo:
PTAL @brianquinlan

internal/sidekick/dart/annotate.go
```diff
@@ -1089,17 +1098,61 @@ func createToJsonLine(field *api.Field, model *api.API) string {
...
 				name, keyEncoder, valueEncoder)
+		} else {
+			rhs = name
 		}
+	default:
+		enc, _ := encoder(field.Typez, name)
+		rhs = enc
+	}
+
+	fieldRequired := field.Behavior != nil && slices.Contains(field.Behavior, api.FieldBehaviorRequired)
```
brianquinlan: The `nil` check is not required.

internal/sidekick/dart/annotate_test.go
```diff
@@ -1694,9 +1695,10 @@ func TestCreateToJsonLine(t *testing.T) {
 				"prefix:google.cloud.foo": "foo",
 			})
 
-			got := createToJsonLine(test.field, model)
-			if diff := cmp.Diff(test.want, got); diff != "" {
-				t.Errorf("mismatch in TestCreateToJsonLine (-want, +got)\n:%s", diff)
+			want := fmt.Sprintf("if (%s.isNotDefault) '%s': %s", fieldName(test.field), test.field.JSONName, test.want)
```
brianquinlan: Can you put this in the test table? Better not to have logic in tests.

internal/sidekick/dart/annotate.go
```diff
@@ -1162,6 +1215,24 @@ func (annotate *annotateModel) buildQueryLines(
 
 	switch {
 	case field.Repeated:
+		if codec.Nullable {
```
brianquinlan: Don't you need tests for this in TestBuildQueryLines_Primitives, etc?

internal/sidekick/dart/annotate.go
```diff
@@ -1089,17 +1098,61 @@ func createToJsonLine(field *api.Field, model *api.API) string {
...
-	enc, _ := encoder(field.Typez, name)
+	// For types that require encoding (like Messages or 64-bit ints),
+	// encoder appends ".toJson()" or ".toString()".
+	// Passing "name?" results in "name?.toJson()" or "name?.toString()".
+	enc, _ := encoder(field.Typez, name+"?")
 	return enc
 }
 
+// createNullableToJsonElement creates a JSON element expression for map literals for nullable fields.
+func createNullableToJsonElement(field *api.Field) string {
```
brianquinlan: Do you still need the "Element" suffix in these names?

internal/sidekick/dart/annotate_test.go
```diff
@@ -1355,7 +1356,7 @@ func TestCreateFromJsonLine(t *testing.T) {
 	}
 }
 
-func TestCreateToJsonLine(t *testing.T) {
+func TestCreateNonNullableToJsonElement(t *testing.T) {
```
brianquinlan: Is `createNullableToJsonElement` tested?

Maybe you should just blackbox test the `ToJson` line? Like make a `createToJson` and just test that. It could call `createNullableToJsonElement` or `createNonNullableToJsonElement` internally.

brianquinlan (changes requested):

internal/sidekick/dart/annotate.go
kevmoo: Or we don't need it at all!

kevmoo: ...repeated fields aren't nullable

internal/sidekick/dart/annotate.go
kevmoo: fixed

kevmoo:
PTAL @brianquinlan !

kevmoo:
Feel free to squash/merge and delete the branch if you're happy

internal/sidekick/dart/annotate.go
```diff
@@ -1089,17 +1098,61 @@ func createToJsonLine(field *api.Field, model *api.API) string {
...
+	enc, _ := encoder(field.Typez, name+"?")
 	return enc
 }
 
+// createNullableToJson creates a JSON element expression for map literals for nullable fields.
+func createNullableToJson(field *api.Field) string {
+	name := fieldName(field)
+	jsonName := field.JSONName
+
+	switch field.Typez {
```
brianquinlan: Are there really two cases here:
1. api.TypezFloat, api.TypezDouble, api.TypezBytes, and canBeNull:
   all of these use `if (%s case final $1?) + <encoder>`
2. everything else

The reason for this special-casing wasn't obvious to me at first. A comment saying something about encoders that are functions can't use null-aware operators.

Or maybe `encoder` should return another bool indicating whether it should use the null-aware encoding or not?

kevmoo: great idea!

kevmoo:
@brianquinlan – PTAL?

internal/sidekick/dart/annotate.go
```diff
@@ -1089,17 +1098,60 @@ func createToJsonLine(field *api.Field, model *api.API) string {
...
+func createNullableToJson(field *api.Field) string {
+	name := fieldName(field)
+	jsonName := field.JSONName
+
+	// Float, Double, and Bytes use function-based encoders (e.g., `encodeDouble`, `encodeBytes`),
+	// and certain fields (like `NullValue` or `Value`) can serialize to `null`.
+	// Standalone function calls are not null-aware expressions, so we cannot use Dart's null-aware
+	// map element syntax `?expression` with them. Instead, we use `if (name case final $1?)`
+	// to safely extract the value before encoding.
+	if field.Typez == api.TypezFloat || field.Typez == api.TypezDouble || field.Typez == api.TypezBytes || canBeNull(field) {
```
brianquinlan, line 1146: I don't like that the knowledge about whether a function is used or not is duplicated between here and `encoder`.

I can fix that in a follow-up PR.

brianquinlan:
I'm not sure why the coverage reports aren't being generated anymore, but net coverage in this directory went down from 91.5% to 91% after this change. There must be uncovered scenarious.

brianquinlan (approved):

PR #5716, opened 2026-04-30
fix(internal/sidekick/dart): fix unused constructor parameters

kevmoo:
Renames the fromJson factory parameter to _ when the message has no fields, avoiding the avoid_unused_constructor_parameters lint warning in generated Dart code.

Adds tests to verify that HasFields correctly identifies empty messages and that the generated code uses _ for the parameter.


natebosch (commented):
LGTM for the change to the generated Dart. I don't understand the implications of the "skip format" stuff in the test so I'll let Brian review that part.

internal/sidekick/dart/generate_test.go
```diff
@@ -132,3 +132,44 @@ func requireProtoc(t *testing.T) {
 		t.Skip("skipping test because protoc is not installed")
 	}
 }
+
```
brianquinlan: Would you mind removing this? I'm not sure if we want to be checking generated text here. Or maybe we should have some golden templates or something.

brianquinlan (approved):

internal/sidekick/dart/generate_test.go
kevmoo: Removing...what? This blank line?

brianquinlan: This test.

PR #5718, opened 2026-04-30
fix(internal/sidekick/gcloud): omit extract_resource_result if false

jameslynnwu:
Omit extract_resource_result from the generated async section if it is false, to match existing generated output.

This change adds omitempty to the ExtractResourceResult field in declarative.go and updates the golden files in internal/surfer.

### DIFF
--- a/internal/surfer/testdata/apis/parallelstore/expected/target/surface/parallelstore/instances/_partials/_export_data_ga.yaml
+++ b/internal/surfer/testdata/apis/parallelstore/expected/target/surface/parallelstore/instances/_partials/_export_data_ga.yaml
@@ -163,4 +163,3 @@
   async:
     collection:
       - parallelstore.projects.locations.operations
-    extract_resource_result: false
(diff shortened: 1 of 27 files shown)


sarahheacock (approved):
LGTM

internal/surfer/testdata/apis/parallelstore/expected/target/surface/parallelstore/instances/_partials/_export_data_ga.yaml
```diff
@@ -163,4 +163,3 @@
   async:
     collection:
       - parallelstore.projects.locations.operations
-    extract_resource_result: false
```
sarahheacock, line 166: This and import_data actually should be set explicitly to false. yaml_command_schema defaults this to true when not provided. The gcloud config value display_operation_result sets this value to false (the logic in the command_builder is just wrong). Not providing it actually breaks this command

PR #5719, opened 2026-04-30
feat(internal/sidekick/dart): make workspace resolution optional

kevmoo:
Wraps resolution: workspace in a conditional in pubspec.yaml.mustache.

Adds use-workspace option to annotateModel in annotate.go, defaulting to true to preserve existing behavior.

### DIFF
--- a/internal/sidekick/dart/annotate.go
+++ b/internal/sidekick/dart/annotate.go
@@ -83,8 +83,9 @@ type modelAnnotations struct {
 	// ["export 'package:google_cloud_gax/gax.dart' show Any", "export 'package:google_cloud_gax/gax.dart' show Status"]
 	Exports []string
 	// A comma-separated list of service fakes, e.g. "FakeCacheService, FakeGenaiService".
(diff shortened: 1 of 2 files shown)


natebosch:
What is the motivation?

kevmoo:
> What is the motivation?

At the moment you can't generate just ONE stand-alone library. It expects a workspace and you end up with a turd project.

natebosch:
When do we want to generate just one standalone library?

I'm not particularly opposed to the change, but my (possibly mistaken) understanding was we are intentionally narrowly targeting exactly the usage we need, and I think our current processes generate more than one.

kevmoo:
> When do we want to generate just one standalone library?
> 
> I'm not particularly opposed to the change, but my (possibly mistaken) understanding was we are intentionally narrowly targeting exactly the usage we need, and I think our current processes generate more than one.

I did just to test the generator with some sample code.

brianquinlan (approved):

PR #5720, opened 2026-04-30
fix(internal/sidekick/dart): remove trailing blank doc comment line

kevmoo:
Makes the blank /// line in templates conditional on having DocLines, avoiding empty doc comment lines when there is no description content.

Adds HasDocLines field to modelAnnotations in Go.


natebosch (approved):
Needs a test?

PR #5723, opened 2026-04-30
feat(internal/librarian/python): create a changelog during add

### LINKED ISSUE #5693: python: new library fails tests due to a lack of changelog
Sample PR: https://github.com/googleapis/google-cloud-python/pull/16856

Failure in docs test:

```text
sphinx.errors.SphinxWarning: /home/runner/work/google-cloud-python/google-cloud-python/packages/google-maps-mapmanagement/docs/index.rst:20:toctree contains reference to nonexisting document 'CHANGELOG'
```

It's unclear to me at the moment whether we need a change in the generator, synthtool, or librarian. Will investigate.

jskeet:
The python.Add function is changed to create a changelog file in the root directory of the new package, and a symlink to it from a docs directory.

Fixes #5693

### DIFF
--- a/internal/librarian/python/add.go
+++ b/internal/librarian/python/add.go
@@ -17,15 +17,29 @@ package python
 import (
 	"errors"
 	"fmt"
+	"os"
+	"path/filepath"
 	"slices"
 
 	"github.com/googleapis/librarian/internal/config"
 	"github.com/googleapis/librarian/internal/serviceconfig"
 )
 
-// defaultVersion is the first version used for a new library.
-// This is set on the initial `librarian add` for a new API.
-const defaultVersion = "0.0.0"
+const (
+	// defaultVersion is the first version used for a new library.
+	// This is set on the initial `librarian add` for a new API.
+	defaultVersion = "0.0.0"
+	// changelog is the name of the changelog file to create. A regular file
+	// is created in the package root, and a symlink is created in the docs
+	// directory.
+	changelog         = "CHANGELOG.md"
+	changelogTemplate = `# Changelog
+
+[PyPI History][1]
+
+[1]: https://pypi.org/project/%s/#history
+`
+)
 
 var (
 	approvedGAPICNamespaces = []string{
@@ -42,7 +56,7 @@ var (
 )
 
 // Add initializes a new Python library with default values.
-func Add(lib *config.Library) (*config.Library, error) {
+func Add(lib *config.Library, output string) (*config.Library, error) {
 	lib.Version = defaultVersion
 	if len(lib.APIs) != 1 {
 		return nil, errNewLibraryMustHaveOneAPI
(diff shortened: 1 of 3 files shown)


internal/librarian/python/add_test.go
```diff
@@ -113,8 +118,9 @@ func TestAdd_Error(t *testing.T) {
 			wantErr: errNewLibraryBadNamespace,
 		},
 	} {
+		output := t.TempDir()
 		t.Run(test.name, func(t *testing.T) {
-			_, gotErr := Add(test.lib)
+			_, gotErr := Add(test.lib, output)
```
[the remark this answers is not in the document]
jskeet: Done.

internal/librarian/python/add_test.go
```diff
@@ -179,3 +185,76 @@ func TestValidateNewAPIs(t *testing.T) {
...
+		t.Errorf("expected changelog to contain %s; was %s", libName, textContent)
+	}
+	linkPath := filepath.Join(output, "docs", changelog)
+	linkInfo, err := os.Lstat(linkPath)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if linkInfo.Mode()&os.ModeSymlink == 0 {
+		t.Errorf("docs file is not a symlink")
+	}
```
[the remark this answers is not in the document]
jskeet: done.

internal/librarian/python/add.go
```diff
@@ -57,6 +71,9 @@ func Add(lib *config.Library) (*config.Library, error) {
 	if !slices.Contains(approvedGAPICNamespaces, namespace) {
 		return nil, fmt.Errorf("%w: unapproved namespace %s derived from API path %s", errNewLibraryBadNamespace, namespace, apiPath)
 	}
+	if err := createChangelog(lib.Name, output); err != nil {
```
noahdietz, line 74: Are we sure we don't want to do this in `generate` gated by an existence check? I'm not opposed to this approach, but `librarian add` changing anything other than `librarian.yaml` seems odd to me?

jskeet: We *can* do - it just feels more natural to do "stuff that we only do once" in add than generate. (I know there's some one-off stuff for Go, but that needs generation to happen first.)

I definitely don't mind moving it to generate if you'd prefer.

noahdietz: Yeah, in Go generate we have a one time initialization of `go.mod` which is gated by an existence check.

I think I'd personally prefer keeping it in `generate` but also happy to chat about it a bit more as a team. Java might have some similar requirement that would help set precedent with Python here but idk.

jskeet: Agree with chatting about it more in the team. I'll create a second PR tomorrow which is based on this, but moves it to generate, so that we can then just merge whichever we prefer.

jskeet:
Converting this to a draft; #5756 is a newer PR which will probably be merged in preference to this one.

noahdietz:
Took the other PR based on generate

PR #5726, opened 2026-04-30
refactor(internal/sidekick/gcloud): rename gcloud package to surfer

julieqiu:
Currently, internal/sidekick/gcloud contains logic primarily used by the surfer CLI. As part of migrating gcloud generation to librarian, we plan to change the interface.

Renaming the existing package to internal/sidekick/surfer to free up the gcloud namespace, so that we can incrementally migrate only the necessary components into the new namespace.


coryan (approved):
You still need to get the CI to pass.

PR #5727, opened 2026-04-30
fix(sidekick/swift): include verb in URL path

coryan:
Some RPCs have a "verb" suffix[^1] that is part of the URL path in the request. I had forgotten to append the verb when it is present.

[^1]: Not to be confused with the HTTP method, often called the verb too.

### DIFF
--- a/internal/sidekick/swift/format_path.go
+++ b/internal/sidekick/swift/format_path.go
@@ -32,7 +32,11 @@ func pathExpression(t *api.PathTemplate) string {
 			count += 1
 		}
 	}
-	return "/" + strings.Join(pathComponents, "/")
+	path := "/" + strings.Join(pathComponents, "/")
+	if t.Verb != nil {
+		path += ":" + *t.Verb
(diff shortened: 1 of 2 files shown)


internal/sidekick/swift/format_path.go
```diff
@@ -32,7 +32,11 @@ func pathExpression(t *api.PathTemplate) string {
 			count += 1
 		}
 	}
-	return "/" + strings.Join(pathComponents, "/")
+	path := "/" + strings.Join(pathComponents, "/")
+	if t.Verb != nil {
```
jameslynnwu, line 36: is there any chance a verb will be set to an empty string?

jameslynnwu (approved):

internal/sidekick/swift/format_path.go
coryan: Good question, no. The parsers reject empty strings. Added a comment so future us don't need to wonder.

PR #5747, opened 2026-04-30
feat(internal/librarian/tidy): remove unused sections in tidy

### LINKED ISSUE #5469: librarian: tidy should remove unused sections
https://github.com/googleapis/google-cloud-go/blob/a1a30091f329a3b22e8510e5089c0d95fc632e9d/librarian.yaml#L21 contains and unused `release` section. 

We should update librarian tidy to remove unused sections from the config.

sofisl:
Update librarian tidy to remove empty release, tools, and default sections from the configuration file.

Pointers to structs with all zero values were previously written back to the YAML file as empty objects (e.g., `release: {}`). This change adds explicit checks to nil out these pointers if their fields are all empty, allowing the `omitempty` tag to work as intended.

In an effort to keep the PR small and focus on actually occurring problems, I did not focus on any of [these edge cases.](https://github.com/googleapis/librarian/issues/5469#issuecomment-4347378514) If they become a problem in the future I will create a separate F/U PR.

Fixes #5469

### DIFF
--- a/internal/librarian/tidy_test.go
+++ b/internal/librarian/tidy_test.go
@@ -797,6 +797,67 @@ func TestTidyLanguageConfig_Rust(t *testing.T) {
 	}
 }
 
+func TestTidy_UnusedSections(t *testing.T) {
+	for _, test := range []struct {
+		name        string
+		cfg         *config.Config
+		wantRelease bool
+		wantTools   bool
+		wantDefault bool
+	}{
+		{
+			name: "empty sections removed",
+			cfg: &config.Config{
+				Language: config.LanguageRust,
+				Sources: &config.Sources{
+					Googleapis: &config.Source{Commit: "commit"},
+				},
+				Release: &config.Release{},
+				Tools:   &config.Tools{},
+				Default: &config.Default{},
+			},
+			wantRelease: false,
+			wantTools:   false,
+			wantDefault: false,
+		},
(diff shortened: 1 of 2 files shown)


internal/librarian/tidy_test.go
```diff
@@ -826,6 +797,67 @@ func TestTidyLanguageConfig_Rust(t *testing.T) {
...
+			wantDefault: true,
+		},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			tempDir := t.TempDir()
+			if err := RunTidyOnConfig(t.Context(), tempDir, test.cfg); err != nil {
+				t.Fatal(err)
+			}
+			got, err := yaml.Read[config.Config](filepath.Join(tempDir, config.LibrarianYAML))
+			if err != nil {
```
sofisl, line 845: For any go experts, I couldn't tell whether this would be better using cmp.Diff, so I asked Jetski:

```
cmp.Diff is typically recommended for comparing complex types (like structs, slices, or maps) where a detailed diff is helpful for debugging. For simple boolean comparisons (like checking if a pointer is nil or not), a direct comparison is usually clearer and more idiomatic in Go.
```

But I know in our docs it says explicitly to use cmp.Diff so wasn't sure.

julieqiu (approved):

PR #5752, opened 2026-05-01
feat(sidekick/surfer): support surface-level cascading hiding via gcloud.yaml

### LINKED ISSUE #5510: surfer: enable hidden commands through gcloud config
Use gcloud config to make the entire command tree hidden.

jameslynnwu:
If a surface is configured as hidden in gcloud.yaml, we statically mark its root group, subgroups, and commands as hidden during tree construction.
  
  This is a simple surface-level cascading model to unblock basic command hiding. In the future, we plan to introduce fine-grained command-group and individual command-level hiding. This will require transitioning sidekick to a dynamic visibility model (such as a bottom-up AST traversal pass at the end of surface building) to resolve group visibility based on its children.
  
  - Propagates RootIsHidden from API config to groups, subgroups, and commands during build.
  - Uses exact FQN service-level scoping to correctly resolve visibility for multi-API packages (e.g. keeping GA policy-bindings public under a shared root).
  - Updates templates to conditionally output Calliope's @base.Hidden decorator for hidden groups and subgroups.
  - Exposes Hidden config in Librarian's SurferAPI schema.
  
Fixes https://github.com/googleapis/librarian/issues/5510


internal/surfer/testdata/apis/parallelstore/expected/current/surface/parallelstore/instances/_partials/_create_ga.yaml
```diff
@@ -17,7 +17,6 @@
 - release_tracks:
     - GA
   auto_generated: true
-  hidden: true
```
sarahheacock, line 20: `hidden: true` is what is generated in `gen_sfc` today. I believe the reason hidden is propagated is because if a user marks one API service as hidden but another as visible (see the multi_service example), then the root command group itself should not be hidden.

In the future, there will be other reasons that subgroups or commands will be hidden outside of the gcloud.yaml config (such as proto visibility labels). We will also need a way to prune command groups that have no visible subcommands.

To address this cleanly, I recommend keeping all resolution logic in surface_builder and keeping the writer completely "dumb":

1. **Leaf-only Tagging:** Only label leaf Commands as hidden (nothing else) when building them.
2. **AST Resolution Pass:** After the command tree is built in surface_builder, run a final bottom-up traversal pass to resolve group visibility (a group is labeled hidden only if all of its children are hidden). This is also the perfect place to prune empty command groups.
3. **Dumb Emission:** Have the writer dumbly output Hidden properties straight to the templates.

I don't love mutating state but surface_builder is the only stateful file. I think a final resolution pass is much faster, simpler, and less prone to redundant recursive lookups than having the writer execute recursive getters during emission.

gen_sfc is similar but it does bottom up and top down traversal at the end. This at least limits us to one bottom up traversal

I need to review the rest but wanted to send off the comment sooner rather than later

internal/librarian/surfer/generate.go
```diff
@@ -88,16 +88,39 @@ func generateAPI(api *config.API, gcloudCfg *config.Surfer, googleapisDir, outDi
...
-	providerCfg := &provider.Config{}
-	if gcloudCfg != nil && gcloudCfg.HelpText != nil {
-		providerCfg.APIs = []provider.API{{
-			Name:     model.Name,
-			HelpText: mapHelpText(gcloudCfg.HelpText),
-		}}
-	}
+	surferAPI := resolveSurferAPI(library, api)
+
+	var providerAPIs []provider.API
```
sarahheacock: This provider is supposed to abstract away where information is coming from. Would recommend some sort of load function to generate this config rather than mapping it inline

internal/librarian/surfer/generate.go
```diff
@@ -88,16 +88,39 @@ func generateAPI(api *config.API, gcloudCfg *config.Surfer, googleapisDir, outDi
...
 
+// resolveSurferAPI returns the surfer-specific configuration for the given API.
+func resolveSurferAPI(library *config.Library, api *config.API) *config.SurferAPI {
+	res := &config.SurferAPI{
+		Path: api.Path,
+	}
+	if library.Surfer == nil {
+		return res
+	}
+	for _, surferAPI := range library.Surfer.SurferAPIs {
```
sarahheacock: Nit: for every API, iterating through the list of APIs to find the corresponding surferApi is repetative. Also, I'm not familiar with the librarian config but this resolution also seems really odd / seems like it should be abstracted.

internal/sidekick/surfer/command.go
```diff
@@ -49,11 +50,6 @@ type Command struct {
 	// Name specifies the name of the command.
 	Name string
 
-	// Hidden specifies whether this command should be hidden from the user in
-	// help text and command listings.
-	// Origin: Hardcoded to `true` by this generator.
-	Hidden bool
```
sarahheacock, line 55: Hidden is still needed for commands regardless of the technical approach. Individual commands can be labeled as hidden through config and visibility labels

internal/sidekick/surfer/command_builder.go
```diff
@@ -110,14 +109,6 @@ func async(method *api.Method, model *api.API, service *api.Service) *Async {
 	return async
 }
 
-func hidden(overrides *provider.Config) bool {
-	if overrides != nil && len(overrides.APIs) > 0 {
```
sarahheacock, line 151: I personally think have the logic here and bubbling is up is the correct approach.

If you do, just taking the first API is wrong. You need to grab the API that corresponds with this method.

You also need to default to false

internal/librarian/surfer/generate.go
jameslynnwu: I agree, but I don't think its an optimization we need to make atm. btw, this "resolve*" is in the other language generators

internal/sidekick/surfer/provider/config.go
```diff
@@ -280,3 +284,24 @@ func ShouldGenerateOperations(c *Config) bool {
...
+
+// FindAPIConfigByService finds the API config that applies to the given service.
+func FindAPIConfigByService(c *Config, service *api.Service) *API {
+	serviceName := service.Name
+	apiVersion := apiVersionFromPackage(service.Package)
+
+	for i := range c.APIs {
+		apiCfg := &c.APIs[i]
+		if apiCfg.Name == serviceName {
+			if apiCfg.APIVersion == "" || apiCfg.APIVersion == apiVersion {
```
sarahheacock: APIVersion shouldn't be an empty string.

internal/sidekick/surfer/provider/config.go
```diff
@@ -280,3 +284,24 @@ func ShouldGenerateOperations(c *Config) bool {
...
+				return apiCfg
+			}
+		}
+	}
+	return nil
+}
+
+// FindAPIConfig finds the API config that applies to the given method.
+func FindAPIConfig(c *Config, method *api.Method) *API {
```
sarahheacock, line 312: Nit: out of scope for this PR. We should follow the same pattern as sidekick accessors through out the provider https://github.com/googleapis/librarian/pull/5788/changes. This will make it easier to create mocks for unit tests and add consistency

internal/sidekick/surfer/group_builder.go
```diff
@@ -34,10 +34,16 @@ func newRootGroup(params *groupParams) *CommandGroup {
 	// TODO (https://github.com/googleapis/librarian/issues/3033): Use service selector
 	// to look up the help text from the gcloud config.
 	rootName := provider.ResolveRootPackage(params.model)
+	hidden := false
```
sarahheacock, line 41: In an original comment, I mentioned bubbling up the hidden state from the commands in the surface_builder. Two reasons...
* Root group does not map neatly to one service (even though the rest of the subgroups do). So you can have one subgroup be hidden and not another. The reason service is passed in for the root group is due to quirk in gen_sfc help text. In which case the root should not be hidden
* There is more than one reason commands can be hidden. If all of the subcommands are hidden due to visibility for example, the shared command groups should be hidden as well.

In the multi_service golden, I believe the test is passing because the service that is not hidden is processed first. If you set root_is_hidden to the first service rather than second, I think the root would be hidden while it should not be

However, I'm ok merging as is since it works for parallelstore. This will just need to be fixed when in a follow up

internal/surfer/testdata/hidden_surface/expected/target/surface/hidden_surface/resources/_partials/_describe_ga.yaml
```diff
@@ -1,54 +0,0 @@
-# -*- coding: utf-8 -*- #
```
sarahheacock, line 1: To ensure we do not move away from target, I've been enabling the target tests here https://github.com/googleapis/librarian/blob/e3ac27146d3760e5e4098ac50e4d9357ce836438/internal/surfer/golden_file_test.go#L39 and keeping the target files for now so they run on presubmit.

Open to other suggestions. But I feel like it's helpful for tracking and making sure we do not diverge while implementing on these features. And then we can delete them.

internal/surfer/testdata/multi_service/input/gcloud.yaml
```diff
@@ -22,3 +22,4 @@ apis:
     api_version: v1
     release_tracks:
       - GA
+    root_is_hidden: true
```
sarahheacock, line 25: I think this is a good addition to the test

internal/sidekick/surfer/provider/config.go
```diff
@@ -295,3 +299,19 @@ func SupportsStarUpdateMasks(c *Config, serviceName, version string) bool {
 	}
 	return *apiConfig.SupportsStarUpdateMasks
 }
+
+// FindAPIConfigByService finds the API config that applies to the given service.
+func FindAPIConfigByService(c *Config, service *api.Service) *API {
```
sarahheacock, line 304: Make sure to merge head and use the API func

internal/sidekick/surfer/group_builder.go
sarahheacock, line 41: Can you add a comment here?

internal/librarian/surfer/testdata/publicca/publicca/external_account_keys/_partials/_create_ga.yaml
```diff
@@ -17,7 +17,6 @@
 - release_tracks:
```
sarahheacock, line 17: Can you enable in golden file test?

sarahheacock (approved):

PR #5766, opened 2026-05-01
refactor(internal/sidekick/surfer): delete surfaceBuilder

julieqiu:
Remove the surfaceBuilder struct and pass its fields (model and config) directly to the buildSurface function to remove an unnecessary layer of abstraction.

Also delete isFlattenedSegment, and use the flattenedSegments map directly to check for existence.

### DIFF
--- a/internal/sidekick/surfer/surface_builder.go
+++ b/internal/sidekick/surfer/surface_builder.go
@@ -19,35 +19,23 @@ import (
 	"github.com/googleapis/librarian/internal/sidekick/surfer/provider"
 )
 
-type surfaceBuilder struct {
-	model  *api.API
-	config *provider.Config
-}
-
-func newSurfaceBuilder(model *api.API, config *provider.Config) *surfaceBuilder {
-	return &surfaceBuilder{
-		model:  model,
-		config: config,
-	}
-}
-
-func (b *surfaceBuilder) build() (*Surface, error) {
+func buildSurface(model *api.API, config *provider.Config) (*Surface, error) {
 	var root *CommandGroup
-	providerTracks := provider.Tracks(provider.APIVersionFromModel(b.model))
+	providerTracks := provider.Tracks(provider.APIVersionFromModel(model))
 	var tracks []ReleaseTrack
 	for _, t := range providerTracks {
 		tracks = append(tracks, ReleaseTrack(t))
 	}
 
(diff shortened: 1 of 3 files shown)


internal/sidekick/surfer/surface_builder.go
```diff
@@ -19,35 +19,23 @@ import (
 	"github.com/googleapis/librarian/internal/sidekick/surfer/provider"
 )
 
-type surfaceBuilder struct {
-	model  *api.API
-	config *provider.Config
-}
-
-func newSurfaceBuilder(model *api.API, config *provider.Config) *surfaceBuilder {
```
sarahheacock, line 27: This method is definitely fair to replace unnecessary

internal/sidekick/surfer/surface_builder.go
```diff
@@ -19,35 +19,23 @@ import (
...
-	return &surfaceBuilder{
-		model:  model,
-		config: config,
-	}
-}
-
-func (b *surfaceBuilder) build() (*Surface, error) {
```
sarahheacock, line 34: The goal of the builder structs is to group together similar functions and pass down the same context to all of the functions. It should be stateless / not a traditional builder that mutates the output as it goes.

Before, there was a large list of arguments that were passed down to multiple functions that were hard to trace.  I think the (b *surfaceBuilder) makes it easier to visually group these functions together. However, if it looks more complicated than it should be, can we have a struct that is passed down to each function? The struct can be improved over time and only exact values needed passed in. Just tracing back through a bunch of functions to find exact value missing can be challenging rather than passing around a pretty, slimmed down struct.

sarahheacock (approved):

internal/sidekick/surfer/surface_builder.go
julieqiu: Ack.

internal/sidekick/surfer/surface_builder.go
julieqiu: I think adding a separate struct for just two fields introduces more complexity than it saves right now. It's fairly easy to move these into a dedicated struct later if the list grows, but for now, the current setup optimizes for making things simple and easy to trace.

sarahheacock: Fair enough

PR #5767, opened 2026-05-01
feat(internal/librarian): limit "add" command to one API path

### LINKED ISSUE #5815: proposal: limit "librarian add" to a single API path
The librarian "add" command currently accepts multiple API paths, but its behavior isn't obvious, and it will be incorrect for at least some languages.

For a start, it only ever adds or updates a single library - so if were to try to clear a backlog by running "librarian add google/developers/knowledge/v1 google/devicesandservices/healthv4" then that will do the wrong thing in basically all cases.

Even if you do something more reasonable, e.g. "librarian add google/cloud/tasks/v3 google/cloud/tasks/v3beta" that's only a suitable thing to do for some languages (those which allow betas, and can have multiple variants in a single library).

Proposal: restrict "librarian add" to accepting a single API path. This will simplify the logic, and avoid any human confusion about how it should be used.

If we find this gets annoying for operators, we can gather information about use cases where we *want* to specify multiple API paths in a single call, and figure out the right way forward. For example, an outer loop that simply iterates over the arguments and calls the inner logic one argument at a time may be sufficient.

jskeet:
Requires that only one argument is specified for "librarian add". This reduces the number of paths that the code was already considering (e.g. detecting a mixture of stable and preview APIs) and closes out possibilities of doing entirely the wrong thing (e.g. creating a library from two unrelated APIs).

Fixes #5815


internal/librarian/add_test.go
```diff
@@ -142,33 +142,28 @@ func TestAddCommand(t *testing.T) {
...
 		{
 			name:    "no args",
-			wantErr: errMissingAPI,
+			wantErr: errWrongAPICount,
 		},
 		{
 			name:     "single API",
-			apis:     []string{"google/cloud/secretmanager/v1"},
+			args:     []string{"google/cloud/secretmanager/v1"},
 			wantName: "google-cloud-secretmanager-v1",
```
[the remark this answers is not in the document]
jskeet: Done.

internal/librarian/add.go
```diff
@@ -75,20 +73,20 @@ A typical librarian workflow for adding a new client library is:
 	librarian generate <library>   # generate the client library`,
 		Action: func(ctx context.Context, c *cli.Command) error {
```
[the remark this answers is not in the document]
jskeet: Rewritten, and generated.

jskeet:
Investigating the test failure now. (I think the test was previously broken...)

internal/librarian/add_test.go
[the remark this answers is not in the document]
jskeet: No, wantAPIs can't be populated here, because tidying removes the API. We end up with a library without any APIs, where the API path would be inferred.

internal/librarian/add.go
```diff
@@ -39,34 +39,31 @@ import (
...
 
 var (
-	errLibraryAlreadyExists      = errors.New("library already exists in config")
-	errAPIAlreadyExists          = errors.New("api already exists in library")
-	errAPIDuplicate              = errors.New("api duplicate in input")
-	errMissingAPI                = errors.New("must provide at least one API")
-	errMixedPreviewAndNonPreview = errors.New("cannot mix preview and non-preview APIs")
-	errPreviewRequiresLibrary    = errors.New("only APIs with an existing Library can have a Preview")
-	errPreviewAlreadyExists      = errors.New("preview library config already exists")
+	errLibraryAlreadyExists   = errors.New("library already exists in config")
```
julieqiu: nit: order alphabetically

julieqiu:
> Requires that only one argument is specified for "librarian add". This reduces the number of paths that the code was already considering (e.g. detecting a mixture of stable and preview APIs) and closes out possibilities of doing entirely the wrong thing (e.g. creating a library from two unrelated APIs).

What was the motivation for this change? Is there an issue we can link to?

julieqiu (approved):

internal/librarian/add.go
```diff
@@ -39,34 +39,31 @@ import (
...
+		Description: `add registers a single API in librarian.yaml.
 
-Each <api> is a path within the configured googleapis source, such as
+The <api> is a path within the configured googleapis source, such as
 "google/cloud/secretmanager/v1". The library name and other defaults are
 derived from the first API path using language-specific rules.
 
-Multiple API paths may be passed to bundle them into a single library. To
-add a preview client of an existing library, prefix every API path with
-"preview/"; preview and non-preview APIs cannot be mixed in one invocation.
```
zhumin8, line 64: It might be useful to keep a line around how to add preview client to existing library?

jskeet:
> What was the motivation for this change? Is there an issue we can link to?

I can create an issue from the conversation in chat.

I can't address the comments at this point before I start my weekend (I have other commitments); will put this into draft and create the issue and fix the comments on Tuesday.

jskeet:
> > Requires that only one argument is specified for "librarian add". This reduces the number of paths that the code was already considering (e.g. detecting a mixture of stable and preview APIs) and closes out possibilities of doing entirely the wrong thing (e.g. creating a library from two unrelated APIs).
> 
> What was the motivation for this change? Is there an issue we can link to?

Filed https://github.com/googleapis/librarian/issues/5815; will link to it in the PR description too.

internal/librarian/add.go
jskeet: Good call - done, while removing the "mixed in one invocation" part which is effectively obsolete.

internal/librarian/add.go
jskeet: Done.

jskeet:
Merging to unblock further Python work - changes requested in review have been implemented.

PR #5775, opened 2026-05-01
refactor(internal/surfer): consolidate testdata protos

### LINKED ISSUE #5744: internal/surfer: consolidate testdata protos into `internal/testdata/googleapis`
protos and service config are stored in both
- `internal/testdata/googleapis`
- `internal/surfer/testdata/apis`

We should consolidate these and move all files into `internal/testdata/googleapis/`

julieqiu:
Move protos and service configs are moved from internal/surfer/testdata/apis to internal/testdata/googleapis so that all googleapis testdata lives in the same location.

Copy oslogin and iam/v3 protos directly from googleapis/googleapis so the testdata matches upstream, and TestGolden passes.

Fixes https://github.com/googleapis/librarian/issues/5744


sarahheacock (approved):

jameslynnwu (changes requested):
Let's update [internal/testdata/googleapis/README.md](https://github.com/googleapis/librarian/blob/main/internal/testdata/googleapis/README.md)

CONTRIBUTING.md
```diff
@@ -445,6 +445,14 @@ tools are not available in the development environment. See
 [doc/onboarding.md](https://github.com/googleapis/librarian/blob/main/doc/onboarding.md)
 for information on how to configure your development environment.
 
+### Testdata
+
+Librarian accepts protos, discovery documentation, OpenAPI specifications, and service configurations as input. These live in `internal/testdata/`.
+
+- `internal/testdata/googleapis`: Contains a subset of [googleapis](https://github.com/googleapis/googleapis) protos. To update, clone the repository and copy the required protos maintaining the directory structure.
```
jameslynnwu, line 452: we might want to create a script for updating this testdata in the future

jameslynnwu (approved):

CONTRIBUTING.md
julieqiu: I think the script would basically just be:

`cp /path/to/googleapis/{api} /path/to/internal/testdata/googleapis/{api}`

So I’m not sure we’d get much value from formalizing it yet, but if the process gets more complicated over time we can definitely add a helper script later.

PR #5780, opened 2026-05-01
feat(internal/legacylibrarian): propagate the release notes again

codyoss:
This change is a partial revert of
5b2f035c7a2cd2679cc4e8034022084d30499aaa. We have decided to have release notes populate with commits again. This should have the effect of significant changes in hand written libraries appearing in the release notes.


internal/legacylibrarian/legacylibrarian/release_stage_test.go
```diff
@@ -1434,6 +1434,20 @@ func TestProcessLibrary_ReleaseOnlyMode(t *testing.T) {
 				Version:          "1.3.0",
 				SourceRoots:      []string{"one-id"},
 				ReleaseTriggered: true,
+				Changes: []*legacyconfig.Commit{
+					{
+						Type:       "chore",
```
codyoss, line 1439: @noahdietz Is this an okay behaviour or should a add some net-new filtering here. These do not show up in the PR body as we already filter them but chore commits are forwarded to the language containers. I am pretty sure in Go we just dropped these on the floor, but I am not sure the same is true in python.

jskeet: @parthea might know off hand, but my sample commit from earlier suggests they're dropped.

noahdietz: If all commits are sent to the language container and it is up to the language container to decide what to do with them, filter or include, then this test expectation seems correct - include the `chore` commits in the commits sent to the container bc they are relevant to the library path.

noahdietz (approved):

zhumin8:
nit: typo in title "propagate"

noahdietz:
@codyoss can you fix the typo in the title/commit message :D

PR #5791, opened 2026-05-02
refactor(internal/sidekick/api): replace path template builders

julieqiu:
Change PathSegment.Literal and PathTemplate.Verb from `*string` to `string`, where the empty string means the field is unset.

Delete the builder methods on PathTemplate, PathVariable, and PathSegment and rewrite all call sites using plain struct literals instead. Once the pointer fields were removed, the builder chain no longer provided any meaningful benefit over directly constructing the structs.

### DIFF
--- a/internal/sample/api.go
+++ b/internal/sample/api.go
@@ -83,11 +83,17 @@ func MethodCreate() *api.Method {
 			Bindings: []*api.PathBinding{
 				{
 					Verb: http.MethodPost,
-					PathTemplate: (&api.PathTemplate{}).
-						WithLiteral("v1").
-						WithLiteral("projects").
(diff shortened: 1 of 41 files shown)


coryan (approved):
> Once the pointer fields were removed, the builder chain no longer provided any meaningful benefit over directly constructing the structs.

And yet the new code is 2200 lines vs. 970...  Not to mention that the builders made it harder to make some mistakes like setting both the `Literal` and `Variable` strings.

```
Segments: []api.PathSegment{
  {Literal: "v1", Variable: .... }
}
```

🤷

julieqiu:
Alright, let's close this PR for now and revisit it later.

Issue #5485, opened 2026-04-20
surfer: fix operation command output


sarahheacock:
If you run operations commands against golden, the resource arguments are slightly off. Update using necessary info from sidekick

PR #5798, opened 2026-05-04
feature(internal/sidekick/surfer): operations commands

sarahheacock:
- Infers resource from operation method patterns to populate resource argument. Resource descriptors are not available for operation mixins
- Adds new mixins file to provide missing information such as documentation.
- Skip return_partial_success field since not available on all APIs

For https://github.com/googleapis/librarian/issues/5485


internal/sidekick/surfer/provider/mixin.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+// OperationMethodDocumentation returns the fallback help text for operations methods.
+// Current documentation is stripped from the proto in the LRO service.
+func OperationMethodDocumentation(methodName string) string {
+	switch methodName {
+	case GetOperation:
+		return "The name of the operation resource."
+	case CancelOperation:
+		return "The name of the operation resource to be cancelled."
```
jameslynnwu, line 48: Canceled is preferred in american english.

internal/sidekick/surfer/provider/mixin.go
```diff
@@ -0,0 +1,157 @@
...
+	"operations": "operation",
+}
+
+// InferOperationResource creates a synthetic resource for operations based on the method's path.
+func InferOperationResource(m *api.Method) (*api.Resource, error) {
+	if m.PathInfo == nil || len(m.PathInfo.Bindings) == 0 {
+		return nil, nil
+	}
+
+	inflections := BuildInflectionMap(m.Model)
```
jameslynnwu: why build this map? can we just have a function that returns the singular tense for a word?

internal/sidekick/surfer/provider/resource.go
```diff
@@ -105,6 +106,11 @@ func IsPrimaryResourceField(field *api.Field, method *api.Method) bool {
 		return true
 	}
 
+	// Fallback for operations methods where the primary resource field is named "name".
+	if IsOperationsResourceField(field, method) {
+		return true
+	}
```
jameslynnwu, line 112: the logic in IsOperationResourceField looks similar to the logic above. also, let's order args method, field.

jameslynnwu (changes requested):

internal/sidekick/surfer/provider/mixin.go
sarahheacock: Based from proto annotations and gen_sfc https://github.com/googleapis/googleapis/blob/master/google/longrunning/operations.proto#L213. We can update but would need to change target

internal/sidekick/surfer/provider/mixin.go
sarahheacock: Just to help with reasoning / testing. Removed due to offline conversation regardless. BuildInflectionMap will not work for our use case since the singular noun is not standardized

internal/sidekick/surfer/provider/resource.go
sarahheacock: Operations uses `name` instead of `parent` for ListOperations which is not AIP complaint https://github.com/googleapis/googleapis/blob/master/google/longrunning/operations.proto#L169. So operations does not get flagged here.

Updated comment in mixins. Fine with going from broader to smaller but would rather do that separately. Current way is inline with the rest of the file. Making it an outliar would be confusing

jameslynnwu: My first thought is whether or not we need IsOperationResourceField. We could do
`IsOperationsMethod(method) && field.Name == "name"`, which mirrors the if statements above.

However, I think this might express the intent better? We might want to make it specific to ListOperations also.
```
if IsCollectionMethod(method) {
  if field.Name == "parent" {
     return true
  } else if IsOperationMethod(method) && field.Name == "name" {
     return true
  }
}
```

re: going broader to smaller in a separate PR. I personally think it makes more sense to go broader to smaller. wdyt?

internal/sidekick/surfer/provider/resource.go
```diff
@@ -173,24 +179,16 @@ func GetResourceForMethod(method *api.Method, model *api.API) *api.Resource {
...
-	}
-
 	// TODO(https://github.com/googleapis/librarian/issues/3363): Avoid this lookup by linking the ResourceReference
 	// to the Resource definition during model creation or post-processing.
 
 	// Use the API model's indexed maps for an efficient lookup.
-	for _, r := range model.ResourceDefinitions {
-		if r.Type == resourceType {
+	for _, r := range GetAllResources(model) {
+		if resourceType != "" && r.Type == resourceType {
```
jameslynnwu: do we actually need to check if resourceType != "" here and if resourceType == "" below?

internal/sidekick/surfer/provider/resource.go
```diff
@@ -336,10 +334,43 @@ func GetLiteralSegments(raw []api.PathSegment) []string {
...
 }
 
+// GetAllResources returns all resource definitions in the model, including
+// file-level definitions, message-level definitions, and synthetic resources
+// inferred from operations methods.
+func GetAllResources(model *api.API) []*api.Resource {
+	var resources []*api.Resource
+	resources = append(resources, model.ResourceDefinitions...)
+
+	// Add message-level resources if not already present
```
jameslynnwu: "if not already present", what would cause them to be or not be present?

I believ message-level resources are already part of model.ResourceDefinitions? https://github.com/googleapis/librarian/blob/7c5dc907a23bb68ad3598955be41689144c34307/internal/sidekick/parser/protobuf.go#L305

internal/sidekick/surfer/provider/resource.go
```diff
@@ -336,10 +334,43 @@ func GetLiteralSegments(raw []api.PathSegment) []string {
...
+	for _, m := range model.Messages {
+		if m.Resource != nil {
+			resources = append(resources, m.Resource)
+		}
+	}
+
+	// Infer operations resources from GetOperation methods
+	for _, s := range model.Services {
+		for _, m := range s.Methods {
+			if m.Name == GetOperation && IsOperationsMethod(m) {
```
jameslynnwu: how is a GetOperation method not an operations method?

internal/sidekick/surfer/provider/mixin.go
```diff
@@ -0,0 +1,146 @@
...
+		return "The name of the operation resource to be cancelled."
+	case DeleteOperation:
+		return "The name of the operation resource to be deleted."
+	case ListOperations:
+		return "The name of the operation's parent resource."
+	}
+	return ""
+}
+
+// InferOperationResource creates a synthetic resource for operations based on the method's path.
```
jameslynnwu: does this function need to be exported?

internal/sidekick/surfer/argument_builder_test.go
```diff
@@ -269,6 +289,56 @@ func TestNewPrimaryResourceArgument(t *testing.T) {
...
+				HelpText:          "The name of the operation resource.",
+				IsPositional:      true,
+				IsPrimaryResource: true,
+				Required:          true,
+				ResourceSpec: &ResourceSpec{
+					Name:       "operation",
+					PluralName: "operations",
+					Collection: "test.projects.locations.operations",
+					Attributes: []Attribute{
+						{ParameterName: "projectsId", AttributeName: "project", Help: "The project id of the {resource} resource.", Property: "core/project"},
```
jameslynnwu, line 335: outside of scope of this PR, but we should use ID instead of id

internal/sidekick/surfer/provider/resource.go
```diff
@@ -336,10 +334,43 @@ func GetLiteralSegments(raw []api.PathSegment) []string {
 	return filtered
 }
 
+// GetAllResources returns all resource definitions in the model, including
+// file-level definitions, message-level definitions, and synthetic resources
+// inferred from operations methods.
+func GetAllResources(model *api.API) []*api.Resource {
```
jameslynnwu: this can probably be unexported?

internal/sidekick/surfer/provider/resource.go
```diff
@@ -336,10 +334,43 @@ func GetLiteralSegments(raw []api.PathSegment) []string {
...
+		}
+	}
+
+	// Infer operations resources from GetOperation methods
+	for _, s := range model.Services {
+		for _, m := range s.Methods {
+			if m.Name == GetOperation && IsOperationsMethod(m) {
+				res, err := InferOperationResource(m)
+				if err != nil {
+					log.Printf("WARNING: failed to infer operations resource for method %q: %v", m.ID, err)
```
jameslynnwu: why isn't this an error?

internal/sidekick/surfer/argument_builder.go
```diff
@@ -96,6 +96,10 @@ func (b *argumentBuilder) isIgnored() bool {
 		switch b.field.Name {
 		case "page_size", "page_token", "filter", "order_by":
 			return true
+		case "return_partial_success":
+			// Field is available in all APIs due to mixin but not all APIs actually
+			// support it. Omitting for now.
```
jameslynnwu, line 101: how does this comment relate to the IsOperationsMethod check?

internal/sidekick/surfer/provider/resource.go
```diff
@@ -373,3 +404,22 @@ func GetPluralResourceTypeName(model *api.API, methodPath []string) string {
 	}
 	return res.Plural
 }
+
+// TODO(https://github.com/googleapis/librarian/issues/3414): Use a dedicated inflection library
+// or make this configuration-driven in the long term.
+// singular converts a plural collection segment name (e.g., "projects", "locations", "operations")
+// into its singular variable form (e.g., "project", "location", "operation").
+func singular(s string) string {
```
jameslynnwu, line 412: IIUC, based on https://google.aip.dev/123, resource patterns will always have the singular version embedded because:
- Well-defined singular and plurals of a resource enable clients to determine the proper name to use in code and documentation.

This is used by GetSingularResourceNameForMethod.

However, expandNestedVariable uses path_templates, which have wildcards in them, so we have to infer. Did the previous inflection map get around this by building a map of singular names using resource patterns?

jameslynnwu (changes requested):

internal/sidekick/surfer/provider/resource.go
sarahheacock: We don't want to halt all generation because we cannot generate operations. In general, we need to put more thought behind it

internal/sidekick/surfer/provider/resource.go
sarahheacock: I could go either way. I initially had it inline similar to your suggestion. But operations is not AIP compliant and I did not want to litter a bunch of exceptions through out the code. However, if it's hard to understand, that's also a problem. I'll default to you on this one.

Re: broader to smaller. I had to look it up to be honest. But I think you're right / it's more common practice. But I think it's worse to toggle around the order of the arguments in the same file. Created follow up https://github.com/googleapis/librarian/issues/5821

internal/sidekick/surfer/provider/resource.go
sarahheacock: "r.Type" is never an empty string so was able to remove there. But we do need the resourceType == "" for the case where a type is just not found in GetAllResources

internal/sidekick/surfer/provider/resource.go
sarahheacock: I think you're right! Removed

internal/sidekick/surfer/provider/resource.go
sarahheacock: You could technically have a GetOperation rpc method in a different service

internal/sidekick/surfer/argument_builder_test.go
sarahheacock: Unfortunately, projectsId is tied to apitools. But to your point, we don't want main logic to tie to fragile heuristics. We should move the "Id" suffix to the writer

internal/sidekick/surfer/provider/mixin.go
sarahheacock: Unexported

internal/sidekick/surfer/provider/resource.go
sarahheacock: Unexported

internal/sidekick/surfer/argument_builder.go
sarahheacock: Made check more explicit, but we only want to filter out for ListOperations method

internal/sidekick/surfer/provider/resource.go
sarahheacock: Yes, that's exactly what the map did. Unfortunately, according the to the AIP, you can have anything between the `{...}` as long as it's snake case and does not have `_id`. The "singular" and "plural" is referring to 

```
message Topic {
  option (google.api.resource) = {
    type: "pubsub.googleapis.com/Topic"
    pattern: "projects/{project}/topics/{topic}"
    singular: "topic" <<<<<
    plural: "topics"  <<<<<
  };

  // Name and other fields...
}
```

We cannot rely on those annotations in this circumstances since it is the type name and not the collection. For example, the follow is also valid https://google.aip.dev/122

```
message UserEvent {
  option (google.api.resource) = {
    type: "example.googleapis.com/UserEvent"
    // Only the collection & resource identfiers in the `pattern` are shortened.    
    pattern: "projects/{project}/users/{user}/events/{event}" <<<<<
    singular: "userEvent" <<<<
    plural: "userEvents"
  };

  string name = 1;
}
```

internal/sidekick/surfer/argument_builder_test.go
jameslynnwu: Sorry, I meant the help text should be "The project ID of the "

internal/sidekick/surfer/provider/resource.go
jameslynnwu: let's use `slog.Warn` then. its more common in the codebase

internal/sidekick/surfer/provider/resource.go
jameslynnwu: We should tackle this in a separate PR, if its feasible.

in the second example, could we use the pattern defined to figure out what the field paths for the wildcards should be when we encounter a path template like `{projects/*/users/*/events*}`

internal/sidekick/surfer/provider/resource.go
sarahheacock: And the only issue is with ListOperations at the moment. IsResourceMethod and IsCollectionMethod are incorrect / rely on rpc strings which we need to get rid of. So it could break later. But keeping the logic simple for now and adding test cases is fine

jameslynnwu: let's inline the check

internal/sidekick/surfer/provider/resource.go
jameslynnwu: Discussed offline, renaming IsOperationsMethod to IsOperationsServiceMethod

internal/sidekick/surfer/provider/resource.go
jameslynnwu: discussed offline.

jameslynnwu (changes requested):

internal/sidekick/surfer/provider/resource.go
sarahheacock: Sorry, my point is that you can't rely on `projects/{project}/users/{user}/events/{event}`. It's the wild west for what is in the brackets. I've seen stuff like `projects/{project_or_number}/users/{user}/events/{user_event}`. All are AIP compliant and messed up gen_sfc. But agreed it's a separate concern

internal/sidekick/surfer/argument_builder_test.go
sarahheacock: Got it. We can do that. But agree it's out of scope

internal/sidekick/surfer/provider/resource.go
sarahheacock: Done

internal/sidekick/surfer/provider/resource.go
sarahheacock: Done

internal/sidekick/surfer/argument_builder.go
jameslynnwu: Update to say

Omitting for now, except for ListOperations.

jameslynnwu (approved):

Issue #5731, opened 2026-04-30
java: generate showcase


zhumin8:
Migration script and librarian logic may need special adjustments for generating this library.
java-showcase used to be part of sdk-platform-java, now directly in google-cloud-java

zhumin8:
In hermetic build, showcase commit is not specified in generation_config.yaml, instead, it is sourced from the pom.xml file:
 
1. The Workflow Triggers Showcase Mode
  In hermetic_library_generation.yaml, the job explicitly sets  showcase_mode: true  when calling the local action  ./sdk-platform-java/.github/scripts .

 2. The Script Fetches the Version from  pom.xml 
  The action calls hermetic_library_generation.sh, which sources showcase_utilities.sh.
  • reads the  gapic-showcase.version  property directly from the file  java-showcase/gapic-showcase/pom.xml with grep.
  • clones and checkout repo at the commit

3. It Integrates Protos into the Input Directory Before Generate
  The script copies the  schema  directory (which contains the protos) from the cloned repository directly into the folder where  googleapis  was cloned:

    cp -r "${showcase_def_dir}/schema" "${api_def_dir}/"

zhumin8:
In migrate script, will source version specified in java-showcase/gapic-showcase/pom.xml to populate librarian.yaml sources.
Then after migrated to librarian generate, librarian.yaml becomes the source of truth for showcase version.

These 2 values risks out of sync, should at least introduce a workflow to check these points to the same commit:
- librarian.yaml sources showcase commit
- java-showcase/gapic-showcase/pom.xml gapic-showcase.version property

zhumin8:
remaining: 
skip pom update (implement excluded_poms)
keep needs to be fixed

zhumin8:
changes in https://github.com/googleapis/google-cloud-java/pull/13030, now version.txt does not have showcase versions. It is used in migrate script to populate librarian.yaml versions. Need a hardcoded value for showcase.

zhumin8:
local tested with no diff

PR #5799, opened 2026-05-04
feat(internal/librarian/java): support showcase as source in generate

zhumin8:
Support showcase as source in generate, pass both showcase dir and googleapis dir to protoc via -I flags, use only showcase dir for copy protos, and finding service configs. Also renaming variables related to googleapisDir for clarity.

For https://github.com/googleapis/librarian/issues/5731

### DIFF
--- a/internal/librarian/java/generate.go
+++ b/internal/librarian/java/generate.go
@@ -103,7 +103,7 @@ type librarySources struct {
 
 	// googleapisDir is the absolute path to the googleapis source directory.
 	// This is always needed (even for showcase) because common proto resources always reside in
-	//  the googleapis repository.
+	// the googleapis repository.
 	googleapisDir string
 
 	// includeDirs contains the ordered list of import directories to pass to protoc via -I flags.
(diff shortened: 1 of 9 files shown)


internal/librarian/java/generate.go
```diff
@@ -94,10 +100,50 @@ func Generate(ctx context.Context, cfg *config.Config, library *config.Library,
 	}); err != nil {
 		return err
 	}
-
 	return nil
 }
 
+type librarySources struct {
```
sofisl: this feels like it *should* be cross-language?

sofisl: Ok I think in reviewing this PR, my main question is around this. Not only is this structure relevant for all languages, it is actually mostly the same for all languages IIUC (i.e., googleapis will be the same for all, no?)

zhumin8: Good point, we actually already have `SourceConfig` for it (used in Dart and Rust). Let me refactor this.
https://github.com/googleapis/librarian/blob/c33f0346a1236fd2c4089c3be17515adec4140bc/internal/sources/sources.go#L34-L51

sofisl (approved):

PR #5804, opened 2026-05-04
feat(internal/librarian/golang): extract Go compiler version from Tools configuration

### LINKED ISSUE #5201: golang: onboarding a new API sets the wrong Go version in go.mod
After doing `librarian add` for a new API, and running `librarian generate` for it, the `go.mod` initialized by `librarian` uses the Go version of the `go` tool used to invoke librarian i.e. 1.26.1. However, google-cloud-go has a policy of setting the Go version in `go.mod` files to the `latest - 1` version as per https://github.com/googleapis/google-cloud-go?tab=readme-ov-file#go-versions-supported.

https://github.com/googleapis/librarian/blob/befad4aa0ce627344d058eccc55f8ac757d5bbd2/internal/librarian/golang/module.go#L198-L203

We need to figure out a way to set the appropriate Go version for the repo.

cc @JoeWang1127

sofisl:
Scanning Tools.Go to find the Go compiler package under golang.org/dl/go. If found, it extracts the base name (e.g. go1.22.3) and uses it to run go mod init and go mod tidy during Go client generation. If not found, it falls back to command.Go.

We verified this by adding the tool in testdata/librarian.yaml and asserting the resolved Go version in new test cases.

Fixes #5201


internal/librarian/generate.go
```diff
@@ -225,7 +225,8 @@ func generateLibraries(ctx context.Context, cfg *config.Config, libraries []*con
 		g, gctx := errgroup.WithContext(ctx)
 		for _, library := range libraries {
 			g.Go(func() error {
-				if err := golang.Generate(gctx, library, src); err != nil {
+				goCmd := golang.GoCommand(cfg.Tools)
```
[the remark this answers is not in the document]
sofisl: done

internal/librarian/golang/generate_test.go
```diff
@@ -1102,3 +1103,47 @@ func setupSnippets(t *testing.T, repoRoot string) {
 		t.Fatal(err)
 	}
 }
+
+func TestGoCommand(t *testing.T) {
```
noahdietz, line 1106: I don't think we need this test in particular. The `TestGoCommand_Default` should be combined with this one, and the test case should just include a `Toosl.Go` entry with `golang.org/dl/go1.22.3` specified to see that it resolves.

internal/librarian/golang/testdata/librarian.yaml
```diff
@@ -25,3 +25,5 @@ tools:
       version: v1.3.0
     - name: google.golang.org/protobuf/cmd/protoc-gen-go
       version: v1.36.11
+    - name: golang.org/dl/go1.22.3
+      version: v0.0.0-20260409215720-060e9b17b980
```
noahdietz: Either we shouldn't set this here OR set it to go1.25.0 (what we actually use in google-cloud-go`) bc it will be used for all Go related tests (`golang/testdata/librarian.yaml` is used as our integration test dependency manifest, right?). 

See my other comment about testing the `GoCommand` logic - I'm not sure it needs to read from a librarian.yaml to be adequately tested.

noahdietz (changes requested):

noahdietz (approved):

noahdietz:
Sweet!!

PR #5805, opened 2026-05-04
feat(sidekick/surfer): prefix nested message flag names with parent names

### LINKED ISSUE #5511: surfer: enable flag naming prefixes
Flag names for nested messages require prefix. Look at parallelstore export data for reference

jameslynnwu:
Surfer now prepends parent field names to generated gcloud flag names for fields inside nested messages, preventing collisions and matching gen_sfc behavior.

The new logic derives the flag name from the API field path and applies rules to strip request message names and explicit body field names, avoiding redundant prefixes. This logic is encapsulated in a new deriveArgName helper function.

Fixes https://github.com/googleapis/librarian/issues/5511

### DIFF
--- a/internal/sidekick/surfer/argument_builder.go
+++ b/internal/sidekick/surfer/argument_builder.go
@@ -31,7 +31,7 @@ type argumentParams struct {
 	model     *api.API
 	service   *api.Service
 	field     *api.Field
-	apiField  []string
+	fieldPath []*api.Field
 }
 
 // newArgument creates a single command-line argument (an `Argument` struct) from the parameters.
@@ -40,11 +40,13 @@ func newArgument(ap *argumentParams) (*Argument, error) {
 	if isArgIgnored(ap.field, ap.method) {
 		return nil, nil
 	}
+	if len(ap.fieldPath) == 0 {
+		return nil, fmt.Errorf("field %q has empty field path", ap.field.Name)
+	}
 
-	// TODO(https://github.com/googleapis/librarian/issues/3414): Abstract away casing logic in the model.
 	arg := &Argument{
-		ArgName:   ap.field.Name,
-		APIField:  ap.apiField,
+		ArgName:   argName(ap),
+		APIField:  apiField(ap),
 		Required:  ap.field.DocumentAsRequired(),
 		Repeated:  repeated(ap.field),
 		Clearable: clearable(ap.field, ap.method),
@@ -258,4 +260,38 @@ func newAttributesFromSegments(segments []api.PathSegment) []Attribute {
 		attributes = append(attributes, attr)
 	}
 	return attributes
+}
+
+func argName(ap *argumentParams) string {
+	parts := ap.fieldPath
+	// Strip explicit body field name (e.g., "instance" when body: "instance").
+	// The explicit body message represents the main resource being acted upon,
+	// so its fields should appear as top-level flags without redundant prefixes.
+	if ap.method.PathInfo != nil && !isBodyWildcard(ap) {
+		if len(parts) > 0 && parts[0].Name == ap.method.PathInfo.BodyFieldPath {
+			parts = parts[1:]
+		}
+	}
+
+	var names []string
+	for _, f := range parts {
+		names = append(names, f.Name)
+	}
+	return strings.Join(names, "_")
+}
+
+// apiField constructs the path segments to the field in the API request message.
+func apiField(ap *argumentParams) []string {
+	var apiFields []string
+	if isBodyWildcard(ap) && ap.method.InputType != nil {
+		apiFields = append(apiFields, ap.method.InputType.Name)
+	}
+	for _, f := range ap.fieldPath {
+		apiFields = append(apiFields, f.JSONName)
+	}
+	return apiFields
+}
+
+func isBodyWildcard(ap *argumentParams) bool {
+	return ap.method.PathInfo != nil && ap.method.PathInfo.BodyFieldPath == "*"
 }
--- a/internal/sidekick/surfer/command_builder.go
+++ b/internal/sidekick/surfer/command_builder.go
@@ -179,15 +179,10 @@ func requestMethod(method *api.Method) string {
 	return ""
 }
 
-type fieldWithPrefix struct {
-	field  *api.Field
-	prefix []string
-}
-
 type classifiedFields struct {
-	primaryField    *fieldWithPrefix
-	resourceIdField *fieldWithPrefix
-	other           []fieldWithPrefix
+	primaryField    []*api.Field
+	resourceIdField []*api.Field
+	other           [][]*api.Field
 }
 
 // newArguments generates the set of arguments for a command by parsing the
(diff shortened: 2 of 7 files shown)


internal/sidekick/surfer/argument_builder.go
```diff
@@ -88,6 +89,39 @@ func (b *argumentBuilder) build() (*Argument, error) {
 	return arg, nil
 }
 
+// deriveArgName computes the flag name for an argument.
+// For nested fields, it prepends parent field names to ensure uniqueness
+// and context, while stripping API request boilerplate.
+func (b *argumentBuilder) deriveArgName() string {
```
sarahheacock: Provider is a bad example but go idiomatic recommends keeping function succinct and dropping verbs. So "argName()" rather than "getArgName()"

internal/sidekick/surfer/argument_builder.go
```diff
@@ -54,9 +54,10 @@ func (b *argumentBuilder) build() (*Argument, error) {
 		return nil, nil
 	}
 
-	// TODO(https://github.com/googleapis/librarian/issues/3414): Abstract away casing logic in the model.
+	argName := b.deriveArgName()
```
sarahheacock: Nit: unnecessary assignment. Can keep inline

internal/sidekick/surfer/argument_builder.go
```diff
@@ -88,6 +89,39 @@ func (b *argumentBuilder) build() (*Argument, error) {
...
+func (b *argumentBuilder) deriveArgName() string {
+	argName := b.field.Name
+	if len(b.apiField) == 0 {
+		return argName
+	}
+
+	parts := b.apiField
+	// Strip the request message name (e.g., CreateInstanceRequest) if it was
+	// prepended by command_builder.go (happens when body: "*").
+	if b.method.InputType != nil && parts[0] == b.method.InputType.Name {
```
sarahheacock: I think this is fine for now, but coupling the CLI flag name calculation with the api_field string heuristics (rather than a semantic source of truth) forces fragile backward-calculations like this.

Instead of passing a slice of strings representing the api_field, we should pass in the list of traversed field descriptors (`[]*api.Field` or equivalent descriptor path) as the source of truth. This would allow us to compute api_field and the CLI arg_name independently and cleanly using the descriptors' actual metadata.

internal/sidekick/surfer/argument_builder.go
```diff
@@ -88,6 +89,39 @@ func (b *argumentBuilder) build() (*Argument, error) {
...
+		if len(parts) > 0 && parts[0] == strcase.ToLowerCamel(b.method.PathInfo.BodyFieldPath) {
+			parts = parts[1:]
+		}
+	}
+	// Only prefix if we are truly dealing with a nested field (len > 1).
+	// For true nested fields, join the parts and convert to consistent snake_case
+	// (which is later converted to kebab-case for CLI flags).
+	// If len == 1, it's a top-level flag, so we preserve original b.field.Name.
+	if len(parts) > 1 {
+		argName = strcase.ToSnake(strings.Join(parts, "_"))
```
sarahheacock: Nit: trying to only use casing in the writer to avoid tracing through multiple string conversions. Is ToSnake necessary?

internal/sidekick/surfer/argument_builder.go
```diff
@@ -259,3 +258,47 @@ func newAttributesFromSegments(segments []api.PathSegment) []Attribute {
...
+	argName := ap.field.Name
+	if len(ap.fieldPath) == 0 {
+		return argName
+	}
+
+	parts := ap.fieldPath
+	// Strip explicit body field name (e.g., "instance" when body: "instance").
+	// The explicit body message represents the main resource being acted upon,
+	// so its fields should appear as top-level flags without redundant prefixes.
+	if ap.method.PathInfo != nil && ap.method.PathInfo.BodyFieldPath != "" && ap.method.PathInfo.BodyFieldPath != "*" {
```
sarahheacock: Nit: do not think `p.method.PathInfo.BodyFieldPath != ""` is necessary. Also, can we use `!isBodyWildcard(ap)`?

internal/sidekick/surfer/argument_builder.go
```diff
@@ -259,3 +258,47 @@ func newAttributesFromSegments(segments []api.PathSegment) []Attribute {
 	}
 	return attributes
 }
+
+func argName(ap *argumentParams) string {
+	argName := ap.field.Name
+	if len(ap.fieldPath) == 0 {
+		return argName
```
sarahheacock: I think if if the length is zero, our code has made a mistake. We should verify in the constructor the length is not 0 and error out.

internal/sidekick/surfer/argument_builder.go
```diff
@@ -259,3 +258,47 @@ func newAttributesFromSegments(segments []api.PathSegment) []Attribute {
...
+	// so its fields should appear as top-level flags without redundant prefixes.
+	if ap.method.PathInfo != nil && ap.method.PathInfo.BodyFieldPath != "" && ap.method.PathInfo.BodyFieldPath != "*" {
+		if len(parts) > 0 && parts[0].Name == ap.method.PathInfo.BodyFieldPath {
+			parts = parts[1:]
+		}
+	}
+	// Only prefix if we are truly dealing with a nested field (len > 1).
+	// For true nested fields, join the parts. Since we are using []*api.Field,
+	// the Field.Name is already snake_case, so we just join them with "_".
+	if len(parts) > 1 {
```
sarahheacock: Same comment as above. We should be strict on our input expectations so we can simplify the code to...

var names []string
for _, f := range parts {
	names = append(names, f.Name)
}
return strings.Join(names, "_")

Falling back to field name when we should never have an empty list could lead to unexpected bugs

internal/sidekick/surfer/command_builder.go
```diff
@@ -179,15 +179,15 @@ func requestMethod(method *api.Method) string {
 	return ""
 }
 
-type fieldWithPrefix struct {
-	field  *api.Field
-	prefix []string
+type fieldWithPath struct {
+	field *api.Field
```
sarahheacock: Nit: `field` is just the last element, correct? Can this be simplified / this struct removed?

internal/sidekick/surfer/command_builder.go
```diff
@@ -276,31 +276,28 @@ func categorizeFields(method *api.Method, model *api.API) (classifiedFields, err
...
 		isBodyWildcard := bodyFieldPath == "*"
 		isBodyField := bodyFieldPath == field.Name || isBodyWildcard
 
-		var prefix []string
-		if isBodyWildcard {
-			prefix = append(prefix, method.InputType.Name)
-		}
+		var path []*api.Field
 
 		if isExpandableMessage && isBodyField {
```
sarahheacock, line 283: Unrelated to this PR / just giving heads up. This line of logic may not be quite right. gen_sfc collapses down arg groups one level at the root level. I did not know that when I added this line so the code may need to be tweaked slightly (and maybe even simplified) when arg groups are added. 

But all of the logic you added in this PR looks great. I'm just referring how we iterate through the first level of arguments in command_builder

sarahheacock (approved):

Issue #5769, opened 2026-05-01
sidekick/gcloud: prototype a gcloud CLI written in Go


julieqiu:
Prototype a gcloud CLI written in Go as a possible approach for gcloud autogen.

- [ ] https://github.com/googleapis/librarian/pull/5812#discussion_r3188420251: Eventually you will have to deal with services that have both global (projects/%s/secrets/%s) and regional (projects/%s/locations/%s/secrets/%s) resources. The ergonomics are better if you branch off location == "". But it gets messy because that requires switching the endpoint at the same time (but not for all services, yuck).

PR #5807, opened 2026-05-04
feat(internal/sidekick/gcloud): generate required flags

julieqiu:
Each generated command now requires one flag per variable in the resource it operates on. For example, a command on an instance gets --project, --location, and --instance. The flag names come from the resource's pattern in the proto annotation.

### DIFF
--- a/internal/sidekick/gcloud/generate.go
+++ b/internal/sidekick/gcloud/generate.go
@@ -55,9 +55,13 @@ type Subgroup struct {
 
 // Command represents a leaf command.
 type Command struct {
+	Flags []Flag
 	Name  string
 	Usage string
 }
+
+// HasFlags reports whether the command has any flags.
+func (c Command) HasFlags() bool { return len(c.Flags) > 0 }
 
 // Generate is the package entry point. It builds the model, renders main.go,
 // writes it, then renders any other generated files via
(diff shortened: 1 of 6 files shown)


coryan:
>   with a trailing "s" trimmed.

That is gross.  Try to use the annotations on the requests or responses, they have the singular and plural name of the type:

https://github.com/googleapis/librarian/blob/aa7040039744516e0d35b5a58a1480c96e2aeb4f/internal/sidekick/api/model.go#L1166-L1171

internal/sidekick/gcloud/generate.go
```diff
@@ -157,3 +159,55 @@ func constructCLIModel(model *api.API) CLIModel {
...
+// segment in the pattern, named after the variable's last FieldPath
+// component. Duplicates (same FieldPath) are skipped.
+func pathFlagsFromSegments(segments []api.PathSegment) []Flag {
+	var flags []Flag
+	seen := map[string]bool{}
+	for _, seg := range segments {
+		if seg.Variable == nil || len(seg.Variable.FieldPath) == 0 {
+			continue
+		}
+		name := seg.Variable.FieldPath[len(seg.Variable.FieldPath)-1]
```
coryan, line 205: I think this is an Okay heuristic, but eventually it will need some rework.  Better to start with something and improve it than being paralyzed while we design the perfect thing, so LGTM.

coryan (approved):

PR #5811, opened 2026-05-04
fix(legacylibrarian): restore chore as noop

### LINKED ISSUE #5786: legacylibrarian: Consider restoring chore as a no-op commit when determining release eligibility
## Background

In "release only" mode, `legacylibrarian release state` currently deems a library as "eligible for release" if there is _any_ commit since the last tag that touches _any_ of the libraries artifacts in _any_ capacity. This is to say that conventional commits are not consulted for determining if the commit contains a "meaningful" change.

https://github.com/googleapis/librarian/blob/9cb2da5a147ff207451d46a1ccebe2e1930b6ff5/internal/legacylibrarian/legacylibrarian/commit_version_analyzer.go#L172-L182

There is still the `release_exclude_paths` (e.g. those added in https://github.com/googleapis/google-cloud-python/pull/16908) to configure excluding files from the release eligibility evaluation.

## Problem

Recent Go and Python releases (fulfilled or otherwise) have resulted in a vast majority of libraries (if not all of them) being bumped and released ([Go example](https://github.com/googleapis/google-cloud-go/pull/14513), [Python example](https://github.com/googleapis/google-cloud-python/pull/16901)).

We'd like to be able to make make benign changes to libraries without triggering a release. For example, changing the License header year _en masse_ should not be bulk release trigger. We'd like to be able to use the conventional commit type `chore` to signify a no-op change that shouldn't trigger a release.

### Respect no-op, use minor

We could change the above code to respect when the `highestChange` is `semver.None` (indicating a no-op) and **not** bump the library, rather than always forcing `semver.Minor`. Update playbooks to classify standard (re)generation PRs as `feat` so that bulk regeneration following API sources updates trigger releases in libraries it touched.

### Change nothing

Keep doing what we are doing with the understanding that any change, even those we know to be benign/non-functional, will trigger a release in the client. We will often have large bulk releases, especially if we regularly update dependencies _en masse_.

### Restore conventional commit semver

Remove the release-only mode gate, effectively restoring the conventional commit semver change decision logic. This isn't really a goal at the moment, we really just want to make more pointed release eligibility decisions.

## Preferred solution

Respect no-op, use minor

noahdietz:
Restores previous `legacylibrarian` behavior that would allow `chore` commits (which map to a `semver.ChangeLevel` of `semver.None`) through as "no-op" changes. After this, a library with only `chore` commits will not be considered eligible for release.  More background/details in #5786.

Updates `librarianops generate` to produce a `feat` commit. Corresponding playbook updates will follow.

Fixes #5786


sofisl (dismissed):

jskeet:
Doesn't this mean that no librarian generation PRs will trigger a release? Or is the intention that Librarian PRs will now have a "feat" commit?

noahdietz:
> Or is the intention that Librarian PRs will now have a "feat" commit?

@jskeet, this! See the `Respect no-op, use minor` section of the issue: https://github.com/googleapis/librarian/issues/5786

jskeet:
> > Or is the intention that Librarian PRs will now have a "feat" commit?
> 
> @jskeet, this! See the `Respect no-op, use minor` section of the issue: #5786

Cool - sorry for missing that before :)

noahdietz:
> > > Or is the intention that Librarian PRs will now have a "feat" commit?
> > 
> > 
> > @jskeet, this! See the `Respect no-op, use minor` section of the issue: #5786
> 
> Cool - sorry for missing that before :)

But now that you got me thinking about it, I also updated `librarianops` to produce a `feat` commit :)

jskeet (approved):

PR #5813, opened 2026-05-04
feat(internal/librarian): support dot notation for source updates

### LINKED ISSUE #5075: librarian: change update command to use hierarchical dot notation

sofisl:
Support dot-separated paths for updating and configuring upstream source repositories in the configuration. The update and config commands now support targets using the sources.`<source>`.`<field>` path notation for conformance, discovery, googleapis, protobuf, and showcase sources. Setting a commit automatically resolves and updates its corresponding checksum atomically using a nested schema-enforced layout.

Follow up to #5705 (specifically https://github.com/googleapis/librarian/pull/5705#discussion_r3168857391):
> refactor this so that we just call setConfigValue to set the config value, so that that operation is consistent for every value that is set in librarian.yaml

Fixes https://github.com/googleapis/librarian/issues/5075


internal/librarian/cfg_value_utility.go
```diff
@@ -28,21 +30,150 @@ var (
...
+	}
+	return "", fmt.Errorf("%w: %s", errUnsupportedPath, path)
+}
+
+// getOrCreateSource returns a Source pointer from the Config, initializing if needed.
+func getOrCreateSource(cfg *config.Config, sourceName string) *config.Source {
+	if cfg.Sources == nil {
+		cfg.Sources = &config.Sources{}
+	}
+	switch sourceName {
```
zhumin8, line 114: This switch looks similar to the ones in getSource, maybe we can extract a helper `getSourcePointer` and simplify getOrCreateSource and getSource

internal/librarian/cfg_value_utility_test.go
```diff
@@ -103,24 +209,30 @@ func TestSetConfigValue(t *testing.T) {
...
+			path:  "sources.googleapis.commit",
+			value: "non-existent-branch",
 		},
 	} {
 		t.Run(test.name, func(t *testing.T) {
 			cfg := &config.Config{
 				Version: "v1.0.0",
 			}
-			_, err := setConfigValue(cfg, test.path, "some-value")
-			if !errors.Is(err, test.wantErr) {
```
zhumin8, line 122: Is this change intentional?

internal/librarian/cfg_value_utility_test.go
```diff
@@ -103,24 +209,30 @@ func TestSetConfigValue(t *testing.T) {
...
+		path  string
+		value string
 	}{
 		{
-			name:    "unsupported path",
-			path:    "unknown.field",
-			wantErr: errUnsupportedPath,
+			name:  "unsupported path",
+			path:  "unknown.field",
+			value: "val",
```
zhumin8: It's a bit unclear what `value: "val",` do, maybe change back to "some-value" or "some-value-not-used"?

internal/librarian/cfg_value_utility_test.go
sofisl: good question! it was because the refactor made it so that we were just surfacing generic network errors. but so then we lose checking for the exact sentinel error (errUnsupportedPath) so I readded it back in!

internal/librarian/cfg_value_utility_test.go
```diff
@@ -103,24 +209,36 @@ func TestSetConfigValue(t *testing.T) {
...
 			name:    "unsupported path",
 			path:    "unknown.field",
+			value:   "some-value-not-used",
 			wantErr: errUnsupportedPath,
 		},
+		{
+			name:    "failed fetch commit",
+			path:    "sources.googleapis.commit",
+			value:   "non-existent-branch",
+			wantErr: nil,
```
zhumin8: nil can be omitted

zhumin8 (approved):
nit: link in description does not bring to discussion
> Follow up to https://github.com/googleapis/librarian/pull/5705 (specifically https://github.com/googleapis/librarian/pull/5705#discussion_r3168857391)

PR #5814, opened 2026-05-05
refactor(internal/sidekick/gcloud): use text/template

julieqiu:
Switch the cli.go template engine from cbroglie/mustache to text/template to support if/else statements in the template. This is useful for upcoming changes and allows us to delete Command.HasFlags.

The rendered file is run through go/format so the golden output stays stable.


coryan (approved):
I am curious how this will evolve.

julieqiu:
> I am curious how this will evolve.

Tickets are being sold for a front row seat.

PR #5822, opened 2026-05-05
feat(internal/sidekick/gcloud): call the Go client in describe command

julieqiu:
A generated describe command used to print the resource path and return. It now creates the Go client for the API, fetches the resource, and prints the response. The generator picks the Go client package and request type from the proto package name.

### DIFF
--- a/internal/sidekick/gcloud/generate.go
+++ b/internal/sidekick/gcloud/generate.go
@@ -38,7 +38,24 @@ var templates embed.FS
 
 // CLIModel represents the data structure for the template.
 type CLIModel struct {
+	// Imports holds the Go imports rendered into the generated main.go.
+	Imports []Import
+
+	// Groups holds the top-level gcloud command groups rendered into main.go.
 	Groups []Group
+}
+
+// Import represents a Go import line in the generated main.go.
+type Import struct {
+	// Alias is the optional package alias; empty when the import has no
+	// alias. For example, "parallelstore" in:
+	//
+	//	parallelstore "cloud.google.com/go/parallelstore/apiv1"
+	Alias string
+
+	// Path is the import path of the package, for example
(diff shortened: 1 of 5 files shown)


internal/sidekick/gcloud/generate.go
```diff
@@ -155,9 +189,98 @@ func constructCLIModel(model *api.API) CLIModel {
...
+		Method:      method.Name,
+		NameField:   "Name",
+		Package:     goClient.Alias,
+		RequestType: goClient.Alias + "pb." + method.InputType.Name,
+	}
+}
+
+// goClientInfo describes the Go client and proto-Go packages for a proto
+// package like "google.cloud.parallelstore.v1".
+type goClientInfo struct {
```
coryan, line 220: Sounds like it is time to split `generate.go`.  I know it is mostly a prototype, but it is growing long in the tooth and mixing code with different purpose.

internal/sidekick/gcloud/testdata/parallelstore/main.go.golden
```diff
@@ -44,7 +47,20 @@ func main() {
...
+									client, err := parallelstore.NewClient(ctx)
+									if err != nil {
+										return err
+									}
+									defer client.Close()
+									resp, err := client.GetInstance(ctx, &parallelstorepb.GetInstanceRequest{Name: name})
+									if err != nil {
+										return err
+									}
+									out, err := protojson.MarshalOptions{Multiline: true, Indent: "  "}.Marshal(resp)
```
coryan, line 59: Ah, clever, so this is the same JSON dialect that the service uses.

coryan (approved):

internal/sidekick/gcloud/generate.go
julieqiu: will do in a follow up

PR #5823, opened 2026-05-05
refactor(sidekick): merge state into API

ianthehat:
It is now a pure implememtation detail of methods on API, with no external references, there is no point it being a unique struct.


coryan (approved):
Nice.  Should we sign "the witch is dead" or something?

PR #5824, opened 2026-05-05
feat(sidekick/swift): generate snippets for all methods

coryan:
Generate at least the skeleton of a snippet for each RPC. They really need the `parent` or `name` field initialized too. Having the skeleton will help us gauge how the code grows as we add more services and and method overloads.

Towards #5243


internal/sidekick/swift/generate.go
```diff
@@ -114,12 +115,24 @@ func (c *codec) generateServices(outdir string, model *api.API, provider languag
...
 		generated := language.GeneratedFile{
-			TemplatePath: "templates/common/snippet.swift.mustache",
+			TemplatePath: "templates/common/client_snippet.swift.mustache",
 			OutputPath:   filepath.Join("Snippets", s.Name+"Quickstart.swift"),
 		}
 		if err := language.GenerateService(outdir, s, provider, generated); err != nil {
 			return err
 		}
+		for _, m := range s.Methods {
+			if !isGeneratedMethod(m) {
```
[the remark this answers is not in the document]
coryan: And yet the code compiles without problems.... LLMs can be as confident when they are right as when they are wrong.  In that way, they resemble this human.

internal/sidekick/language/generate.go
```diff
@@ -57,6 +57,11 @@ func GenerateService(outDir string, service *api.Service, provider TemplateProvi
 	return generateElement(outDir, service, provider, gen)
 }
 
+// GenerateMethod generates a single file using the api.Method model.
+func GenerateMethod(outDir string, method *api.Method, provider TemplateProvider, gen GeneratedFile) error {
+	return generateElement(outDir, method, provider, gen)
```
julieqiu, line 62: it seems like you could just put the code for generateElement here

julieqiu (approved):

internal/sidekick/language/generate.go
coryan: Wouldn't that dup the code? I also have `GenerateMessage()` and `GenerateEnum()` and I would prefer to not use `any` in a public API.

PR #5825, opened 2026-05-05
feat(internal/librarian/golang): run go mod tidy after generating the Go library module

### LINKED ISSUE #5519: golang: go mod tidy should be run post-generation by librarian
After doing a standard `update` & `generate` for `google-cloud-go`, my PR failed a `vet` check in one GAPIC module ([check](https://github.com/googleapis/google-cloud-go/actions/runs/24746094862/job/72398010386)) due to a missing direct dependency in the `go.mod` of a single submodule that was updated in the PR.

The API that changed and was regenerated added a new direct dependency to their API surface - a common type from google/rpc/status.proto.

This was fixed by doing `cd {module} && go mod tidy` (using the target version of Go which was Go 1.25.0 as of writing this).

It would be great if `librarian` did this as a post generate step in the client module output dir after regenerating each library.

I think this depends on #5201 being fixed as well, where the target version of the Go tool used by `google-cloud-go` can be executed instead of the version of the Go tool used to execute librarian.

sofisl:
This change modifies Librarian to run 'go mod tidy' in the output directory after generating a Go client library to ensure the newly generated Go client libraries are left in a tidied state.

Fixes https://github.com/googleapis/librarian/issues/5519


internal/librarian/golang/module.go
```diff
@@ -260,3 +261,49 @@ func findSnippetDirectory(library *config.Library, goAPI *config.GoAPI, output s
...
+	if _, err := os.Stat(filepath.Join(outDir, "go.mod")); err == nil {
+		if err := command.RunInDir(ctx, outDir, goCmd, "mod", "tidy"); err != nil {
+			return fmt.Errorf("failed to tidy module %s: %w", outDir, err)
+		}
+	}
+	return nil
+}
+
+// TidySnippets tidies all unique snippets modules that exist in the given libraries.
+func TidySnippets(ctx context.Context, libraries []*config.Library, goCmd string) error {
```
sofisl: we're separating `Tidy Module` and `TidySnippets` out because of a possible race condition, see https://github.com/googleapis/librarian/pull/5825#discussion_r3190474098

internal/librarian/generate_test.go
```diff
@@ -391,6 +392,160 @@ libraries:
 	}
 }
 
+func TestGenerate_Go(t *testing.T) {
```
sofisl: Adding this test or else our coverage dips. In general, seems like a good test to make sure we're calling the things we expect to call during execution.

internal/librarian/generate_test.go
```diff
@@ -391,6 +391,95 @@ libraries:
 	}
 }
 
+func setupGoMockEnvironment(t *testing.T, tempDir string) {
```
sofisl: I will be the first to admit this is a really ugly function setup. Here's where we're at:
- we don't really have an E2E test for the whole generate function. adding the functions in this PR spills us over into <79% test coverage, so this PR forces us adding this E2E test
- it would be much cleaner/simpler if we did something like 
```
type Runner interface {
    Run(ctx context.Context, dir string, cmd string, args ...string) error
}
```
so that we could mock the CLI better. But I think this would require a big refactor...which I'm open to doing if people agree!

internal/librarian/golang/module.go
noahdietz: We actually do not want to tidy the `internal/generated/snippets/go.mod`. It's kind of a brittle thing filled with local replacement directives. If we tidy it, it will change a bunch of stuff (you can see for yourself in by trying to do so in google-cloud-go). Let's remove this for now

internal/librarian/generate.go
```diff
@@ -239,12 +239,18 @@ func generateLibraries(ctx context.Context, cfg *config.Config, libraries []*con
 				if err := golang.Generate(gctx, library, src, goCmd); err != nil {
 					return fmt.Errorf("generate library %q (%s): %w", library.Name, cfg.Language, err)
 				}
+				if err := golang.TidyModule(gctx, library, goCmd); err != nil {
```
noahdietz: I think this should be done inside of `golang.Generate` e.g. right **after the block** where we would initialize a new `go.mod`: https://github.com/googleapis/librarian/blob/bac78033c5eecc58bfa3ee65e288e5750a2164d8/internal/librarian/golang/generate.go#L113-L121

This would at least get us out of changing the `generate_test.go`

internal/librarian/golang/module.go
```diff
@@ -260,3 +261,49 @@ func findSnippetDirectory(library *config.Library, goAPI *config.GoAPI, output s
...
 	return snippetDir
 }
+
+// TidyModule tidies the Go module by running 'go mod tidy' in the output directory if it exists.
+func TidyModule(ctx context.Context, library *config.Library, goCmd string) error {
+	outDir, err := filepath.Abs(library.Output)
+	if err != nil {
+		return fmt.Errorf("failed to get absolute path of output directory: %w", err)
+	}
+	if _, err := os.Stat(filepath.Join(outDir, "go.mod")); err == nil {
```
noahdietz: If we move this to after this: https://github.com/googleapis/librarian/blob/bac78033c5eecc58bfa3ee65e288e5750a2164d8/internal/librarian/golang/generate.go#L113-L121

we will be able to drop this conditional

sofisl:
@noahdietz thank you for the pointers and the helpful comments :)

PR #5830, opened 2026-05-05
refactor(internal/sidekick/surfer): remove group builder object oriented boiler plate

sarahheacock:
Remove the object-oriented group builder and pass around stateless group context. This keeps the codebase consistent withe surface_builder and eliminates confusion about internal state.

### DIFF
--- a/internal/sidekick/surfer/group_builder.go
+++ b/internal/sidekick/surfer/group_builder.go
@@ -23,24 +23,17 @@ import (
 	"github.com/iancoleman/strcase"
 )
 
-type groupBuilder struct {
-	model   *api.API
-	config  *provider.Config
-	service *api.Service
+// GroupContext contains the context required to build a command group.
+type GroupContext struct {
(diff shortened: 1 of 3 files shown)


internal/sidekick/surfer/group_builder.go
```diff
@@ -23,24 +23,17 @@ import (
 	"github.com/iancoleman/strcase"
 )
 
-type groupBuilder struct {
+// groupContext contains the context required to build a command group.
+type groupContext struct {
```
jameslynnwu: We might want to reserve *context for things related to golang context.

jameslynnwu (approved):

internal/sidekick/surfer/group_builder.go
sarahheacock: Changed to `Params` after getting feedback from team. While context is not reserved, it is confusing / unidiomatic

PR #5836, opened 2026-05-05
fix(internal/librarian/java): remove keep files from staging dir before running owlbot.py

### LINKED ISSUE #5834: java: some files in keep are still overridden by generated code
Observed in local testing, seems that some files in keep are still overriden by generated code, for example:
`stub/Version.java` files, these are newly added by the generator but intended to be generated only once (then managed by release please and get updated versions). 

google-cloud-java/java-bigquerydatapolicy/google-cloud-bigquerydatapolicy/src/main/java/com/google/cloud/bigquery/datapolicies/v1beta1/stub/Version.java

zhumin8:
Remove keep files from staging dir in post processing before running owlbot.py. 
This is needed because although keep files are not removed in clean phase, when running owlbot.py, synthtool move would blindly override if exists. In the case that a file is generated but in keep list, it fails to keep and the generated one gets in final output.

Fix #5834


internal/librarian/java/postprocess_test.go
```diff
@@ -890,3 +890,54 @@ func TestCopyFiles_Error(t *testing.T) {
...
+		Keep: []string{
+			"google-cloud-lib/src/main/java/com/google/kept/File.java",
+			"google-cloud-lib/src/main/java/com/google/keptdir/", // Test with trailing slash
+		},
+	}
+	if err := removeKeptFilesFromStaging(library, outDir); err != nil {
+		t.Fatalf("removeKeptFilesFromStaging failed: %v", err)
+	}
+
+	if _, err := os.Stat(keptFile); !os.IsNotExist(err) {
```
sofisl:
```suggestion
    if _, err := os.Stat(keptFile; errors.Is(err, fs.ErrNotExist) {
```

same for all of the os.IsNotExists below

internal/librarian/java/postprocess_test.go
```diff
@@ -890,3 +890,54 @@ func TestCopyFiles_Error(t *testing.T) {
...
+	// 1. Explicitly kept file
+	keptFile := filepath.Join(stagingDir, "v1", "google-cloud-lib", "src", "main", "java", "com", "google", "kept", "File.java")
+	// 2. File matching versionRegexp (not explicitly in Keep)
+	versionFile := filepath.Join(stagingDir, "v1", "google-cloud-lib", "src", "main", "java", "com", "google", "cloud", "lib", "v1", "stub", "Version.java")
+	// 3. Regular file (should be preserved in staging)
+	regularFile := filepath.Join(stagingDir, "v1", "google-cloud-lib", "src", "main", "java", "com", "google", "cloud", "lib", "v1", "stub", "Regular.java")
+	// 4. File inside an explicitly kept directory
+	keptDirFile := filepath.Join(stagingDir, "v1", "google-cloud-lib", "src", "main", "java", "com", "google", "keptdir", "SubDir", "File.java")
+
+	for _, f := range []string{keptFile, versionFile, regularFile, keptDirFile} {
```
sofisl: use full names (file)

internal/librarian/java/postprocess.go
```diff
@@ -404,3 +412,54 @@ func copyProtos(protoSourceDir string, protos []string, destDir string) error {
...
+// It strips this first component (the API base like "v1") from the relative
+// path to reconstruct the expected path relative to the library root, which is
+// then matched against the library's Keep configuration.
+func removeKeptFilesFromStaging(library *config.Library, outDir string) error {
+	stagingDir := stagingDir(outDir)
+	if _, err := os.Stat(stagingDir); os.IsNotExist(err) {
+		return nil
+	}
+	keepSet := make(map[string]bool)
+	for _, k := range library.Keep {
```
sofisl: use keep

sofisl (commented):
Jetski helped me notice that there might be a windows path separator bug here. While we convert file paths to forward slashes (relSlash) before checking them, the directory checks in clean.go on lines 87 and 117 (keepSet[rel]) still use the raw path (rel).

On Windows, rel will have backslashes (like a\b\c), but the configuration keys are stored with forward slashes (a/b/c). This mismatch means kept directories won't be skipped or preserved on Windows. We should normalize keepSet keys and use relSlash for all lookups.

sofisl (approved):

PR #5842, opened 2026-05-05
feat(internal/librarian/golang): support GOTOOLCHAIN

### LINKED ISSUE #5519: golang: go mod tidy should be run post-generation by librarian

julieqiu:
Support for specifying a Go toolchain version is added to librarian.yaml under default.go.toolchain. This allows specifying a specific Go version when initializing module by leveraging the GOTOOLCHAIN environment variable (https://go.dev/blog/toolchain), as opposed to downloading a separate binary via golang.org/dl/goVERSION. 

This ensures that the generated go.mod has the requested version and that the correct toolchain is used during generation.

For https://github.com/googleapis/librarian/issues/5201
For https://github.com/googleapis/librarian/issues/5246
Fixes https://github.com/googleapis/librarian/issues/5519

### DIFF
--- a/internal/librarian/generate.go
+++ b/internal/librarian/generate.go
@@ -232,11 +232,14 @@ func generateLibraries(ctx context.Context, cfg *config.Config, libraries []*con
 		}
 		return g.Wait()
 	case config.LanguageGo:
-		goCmd := golang.GoCommand(cfg.Tools)
+		var toolchain string
+		if cfg.Default != nil && cfg.Default.Go != nil {
+			toolchain = cfg.Default.Go.Toolchain
+		}
 		g, gctx := errgroup.WithContext(ctx)
 		for _, library := range libraries {
 			g.Go(func() error {
-				if err := golang.Generate(gctx, library, src, goCmd); err != nil {
+				if err := golang.Generate(gctx, library, src, toolchain); err != nil {
 					return fmt.Errorf("generate library %q (%s): %w", library.Name, cfg.Language, err)
 				}
 				return nil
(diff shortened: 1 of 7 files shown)


internal/librarian/generate.go
```diff
@@ -232,11 +232,14 @@ func generateLibraries(ctx context.Context, cfg *config.Config, libraries []*con
...
-		goCmd := golang.GoCommand(cfg.Tools)
+		var toolchain string
+		if cfg.Default != nil && cfg.Default.Go != nil {
+			toolchain = cfg.Default.Go.Toolchain
+		}
 		g, gctx := errgroup.WithContext(ctx)
 		for _, library := range libraries {
 			g.Go(func() error {
-				if err := golang.Generate(gctx, library, src, goCmd); err != nil {
+				if err := golang.Generate(gctx, library, src, toolchain); err != nil {
```
noahdietz, line 242: (missed this in previous review) Do we need this here or can we keep the `Generate` signature cleaner, and just load the toolchain there?

noahdietz:
Also addresses a need for #5246, having a `GoDefault` config section

internal/librarian/generate.go
julieqiu: I think that would still require passing in cfg.Default.Go, so we’d still end up needing the extra argument. That said, I do agree we should optimize for keeping the interface consistent across languages. Let me file a follow-up to rethink the shape of this a bit more holistically.

julieqiu: https://github.com/googleapis/librarian/issues/5847

noahdietz (approved):

PR #5843, opened 2026-05-05
feat(internal/sidekick/surfer): operations wait command

### LINKED ISSUE #5485: surfer: fix operation command output

sarahheacock:
Build synthetic wait command from getOperation method. Enable method_operation golden

Fix #5485


internal/sidekick/surfer/command_builder.go
```diff
@@ -53,6 +53,42 @@ func buildCommand(method *api.Method, overrides *provider.Config, model *api.API
...
+	var waitArgs []Argument
+	if arg != nil {
+		arg.HelpText = "The name of the operation resource to wait on."
+		waitArgs = []Argument{*arg}
+	}
+
+	return &Command{
+		Name:   "wait",
+		Hidden: hidden(overrides),
+		HelpText: HelpText{
```
jameslynnwu, line 78: is this the help text from gen_sfc? it seems like its missing words

internal/sidekick/surfer/command_builder.go
jameslynnwu, line 273: outside of scope of this PR. classified vs categorized. use consistent words

https://github.com/googleapis/librarian/issues/5852

internal/sidekick/surfer/command_builder.go
```diff
@@ -53,6 +53,42 @@ func buildCommand(method *api.Method, overrides *provider.Config, model *api.API
...
+		return nil, err
+	}
+
+	apiVersion, err := provider.APIVersionFromMethod(getMethod)
+	if err != nil {
+		return nil, err
+	}
+
+	var waitArgs []Argument
+	if arg != nil {
```
jameslynnwu: is it an error if the arg is nil?

jameslynnwu (changes requested):

internal/sidekick/surfer/command_builder.go
sarahheacock: Ack

internal/sidekick/surfer/command_builder.go
sarahheacock: Agreed unfortunately. It is what is matching gen_sfc (the target golden is enabled in this PR). We can and should override in a follow up but would require adjusting the targets

internal/sidekick/surfer/command_builder.go
sarahheacock: Yes, good point. We should return an error here and skip generating this command in surface_builder. We shouldn't output a faulty command. Updated

jameslynnwu (approved):

PR #5849, opened 2026-05-05
feature(internal/sidekick/surfer): implement locations mixin commands

### LINKED ISSUE #5486: surfer: validate location command output
Validate location commands are correct. If not, make appropriate updated and fix golden expected output

sarahheacock:
Implement locations mixins commands for surfer
- Generate synthetic location resource from GetLocation rpc to get collection pattern
- Generation synthetic location proto documentation for help text

Fix #5486


internal/sidekick/surfer/provider/resource.go
```diff
@@ -341,14 +347,15 @@ func GetLiteralSegments(raw []api.PathSegment) []string {
...
 }
 
-// getAllResources returns all resource definitions in the model, including
-// file-level definitions, message-level definitions, and synthetic resources
-// inferred from operations methods.
+// getAllResources returns all resource definitions in the model.
+// This acts as the single source of truth for both path-based and method-based
+// resource lookups, ensuring that synthetic mixin resources (operations, locations)
+// resolve to the exact same definitions across all entry points.
 func getAllResources(model *api.API) []*api.Resource {
```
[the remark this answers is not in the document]
sarahheacock: Known limitation and will be addressed in sidekick

internal/sidekick/surfer/provider/resource.go
```diff
@@ -341,36 +348,56 @@ func GetLiteralSegments(raw []api.PathSegment) []string {
...
+					}
 				}
 			}
 		}
 	}
+}
 
-	return resources
+// mixinSyntheticResource attempts to infer a synthetic resource definition from a mixin method.
+func mixinSyntheticResource(m *api.Method) *api.Resource {
```
jameslynnwu: mixinSyntheticResource should have similar naming to inferOperationResource

internal/sidekick/surfer/provider/mixin.go
```diff
@@ -54,8 +64,54 @@ func OperationMethodDocumentation(methodName string) string {
...
+	}
+	return ""
+}
+
 // inferOperationResource creates a synthetic resource for operations based on the method's path.
 func inferOperationResource(m *api.Method) (*api.Resource, error) {
+	// For each method pattern like /v1/{name=projects/*/locations/*/operations/*}, create a corresponding
+	// resource pattern like /projects/{project}/locations/{location}/operations/{operation}
+	patterns, err := resourcePatterns(m)
+	if err != nil || len(patterns) == 0 {
```
jameslynnwu: should len(patterns) == 0 be considered an error or not? If its an error, can we move that check into resourcePatterns?

internal/sidekick/surfer/provider/mixin.go
```diff
@@ -54,8 +64,54 @@ func OperationMethodDocumentation(methodName string) string {
...
+		return "The name of the location resource."
+	case ListLocations:
+		return "The name of the location's parent resource."
+	}
+	return ""
+}
+
 // inferOperationResource creates a synthetic resource for operations based on the method's path.
 func inferOperationResource(m *api.Method) (*api.Resource, error) {
+	// For each method pattern like /v1/{name=projects/*/locations/*/operations/*}, create a corresponding
```
jameslynnwu: this comment can be moved to resourcePatterns?

jameslynnwu (changes requested):

internal/sidekick/surfer/provider/mixin.go
sarahheacock: Idk to be honest. If a method does not have bindings, we have bigger problems in general. I think your question is important and we need to decide. But I think we might need to punt it

internal/sidekick/surfer/provider/mixin.go
sarahheacock: Updated

internal/sidekick/surfer/provider/mixin.go
jameslynnwu: If that's the case, we should get rid of the following checks and return an error. (or should we let a nil pointer error happen). Why do you think its something we need to punt on?

In resourcePatterns:
```
	if m.PathInfo == nil || len(m.PathInfo.Bindings) == 0 {
		return nil, nil
	}
```

and expandBinding:
```
if binding == nil || binding.PathTemplate == nil {
		return nil, nil
```

And we should handle case when we're not able to extractBindings or return any patterns.

sarahheacock: Talked offline

I want to avoid littering the deep helper methods with defensive nil checks and unhelpful error returns that we should never realistically reach. Ideally, these would the scenarios:
* **Internal Invariant Protections (Tier 1):** If we hit an structurally impossible state in our own AST representation, we should return a hard error to catch developer/generator bugs.
* **Source Compliance Validation (Tier 2):** Validation of configs/protos should be centralized at the boundary/loading phase. Deep generator logic should assume compliant ASTs.
* **Feature Gaps (Tier 3):** If an input is valid but lacks optional metadata (like a method without HTTP bindings), we should gracefully log some sort of warning.

However, I do not know how to handle tier 2. Ideally, it should be handled in sidekick at the source. Methods without http bindings should be expected and are tier 3 for gcloud at least.

James determined that PathInfo should not be nil but could be empty by looking at source code. We agreed to remove the defensive programming in the short term but we need to think about a sustainable pattern long-term. One that does not require reverse engineering sidekick.

jameslynnwu: discussed offline

jameslynnwu (approved):

PR #5851, opened 2026-05-06
refactor(surfer): configure surfer overrides per API and abstract config converter

jameslynnwu:
This change migrates gcloud surfer overrides in librarian.yaml to be configured per-API rather than globally, allowing fine-grained help text control in libraries with multiple APIs.

Specifically:
- Updates config.Surfer schema in language.go to support a SurferAPIs slice and moves GcloudHelpTextRules entirely into SurferAPI (removing global HelpText).
- Implements provider.FromLibrarianConfig inside the sidekick/surfer/provider package to abstract the conversion and mapping of Librarian configs into provider.Config, keeping the provider decoupled from Librarian.
- Refactors Librarian's generate.go to delegate configuration generation directly to provider.FromLibrarianConfig, simplifying generation logic.
- Adds targeted unit test TestFromLibrarianConfig to config_test.go to verify the conversion and override mapping logic.

For https://github.com/googleapis/librarian/issues/5511

### DIFF
--- a/internal/sidekick/surfer/provider/config.go
+++ b/internal/sidekick/surfer/provider/config.go
@@ -322,12 +322,12 @@ func mapHelpText(ht *libconfig.GcloudHelpTextRules) *HelpTextRules {
 		return nil
 	}
 	return &HelpTextRules{
-		MethodRules: mapRules(ht.MethodRules),
-		FieldRules:  mapRules(ht.FieldRules),
+		MethodRules: mapHelpTextRules(ht.MethodRules),
+		FieldRules:  mapHelpTextRules(ht.FieldRules),
 	}
 }
 
-func mapRules(src []*libconfig.GcloudHelpTextRule) []*HelpTextRule {
+func mapHelpTextRules(src []*libconfig.GcloudHelpTextRule) []*HelpTextRule {
 	var dst []*HelpTextRule
 	for _, r := range src {
 		dst = append(dst, &HelpTextRule{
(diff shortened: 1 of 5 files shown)


internal/sidekick/surfer/provider/config.go
```diff
@@ -280,3 +284,60 @@ func ShouldGenerateOperations(c *Config) bool {
...
+
+	return &Config{
+		APIs: []API{{
+			Name:     modelName,
+			HelpText: helpText,
+		}},
+	}
+}
+
+func findSurferAPI(library *libconfig.Library, api *libconfig.API) *libconfig.SurferAPI {
```
sarahheacock: I believe this should be findSurferAPIs / return list? Parallelstore only has one api (service proto). But other apis like iam have multiple https://github.com/googleapis/googleapis/tree/master/google/iam/v3.

sarahheacock: Nit: the provider package has poor naming already (will add a markdown file correcting it later). In go, you want to typically drop verbs like `Get` and `Find`. Finder functions use `For` so this would be `func surferAPIFor(library *libconfig.Library, api *libconfig.API) *libconfig.SurferAPI`

jameslynnwu: I believe there should only be one libconfig.SurferAPI (surfer overrides for an API) per API. iam, even if there are many libconfig.APIs for it, should have a single libconfig.SurferAPI for it.

wdyt?

sarahheacock: Chatted offline. The api names do not map neatly over. So we're just porting over the gcloud config and fixing over time

PR #5853, opened 2026-05-06
refactor(internal/sidekick/gcloud): generate one multi-surface gcloud binary

julieqiu:
Previously, the Generate produced a separate main.go for each API. Now it takes the full set of APIs, and creates one Go package per surface under internal/generated and a single top-level main under cmd/gcloud that pulls each surface in as a subcommand.

Add publicca as a second surface in the tests so the multi-surface path is covered. Add a -update flag for refreshing the golden files.

Delete README generation since that isn't necessary for gcloud.


coryan (approved):
Is the idea that you will have a single `gcloud` binary with all the subcommands linked in? Fine to start, but I think you should consider a structure where `gcloud foo` just calls `gcloud-foo` and `gcloud-foo` is a separate binary that must meet some requirements.

That way you can later switch some binaries with Rust, or Python, or a hand-crafted thing, or just keep them small so you don't have to install one giant honking binary just to access one secret in secret manager.

julieqiu:
> Is the idea that you will have a single `gcloud` binary with all the subcommands linked in? Fine to start, but I think you should consider a structure where `gcloud foo` just calls `gcloud-foo` and `gcloud-foo` is a separate binary that must meet some requirements.
> 
> That way you can later switch some binaries with Rust, or Python, or a hand-crafted thing, or just keep them small so you don't have to install one giant honking binary just to access one secret in secret manager.

Yes exactly, I wanted to build out the single binary prototype first before doing something more complicated.

PR #5854, opened 2026-05-06
feat(internal/librarian/gcloud): derive API path and output

julieqiu:
librarian add, generate and tidy now supports deriving the API path and output based on the name.

### DIFF
--- a/internal/librarian/gcloud/default.go
+++ b/internal/librarian/gcloud/default.go
@@ -22,6 +22,10 @@ import (
 // DeriveAPIPath returns the canonical API path for a gcloud library name. For
 // example: accessapproval -> google/cloud/accessapproval/v1.
 func DeriveAPIPath(name string) string {
+	// TODO(https://github.com/googleapis/librarian/issues/5862): use the
+	// actual API version from the service configuration or the library's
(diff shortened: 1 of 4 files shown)


internal/librarian/gcloud/default.go
```diff
@@ -0,0 +1,55 @@
...
+
+import (
+	"path/filepath"
+	"strings"
+)
+
+// DeriveAPIPath returns the canonical API path for a gcloud library name. For
+// example: accessapproval -> google/cloud/accessapproval/v1.
+func DeriveAPIPath(name string) string {
+	return "google/cloud/" + name + "/v1"
```
coryan, line 29: That "v1" cannot be right in all cases.  We have services that have more than one version (and all are needed) and services that only start at v2 or v4.  If this is just a placeholder, can you add a TODO and a tracking bug?

coryan (approved):

internal/librarian/gcloud/default.go
julieqiu: I'm planning to override this in librarian.yaml for now with apis.path. I expect we will do something smarter later on. Will leave a TODO.

PR #5856, opened 2026-05-06
feat(internal/config): add gcloud.client_path override for Library

julieqiu:
The gcloud generator derives the GAPIC Go client import path from the proto package using the format cloud.google.com/go/<short>/api<version>. Some surfaces do not follow that shape, so add an option for an override.

### DIFF
--- a/doc/config-schema.md
+++ b/doc/config-schema.md
@@ -125,6 +125,7 @@ This document describes the schema for the librarian.yaml.
 | `specification_format` | string | Specifies the API specification format. Valid values are "protobuf" (default) or "discovery". |
 | `dart` | [DartPackage](#dartpackage-configuration) (optional) | Contains Dart-specific library configuration. |
 | `dotnet` | [DotnetPackage](#dotnetpackage-configuration) (optional) | Contains .NET-specific library configuration. |
+| `gcloud` | [GcloudCommand](#gcloudcommand-configuration) (optional) | Contains gcloud-specific library configuration. |
 | `go` | [GoModule](#gomodule-configuration) (optional) | Contains Go-specific library configuration. |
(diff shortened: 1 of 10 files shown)


doc/config-schema.md
```diff
@@ -230,6 +231,12 @@ This document describes the schema for the librarian.yaml.
 | `to` | string |  |
 | `wire_name` | string |  |
 
+## GcloudCommand Configuration
+
+| Field | Type | Description |
+| :--- | :--- | :--- |
+| `client_import_path` | string | Overrides the GAPIC Go client import path that would otherwise be derived from the proto package. Set this when the proto package and the published Go GAPIC location diverge. For example, the proto package google.cloud.recaptchaenterprise.v1 publishes its Go client at "cloud.google.com/go/recaptchaenterprise/v2/apiv1". |
```
coryan, line 238: "v2/apiv1" ... Oh the horrors...

internal/sidekick/gcloud/client.go
```diff
@@ -60,6 +64,36 @@ func goClientPackage(protoPkg string) *goClientInfo {
 	}
 }
 
+// clientInfoFromPath derives a goClientInfo from an explicit GAPIC Go client
+// import path. It returns nil when the path's final segment is not an
+// "api<version>" segment or when no usable alias can be found.
```
coryan, line 69: Should those be errors so the caller can tell what the problem was?

coryan (approved):

internal/sidekick/gcloud/client.go
julieqiu: Filed an issue and will do in a follow up since the refactoring is less trivial than I was expecting.

PR #5860, opened 2026-05-06
feat(internal/librarian/python): parallelize Python generation

### LINKED ISSUE #3728: python: ensure generation works in parallel
Currently Python generation creates an owl-bot-staging directory, generates into that, moves files around, and then cleans it up. I'm pretty sure that's going to fail badly when we try to run it in parallel.

Ideally we'd eliminate a lot of the staging aspect entirely, but if that's too hard to do right away, we should stage into a uniquely-named temporary directory. (Even that may be fiddly - we may need to generate into tmpxyz/owl-bot-staging...)

jskeet:
Generation of each package is now performed within a package-specific tmp directory, with a symlink to fulfil synthtool expectations. The code calling python.Generate within internal/librarian is now parallelized.

See go/sdk:librarian-python-parallel-generation for more details around the design of the tmp directory.

Fixes #3728


internal/librarian/python/generate.go
```diff
@@ -84,7 +84,16 @@ func Generate(ctx context.Context, cfg *config.Config, library *config.Library,
...
+	// The "generation root" is a tmp directory created within the package
+	// directory, to isolate it from other generation operations which may
+	// happen in parallel. It is deleted by cleanUpFilesAfterPostProcessing.
+	generationRoot := ""
+	if len(library.APIs) > 0 {
+		generationRoot, err = prepareGenerationRoot(outdir)
+		if err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
jskeet: When generation fails, it can actually be useful to still have the results of the failed generation - it's easier to figure out what went wrong.
(This is also consistent with the current code, which doesn't remove owl-bot-staging on error.)

internal/librarian/python/generate.go
```diff
@@ -134,6 +143,26 @@ func Generate(ctx context.Context, cfg *config.Config, library *config.Library,
...
+// concerned, such that packages/{xyz}/tmp/packages/{xyz} is a symlink back to
+// packages/{xyz}, and we generate into packages/{xyz}/tmp/owl-bot-staging.
+// This allows the post-processor to operate on packages/{xyz}/tmp (which in
+// turn allows us to run everything in parallel) without changing the
+// post-processor itself.
+// See go/sdk:librarian-python-parallel-generation for more details.
+func prepareGenerationRoot(packageRoot string) (string, error) {
+	packageName := filepath.Base(packageRoot)
+	generationRoot := filepath.Join(packageRoot, "tmp")
+	if err := os.MkdirAll(filepath.Join(generationRoot, "packages"), 0755); err != nil {
```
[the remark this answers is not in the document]
jskeet: Again, the approach is consistent with the current code, which doesn't attempt to clean up previous failures.

internal/librarian/python/generate.go
```diff
@@ -84,7 +84,16 @@ func Generate(ctx context.Context, cfg *config.Config, library *config.Library,
 	// Some aspects of generation currently require the repo root. Compute it
 	// once here and pass it down.
 	repoRoot := filepath.Dir(filepath.Dir(outdir))
-
+	// The "generation root" is a tmp directory created within the package
+	// directory, to isolate it from other generation operations which may
+	// happen in parallel. It is deleted by cleanUpFilesAfterPostProcessing.
+	generationRoot := ""
```
noahdietz:
```suggestion
	var generationRoot string
```
initializes to empty string

internal/librarian/python/generate.go
```diff
@@ -134,6 +143,26 @@ func Generate(ctx context.Context, cfg *config.Config, library *config.Library,
...
+// turn allows us to run everything in parallel) without changing the
+// post-processor itself.
+// See go/sdk:librarian-python-parallel-generation for more details.
+func prepareGenerationRoot(packageRoot string) (string, error) {
+	packageName := filepath.Base(packageRoot)
+	generationRoot := filepath.Join(packageRoot, "tmp")
+	if err := os.MkdirAll(filepath.Join(generationRoot, "packages"), 0755); err != nil {
+		return "", err
+	}
+	if err := os.Symlink("../..", filepath.Join(generationRoot, "packages", packageName)); err != nil {
```
noahdietz, line 160: Does the `"packages"` part matter WRT previews being under `preview-packages`? I'm guessing not bc it just symlinks back to the actual which is `preview-packages/`.

noahdietz (approved):

internal/librarian/python/generate.go
jskeet: Will do that in the morning.

internal/librarian/python/generate.go
jskeet: It's this that actually means post-processing will work for this when it wouldn't now - because synthtool expects a directory called packages in some places. So it'll modify the right directory, despite "aiming at" the wrong directory :)

noahdietz: beautiful!!!

Issue #5072, opened 2026-04-04
sidekick: generate cross-reference links in documentation comments


coryan:
The documentation comments include things like:

```swift
/// Creates a new
/// [ExternalAccountKey][google.cloud.security.publicca.v1.ExternalAccountKey] in
/// a given project.
```

The generator should enrich the comments with reference-style links:

```swift
/// Creates a new
/// [ExternalAccountKey][google.cloud.security.publicca.v1.ExternalAccountKey] in
/// a given project.
///
/// [google.cloud.security.publicca.v1.ExternalAccountKey]: `ExternalAccountKey`
```

*OR* (if this is easier), convert the cross-reference links in place:

```swift
/// Creates a new `ExternalAccountKey` in a given project.
```

The first approach is used by Rust, the second approach is used by Dart.

The first approach is known to work with external types.  Maybe the second one does too?

PR #5863, opened 2026-05-06
refactor(sidekick): common `ExtractCrossReferenceLinks`

coryan:
Swift is going to need the cross-reference links found in the service specification comments (see aip.dev/192), and then will use these cross reference links to create proper Swifty links to the right things.

This refactors some of the code to a common place.

Towards #5072

### DIFF
--- a/internal/sidekick/language/extract_xref_links.go
+++ b/internal/sidekick/language/extract_xref_links.go
@@ -1,0 +1,86 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package language
+
+import (
+	"bytes"
+	"regexp"
+	"sort"
+
+	"github.com/yuin/goldmark/ast"
+)
+
+// ExtractCrossReferenceLinks returns the cross-reference links found in `doc` and `source`.
+//
+// Google Cloud documentation comments include cross-reference links:
+//
+// https://google.aip.dev/192#cross-references
+//
+// That is, markdown reference-style links in the form `[Title][Definition]` where `Title` is the text that should
+// appear in the documentation and `Definition` is the name of a Protobuf entity, e.g., `google.longrunning.Operation`.
+//
+// The cross reference links can be of the form `[Title][]` when both the title and definitions match. And may also
+// be relative to the current entity (e.g. `Operation` instead of the fully qualified `google.longrunning.Operation`).
+func ExtractCrossReferenceLinks(doc ast.Node, source []byte) []string {
+	protobufLinks := map[string]bool{}
+	ast.Walk(doc, func(node ast.Node, entering bool) (ast.WalkStatus, error) {
+		switch node.Kind() {
+		case ast.KindParagraph:
+			text := node.Lines().Value(source)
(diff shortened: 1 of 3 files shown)


internal/sidekick/language/extract_xref_links.go
```diff
@@ -0,0 +1,89 @@
...
+	})
+	var sortedLinks []string
+	for link := range protobufLinks {
+		sortedLinks = append(sortedLinks, link)
+	}
+	sort.Strings(sortedLinks)
+	return sortedLinks
+}
+
+var commentCrossReferenceLink = regexp.MustCompile(
```
julieqiu: this makes my eyes hurt. I dislike regex, so I would probably write it like this:

```
func extractProtoLinks(paragraph []byte, links map[string]bool) {
	for i := 0; i < len(paragraph); i++ {
		if paragraph[i] != '[' {
			continue
		}

		// Look for [Identifier][]
		if link, ok := parseImpliedLink(paragraph[i:]); ok {
			links[link] = true
			continue
		}

		// Look for ][Identifier]
		// We check for the preceding ']' to identify the cross-reference style
		if i > 0 && paragraph[i-1] == ']' {
			if link, ok := parseCrossReference(paragraph[i:]); ok {
				links[link] = true
			}
		}
	}
}

// isProtoIdent returns true if the byte is [A-Za-z0-9_.]
func isProtoIdent(b byte) bool {
	return (b >= 'a' && b <= 'z') || (b >= 'A' && b <= 'Z') || 
	       (b >= '0' && b <= '9') || b == '_' || b == '.'
}

func parseImpliedLink(data []byte) (string, bool) {
	// Pattern: [Identifier][]
	// data[0] is '['
	i := 1
	for i < len(data) && isProtoIdent(data[i]) {
		i++
	}
	// Must have found at least one ident character and end with "][]"
	if i > 1 && i+2 < len(data) && data[i] == ']' && data[i+1] == '[' && data[i+2] == ']' {
		return string(data[1:i]), true
	}
	return "", false
}

func parseCrossReference(data []byte) (string, bool) {
	// Pattern: [Identifier]
	// data[0] is '['
	i := 1
	for i < len(data) && isProtoIdent(data[i]) {
		i++
	}
	if i > 1 && i < len(data) && data[i] == ']' {
		return string(data[1:i]), true
	}
	return "", false
}
```

but if you love regex, you could write it like this:

```
const protoIdent = `[A-Za-z][A-Za-z0-9_]*`
const protoName = protoIdent + `(\.` + protoIdent + `)*`

var (
	commentCrossReferenceLink = regexp.MustCompile(`\]\[(` + protoName + `)\]`)
	commentImpliedReferenceLink = regexp.MustCompile(`\[(` + protoName + `)\]\[\]`)
)

func extractProtoLinks(paragraph []byte, links map[string]bool) {
	for _, m := range commentCrossReferenceLink.FindAllSubmatch(paragraph, -1) {
		links[string(m[1])] = true
	}
	for _, m := range commentImpliedReferenceLink.FindAllSubmatch(paragraph, -1) {
		links[string(m[1])] = true
	}
}
```

or another option:

```
var (
	explicitLink = regexp.MustCompile(`\]\[([^.][^]]*)\]`)
	impliedLink  = regexp.MustCompile(`\[([^.][^]]*)\]\[\]`)
)

func extractProtoLinks(p []byte, links map[string]bool) {
	for _, m := range explicitLink.FindAllSubmatch(p, -1) {
		if validProtoName(m[1]) {
			links[string(m[1])] = true
		}
	}
	for _, m := range impliedLink.FindAllSubmatch(p, -1) {
		if validProtoName(m[1]) {
			links[string(m[1])] = true
		}
	}
}

func validProtoName(b []byte) bool {
	parts := bytes.Split(b, []byte("."))
	for _, p := range parts {
		if len(p) == 0 || !isIdent(p) {
			return false
		}
	}
	return true
}
```

julieqiu (approved):

internal/sidekick/language/extract_xref_links.go
coryan: Thanks. I used the last option.

PR #5865, opened 2026-05-06
cleanup(sidekick): use `Scopes()` method

coryan:
Add a `Scopes()` method to the elements that lacked them, and then use the new method. This will be more useful when we start using these methods for the Swift codec.

Towards #5072

### DIFF
--- a/internal/sidekick/api/scopes.go
+++ b/internal/sidekick/api/scopes.go
@@ -53,7 +53,7 @@ func (x *Service) Scopes() []string {
 func (x *Message) Scopes() []string {
 	localScope := strings.TrimPrefix(x.ID, ".")
 	if x.Parent == nil {
-		return []string{localScope, x.Package}
+		return []string{localScope, x.Package} // simplify some test set-up
 	}
 	return append([]string{localScope}, x.Parent.Scopes()...)
 }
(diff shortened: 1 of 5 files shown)


julieqiu (approved):

internal/sidekick/api/scopes.go
```diff
@@ -62,15 +62,48 @@ func (x *Message) Scopes() []string {
...
 	return x.Parent.Scopes()
 }
+
+// Scopes returns the scopes for a method.
+func (x *Method) Scopes() []string {
+	if x.SourceService == nil {
+		return []string{} // simplify some test set-up
+	}
+	return x.SourceService.Scopes()
+}
```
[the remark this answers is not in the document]
coryan: Fixed.

internal/sidekick/api/scopes.go
```diff
@@ -62,15 +62,57 @@ func (x *Message) Scopes() []string {
...
+	if x.Parent != nil {
+		return x.Parent.Scopes()
+	}
+	return fallbackScopes(x.ID)
+}
+
+// Scopes returns the scopes for a method.
+func (x *Method) Scopes() []string {
+	if x.SourceService != nil {
+		// For mixing methods, the local names probably refer to symbols in the source service.
```
amanda-tarafa, line 89: But the method, at least in Rust, will be on the API service not on the source service, right? So the symbols from the source service won't be in scope where the method actually is?

I think this is what's happening in https://github.com/googleapis/google-cloud-rust/pull/5599#discussion_r3204561227.

coryan: Let's continue the discussion on https://github.com/googleapis/librarian/issues/5906

Issue #5730, opened 2026-04-30
java: generate for iam and iam-policy


zhumin8:
migration script and librarian logic may need special adjustments for generating these 2 libraries.

zhumin8:
For `iam`, generate finishes with no failures with below config, artifact overrides and proto_grpc_only need to be hardcoded in migrate.
The generate result is missing *Name.java classes in v3 and v3beta: 
```
java-iam/proto-google-iam-v3/src/main/java/com/google/iam/v3/FolderLocationName.java
java-iam/proto-google-iam-v3/src/main/java/com/google/iam/v3/LocationName.java
java-iam/proto-google-iam-v3/src/main/java/com/google/iam/v3/OrganizationLocationName.java
java-iam/proto-google-iam-v3/src/main/java/com/google/iam/v3/PolicyBindingName.java
java-iam/proto-google-iam-v3/src/main/java/com/google/iam/v3/PrincipalAccessBoundaryPolicyName.java
java-iam/proto-google-iam-v3beta/src/main/java/com/google/iam/v3beta/AccessPolicyName.java
java-iam/proto-google-iam-v3beta/src/main/java/com/google/iam/v3beta/FolderLocationName.java
java-iam/proto-google-iam-v3beta/src/main/java/com/google/iam/v3beta/LocationName.java
java-iam/proto-google-iam-v3beta/src/main/java/com/google/iam/v3beta/OrganizationLocationName.java
java-iam/proto-google-iam-v3beta/src/main/java/com/google/iam/v3beta/PolicyBindingName.java
java-iam/proto-google-iam-v3beta/src/main/java/com/google/iam/v3beta/PrincipalAccessBoundaryPolicyName.java
```
The missing  *Name.java  files (e.g.,  PolicyBindingName.java ) are Resource Name helper classes. They are generated by the GAPIC generator plugin ( protoc-gen-java_gapic ) when it processes  google.api.resource  annotations in the proto files. So setting strictly `proto_grpc_only: true` is the issue.


```
  - name: iam
    version: 1.65.0
    apis:
      - path: google/iam/v3
      - path: google/iam/v2
      - path: google/iam/v1
      - path: google/iam/v3beta
      - path: google/iam/v2beta
    java:
      api_description_override: Manages access control for Google Cloud Platform resources
      client_documentation_override: https://cloud.google.com/java/docs/reference/proto-google-iam-v1/latest/overview
      distribution_name_override: com.google.api.grpc:proto-google-iam-v1
[...]

PR #5867, opened 2026-05-06
feat(internal/librarian/java): allow more granular generate control

zhumin8:
Allow more granular generate control with configs. Instead of `proto_grpc_only` , introduce `generate_gapic`, `generate_proto_grpc` and `generate_resource_names` to control what to generate. Default to generate all.

This is to allow more control to generate `iam` and `iam-policy`. For context, these two libraries are generated from same set of API paths, `iam` contains all grpc and proto modules plus resource name classes (generated in gapic phase, but relocated in proto module), `iam-policy` contains generated gapic modules.

For #5730


tool/cmd/migrate/java.go
```diff
@@ -803,3 +805,25 @@ func extractVersionFromPOM(pomPath string) (string, error) {
...
+	}
+	if !slices.Contains(explicitPaths, api.Path) {
+		return
+	}
+	if libraryName == "iam" {
+		api.GenerateGAPIC = new(false)
+		api.GenerateProtoGRPC = new(true)
+		api.GenerateResourceNames = new(true)
+	}
+	if libraryName == "iam-policy" {
```
sofisl: same thing about hardcoded, maybe we could have a global-level variable or something for mixins?

tool/cmd/migrate/java.go
```diff
@@ -803,3 +805,25 @@ func extractVersionFromPOM(pomPath string) (string, error) {
 	}
 	return string(match[1]), nil
 }
+
+func applyJavaIAMSpecialOverrides(libraryName string, api *config.JavaAPI) {
+	explicitPaths := []string{
```
sofisl: should this be a param to this function to make it more testable?

tool/cmd/migrate/java.go
```diff
@@ -803,3 +805,25 @@ func extractVersionFromPOM(pomPath string) (string, error) {
 	}
 	return string(match[1]), nil
 }
+
```
sofisl, line 822: comment for this function?

internal/librarian/java/generate.go
```diff
@@ -409,3 +411,24 @@ func filterProtos(fullPaths []string, relExcludes []string, root string) []strin
 	}
 	return filtered
 }
+
+func shouldGenerateGAPIC(ja *config.JavaAPI) bool {
```
sofisl: I think prefer full names to abbreviated names per docs (ja vs javaApi)

tool/cmd/migrate/java_test.go
```diff
@@ -1528,3 +1530,63 @@ func TestInsertMarkers_Full(t *testing.T) {
...
+			apiPath:     "google/iam/v2",
+			wantGAPIC:   new(true),
+			wantProto:   new(false),
+			wantResName: new(false),
+		},
+		{
+			name:        "iam non-special path no override",
+			libraryName: "iam",
+			apiPath:     "google/iam/v1",
+			wantGAPIC:   nil,
```
sofisl: do we need the nil cheks? same below

tool/cmd/migrate/java.go
zhumin8: refactored to use java_module.go for hardcoded values.

tool/cmd/migrate/java.go
zhumin8: refactored to use java_module.go for hardcoded values.

sofisl (approved):

PR #5870, opened 2026-05-06
feat(internal/sidekick/gcloud): support list methods in generator

julieqiu:
The gcloud CLI generator is updated to support List methods. This enables generated commands to invoke the API and iterate through results rather than displaying a placeholder message.

### DIFF
--- a/internal/sidekick/gcloud/client.go
+++ b/internal/sidekick/gcloud/client.go
@@ -94,25 +94,36 @@ func clientInfoFromPath(clientImportPath string) *goClientInfo {
 	}
 }
 
-// buildClientCall returns a ClientCall for an AIP-131 Get method when the
-// model maps to a standard GAPIC Go package and the command composes a
-// resource path. It returns nil otherwise so the command keeps its
-// print-only action.
+// buildClientCall returns a ClientCall for an AIP-131 Get or AIP-132 List
+// method when the model maps to a standard GAPIC Go package and the command
+// composes a resource path. It returns nil otherwise so the command keeps
+// its print-only action.
 func buildClientCall(method *api.Method, goClient *goClientInfo, hasPath bool) *ClientCall {
 	if goClient == nil || !hasPath {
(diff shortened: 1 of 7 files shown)


internal/sidekick/gcloud/client.go
```diff
@@ -94,25 +94,36 @@ func clientInfoFromPath(clientImportPath string) *goClientInfo {
...
+
+	switch {
+	case provider.IsGet(method):
+		return &ClientCall{
+			Method:      method.Name,
+			NameField:   "Name",
+			Package:     goClient.Alias,
+			RequestType: goClient.Alias + "pb." + method.InputType.Name,
+		}
+	case provider.IsList(method):
```
coryan, line 117: `IsList()` was introduce to write samples. It is correct, by the definition in [aip.dev/132](https://aip.dev/132), but there is also [aip.dev/158](https://aip.dev/158) which defines what methods produce paginated APIs and which methods return an iterator (well, there is [aip.dev/4223](https://google.aip.dev/client-libraries/4233) too, to make things more complicated).

In other words, you could have a method that is a paginated API and is not a list API.  Is it likely? No. Do I want to bet that there are none of these in GCP? No. Do I want to bet that there are some such methods in GCP? Yes, but not more than 2 pennies.

You may want to use 

https://github.com/googleapis/librarian/blob/f1acf900023a52174edadeb62092c0780c807879/internal/sidekick/api/model.go#L437-L438

Which is a more generous definition of "I can use this method to list things".

Or maybe simplify all this code to page it easy for all of ourselves to deal with the absurd complexity.

coryan (approved):

internal/sidekick/gcloud/client.go
julieqiu: thanks! filed https://github.com/googleapis/librarian/issues/5882 - will revisit once we're generating surfaces beyond parallelstore

PR #5871, opened 2026-05-06
refactor(sidekick): use string for PathSegment.Literal

ianthehat:
This change converts the Literal field in the PathSegment struct from `*string` to just `string`. This simplifies object construction and eliminates the need for safe pointer handling (nil checks and dereferences) throughout the codebase.

### DIFF
--- a/internal/sidekick/api/model.go
+++ b/internal/sidekick/api/model.go
@@ -741,8 +741,8 @@ func (template *PathTemplate) FlatPath() string {
 	sep := ""
 	for _, segment := range template.Segments {
 		buffer.WriteString(sep)
-		if segment.Literal != nil {
-			buffer.WriteString(*segment.Literal)
+		if segment.Literal != "" {
+			buffer.WriteString(segment.Literal)
 		} else if segment.Variable != nil {
 			fmt.Fprintf(&buffer, "{%s}", strings.Join(segment.Variable.FieldPath, "."))
 		}
@@ -753,7 +753,7 @@ func (template *PathTemplate) FlatPath() string {
 
 // PathSegment is a segment of a path.
 type PathSegment struct {
-	Literal  *string
+	Literal  string
 	Variable *PathVariable
 }
 
@@ -773,7 +773,7 @@ func NewPathVariable(fields ...string) *PathVariable {
 
 // WithLiteral adds a literal to the path template.
 func (p *PathTemplate) WithLiteral(l string) *PathTemplate {
-	p.Segments = append(p.Segments, PathSegment{Literal: &l})
+	p.Segments = append(p.Segments, PathSegment{Literal: l})
 	return p
 }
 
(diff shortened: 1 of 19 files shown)


ianthehat:
I think this is a safe, correct and good change, but I was worried that it might break a template without failing a test, is that likely?

internal/sidekick/parser/httprule/http_rule_parser.go
```diff
@@ -204,7 +204,7 @@ func parseLiteralSegment(literalSegment string) (*api.PathSegment, int, error) {
 		return nil, 0, err
 	}
 	return &api.PathSegment{
-		Literal: (*string)(literal),
+		Literal: string(*literal),
```
coryan, line 207: Seems like `parseLiteral()` should change to return `string`?

internal/sidekick/api/xref.go
```diff
@@ -411,12 +411,11 @@ func toResourceNamePattern(pattern ResourcePattern, skipLast bool) *ResourceName
...
 		s := pattern[i]
-		if s.Literal != nil && strings.HasPrefix(*s.Literal, "//") {
+		if strings.HasPrefix(s.Literal, "//") {
 			continue
 		}
-		seg := ResourceNameSegment{}
-		if s.Literal != nil {
-			seg.Literal = *s.Literal
+		seg := ResourceNameSegment{
+			Literal: s.Literal,
```
coryan, line 418: Shouldn't this be:


```suggestion
        if s.Literal != "" {
		    seg := ResourceNameSegment{
			    Literal: s.Literal,
			}
```

 the `s` thing is supposed to be a sum type, and `nil` was a marker for "not set", now the marker is  `""`. Right?

internal/sidekick/api/model.go
```diff
@@ -753,7 +753,7 @@ func (template *PathTemplate) FlatPath() string {
 
 // PathSegment is a segment of a path.
 type PathSegment struct {
```
coryan, line 755: So what is the correct way to model "Either Literal or Variable are set but not both"?

coryan (changes requested):
I think maybe I found a problem.  Separately, I searched for Literal and none of the templates use it directly.  If they did, I think `{{#Literal}}` and `{{Literal}}` would still work the same.

internal/sidekick/api/xref.go
ianthehat: This should match the previous behavior, if the literal is the empty string, then it is the same as the default value of the initialized struct.

internal/sidekick/parser/httprule/http_rule_parser.go
ianthehat: Yes, but that also impacts parseVerb, which is problematic.
Changing Verb to plain string is the next PR in the stack and it is easy to fix after that one.

internal/sidekick/api/model.go
ianthehat: I do wish there were sum types...
There is not really a correct way, there are a few ways and which one is best depends on circumstance.
If it was really important to enforce it, we would probably hide it behind a struct that carried a string and a tag.

internal/sidekick/api/xref.go
coryan: Ack, I see now.

coryan (approved):

PR #5872, opened 2026-05-06
fix(tool/cmd/migrate): fix and override for showcase for java

zhumin8:
The hardcoded overrides added in https://github.com/googleapis/librarian/pull/5801 was never reached because showcase BUILD.bazel does not contain standard java rules. Add logic to hardcode relevant configs according to [BUILD.partial.bazel](https://github.com/googleapis/google-cloud-java/blob/2cecd0caf4f15da2d894a70077ccb0d0fd69d296/java-showcase/scripts/resources/BUILD.partial.bazel#L18-L24).
Also added a warning in case similar situation happens (this is the only one).

For #5731

### DIFF
--- a/tool/cmd/migrate/java.go
+++ b/tool/cmd/migrate/java.go
@@ -46,6 +46,8 @@ const (
 
 	managedModulesStartMarker = "<!-- {x-generated-modules-start} -->"
 	managedModulesEndMarker   = "<!-- {x-generated-modules-end} -->"
+
+	showcaseLibraryName = "showcase"
 )
 
 var (
@@ -254,8 +256,8 @@ func buildConfig(gen *GenerationConfig, repoPath string, src, showcaseSrc *confi
 		var apis []*config.API
 		var javaAPIs []*config.JavaAPI
 		var roots []string
-		if name == "showcase" {
-			roots = []string{"showcase", "googleapis"}
+		if name == showcaseLibraryName {
+			roots = []string{showcaseLibraryName, "googleapis"}
 		}
 		for _, g := range l.GAPICs {
 			if g.ProtoPath == "" {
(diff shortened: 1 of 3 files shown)


zhumin8:
The migrated config for showcase now is: 

```
  - name: showcase
    version: 0.0.1-SNAPSHOT
    apis:
      - path: schema/google/showcase/v1beta1
    keep:
      - gapic-showcase/src/test
      - gapic-showcase/src/main/java/com/google/showcase/v1beta1/Version.java
    roots:
      - showcase
      - googleapis
    java:
      api_description_override: Showcase module
      distribution_name_override: com.google.cloud:gapic-showcase
      excluded_poms: gapic-showcase-bom
      library_type_override: OTHER
      name_pretty_override: Showcase
      java_apis:
        - path: schema/google/showcase/v1beta1
          additional_protos:
            - google/cloud/location/locations.proto
            - google/iam/v1/iam_policy.proto
          omit_common_resources: true
          grpc_artifact_id_override: grpc-gapic-showcase-v1beta1
          proto_artifact_id_override: proto-gapic-showcase-v1beta1
      product_documentation_override: https://cloud.google.com/dummy
      skip_api_id: true
```

tool/cmd/migrate/java.go
```diff
@@ -253,18 +253,30 @@ func buildConfig(gen *GenerationConfig, repoPath string, src, showcaseSrc *confi
 		version := versions[artifactID]
 		var apis []*config.API
 		var javaAPIs []*config.JavaAPI
+		var roots []string
+		if name == "showcase" {
```
sofisl: should this be a global const? not sure how prevalent that is in go, in js we'd list it in the top of the file just to be aware of all the changes in case this name ever changes

sofisl (approved):

PR #5880, opened 2026-05-07
docs(internal/librarian/surfer): document surfer generation limitation

jameslynnwu:
Add a comment to Generate explaining that generating gcloud commands using surfer from librarian is not supported at the moment.

The current implementation of provider.Config{} creation is incomplete and incorrect, so we document this limitation to avoid confusion.

### DIFF
--- a/internal/librarian/surfer/generate.go
+++ b/internal/librarian/surfer/generate.go
@@ -88,6 +88,9 @@ func generateAPI(api *config.API, gcloudCfg *config.Surfer, googleapisDir, outDi
 	if err != nil {
 		return err
 	}
+	// Generating gcloud commands using surfer from librarian is not supported at the
+	// moment. The current implementation is incomplete and incorrect in how it
+	// initializes provider.Config{}.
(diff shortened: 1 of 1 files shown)


sarahheacock (approved):

internal/librarian/surfer/generate.go
```diff
@@ -88,6 +88,9 @@ func generateAPI(api *config.API, gcloudCfg *config.Surfer, googleapisDir, outDi
 	if err != nil {
 		return err
 	}
+	// Generating gcloud commands using surfer from librarian is not supported at the
+	// moment. The current implementation is incomplete and incorrect in how it
+	// initializes provider.Config{}.
 	providerCfg := &provider.Config{}
 	if gcloudCfg != nil && gcloudCfg.HelpText != nil {
 		providerCfg.APIs = []provider.API{{
```
sarahheacock, line 96: I would personally recommend just passing in an empty config since we know this is not correct

PR #5888, opened 2026-05-07
feat(internal/sidekick/gcloud): add --limit flag to list commands

julieqiu:
Add support for `--limit=N` for list commands. The flag is added only to commands whose ClientCall has Paged set, and the iterator loop in the generated action breaks once the count reaches the limit. The default zero value means unlimited, which matches gcloud's behavior.

### DIFF
--- a/internal/sidekick/gcloud/command.go
+++ b/internal/sidekick/gcloud/command.go
@@ -36,9 +36,6 @@ type ClientCall struct {
 	// IsDelete reports whether the method is a standard Delete method.
 	IsDelete bool
 
-	// IsList reports whether the method is a standard List method.
-	IsList bool
-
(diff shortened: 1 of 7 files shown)


internal/sidekick/gcloud/command.go
```diff
@@ -57,6 +54,10 @@ type ClientCall struct {
 	// `client, err := <Package>.NewClient(ctx)`.
 	Package string
 
+	// Paged reports whether the method returns a paginated iterator (e.g.,
```
coryan, line 57: Much better name, I might steal it.

coryan (approved):

internal/sidekick/gcloud/command.go
julieqiu: ™️

Issue #5858, opened 2026-05-06
surfer: disable star update mask


sarahheacock:
Use gcloud config to disable star update mask

PR #5891, opened 2026-05-07
feature(internal/sidekick/surfer): add accessor for star update

sarahheacock:
* Make star update mask optional so we can default to True
* Add accessor for getting the correct api config
* Remove dead ApiVersion function  

For #5858

### DIFF
--- a/internal/sidekick/surfer/provider/config.go
+++ b/internal/sidekick/surfer/provider/config.go
@@ -286,9 +286,6 @@ func ShouldGenerateOperations(c *Config) bool {
 // SupportsStarUpdateMasks returns true if the API/Service supports '*' updateMasks.
 // It matches the service name and API version and defaults to true if omitted.
 func SupportsStarUpdateMasks(c *Config, serviceName, version string) bool {
-	if c == nil {
-		return true
-	}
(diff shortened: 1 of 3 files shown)


internal/sidekick/surfer/provider/config.go
```diff
@@ -280,3 +282,19 @@ func ShouldGenerateOperations(c *Config) bool {
 	}
 	return *c.GenerateOperations
 }
+
+// SupportsStarUpdateMasks returns true if the API/Service supports '*' updateMasks.
+// It matches the service name and API version and defaults to true if omitted.
+func SupportsStarUpdateMasks(c *Config, serviceName, version string) bool {
+	if c == nil {
```
jameslynnwu: this check shouldn't be necessary?

jameslynnwu (approved):

internal/sidekick/surfer/provider/config.go
sarahheacock: Removed for now. We're still undecided how to handle nils going forward since we do not want them everywhere. But theoretically, a config should be optional so we can autogenerate as many as possible without intervention

PR #5903, opened 2026-05-07
feat(internal/librarian/java): add config for protos to generate with proto module and copy

### LINKED ISSUE #5685: java(os-login diff): common.proto is not copied into proto-google-cloud-os-login-v1/src/main/proto
In current hermetic build, common.proto is hardcoded in script: 
https://github.com/googleapis/google-cloud-java/blob/85f51b299e187645add14d39fa38a738df9b3574/sdk-platform-java/hermetic_build/library_generation/generate_library.sh#L157-L161
It is added as if part of API path protos, so included in both code generation, and eventually copied over to:

https://github.com/googleapis/google-cloud-java/blob/main/java-os-login/proto-google-cloud-os-login-v1/src/main/proto/google/cloud/oslogin/common/common.proto

However, in librarian, we added this common.proto as additional_protos, similar to MIXINs like location and iam, so they are included in code generation, but not included in copy to proto-google-cloud-os-login-v1/src/main/proto.
```
  - name: os-login
    version: 2.88.0-SNAPSHOT
    apis:
      - path: google/cloud/oslogin/v1
      - path: google/cloud/oslogin/v1beta
    java:
      api_description_override: manages OS login configuration for Directory API users.
      issue_tracker_override: https://issuetracker.google.com/savedsearches/559755
      name_pretty_override: Cloud OS Login
      java_apis:
        - path: google/cloud/oslogin/v1
          additional_protos:
            - google/cloud/oslogin/common/common.proto
          samples: false
        - path: google/cloud/oslogin/v1beta
          additional_protos:
            - google/cloud/oslogin/common/common.proto
          samples: false
      product_documentation_override: https://cloud.google.com/compute/docs/oslogin/
```

@blakeli0 What is the intention behind this difference: 
- we don't include proto files from common_resources.proto, locations.proto in the copied proto dir, 
- but includes this oslogin/common/common.proto in there
- For a similar hardcoded case operation_metadata.proto, it is explicitly excluded from copy in:  https://github.com/googleapis/google-cloud-java/blob/85f51b299e187645add14d39fa38a738df9b3574/sdk-platform-java/hermetic_build/library_generation/generate_library.sh#L259-L262

zhumin8:
Introduce a config `additional_protos_to_generate_and_copy` to fulfill generation of os-login. This is different from the existing `additional_protos` used for mixed-ins, in additional to be included in gapic generation phase, this needs to be included in proto generation phase and proto copy.

Fix https://github.com/googleapis/librarian/issues/5685


internal/librarian/java/postprocess.go
```diff
@@ -47,15 +47,16 @@ var (
...
-	includeSamples bool
+	cfg                               *config.Config
+	library                           *config.Library
+	javaAPI                           *config.JavaAPI
+	metadata                          *repoMetadata
+	outDir                            string
+	apiBase                           string
+	protoSourceDir                    string
+	apiProtos                         []string
+	additionalProtosToGenerateAndCopy []string
```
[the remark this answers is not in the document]
zhumin8: Practically this won't happen, because showcase does not have this config. But this does feel wrong.
Moved logic that derives relative path prior to postprocess, so that less params needed.

internal/librarian/java/generate.go
```diff
@@ -205,6 +211,35 @@ func deriveAdditionalProtoPaths(javaAPI *config.JavaAPI, googleapisDir string) [
...
 
+func deriveAbsoluteProtoPaths(paths []string, baseDir string) []string {
+	var absPaths []string
+	for _, p := range paths {
+		absPaths = append(absPaths, filepath.Join(baseDir, filepath.FromSlash(p)))
+	}
+	return absPaths
+}
+
+func deriveProtosToCopy(apiProtos []string, primaryDir string, additionalRel []string, googleapisDir string) ([]protoFileToCopy, error) {
```
sofisl, line 224: I wonder if it's worth adding a comment to this function or renaming it. it seems like this function and the one above map 1:1 with `additionalProtos` and `additionalProtosToGenerateAndCopyAbs` but the naming and/or comments don't reflect it.

internal/librarian/java/generate.go
```diff
@@ -205,6 +211,35 @@ func deriveAdditionalProtoPaths(javaAPI *config.JavaAPI, googleapisDir string) [
...
+	var res []protoFileToCopy
+	for _, p := range apiProtos {
```
sofisl: prefer full names

internal/librarian/java/postprocess.go
```diff
@@ -406,19 +410,14 @@ func deriveLastReleasedVersion(v string) (string, error) {
...
-func copyProtos(protoSourceDir string, protos []string, destDir string) error {
-	for _, proto := range protos {
-		// Calculate relative path from protoSourceDir to preserve directory structure
-		rel, err := filepath.Rel(protoSourceDir, proto)
-		if err != nil {
-			return fmt.Errorf("failed to calculate relative path for %s: %w", proto, err)
-		}
-		target := filepath.Join(destDir, rel)
+func copyProtos(protos []protoFileToCopy, destDir string) error {
+	for _, p := range protos {
```
sofisl: prefer full names

internal/librarian/java/generate.go
zhumin8: added comment, hope it clarifies.

sofisl (approved):

PR #5905, opened 2026-05-07
feat(internal/librarian/java): generate owlbot.py for newly added library

### LINKED ISSUE #5679: java: generate owlbot.py for newly added library
After librarian add adds in a config for a new library in librarian.yaml, when generate runs, it should add a owlbot.py from template to allow for readme.md generation and restructure by owlbot.

sofisl:
Ensures that when a new Java client library is generated for the first time, the post-processor dynamically templates and writes `owlbot.py` if it is missing from the output directory. This aligns Librarian with the legacy Python hermetic build's prerequisite setup and prevents "owlbot.py not found" errors.

Fixes https://github.com/googleapis/librarian/issues/5679


internal/librarian/java/template/owlbot_py.tmpl
```diff
@@ -0,0 +1,38 @@
+# Copyright 2026 Google LLC
```
sofisl, line 1: Do these dates ever need to be updated? if so, should that be done manually or automatically? or is there a separate process to update copyright dates in our libraries?

internal/librarian/java/postprocess.go
```diff
@@ -474,3 +470,24 @@ func removeKeptFilesFromStaging(library *config.Library, outDir string) error {
...
+		file, err := os.Create(owlbotPath)
+		if err != nil {
+			return fmt.Errorf("failed to create owlbot.py: %w", err)
+		}
+		defer file.Close()
+		// Execute the template to write the default owlbot.py script content.
+		if err := templates.ExecuteTemplate(file, "owlbot_py.tmpl", nil); err != nil {
+			return fmt.Errorf("failed to write owlbot.py template: %w", err)
+		}
+	}
```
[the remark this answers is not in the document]
sofisl: I think this is the point of the PR... @zhumin8 lmk if I'm misunderstanding otherwise I'll just resolve this.

internal/librarian/java/postprocess.go
```diff
@@ -474,3 +470,45 @@ func removeKeptFilesFromStaging(library *config.Library, outDir string) error {
...
+	// If owlbot.py does not exist, generate it from the embedded template.
+	if _, err := os.Stat(owlbotPath); errors.Is(err, fs.ErrNotExist) {
+		file, err := os.Create(owlbotPath)
+		if err != nil {
+			return fmt.Errorf("failed to create owlbot.py: %w", err)
+		}
+		defer file.Close()
+
+		data := struct {
+			ShouldIncludeTemplates bool
```
sofisl: @zhumin8 I feel like this should be plumbed from somewhere (like the librarian.yaml)? but I couldn't find the appropriate data value to route it through. or should it actually be hardcoded? that feels wrong to me.

internal/librarian/java/template/owlbot_py.tmpl
zhumin8: No we don't have a separate process to update. Currently in google-cloud-java, copyright header are updated each year to current for generated files. For these that are generated once, they are left with the original year ([example](https://github.com/googleapis/google-cloud-java/blob/main/java-apigee-registry/owlbot.py)).
Both strategy are okay according to go/copyright#the-year and [our policy](https://g3doc.corp.google.com/company/teams/cloud-sdk/client-libraries/team/generators/index.md#copyright-headers). Not sure if it worth to unify copyright strategy across languages in the future?

internal/librarian/java/testdata/postprocess/owlbot.py.golden
```diff
@@ -0,0 +1,38 @@
...
+        "CODE_OF_CONDUCT.md",
+        "CONTRIBUTING.md",
+        "LICENSE",
+        "SECURITY.md",
+        "java.header",
+        "license-checks.xml",
+        "renovate.json",
+        ".gitignore",
+    ],
+)
```
zhumin8: note there has been a recent change in google-cloud-java, https://github.com/googleapis/google-cloud-java/pull/13028/changes
we should follow the same

internal/librarian/java/postprocess.go
zhumin8: I think we don't need this variable at all. In hermetic build, the only call explicitly sets it to true [here](https://github.com/googleapis/google-cloud-java/blob/ddef7d255af64698be423cc991a963809c217fca/sdk-platform-java/hermetic_build/library_generation/utils/utilities.py#L331).
(I suspect this is leftover in attempts to generate owlbot.py for existing files.)

Anyway, there are 2 reasons we need a owlbot.py for new libraries now: 
- it moves code staging dir
- it generates README.md from template

So we need the `java.common_templates()` section for all new libraries.

internal/librarian/java/template/owlbot_py.tmpl
sofisl: I'll bring it up in the next meeting!

internal/librarian/java/postprocess.go
```diff
@@ -474,3 +470,43 @@ func removeKeptFilesFromStaging(library *config.Library, outDir string) error {
...
+				".gitignore",
+			},
+		}
+
+		// Execute the template to write the default owlbot.py script content.
+		if err := templates.ExecuteTemplate(file, "owlbot_py.tmpl", data); err != nil {
+			return fmt.Errorf("failed to write owlbot.py template: %w", err)
+		}
+	}
+	return nil
```
zhumin8, line 503: I think this func is silently ignoring any errors returned by os.Stat that are not "not exist" errors.

internal/librarian/java/postprocess.go
```diff
@@ -474,3 +470,43 @@ func removeKeptFilesFromStaging(library *config.Library, outDir string) error {
...
+// OwlBot post-processing and README generation to complete successfully.
+func createOrVerifyOwlbotPy(outDir string) error {
+	owlbotPath := filepath.Join(outDir, "owlbot.py")
+	// If owlbot.py does not exist, generate it from the embedded template.
+	if _, err := os.Stat(owlbotPath); errors.Is(err, fs.ErrNotExist) {
+		file, err := os.Create(owlbotPath)
+		if err != nil {
+			return fmt.Errorf("failed to create owlbot.py: %w", err)
+		}
+		defer file.Close()
```
zhumin8: Use a named return error to capture and propagate errors returned by `file.Close()` during defer, otherwise close errors might be silently discarded

internal/librarian/java/postprocess.go
```diff
@@ -474,3 +470,43 @@ func removeKeptFilesFromStaging(library *config.Library, outDir string) error {
...
+
```
zhumin8: nit: empty lines

internal/librarian/java/postprocess_test.go
```diff
@@ -941,3 +936,45 @@ func TestRemoveKeptFilesFromStaging(t *testing.T) {
...
 	}
 }
+
+// TestCreateOrVerifyOwlbotPy verifies that the createOrVerifyOwlbotPy function
+// successfully creates the owlbot.py post-processing script when it is missing
+// and generates a valid script that matches the expected golden content.
+func TestCreateOrVerifyOwlbotPy(t *testing.T) {
+	t.Parallel()
+	outDir := t.TempDir()
+	// 1. Call createOrVerifyOwlbotPy to trigger default template generation.
```
zhumin8: nit: comment with not much value

internal/librarian/java/postprocess_test.go
```diff
@@ -941,3 +936,45 @@ func TestRemoveKeptFilesFromStaging(t *testing.T) {
...
+	owlbotPath := filepath.Join(outDir, "owlbot.py")
+	if _, err := os.Stat(owlbotPath); err != nil {
+		t.Errorf("expected owlbot.py to be generated: %v", err)
+	}
+	// 3. Read the generated script's contents.
+	gotContent, err := os.ReadFile(owlbotPath)
+	if err != nil {
+		t.Fatal(err)
+	}
+
```
zhumin8: nit: empty lines, same below

internal/librarian/java/testdata/postprocess/owlbot.py.golden
```diff
@@ -0,0 +1,37 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import synthtool as s
+from synthtool.languages import java
+
```
zhumin8, line 17: more of a note: owlbot.py in google-cloud-java has one additional empty line here (e.g. [java-secretmanager/owlbot.py](https://github.com/googleapis/google-cloud-java/blob/main/java-secretmanager/owlbot.py)), not sure if it was intentional. 
I think it should be fine though, as this is only adding for new libraries, as long as it does not trigger any formatter failure (I did not find any Python Linters/Formatters in google-cloud-java).

internal/librarian/java/postprocess_test.go
```diff
@@ -941,3 +936,45 @@ func TestRemoveKeptFilesFromStaging(t *testing.T) {
...
+	// 4. Verify that the generated script's contents exactly match the expected golden content.
+	wantContent, err := os.ReadFile(goldenPath)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if diff := cmp.Diff(string(wantContent), string(gotContent)); diff != "" {
+		t.Errorf("generated owlbot.py content mismatch (-want +got):\n%s\n\nHint: run 'go test ./internal/librarian/java -v -update' to update golden files.", diff)
```
zhumin8:
```suggestion
t.Errorf("mismatch (-want +got):\n%s", diff)
```
Shorten error message, ref https://github.com/googleapis/librarian/blob/main/doc/howwewritego.md#use-cmpdiff-for-comparisons

internal/librarian/java/postprocess.go
```diff
@@ -474,3 +470,43 @@ func removeKeptFilesFromStaging(library *config.Library, outDir string) error {
...
 	})
 }
+
+// createOrVerifyOwlbotPy ensures that the post-processing script (owlbot.py) exists
+// in the library's output directory. If it is missing (which is typical for newly added
+// client libraries), it automatically creates it from an embedded template to allow
+// OwlBot post-processing and README generation to complete successfully.
+func createOrVerifyOwlbotPy(outDir string) error {
+	owlbotPath := filepath.Join(outDir, "owlbot.py")
+	// If owlbot.py does not exist, generate it from the embedded template.
```
zhumin8: nit: this comment don't add much value

internal/librarian/java/postprocess.go
```diff
@@ -474,3 +470,43 @@ func removeKeptFilesFromStaging(library *config.Library, outDir string) error {
...
+				"LICENSE",
+				"SECURITY.md",
+				"java.header",
+				"license-checks.xml",
+				"renovate.json",
+				".gitignore",
+			},
+		}
+
+		// Execute the template to write the default owlbot.py script content.
```
zhumin8: nit: empty line and remove this comment

internal/librarian/java/testdata/postprocess/owlbot.py.golden
sofisl: added empty line

internal/librarian/java/postprocess.go
```diff
@@ -480,3 +476,47 @@ func removeKeptFilesFromStaging(library *config.Library, outDir string) error {
...
+				"samples/*",
+				"CODE_OF_CONDUCT.md",
+				"CONTRIBUTING.md",
+				"LICENSE",
+				"SECURITY.md",
+				"java.header",
+				"license-checks.xml",
+				"renovate.json",
+				".gitignore",
+			},
```
zhumin8: what about moving this into the template?

internal/librarian/java/postprocess.go
```diff
@@ -480,3 +476,47 @@ func removeKeptFilesFromStaging(library *config.Library, outDir string) error {
...
+// in the library's output directory. If it is missing (which is typical for newly added
+// client libraries), it automatically creates it from an embedded template to allow
+// OwlBot post-processing and README generation to complete successfully.
+func createOrVerifyOwlbotPy(outDir string) (err error) {
+	owlbotPath := filepath.Join(outDir, "owlbot.py")
+	if _, statErr := os.Stat(owlbotPath); statErr != nil {
+		if !errors.Is(statErr, fs.ErrNotExist) {
+			return fmt.Errorf("failed to stat owlbot.py: %w", statErr)
+		}
+		file, createErr := os.Create(owlbotPath)
```
zhumin8: I just learned this: A file created via os.Create will not be executable by default.
We should 
```
// Replace os.Create(owlbotPath) with an explicit permission set
file, createErr := os.OpenFile(owlbotPath, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0755)
```
Because owlbot.py needs to be executed later in postprocess.

Also, instead of os.Stat and then os.Create, you can tell the OS to only create the file if it does not already exist.

This is done using the os.O_EXCL (exclusive) flag combined with os.O_CREATE. The operating system guarantees that this check-and-creation happens as a single, indivisible (atomic) operation.
```
func createOrVerifyOwlbotPy(outDir string) (err error) {
	owlbotPath := filepath.Join(outDir, "owlbot.py")

	// Open with Create and Exclusive flags. 
	// If the file already exists, this returns an os.ErrExist error instantly.
	file, createErr := os.OpenFile(owlbotPath, os.O_RDWR|os.O_CREATE|os.O_EXCL, 0755)
	
	if errors.Is(createErr, fs.ErrExist) {
		// Happy path: The file already exists, so "verification" is successful.
		return nil 
	}
	if createErr != nil {
		// An actual error occurred (e.g., permission denied, directory missing).
		return fmt.Errorf("failed to create owlbot.py: %w", createErr)
	}
	
	// Safely close the file when done writing
	defer func() {
		if closeErr := file.Close(); closeErr != nil && err == nil {
			err = fmt.Errorf("failed to close owlbot.py: %w", closeErr)
		}
	}()

	// ... Proceed to execute the template and write to the file ...
}
```

zhumin8 (approved):

PR #5919, opened 2026-05-08
doc(sidekick/api): fix bad comment

coryan:
The comment was making Gemini confused.

### DIFF
--- a/internal/sidekick/api/model.go
+++ b/internal/sidekick/api/model.go
@@ -1156,13 +1156,32 @@ type Resource struct {
 }
 
 // ResourcePattern is a sequence of path segments that defines the structure of a resource's unique identifier.
+//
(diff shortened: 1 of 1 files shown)


internal/sidekick/api/model.go
```diff
@@ -1160,7 +1160,7 @@ type ResourcePattern []PathSegment
 
 // ResourceNameSegment is a segment of a resource name pattern.
 type ResourceNameSegment struct {
-	// Literal is the literal part of the segment, including any separators.
+	// Literal is the literal part of the segment.
```
jameslynnwu, line 1177: These comments are comments.

Can we add an example to the ResourceNameSegment docstring that illustrates what literal and variable mean?

jameslynnwu (approved):

internal/sidekick/api/model.go
coryan: Done.

PR #5926, opened 2026-05-08
fix(internal/librarian/java): ensure missing preserved files are generated

### LINKED ISSUE #5925: java: missing keep files should be generated for the first time
Found when testing for https://github.com/googleapis/librarian/issues/5820, if a file is on keep list but does not exists in filesystem yet, it won't be in the output even if it is generated. 
Use case are for files that intend to be generated once and not updated in re-generation, like Version.java. 

For existing libraries it works fine, as Version.java is present and not removed in clean phase.
However, for generating new libraries, the intention is to generate Version.java for the first time. Current logic prevents Version.java to be moved to output location.
We should allow

zhumin8:
Allow files in keep list that is generated to be in final output if it does not exist in filesystem to start.

Updated  logic in postprocess that removes kept files from staging to check whether a preserved file or directory actually exists in the target repository before removing it from the temporary staging area. This ensures that explicitly kept files (such as `Version.java`) are successfully generated and copied over initially if they are not already present, while still protecting existing customized files from being overwritten.

Fix #5925

### DIFF
--- a/internal/librarian/java/postprocess.go
+++ b/internal/librarian/java/postprocess.go
@@ -464,8 +464,8 @@ func removeKeptFilesFromStaging(library *config.Library, outDir string) error {
 					if err := os.RemoveAll(path); err != nil {
 						return fmt.Errorf("failed to remove kept dir %s from staging: %w", path, err)
 					}
-					return filepath.SkipDir
 				}
+				return filepath.SkipDir
 			}
 			return nil
 		}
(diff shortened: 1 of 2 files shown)


suztomo (approved):
Can you create keep/delete conditions (the keep list, the generated markers, exclusions in owlbot.py, etc.) in a section in your document with use cases?

PR #5934, opened 2026-05-08
ci(rust): replace install-taplo with librarian install

suztomo:
Replaces custom install-taplo action with librarian install rust. This is to measure installation time.


suztomo:
It hit timeout of 5 minutes https://github.com/googleapis/librarian/actions/runs/25571804359/job/75069127354?pr=5934.

PR #5948, opened 2026-05-11
fix(internal/librarian/java): do not identify missing modules if skip_pom_updates

zhumin8:
When skip_pom_updates is true,  identify missing modules step should return nil, so that no new entries to versions.txt are added. This was missed in original #5660. This check centralized inside `discoverModules`.

`IdentifyMissingModules` takes in srcs instead of googleapisDir to correctly process showcase.

For #5731


internal/librarian/java/pom.go
```diff
@@ -245,8 +249,10 @@ func syncPOMs(library *config.Library, libraryDir, monorepoVersion string, metad
 // IdentifyMissingModules identifies all expected proto-*, grpc-*, client, BOM and Parent modules
 // for the given library based on its configuration and checks for pom.xml presence
 // on the filesystem. It returns a list of artifact IDs for the missing modules.
-func IdentifyMissingModules(library *config.Library, libraryDir, googleapisDir string) ([]string, error) {
-	transports, err := loadTransports(library, googleapisDir)
+func IdentifyMissingModules(library *config.Library, libraryDir string, srcs *sources.Sources) ([]string, error) {
+	srcCfg := sources.NewSourceConfig(srcs, library.Roots)
```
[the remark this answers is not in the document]
zhumin8: This is fine, as loadTransports  is relatively fast, and SkipPOMUpdates is only configured for a handful of libraries.

internal/librarian/java/pom_test.go
```diff
@@ -579,3 +580,24 @@ func TestIdentifyMissingModules(t *testing.T) {
 		})
 	}
 }
+
+func TestIdentifyMissingModules_SkipPOMUpdates(t *testing.T) {
```
sofisl, line 584:
```suggestion
func TestIdentifyMissingModules_SkipPOMUpdates(t *testing.T) {
	library := &config.Library{
		Name:    "secretmanager",
		Version: "1.2.3",
		APIs: []*config.API{
			{Path: "google/cloud/secretmanager/v1"},
		},
		Java: &config.JavaModule{
			SkipPOMUpdates: true,
		},
	}
	tmpDir := t.TempDir()
	got, err := IdentifyMissingModules(library, tmpDir, &sources.Sources{Googleapis: "invalid-dir"})
	if err != nil {
		t.Fatal(err)
	}
	var want []string
	if diff := cmp.Diff(want, got); diff != "" {
		t.Errorf("IdentifyMissingModules() mismatch (-want +got):\n%s", diff)
	}
}
```

this will ensure consistency in using cmp.diff which the rest of the test file uses

internal/librarian/java/pom.go
```diff
@@ -245,8 +249,10 @@ func syncPOMs(library *config.Library, libraryDir, monorepoVersion string, metad
...
+	primaryDir := srcCfg.Root(srcCfg.ActiveRoots[0])
```
noahdietz, line 254: Nit: Just to be safe, are we sure `srcCfg.ActiveRoots` will be non-`nil` and `len >= 1`?

noahdietz (approved):

internal/librarian/java/pom.go
zhumin8: Yes, `sources.NewSourceConfig` makes sure there is at least 1 active root:
https://github.com/googleapis/librarian/blob/4520acc9ca77adeb92fda560554dddb25cd967ed/internal/sources/sources.go#L47-L49

Issue #5894, opened 2026-05-07
preview: legacylibrarian release preview support


noahdietz:
Python preview generation puts Preview clients in the top-level `./preview-packages/` directory with an identical name as its stable counterpart. Go preview generation puts Preview clients in `./preview/internal` directory with an identical name as its stable coutnerpart.

We'd like to be able to release a preview client, as identified by suffixed name e.g. `{package}-preview`. We may need a `state.yaml` entry with this Preview suffixed ID.

This will require changes in legacylibrarian release stage and potentially in publish.

noahdietz:
We are going to move back to release-please support so closing this.

Python had one PR open to finish legacylibrarian support, but I've closed it.If we need to, we can reopen easily.

Issue #4866, opened 2026-03-26
previews(python): generate preview clients into preview directory


noahdietz:
When a library entry has a `preview` subsection (#4863), force the output directory to be `packages/{api-pkg-kebab}-preview` OR consider a separate subdirectory `preview-packages/{api-pkg-kebab}`.

Issue #4865, opened 2026-03-26
previews(go): generate preview clients into internal preview subdirectory


noahdietz:
When a library entry has a `preview` subsection (#4863), force the output directory to be `internal/generated/preview` and disable snippet generation.

PR #5963, opened 2026-05-11
feat(librarian): add preview entry to .librarian/state.yaml for releasing

noahdietz:
Since #4866, we add & generate Preview Python libraries to the `preview-packages/` subdirectory, and since #4865 we add & generate Preview Go lirbaries to `preview/internal/`. In order for us to be able to release a Preview library (at least while still using `legacylibrarian`) we also need to ensure there is a `.librarian/state.yaml` entry for the preview library when `librarian add preview/...` is run. Preview libraries are addressable by the `{name}-preview` form, so the same is used for the `ID`.

Subsequent changes may come for the `release_stage.go` code to interpret this as necessary.

Updates #5894


internal/librarian/add.go
```diff
@@ -338,3 +355,35 @@ func createLegacyLibrary(language string, lib *config.Library) *legacyconfig.Lib
...
+	}
+	legacyLib := &legacyconfig.LibraryState{
+		ID:      stableName + "-preview",
+		Version: previewLib.Version,
+		APIs:    libAPIs,
+	}
+	switch language {
+	case config.LanguageGo:
+		legacyLib.SourceRoots = []string{
+			fmt.Sprintf("preview/internal/%s", stableName),
```
jskeet, line 372: Will we actually get here in Go at the moment? We're expecting the preview to be on a different branch for now, right?

jskeet: (This is fine in terms of where we expect it to end later, of course. Not suggesting anything is changed, just checking my understanding.)

jskeet (approved):

internal/librarian/add.go
noahdietz: Yes we will, all code will exist on `main` and existing `librarian generate` impl for go will emit to `preview/internal/{module}`

noahdietz:
@jskeet re-requested review bc I had a small code change to what tag format was specified in state.yaml for previews

jskeet (approved):
Yup, the change for hard-coding the ID makes sense. Apologies for not spotting that before.

noahdietz:
> Yup, the change for hard-coding the ID makes sense. Apologies for not spotting that before.

All good! I was running through my playbook and noticed the potential discrepancy :)

PR #5976, opened 2026-05-12
refactor(internal/config): embed JavaAPI config directly under API entry

### LINKED ISSUE #5974: java: remove redundant path duplication between apis and java.java_apis
Similar to go.go_apis issue https://github.com/googleapis/librarian/issues/5471, we should restructure the config to move these under each API entry to eliminate duplication and prevent drift.

This technically can be done after migrate, but it is easier before migrate.

zhumin8:
The config is restructured to embed Java-specific API settings directly inside each API entry. Previously, the configuration required duplicating the same API path in both the general APIs list and the Java-specific APIs list. Moving the Java configuration under the API struct eliminates this duplication and prevents configuration drift.

Additionally, the configuration tidy logic now automatically cleans up any empty Java configurations to maintain a minimal file output.

Fix https://github.com/googleapis/librarian/issues/5974


internal/librarian/java/defaults.go
```diff
@@ -51,9 +53,8 @@ func Fill(library *config.Library) (*config.Library, error) {
 		if javaAPI.GenerateResourceNames == nil {
 			javaAPI.GenerateResourceNames = new(true)
 		}
-		javaAPIs = append(javaAPIs, javaAPI)
+		api.Java = javaAPI
```
noahdietz: Nit: Is this reassignment necessary? If `api.Java` is a `*config.JavaAPI`, then the `javaAPI := api.Java` is already causing the subsequent code to edit the pointer directly.

internal/librarian/java/defaults.go
```diff
@@ -63,25 +64,48 @@ func Tidy(library *config.Library) *config.Library {
 	if library.Output == deriveOutput(library.Name) {
 		library.Output = ""
 	}
-	if library.Java != nil {
-		for _, javaAPI := range library.Java.JavaAPIs {
-			if javaAPI.Samples != nil && *javaAPI.Samples {
-				javaAPI.Samples = nil
+	for _, api := range library.APIs {
+		if api.Java != nil {
```
noahdietz: Nit: I would invert this to be

```go
if api.Java == nil {
  continue
}
```

That will reduce indentation of all contained code, making it more readable.

Not sure how that will interact with the `isEmptyJavaAPI` check though

internal/librarian/java/defaults.go
```diff
@@ -97,22 +121,21 @@ var (
...
-			continue
-		}
-		for _, p := range javaAPI.AdditionalProtos {
-			if p == commonResourcesProto {
-				return fmt.Errorf("%s: %w", javaAPI.Path, ErrOmitCommonResourcesConflict)
+	for _, api := range library.APIs {
+		if api.Java != nil {
+			if !api.Java.OmitCommonResources {
+				continue
+			}
```
noahdietz: Nit, similar thing here re: inverting nil check and de-denging nested logic, could we do:

```go
if api.Java == nil || !api.java.OmitCommonResources {
  continue
}

for _, p := range api.Java.AdditionalProtos { ...
```

internal/librarian/java/generate.go
```diff
@@ -369,24 +369,6 @@ func collectJavaFiles(root string) ([]string, error) {
 	return files, err
 }
 
-// ResolveJavaAPI returns the Java-specific configuration for the given API.
-// TODO(https://github.com/googleapis/librarian/issues/5050):
```
noahdietz, line 373: Can this bug be closed now?

noahdietz (approved):

internal/librarian/java/generate.go
zhumin8: No. It's also referenced here: https://github.com/googleapis/librarian/blob/main/internal/librarian/java/maven_coordinate.go#L38

internal/librarian/java/defaults.go
zhumin8: Good catch. It's redundant now.

Issue #5970, opened 2026-05-12
nodejs: Consider stopping inferring API paths from package names


jskeet:
Inferring API paths from package names does not work at the moment - the package name never contains the variant (v1, v1beta etc) whereas the API path pretty much always should.

If we stop inferring the API paths (in the same way as Go and Python) then we don't need to "know" a pattern for handwritten library output directories - we'll just be able to see whether the are any API paths.

(I'm happy to make the change; assigned to @quirogas to check the plan first.)

quirogas:
This undesired behavior is triggered when we attempt to generate handwritten libraries. For example, adding `google-cloud-dns` to `librarian.yaml` and then running librarian generate google-cloud-dns results in the following output:

```console
librarian generate google-cloud-dns
YYYY/MM/DD HH:MM:SS librarian: generate library "google-cloud-dns" (nodejs): failed to generate api "google/cloud/dns": no protos found in api "google/cloud/dns"
```

#### librarian.yaml (snippet)
```yaml
# ...
libraries:
  # TODO(https://github.com/googleapis/librarian/issues/5953): investigate why google-cloud-dns get's added and determine if it should be deleted
  - name: google-cloud-dns
    version: 5.3.2
    copyright_year: "2014"
    keep:
      - .gitignore
      - .jsdoc.js
      - .mocharc.js
      - .nycrc
      - .repo-metadata.json
      - CHANGELOG.md
      - README.md
      - linkinator.config.json
      - package.json
      - tsconfig.json
      - samples/.eslintrc.yml
      - samples/README.md
      - samples/package.json
      - samples/quickstart.js
      - samples/test/quickstart.test.js
      - system-test/data/zonefile.zone
      - system-test/dns.ts
    nodejs:
      dependencies:
        "@google-cloud/common": ^6.0.0
        "@google-cloud/paginator": ^6.0.0
        "@google-cloud/promisify": ^5.0.0
        arrify: ^2.0.0
        dns-zonefile: 0.2.10
        lodash.groupby: ^4.6.0
        string-format-obj: ^1.1.1
# ...
```

PR #5980, opened 2026-05-13
feat(internal/librarian/nodejs): stop inferring API paths from package names

quirogas:
This pull request refactors the Node.js migration engine in `tool/cmd/migrate/nodejs.go` to stop inferring API paths from npm package names.

### Changes
- **Eliminated Package Name Inference**: `parseOwlBotAPIPaths` no longer attempts to match npm package names between `package.json` and Bazel `package_name` attributes.
- **Metadata-Driven Discovery**: Discovers explicit API version paths by inspecting `.OwlBot.yaml` `api-name` metadata and walking local version directories for valid `nodejs_gapic_library` rules.
- **Handwritten Library Skipping**: Skips purely handwritten packages during migration runs, omitting them from `librarian.yaml`.

Updates #5970


jskeet:
Initial response, when I haven't actually read the code yet - this isn't the aspect of inference I was referring to in #5970. As far as I've seen so far, the results of migration look good... what's problematic is that when there are no API paths present in librarian.yaml, at *generation* time Librarian decides to infer an API path. So google-cloud-dns is generated *as if* it has an API path of google/cloud/dns, even though that doesn't exist and *isn't* in librarian.yaml. That was the inference I was proposing to remove.

It's entirely possible that the code in this PR should *also* be included - but we probably ought to talk about it, particularly the omission of handwritten libraries from librarian.yaml. (Eventually, they should be present - we may want to omit them from the initial migration though, while we're using release-please.)

quirogas:
For posterity: after discussing this with @jskeet, we agreed that the issue described in #5970 will not be addressed by this PR. However, this PR prevents the issue from occurring during `librarian generate --all` calls by removing handwritten libraries from librarian.yaml. Consequently, the issue described in #5970 will resurface once we add handwritten libraries back to the file.

GautamSharda (approved):

PR #5982, opened 2026-05-13
refactor: remove Config.Release

### LINKED ISSUE #4910: config: remove release section from librarian.yaml
The release section in librarian.yaml is currently only used by google-cloud-rust and supports these fields:

```
  release:
    branch: main
    ignored_changes:
      - .repo-metadata.json
      - .sidekick.toml
    remote: upstream
    tools:
      cargo:
        - name: cargo-semver-checks
          version: 0.46.0
        - name: cargo-workspaces
          version: 0.4.0
```

https://github.com/googleapis/google-cloud-rust/blob/9b9d51e3f30db158936a46451c293692efa3fd5d/librarian.yaml#L34-L45

Each field seems like it can be removed for the following reasons:

  - **branch:** all languages will release from main, so this can be hardcoded as a default
  - **remote:** all languages use upstream, so this can be hardcoded as a default
  - **tools:** this is being moved to a separate top-level section (#4848)
  - **ignored_changes:** .repo-metadata.json should be ignored by all languages when deciding if we want to bump the version, so it should be a hardcoded default rather than per-repo config. .sidekick.toml is no longer used
  - **preinstalled:** not used
  - **roots_pem:** not used

noahdietz:
Removes the `type Release`, which was a top-level `Config` property for `librarian.yaml`, and all references to it.

Also removes the `type Tool`, which was only referenced from `Release` and is effectively replaced by language-specific `Tools` properties e.g. `CargoTool`, `GoTool`, etc.

Refactors tag, bump, and publish code accordingly. Most notable refactoring here is that we will no longer be able to define a "preinstalled" tool/binary (e.g. `git` or `cargo`). As per https://github.com/googleapis/librarian/issues/4910#issuecomment-4407139116 we are no longer using that config anyways and our playbooks have the appropriate one-time setup requirements documented. We otherwise don't have a strong use case (yet) for continuing to support the use case of pointing at a preinstall tool that isn't in the `PATH`. This may change in the future, but it wouldn't be part of the `Release` section anyways.

Furthermore, it refactors a lot of the rust publish testing code to use `t.SetEnv`-based PATH overrides with a per-test bash script for mocking installed tools rather than plumbing "command overrides" through the codebase to override using a preinstalled tool. This allows us to "just use" the binary we want to use in the code, but requires that we mock it in all of our tests (which we were already doing to some extent).

Fixes #4910

### DIFF
--- a/internal/command/command_test.go
+++ b/internal/command/command_test.go
@@ -25,7 +25,6 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
-	"github.com/googleapis/librarian/internal/config"
 )
 
 const (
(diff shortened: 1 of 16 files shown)


internal/librarian/rust/verify_cargo_tools_test.go
```diff
@@ -15,98 +15,125 @@
...
 	"github.com/googleapis/librarian/internal/testhelper"
 )
 
 func TestCargoPreFlightSuccess(t *testing.T) {
-	testhelper.RequireCommand(t, "cargo")
-	tools := []config.Tool{
+	tmpDir := t.TempDir()
+	cargoScript := filepath.Join(tmpDir, "cargo")
+	os.WriteFile(cargoScript, []byte("#!/bin/sh\nexit 0"), 0755)
+	t.Setenv("PATH", tmpDir+string(os.PathListSeparator)+os.Getenv("PATH"))
```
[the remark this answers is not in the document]
noahdietz: Good call, yes. done

internal/command/command_test.go
```diff
@@ -116,42 +115,38 @@ func TestOutput_Error(t *testing.T) {
 
 func TestGetExecutablePath(t *testing.T) {
 	tests := []struct {
```
julieqiu: nit: use `for _, test := range []struct { ... }{ ... }`

julieqiu (approved):

internal/command/command_test.go
noahdietz: Done!

PR #5983, opened 2026-05-13
fix(internal/librarian/java): collectModules should respect configs to skip certain modules

### LINKED ISSUE #5730: java: generate for iam and iam-policy

zhumin8:
Respect configs (generate_proto_grpc, generate_gapic and generate_resource_names) to skip certain module generation when collecting expected modules. `collectModules` is used to identify new modules to add to versions.txt and to update pom.xml.
context https://github.com/googleapis/librarian/issues/5730#issuecomment-4442714032

Fix #5730


internal/librarian/java/pom.go
```diff
@@ -158,18 +165,20 @@ func discoverModules(library *config.Library, libraryDir string, transports map[
 		}
 	}
 	// Client module
-	clientDir := filepath.Join(libraryDir, libCoord.GAPIC.ArtifactID)
-	isClientMissing, err := isPOMMissing(clientDir)
-	if err != nil {
-		return nil, err
+	if shouldGenerateClient {
+		clientDir := filepath.Join(libraryDir, libCoord.GAPIC.ArtifactID)
```
sofisl, line 169: is it possible this isn't a path? if so should we check that before `isPOMMissing`?

zhumin8: No, I don't think we need an extra check before `isPOMMissing`.
`libCoord.GAPIC.ArtifactID` is ensured to be non-empty, so `clientDir` is a well-formed path string.
If  clientDir  does not exist on disk,  `os.Stat` inside `isPOMMissing` evaluates to  `fs.ErrNotExist` .

sofisl (approved):

PR #5993, opened 2026-05-13
fix(swift): add pagination overloads to service protocol

chingor13:
The paginated overloads for RPCs should be included in the service's protocol functions.

Note: a protocol return type within a protocol MUST be `any` because there is no way for the compiler to infer the return type. This also means that the implementation cannot specify `some` because it does not match the function signature.


coryan (approved):
Thanks!

PR #6000, opened 2026-05-14
fix(internal/librarian/nodejs): fix library name derivation

jskeet:
The correct library name is now derived from the API name, excluding
any trailing version. Library names in Node don't include the API
variant/version - so google/cloud/functions/v1 is in
google-cloud-functions, for example. This logic is currently identical
to the Python logic, but is maintained separately so that they can
diverge easily if necessary.

### DIFF
--- a/internal/librarian/add.go
+++ b/internal/librarian/add.go
@@ -139,6 +139,8 @@ func deriveLibraryName(language string, api string) string {
 		return golang.DefaultLibraryName(api)
 	case config.LanguageJava:
 		return java.DefaultLibraryName(api)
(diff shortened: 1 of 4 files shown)


jskeet:
~~Note: only the last commit of this is new - prior commits are from #5998; I wanted to include the new function in the same file that PR introduces. Once #5998 is merged, I'll either rebase or merge so that the real diff is clear.~~

(Now rebased.)

jskeet:
@quirogas This is now ready for review, having merged the previous PR.

JoeWang1127 (approved):

PR #6016, opened 2026-05-14
refactor(internal/librarian/java): consolidate AdditionalProtos config for fine-grained generation control

### LINKED ISSUE #6001: java: filestore generation diff
```
        deleted:    java-filestore/proto-google-cloud-filestore-v1/src/main/java/com/google/cloud/common/OperationMetadata.java
        deleted:    java-filestore/proto-google-cloud-filestore-v1/src/main/java/com/google/cloud/common/OperationMetadataOrBuilder.java
        deleted:    java-filestore/proto-google-cloud-filestore-v1/src/main/java/com/google/cloud/common/OperationMetadataOuterClass.java
        deleted:    java-filestore/proto-google-cloud-filestore-v1beta1/src/main/java/com/google/cloud/common/OperationMetadata.java
        deleted:    java-filestore/proto-google-cloud-filestore-v1beta1/src/main/java/com/google/cloud/common/OperationMetadataOrBuilder.java
        deleted:    java-filestore/proto-google-cloud-filestore-v1beta1/src/main/java/com/google/cloud/common/OperationMetadataOuterClass.java
```

additional proto "google/cloud/common/operation_metadata.proto" needs to generate but not copy.

zhumin8:
Replaces `additional_protos` (string list) and `additional_protos_to_generate_and_copy` in the Java API configuration with a single structured `additional_protos` field. Each entry now accepts boolean flags (`generate_proto_classes` and `copy_to_output`) to decouple standard Protocol Buffer class generation from source file copying.

This structured approach natively supports a new variation needed for filestore case: generating proto classes without copying the source `.proto` files to the output repository.

Also updated migrate tool hardcoded value for "google/cloud/filestore".

Fixes #6001

### DIFF
--- a/internal/librarian/java/defaults.go
+++ b/internal/librarian/java/defaults.go
@@ -92,7 +92,6 @@ func isEmptyJavaAPI(j *config.JavaAPI) bool {
 	}
 	return !j.Monolithic &&
 		len(j.AdditionalProtos) == 0 &&
-		len(j.AdditionalProtosToGenerateAndCopy) == 0 &&
 		!j.OmitCommonResources &&
 		len(j.ExcludedProtos) == 0 &&
 		len(j.SkipProtoClassGeneration) == 0 &&
@@ -132,7 +131,7 @@ func Validate(library *config.Library) error {
 			continue
 		}
 		for _, p := range api.Java.AdditionalProtos {
-			if p == commonResourcesProto {
+			if p != nil && p.Path == commonResourcesProto {
 				return fmt.Errorf("%s: %w", api.Path, ErrOmitCommonResourcesConflict)
 			}
 		}
--- a/internal/librarian/java/defaults.go
+++ b/internal/librarian/java/defaults.go
@@ -16,6 +16,7 @@ package java
 
 import (
 	"fmt"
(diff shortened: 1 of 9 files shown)


internal/librarian/java/defaults.go
```diff
@@ -131,8 +130,8 @@ func Validate(library *config.Library) error {
 		if api.Java == nil || !api.Java.OmitCommonResources {
 			continue
 		}
-		for _, p := range api.Java.AdditionalProtos {
-			if p == commonResourcesProto {
+		for _, proto := range api.Java.AdditionalProtos {
+			if proto != nil && proto.Path == commonResourcesProto {
```
sofisl, line 138: do we need to check other cases for proto (i.e., if proto.Path === "") or something like that?

sofisl:
fix should be fixes in PR description

internal/librarian/java/defaults.go
zhumin8: Good point. I updated tidy to filter out empty AdditionalProtos, we don't need additional checks here.

sofisl (approved):

PR #6018, opened 2026-05-14
fix(legacylibrarian): sync preview versions

noahdietz:
If a Library has a Preview client, make sure its version is synced as well.

Preview clients will have state.yaml entries upon add as of #5963.

Towards #5894

### DIFF
--- a/internal/legacylibrarian/legacylibrarian/release_stage.go
+++ b/internal/legacylibrarian/legacylibrarian/release_stage.go
@@ -404,6 +404,9 @@ func syncVersion(state *legacyconfig.LibrarianState, cfg *config.Config) (*confi
 	return cfg, nil
 }
 
+// syncLibraryVersion checks that the verison in librarian.yaml is not greater than the same in
+// .librarian/state.yaml, which would imply version drift. Once that check is complete, it sets the library's
+// version in librarian.yaml to match that of .librarian/state.yaml, which is the current source of truth.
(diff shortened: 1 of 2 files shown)


internal/legacylibrarian/legacylibrarian/release_stage.go
```diff
@@ -388,17 +388,30 @@ func (r *stageRunner) updateLibrarianYAML(ctx context.Context) error {
...
+			if err != nil {
+				return nil, err
+			}
 		}
-		lib.Version = legacyLibrary.Version
 	}
 	return cfg, nil
 }
+
+func syncLibraryVersion(name string, state *legacyconfig.LibrarianState, currentVersion string) (string, error) {
```
jskeet, line 410: Could you add a comment here? The name suggests an "action" (mutating state) but it looks like it's more of a "computation" and I'm not quite sure what the intention is. (After reading it through more, I think I understand, but it took longer than I'd like.)

jskeet (approved):
One suggestion for a comment, but otherwise LGTM.

internal/legacylibrarian/legacylibrarian/release_stage.go
noahdietz: Yep! done!

noahdietz: I just moved this code around, but it basically protects against version drift (i.e. librarian.yaml got bumped but state.yaml didn't), and if all checks out, then forces librarian.yaml version to match that of state.yaml

PR #6021, opened 2026-05-15
feat(internal/librarian/nodejs): remove synthtool

quirogas:
`google-cloud-node` removed `synthtool` entirely, which simplifies the container environment, reduces build dependencies, and eliminates unnecessary execution overhead. This change removes all `synthtool` code from `generate`.


jskeet (approved):
I'll follow up with a PR to remove the pip-handling code from install.go.

PR #6022, opened 2026-05-15
chore(internal/librarian/nodejs): remove pip installation code

jskeet:
We no longer use Python at all during Node generation, so we don't need to handle Tools.Pip in the Install function.


quirogas (approved):

jskeet:
Currently building a Docker image locally, then will run generation, and merge if that's successful.

PR #6029, opened 2026-05-15
feat(internal/config): add default_enabled_generator_features to Go configuration

suztomo:
This change introduces default_enabled_generator_features in librarian.yaml's default.go section and updates Librarian to append these defaults to APIs' `go.enabled_generator_features` list.

It implements a `union` helper to merge default generator features with API-specific ones while de-duplicating elements and preserving their original insertion order. This custom helper is necessary as the Go standard library does not provide a built-in order-preserving de-duplication helper for slices. Preserving the order of generator features is important because it ensures the generated `protoc` command line remains consistent across runs.

The companion pull request is https://github.com/googleapis/google-cloud-go/pull/14600. The librarian change needs to happen first. After that, I'll update the librarian version in `librarian.yaml` in google-cloud-go with a pseudo version with the default value added there.

Fixes https://github.com/googleapis/librarian/issues/5246


internal/librarian/library.go
```diff
@@ -54,6 +57,22 @@ func fillDefaults(lib *config.Library, d *config.Default) *config.Library {
...
+// fillGo populates empty Go-specific fields in lib from the provided default.
+func fillGo(lib *config.Library, d *config.Default) *config.Library {
+	if lib.Go == nil {
+		lib.Go = &config.GoModule{}
+	}
+	for _, api := range lib.APIs {
+		if api.Go == nil {
+			api.Go = &config.GoAPI{}
+		}
+		if len(api.Go.EnabledGeneratorFeatures) == 0 {
```
suztomo: This should be union.

suztomo: Fixed.

internal/librarian/library.go
```diff
@@ -54,6 +57,22 @@ func fillDefaults(lib *config.Library, d *config.Default) *config.Library {
...
+			api.Go.EnabledGeneratorFeatures = d.Go.DefaultEnabledGeneratorFeatures
+		}
+	}
```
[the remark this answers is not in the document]
suztomo: Fixed.

internal/librarian/library_test.go
```diff
@@ -502,6 +502,104 @@ func TestFillDefaults_Python(t *testing.T) {
...
+						Go: &config.GoAPI{
+							ImportPath:               "foo/apiv1",
+							EnabledGeneratorFeatures: []string{"F_one", "F_two"},
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "existing features union-ed with defaults",
```
JoeWang1127, line 563: Could you add a test to verify the dedupliation?

internal/librarian/library.go
```diff
@@ -54,6 +57,38 @@ func fillDefaults(lib *config.Library, d *config.Default) *config.Library {
 	return lib
 }
 
+// fillGo populates empty Go-specific fields in lib from the provided default.
+func fillGo(lib *config.Library, d *config.Default) *config.Library {
+	for _, api := range lib.APIs {
+		if api.Go == nil {
+			api.Go = &config.GoAPI{}
+		}
+		api.Go.EnabledGeneratorFeatures = union(api.Go.EnabledGeneratorFeatures, d.Go.DefaultEnabledGeneratorFeatures)
```
JoeWang1127, line 69: Is there a case that `d.Go` is nil?

.github/workflows/go.yaml
```diff
@@ -28,7 +28,7 @@ jobs:
         run: go run ./tool/cmd/coverage ./internal/librarian/golang
   integration:
     runs-on: ubuntu-24.04
-    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+    if: false # Temporarily disable google-cloud-go test until https://github.com/googleapis/google-cloud-go/pull/14600 is merged
```
JoeWang1127: Why do you need to disable the test?

I think you can 1) merge this pull request 2) update librarian version in google-cloud-go 3) modify config in librarian.yaml.

However, default section needs to allow nil GoModule.

internal/librarian/library.go
suztomo: Added a if-statement:

```
	if d == nil || d.Go == nil {
		return lib
```

internal/librarian/library_test.go
suztomo: Added a case "duplicates are de-duplicated in union. The defaults come at the end."

.github/workflows/go.yaml
suztomo: That's right, because this adding the `default_enabled_generator_features` field is not a breaking change. I reverted this.

suztomo (commented):
@JoeWang1127 PTAL

JoeWang1127 (approved):
Could you add a test to verify nil GoDefault?

JoeWang1127:
Could you update the pull request description?

suztomo:
Updated the description:

- The defaults are appended to the `go.enabled_generator_features`.
- The integration test with google-cloud-go is not skipped.

Issue #6012, opened 2026-05-14
java: investigate java firestore diffs


sofisl:
```
(.venv) sofialeon@sofialeon:~/google-cloud-java$ git status
On branch checkDifferences
Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        deleted:    java-firestore/google-cloud-firestore/src/main/resources/META-INF/native-image/com.google.cloud/google-cloud-firestore/reflect-config.json
        modified:   java-firestore/google-cloud-firestore/src/test-jdk17/java/com/google/cloud/firestore/RecordDocumentReferenceTest.java
        modified:   java-firestore/google-cloud-firestore/src/test-jdk17/java/com/google/cloud/firestore/RecordMapperTest.java
        modified:   java-firestore/google-cloud-firestore/src/test-jdk17/java/com/google/cloud/firestore/RecordTestHelper.java
        deleted:    java-firestore/google-cloud-firestore/src/test/resources/META-INF/native-image/reflect-config.json
        deleted:    java-firestore/proto-google-cloud-firestore-admin-v1/src/main/java/com/google/firestore/admin/v1/ParentName.java
        deleted:    java-firestore/proto-google-cloud-firestore-v1/src/main/java/com/google/firestore/v1/AnyPathName.java
        deleted:    java-firestore/proto-google-cloud-firestore-v1/src/main/java/com/google/firestore/v1/DatabaseRootName.java
        deleted:    java-firestore/proto-google-cloud-firestore-v1/src/main/java/com/google/firestore/v1/DocumentPathName.java
        deleted:    java-firestore/proto-google-cloud-firestore-v1/src/main/java/com/google/firestore/v1/DocumentRootName.java
        deleted:    java-firestore/proto-google-cloud-firestore-v1/src/main/proto/firestore/BUILD.bazel
        deleted:    java-firestore/proto-google-cloud-firestore-v1/src/main/proto/firestore/bundle.proto
        modified:   java-firestore/samples/snippets/src/main/java/com/example/firestore/Quickstart.java

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        librarian.yaml

no changes added to commit (use "git add" and/or "git commit -a")
```

sofisl:
@zhumin8 The files deleted fall into two classes:
1. legacy resource name helper classes (like DocumentRootName.java) which are not automatically generated by the java generator
[...]

PR #6036, opened 2026-05-15
chore(tool/cmd/migrate): add firestore native-image configurations to java keep overrides

sofisl:
Add the paths for main and test reflect-config.json files in google-cloud-firestore to the hardcoded keepOverrides list for Java libraries. This ensures these GraalVM native-image configuration files are preserved when running the migrate tool.

For #6012

### DIFF
--- a/tool/cmd/migrate/java_module.go
+++ b/tool/cmd/migrate/java_module.go
@@ -77,6 +77,10 @@ var (
 			"gapic-showcase/src/test",
 			"gapic-showcase/src/main/java/com/google/showcase/v1beta1/Version.java",
 		},
+		"firestore": {
+			"google-cloud-firestore/src/main/resources/META-INF/native-image/com.google.cloud/google-cloud-firestore/reflect-config.json",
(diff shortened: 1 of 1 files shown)


sofisl:
@zhumin8, Is keepOverride intentionally supposed to overrwrite whatever is in the "keep" section of librarian.yaml? right now it's just overwriting, so when i run migrate with this PR, it erases my changes in https://github.com/googleapis/librarian/pull/6017. just wanting to make sure IIUC, should I:

a) just manually list all of the hardcoded keep files in keepOverrides, essentially nullifying #6017, 
b) fix this bug and "append" keepOverrides rather than actually overwrite the keep section of librarian.yaml,
c) something else.

zhumin8 (approved):
Maybe do a separate list that appends to migrated keep?
`showcase`, and `aiplatform` works fine with only these listed, I don't think we want to bring in unnecessary keep items.

approving anyway in case you want to merge this in (no harm to other libraries).

PR #6042, opened 2026-05-18
feat(internal/librarian/nodejs): perform selective cleaning

jskeet:
Only files which are expected to be generated are deleted. These can still be preserved using the keep list.


internal/librarian/nodejs/clean.go
```diff
@@ -0,0 +1,120 @@
...
+	for _, file := range simpleFilesToClean {
+		if slices.Contains(lib.Keep, file) {
+			continue
+		}
+		if err := os.Remove(filepath.Join(lib.Output, file)); err != nil && !errors.Is(err, fs.ErrNotExist) {
+			return err
+		}
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
jskeet: Done, although it seems unlikely to be a performance bottleneck.

internal/librarian/nodejs/clean.go
```diff
@@ -0,0 +1,120 @@
...
+		}
+		if d.IsDir() || !strings.HasSuffix(d.Name(), suffix) {
+			return nil
+		}
+		if marker != "" {
+			content, err := os.ReadFile(path)
+			if err != nil {
+				return err
+			}
+			if !strings.Contains(string(content), marker) {
```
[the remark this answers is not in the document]
jskeet: Done.

internal/librarian/nodejs/clean.go
```diff
@@ -0,0 +1,120 @@
...
+		return err
+	}
+	for _, sourceDirectory := range sourceDirectoriesToClean {
+		if err := cleanFiles(lib, sourceDirectory, ".js", autogeneratedMarker); err != nil {
+			return err
+		}
+		if err := cleanFiles(lib, sourceDirectory, ".ts", autogeneratedMarker); err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
jskeet: While this is true, I'd far rather keep this simple.

internal/librarian/nodejs/clean.go
```diff
@@ -0,0 +1,122 @@
...
+		"protos/protos.json",
+		"samples/package.json",
+		"package.json",
+		"tsconfig.json",
+		"webpack.config.js",
+		".eslintignore",
+		".gitignore",
+		".nycrc",
+		"README.md",
+		".OwlBot.yaml",
```
JoeWang1127, line 45: Do we need to add a TODO to remove this line after migration?

internal/librarian/nodejs/clean.go
```diff
@@ -0,0 +1,122 @@
...
+			continue
+		}
+		if err := os.Remove(filepath.Join(lib.Output, file)); err != nil && !errors.Is(err, fs.ErrNotExist) {
+			return err
+		}
+	}
+	return nil
+}
+
+func cleanFiles(lib *config.Library, keepSet map[string]bool, subDir, suffix string, marker []byte) error {
```
JoeWang1127, line 97: Could you add a comment to this function?

JoeWang1127, line 97: Why not using `dir` in the parameter list? I think the subDir is confusing, we should construct the directory before calling this function.

internal/librarian/nodejs/clean_test.go
```diff
@@ -0,0 +1,138 @@
...
+		got := err != nil
+		want := slices.Contains(wantDeleted, file)
+		if got != want {
+			t.Errorf("file %s deleted: got %t, want %t", file, got, want)
+		}
+	}
+}
+
+func createFileAndDirectories(t *testing.T, path, content string) {
+	if err := os.MkdirAll(filepath.Dir(path), 0755); err != nil {
```
JoeWang1127, line 133: Add `t.Helper()`.

internal/librarian/nodejs/clean_test.go
```diff
@@ -0,0 +1,138 @@
...
+				Output: filepath.Join(repoRoot, "packages", "test"),
+				Keep:   test.keep,
+			}
+			for _, file := range test.setupFiles {
+				createFileAndDirectories(t, filepath.Join(lib.Output, file), test.contentMap[file])
+			}
+			if err := Clean(lib); err != nil {
+				t.Fatal(err)
+			}
+			verifyFileDeletions(t, lib.Output, test.setupFiles, test.wantDeleted)
```
JoeWang1127, line 110: Why not writing the verification logic directly? This is the only place using this function.

jskeet:
Will discuss all the comments with @quirogas.

internal/librarian/nodejs/clean.go
jskeet: Done.

internal/librarian/nodejs/clean.go
jskeet: That would require calling filepath.join in each of the call sites, which I think would be more confusing/ugly.

jskeet: Done.

internal/librarian/nodejs/clean_test.go
jskeet: It's only used here right now, but I can definitely imagine it being used more in later tests. (These functions were copied from the Python tests...) I'd prefer to keep them separate.

internal/librarian/nodejs/clean_test.go
jskeet: Done.

internal/librarian/nodejs/clean.go
```diff
@@ -0,0 +1,129 @@
...
+	// Files to delete without any checking, relative to the package directory.
+	simpleFilesToClean = []string{
+		"protos/protos.d.ts",
+		"protos/protos.js",
+		"protos/protos.json",
+		"samples/package.json",
+		"package.json",
+		"tsconfig.json",
+		"webpack.config.js",
+		".eslintignore",
```
quirogas, line 41: Nit: We could potentially bundle some of these other files into https://github.com/googleapis/librarian/issues/6049, as they are no longer generated by the generator and will be cleaned up in the future.

quirogas (approved):

internal/librarian/nodejs/clean.go
jskeet: Have made the issue broader, will move the comment up and change it slightly.

PR #6064, opened 2026-05-18
Revert "feat(tool/cmd/migrate): add keep list appends support for Java modules"

sofisl:
Reverts googleapis/librarian#6062

Fixes #6063


JoeWang1127 (approved):

suztomo:
The error is on TestTagCommand but the reverted test case seems to be in TestBuildConfig.

PR #6069, opened 2026-05-18
feat(internal/librarian): java add to resolve mixin protos

### LINKED ISSUE #6034: java: add to populate additonal_protos
`additional_protos` are some times needed for mixins. 
Prior to librarian, Java hermetic build infers it from BUILD.bazel at runtime. After migrate to librarian.yaml, as no BUILD.bazel is read, this value needs to be added for new libraries at time of onboarding by hand.
This can be interruptive to on-call person. It would be smoother if this can be automatically populated.

Rust already do the similar, information is inferred directly from the protos. We should be able to re-use logic that are language agnostic.

zhumin8:
When running librarian add, it now resolves mixins to be added to `additional_protos` from parsing service config apis. This eliminates the need to manually enter these values when adding new libraries or new API path.

Fixes https://github.com/googleapis/librarian/issues/6034

### DIFF
--- a/internal/librarian/add.go
+++ b/internal/librarian/add.go
@@ -110,6 +110,16 @@ func runAdd(ctx context.Context, cfg *config.Config, api string) error {
 
 func resolveDependencies(ctx context.Context, cfg *config.Config, name string) (*config.Config, error) {
 	switch cfg.Language {
+	case config.LanguageJava:
+		lib, err := FindLibrary(cfg, name)
+		if err != nil {
+			return nil, err
+		}
+		sources, err := LoadSources(ctx, cfg.Sources)
+		if err != nil {
+			return nil, err
+		}
+		return java.ResolveDependencies(ctx, cfg, lib, sources)
 	case config.LanguageRust:
 		lib, err := FindLibrary(cfg, name)
 		if err != nil {
--- a/internal/librarian/add.go
+++ b/internal/librarian/add.go
@@ -119,7 +119,7 @@ func resolveDependencies(ctx context.Context, cfg *config.Config, name string) (
 		if err != nil {
 			return nil, err
 		}
-		return java.ResolveDependencies(ctx, cfg, lib, sources)
+		return java.ResolveMixinDependencies(cfg, lib, sources)
 	case config.LanguageRust:
 		lib, err := FindLibrary(cfg, name)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


internal/librarian/add_test.go
```diff
@@ -1021,3 +1021,56 @@ func TestSyncToStateYAML_Error(t *testing.T) {
 		})
 	}
 }
+
+func TestAddLibraryCommand_Java(t *testing.T) {
+	copyrightYear := strconv.Itoa(time.Now().Year())
+	_ = copyrightYear // might not be used if it is reset to empty
```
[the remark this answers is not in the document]
zhumin8: copyrightYear not used for java. Should remove this.

internal/librarian/add.go
```diff
@@ -110,6 +110,16 @@ func runAdd(ctx context.Context, cfg *config.Config, api string) error {
 
 func resolveDependencies(ctx context.Context, cfg *config.Config, name string) (*config.Config, error) {
 	switch cfg.Language {
+	case config.LanguageJava:
```
sofisl, line 114: Java and Rust are the same. we could consolidate this into a function for DRY, or consolidate the test cases for now.

internal/librarian/java/resolve_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+func TestResolveDependencies(t *testing.T) {
+	googleapisDir, err := filepath.Abs("../../testdata/googleapis")
+	if err != nil {
+		t.Fatal(err)
+	}
+	srcs := &sources.Sources{
+		Googleapis: googleapisDir,
+	}
+	for _, tt := range []struct {
```
sofisl: please rename to full name, test.

internal/librarian/java/resolve_test.go
```diff
@@ -0,0 +1,134 @@
...
+		name    string
+		lib     *config.Library
+		want    []*config.AdditionalProto
+		wantErr bool
```
sofisl: should we remove? doesn't seem to be used anywhere, and we'd probably want to test it separately anyways

internal/librarian/add.go
zhumin8: Not exactly, but I can extract a helper for setup.

sofisl (approved):

PR #6078, opened 2026-05-19
feat(internal/librarian/java): support non-cloud GroupID inference

### LINKED ISSUE #6047: java: populate default group_id and distribution name for non-cloud API path
When adding new API path (new library), should populate default group_id / distribution name.
Currently, it requires setting up DistributionNameOverride (manual step), otherwise it is defaulting to "com.google.cloud: cloud-name".

Ideally do this before migrate, so that adding a new non-cloud API requires less manual step. May defer to after migration according to workload.

zhumin8:
Java library onboarding now automatically infers the Maven GroupID and distribution name override for recognized non-cloud APIs (Shopping, Maps, and Ads) based on their API path prefixes. This reduces manual configuration steps and ensures consistency.
Referenced existing rules recorded in [guide](https://github.com/googleapis/google-cloud-java/blob/main/generation/new_client_hermetic_build/README.md#special-case-example-google-maps).

If a new library is added with an unrecognized non-cloud API path, prints a warning and sets a fake group id. When running generate on this fake group id, returns error.

Fixes https://github.com/googleapis/librarian/issues/6047

### DIFF
--- a/internal/librarian/java/add.go
+++ b/internal/librarian/java/add.go
@@ -34,6 +34,7 @@ var knownPrefixes = []string{
 var nonCloudGroupIDs = map[string]string{
 	"google/shopping/": "com.google.shopping",
 	"google/maps/":     "com.google.maps",
+	"google/ads/":      "com.google.api-ads",
 }
 
 const defaultVersion = "0.1.0-SNAPSHOT"
--- a/internal/librarian/java/add.go
+++ b/internal/librarian/java/add.go
@@ -31,12 +31,6 @@ var knownPrefixes = []string{
 	"google/",
 }
 
-var nonCloudGroupIDs = map[string]string{
-	"google/shopping/": "com.google.shopping",
-	"google/maps/":     "com.google.maps",
-	"google/ads/":      "com.google.api-ads",
-}
-
 const defaultVersion = "0.1.0-SNAPSHOT"
 
 // Add initializes a new Java library with default values.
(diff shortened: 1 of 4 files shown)


internal/librarian/java/add.go
```diff
@@ -38,6 +45,29 @@ func Add(lib *config.Library) *config.Library {
...
+				if lib.Java == nil {
+					lib.Java = &config.JavaModule{}
+				}
+				lib.Java.GroupID = groupID
+				lib.Java.DistributionNameOverride = groupID + ":google-" + lib.Name
+				matched = true
+				break
+			}
+		}
+		if !matched && !strings.HasPrefix(apiPath, "google/cloud/") {
```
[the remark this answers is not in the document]
zhumin8: This is not the intend.

internal/librarian/java/add.go
```diff
@@ -38,6 +45,29 @@ func Add(lib *config.Library) *config.Library {
...
+			if strings.HasPrefix(apiPath, prefix) {
+				if lib.Java == nil {
+					lib.Java = &config.JavaModule{}
+				}
+				lib.Java.GroupID = groupID
+				lib.Java.DistributionNameOverride = groupID + ":google-" + lib.Name
+				matched = true
+				break
+			}
+		}
```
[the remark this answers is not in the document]
zhumin8: done

internal/librarian/java/add.go
```diff
@@ -38,6 +39,34 @@ func Add(lib *config.Library) *config.Library {
...
+	switch {
+	case strings.HasPrefix(apiPath, "google/shopping/"):
+		return setJavaConfig(lib, "com.google.shopping")
+	case strings.HasPrefix(apiPath, "google/maps/"):
+		return setJavaConfig(lib, "com.google.maps")
+	case strings.HasPrefix(apiPath, "google/ads/"):
+		return setJavaConfig(lib, "com.google.api-ads")
+	}
+	if !strings.HasPrefix(apiPath, "google/cloud/") {
+		log.Printf(
```
JoeWang1127, line 58: Waring may not enough as user can ignore it.

How about returning an error to force the user to specify the group id?

zhumin8: I choose warning intentially. This occurs when running `librarian add`, returning error means user cannot use `librarian add` and need to add the how config by hand instead. 
In my mind, the user experience is:
- user tries to add a non-cloud API that don't have a known group_id, `librarian add` adds the basic and prints out this warning
- user edit the librarian.yaml and add group_id, distribution_name
- user run `librarian generate`

I do get your concern, if an on-call person is running these command and missed the warning, he/she may accidentally generate the new client with default com.google.cloud group id.
What about in addition to the warning, we populate a fake group id. Then in generate code, return error if running into this fake id?

JoeWang1127: > What about in addition to the warning, we populate a fake group id. Then in generate code, return error if running into this fake id?

SGTM.

JoeWang1127 (approved):

PR #6083, opened 2026-05-19
feat(internal/librarian/nodejs): tidy empty NodeJS config

### LINKED ISSUE #6077: nodejs: tidy "empty" node-specific configuration
Now that we don't have the additional protoc value for `metadata`, most of our librarians in librarian.yaml include:

```yaml
nodejs: {}
```

We should add a `Tidy` function to omit configuration. Rather than looking at individual fields etc, we can just serialize the configuration and if it ends up as `{}` we'll set it to nil.

jskeet:
Tidying a NodeJS library omits the Node-specific configuration if it is effectively empty, i.e. it serializes to {}.

This change uses YAML serialization to avoid brittle field-by-field checks, which in turn means that tidy functions need to potentially return errors. (Actually provoking errors during YAML serialization is harder, so there are no tests for that case.)

The check for an empty-when-serialized value is generalized as `yaml.Empty`, which is used to simplify Java code and make it more robust.

Fixes #6077


internal/yaml/yaml.go
```diff
@@ -98,6 +98,16 @@ func Write(path string, v any) error {
...
 }
 
+// Empty returns whether the given value serializes to an empty YAML object
+// (i.e. "{}" with a line break).
+func Empty(v any) (bool, error) {
+	data, err := Marshal(v)
+	if err != nil {
+		return false, err
+	}
+	return string(data) == "{}\n", nil
```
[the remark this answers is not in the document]
jskeet: Not changing this, as we want the behavior to be absolutely consistent between how we marshal the value here and how we end up saving the value. Our configurations aren't huge, and tidy isn't particularly performance-sensitive.

internal/librarian/tidy.go
```diff
@@ -230,12 +238,12 @@ func deleteEmptyRustModules(modules []*config.RustModule) []*config.RustModule {
 	return slices.DeleteFunc(modules, isEmptyRustModule)
 }
 
-func tidyRustConfig(lib *config.Library) *config.Library {
+func tidyRustConfig(lib *config.Library) (*config.Library, error) {
```
jskeet: Not sure how to fix the linter error here. We need it to have this signature to use it in `languageTidiers` - unless we were to add an anonymous function in there.

Alternatively, we could potentially move this to a `rust.Tidy` function - that doesn't seem to trip up the linter, and it would be more consistent. Thoughts welcome.

jskeet: Now moved to rust.Tidy.

zhumin8 (commented):
LGTM from java changes. No behavior change and improvements in mantainance.

quirogas (approved):

PR #6086, opened 2026-05-19
fix(internal/librarian/java): respect partial generation flags in clean

### LINKED ISSUE #5430: java: check if clean logic correctly supports proto-only case
> Do we need to add this pattern if the api is proto-only?
> 
> Can we move this line to the for loop in line 60? 

 _Originally posted by @JoeWang1127 in [#5418](https://github.com/googleapis/librarian/pull/5418/changes#r3094230939)_

zhumin8:
The clean logic in the Java librarian now respects partial generation flags (GenerateProto, GenerateGRPC, and GenerateGAPIC) configured for APIs. 
Previously, the clean step unconditionally targeted all modules for cleaning. Although in reality, we don't have cases that non-generated module exists, so no generation diff caused by this. But in theory this could lead to unexpected deletion of directories, or redundant cleaning attempts for disabled modules in configurations.

Fixes #5430

### DIFF
--- a/internal/librarian/java/clean.go
+++ b/internal/librarian/java/clean.go
@@ -80,13 +80,13 @@ func cleanPatterns(library *config.Library) map[string]bool {
 		javaAPI := api.Java
 		version := filepath.Base(api.Path)
 		apiCoordinates := DeriveAPICoordinates(libraryCoordinates, version, javaAPI)
-		if apiCoordinates.Proto.ArtifactID != "" {
+		if apiCoordinates.Proto.ArtifactID != "" && shouldGenerateProto(javaAPI) {
 			patterns[filepath.Join(apiCoordinates.Proto.ArtifactID, "src")] = false
 		}
-		if apiCoordinates.GRPC.ArtifactID != "" {
+		if apiCoordinates.GRPC.ArtifactID != "" && shouldGenerateGRPC(javaAPI) {
 			patterns[filepath.Join(apiCoordinates.GRPC.ArtifactID, "src")] = false
 		}
-		if apiCoordinates.GAPIC.ArtifactID != "" {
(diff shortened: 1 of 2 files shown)


sofisl (approved):

internal/librarian/java/clean.go
```diff
@@ -80,13 +80,13 @@ func cleanPatterns(library *config.Library) map[string]bool {
...
+		if apiCoordinates.Proto.ArtifactID != "" && shouldGenerateProto(javaAPI) {
 			patterns[filepath.Join(apiCoordinates.Proto.ArtifactID, "src")] = false
 		}
-		if apiCoordinates.GRPC.ArtifactID != "" {
+		if apiCoordinates.GRPC.ArtifactID != "" && shouldGenerateGRPC(javaAPI) {
 			patterns[filepath.Join(apiCoordinates.GRPC.ArtifactID, "src")] = false
 		}
-		if apiCoordinates.GAPIC.ArtifactID != "" {
+		if apiCoordinates.GAPIC.ArtifactID != "" && shouldGenerateGAPIC(javaAPI) {
 			patterns[filepath.Join(apiCoordinates.GAPIC.ArtifactID, "src")] = true
```
JoeWang1127, line 97: Why set to `true` while the other two set to `false`? The conditions seem similar.

internal/librarian/java/clean.go
JoeWang1127, line 80: Could you add comment for this function? What does the return value mean?

zhumin8: done.

internal/librarian/java/clean.go
zhumin8: See comment added. `true` means use marker-based cleaning

JoeWang1127 (approved):

PR #6073, opened 2026-05-18
feat(tool/cmd/migrate): add keep list appends support for Java modules

sofisl:
A keepAppends map is added to java_module.go which permits appending specific files to a library's keep list, instead of completely overriding it. For the firestore library, the hardcoded native-image configuration files are added to this map and appended to the final keep list during migration.

I checked, these files are static (created [here](https://github.com/googleapis/java-firestore/pull/1419) and [here](https://github.com/googleapis/java-firestore/pull/1728/) so they are OK to just hardcode (they are not being "updated" anywhere)

Third PR since this was hit by two flakes (#6058 and #6071) which I have attempted to solve those issues separately.

For https://github.com/googleapis/librarian/issues/6012

PR #6093, opened 2026-05-19
chore: disable grpc generation by default

### LINKED ISSUE #6090: java: generate accesscontextmanager attempts to update pom.xml files
The generation fails with
```
╰─$ librarian generate accesscontextmanager
2026/05/19 16:47:15 librarian: generate library "accesscontextmanager" (java): failed to generate or update pom.xml files: failed to update client pom.xml google-identity-accesscontextmanager: missing start marker "<!-- {x-generated-proto-dependencies-start} -->"
```

Context: the error is complaining about a marker we did not add (running  `migrate -insert-markers .` will add the markers). The real issue here is that it is attempting to update client pom.xml. 
Likely related to https://github.com/googleapis/librarian/pull/6060. Perhaps need to adjust config following that change.

sofisl:
Disables gapic generation by default if `ProtoGRPCOnly` is true unless it is `common-protos` or `iam`. This makes it so that librarian generate does not try to create pom.xml files for these non-libraries.

For #6079
Fixes #6090


zhumin8 (approved):
nit: can you fix the PR description now that #6073 is already merged? 
And I think this fixes https://github.com/googleapis/librarian/issues/6090

PR #6105, opened 2026-05-20
feat(sidekick/rust): propagate LRO telemetry attributes via task-local context

haphungw:
Extract LRO attempt counts and done/logical error status details ambiently from `tokio::task-local!` context, recording them directly onto active T3/polling spans without polluting handwritten traits or generated client interfaces.

Also add logic for polling span detection for both aip151 and discovery api.

### DIFF
--- a/internal/sidekick/rust/annotate_method.go
+++ b/internal/sidekick/rust/annotate_method.go
@@ -44,7 +44,7 @@ type methodAnnotation struct {
 	HasResourceNameGeneration bool
 	ResourceNameTemplateGrpc  string
 	GrpcResourceNameArgs      []string
-	IsLongRunningPoll         bool
+	IsLroPoller               bool
 }
 
 // HasGrpcResourceNameArgs returns true if the method has gRPC resource name arguments.
@@ -312,7 +312,7 @@ func (c *codec) annotateMethod(m *api.Method) (*methodAnnotation, error) {
 		RoutingRequired:           c.routingRequired,
 		DetailedTracingAttributes: c.detailedTracingAttributes,
 		InternalBuilders:          c.internalBuilders,
-		IsLongRunningPoll:         isLongRunningPoll(m),
+		IsLroPoller:               isLongRunningPoll(m),
 	}
 
 	if err := c.annotateResourceNameGeneration(m, annotation); err != nil {
@@ -630,7 +630,5 @@ func isIdempotent(p *api.PathInfo) string {
 }
 
 func isLongRunningPoll(m *api.Method) bool {
-	return m.SourceServiceID == ".google.longrunning.Operations" &&
-		m.Name == "GetOperation" &&
-		m.Service.Package != "google.longrunning"
+	return m.IsLroPoller
 }
--- a/internal/sidekick/rust/annotate_method.go
+++ b/internal/sidekick/rust/annotate_method.go
@@ -44,7 +44,7 @@ type methodAnnotation struct {
 	HasResourceNameGeneration bool
 	ResourceNameTemplateGrpc  string
 	GrpcResourceNameArgs      []string
-	IsGetOperation            bool
+	IsLongRunningPoll         bool
 }
 
 // HasGrpcResourceNameArgs returns true if the method has gRPC resource name arguments.
@@ -312,7 +312,7 @@ func (c *codec) annotateMethod(m *api.Method) (*methodAnnotation, error) {
 		RoutingRequired:           c.routingRequired,
 		DetailedTracingAttributes: c.detailedTracingAttributes,
 		InternalBuilders:          c.internalBuilders,
-		IsGetOperation:            m.SourceServiceID == "google.longrunning.Operations" && m.Name == "GetOperation",
+		IsLongRunningPoll:         isLongRunningPoll(m),
 	}
 
 	if err := c.annotateResourceNameGeneration(m, annotation); err != nil {
@@ -627,4 +627,10 @@ func isIdempotent(p *api.PathInfo) string {
 		}
 	}
 	return "true"
+}
+
+func isLongRunningPoll(m *api.Method) bool {
+	return m.SourceServiceID == ".google.longrunning.Operations" &&
+		m.Name == "GetOperation" &&
+		m.Service.Package != "google.longrunning"
 }
--- a/internal/sidekick/rust/annotate_method.go
+++ b/internal/sidekick/rust/annotate_method.go
@@ -312,7 +312,7 @@ func (c *codec) annotateMethod(m *api.Method) (*methodAnnotation, error) {
 		RoutingRequired:           c.routingRequired,
 		DetailedTracingAttributes: c.detailedTracingAttributes,
 		InternalBuilders:          c.internalBuilders,
-		IsLroPoller:               isLongRunningPoll(m),
+		IsLroPoller:               m.IsLroPoller,
 	}
 
 	if err := c.annotateResourceNameGeneration(m, annotation); err != nil {
@@ -627,8 +627,4 @@ func isIdempotent(p *api.PathInfo) string {
 		}
 	}
 	return "true"
-}
-
-func isLongRunningPoll(m *api.Method) bool {
-	return m.IsLroPoller
 }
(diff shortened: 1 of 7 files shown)


internal/sidekick/rust/templates/crate/src/tracing.rs.mustache
```diff
@@ -79,8 +72,59 @@ where T: super::stub::{{Codec.Name}} + std::fmt::Debug + Send + Sync {
             metric: self.duration.clone(),
             info: *info::INSTRUMENTATION_CLIENT_INFO,
             method: "client::{{Codec.ServiceNameToPascal}}::{{Codec.Name}}",
-            self.inner.{{Codec.Name}}(req, options));
+            self.inner.{{Codec.Name}}(req.clone(), options.clone()));
```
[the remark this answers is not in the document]
haphungw: done

internal/sidekick/rust/templates/crate/src/tracing.rs.mustache
```diff
@@ -79,8 +72,59 @@ where T: super::stub::{{Codec.Name}} + std::fmt::Debug + Send + Sync {
...
+                            };
+                            _span.record("gcp.longrunning.status_code", code);
+                            if let Some(err) = &op.result {
+                                if let {{Codec.LongrunningModelModule}}::operation::Result::Error(status) = err {
+                                    _span.record("otel.status_code", "ERROR");
+                                    _span.record("otel.status_description", &status.message);
+                                    _span.record("rpc.response.status_code", status.code);
+                                    _span.record("error.type", google_cloud_gax::error::rpc::Code::from(status.code).to_string());
+                                }
+                            }
```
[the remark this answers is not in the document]
haphungw: done

internal/sidekick/rust/templates/crate/src/tracing.rs.mustache
```diff
@@ -63,12 +63,6 @@ impl<T> super::stub::{{Codec.Name}} for {{Codec.Name}}<T>
 where T: super::stub::{{Codec.Name}} + std::fmt::Debug + Send + Sync {
     {{#Codec.Methods}}
 
-    {{#Codec.DetailedTracingAttributes}}
-    #[tracing::instrument(level = tracing::Level::DEBUG, ret)]
-    {{/Codec.DetailedTracingAttributes}}
-    {{^Codec.DetailedTracingAttributes}}
-    #[tracing::instrument(ret)]
-    {{/Codec.DetailedTracingAttributes}}
```
coryan, line 71: Why are we removing these?  We need them for debug logging, no?

internal/sidekick/rust/templates/crate/src/tracing.rs.mustache
```diff
@@ -80,7 +74,52 @@ where T: super::stub::{{Codec.Name}} + std::fmt::Debug + Send + Sync {
             info: *info::INSTRUMENTATION_CLIENT_INFO,
             method: "client::{{Codec.ServiceNameToPascal}}::{{Codec.Name}}",
             self.inner.{{Codec.Name}}(req, options));
+        {{#Codec.HasLROs}}
+        {{#Codec.IsLongRunningPoll}}
+        #[cfg(google_cloud_unstable_tracing)]
+        {
+            if let Ok(attempt) = google_cloud_lro::POLL_ATTEMPT_COUNT.try_with(|c| *c) {
```
coryan, line 86: Some thoughts:

1) Can we reuse the request recorder for this instead of introducing a new task local thing?
2) Since we are using poller decorators for tracing, can we keep the counter in the decorator?

internal/sidekick/rust/annotate_method.go
```diff
@@ -626,3 +628,9 @@ func isIdempotent(p *api.PathInfo) string {
 	}
 	return "true"
 }
+
+func isLongRunningPoll(m *api.Method) bool {
+	return m.SourceServiceID == ".google.longrunning.Operations" &&
+		m.Name == "GetOperation" &&
+		m.Service.Package != "google.longrunning"
+}
```
coryan: I am not sure this works for discovery-style pollers.

internal/sidekick/rust/templates/crate/src/tracing.rs.mustache
haphungw: I misunderstood. I thought both tracing and logging go together.

internal/sidekick/rust/templates/crate/src/tracing.rs.mustache
haphungw: 1. We can't reuse request recorder because 
- the LRO crate is public-facing and doesn't depend on gaxi where the recorder lives. 
- the request recorder is scoped to a single request lifecycle (T3), while LRO polling is a loop (T2) that runs multiple independent request attempts over time.

2. I did manage to clean this up by tracking the counter inside `Tracing` poller decorator struct.

internal/sidekick/rust/annotate_method.go
haphungw: thanks for pointing out. this one should.

```
func isLongRunningPoll(m *api.Method) bool {
	if m.SourceServiceID == ".google.longrunning.Operations" &&
		m.Name == "GetOperation" &&
		m.Service.Package != "google.longrunning" {
		return true
	}
	return m.Name == "getOperation"
}
```

internal/sidekick/rust/templates/crate/src/tracing.rs.mustache
coryan: Can you restore the attribute?

internal/sidekick/rust/templates/crate/src/tracing.rs.mustache
coryan: 2. SGTM, thanks.

internal/sidekick/rust/annotate_method.go
coryan: That is not quite right, the LROs are configured:

https://github.com/googleapis/librarian/blob/88e1baa02a2c639706eccfe1e423bd17f239ecc8/internal/sidekick/api/discovery.go#L21-L49

Maybe we should add a `IsPoller` attribute to `api.Method` and set the value during the `xref.go` phase?

internal/sidekick/rust/templates/crate/src/tracing.rs.mustache
haphungw: ack, i will. i'm finalizing some details.

internal/sidekick/rust/annotate_method.go
haphungw: i see, thanks

internal/sidekick/rust/templates/crate/src/tracing.rs.mustache
coryan: > I did manage to clean this up by tracking the counter inside Tracing poller decorator struct.

Did that happen?

internal/sidekick/rust/annotate_method.go
```diff
@@ -626,3 +628,7 @@ func isIdempotent(p *api.PathInfo) string {
 	}
 	return "true"
 }
+
+func isLongRunningPoll(m *api.Method) bool {
+	return m.IsLroPoller
```
coryan: Do we still need this function? Seems like overkill to access a bool?

internal/sidekick/rust/annotate_method.go
```diff
@@ -311,6 +312,7 @@ func (c *codec) annotateMethod(m *api.Method) (*methodAnnotation, error) {
 		RoutingRequired:           c.routingRequired,
 		DetailedTracingAttributes: c.detailedTracingAttributes,
 		InternalBuilders:          c.internalBuilders,
+		IsLroPoller:               isLongRunningPoll(m),
```
coryan: Do we need this if the `*api.Method` already has the same flag?

coryan (changes requested):
I think you need to clean up some intermediate changes.

internal/sidekick/rust/templates/crate/src/tracing.rs.mustache
haphungw: The counter tracking state is inside the `Tracing` decorator. `lib.rs` is clean.
`get_operation` still needs to read from task local `POLL_ATTEMPT_COUNT` to record the count.

internal/sidekick/rust/annotate_method.go
haphungw: done

internal/sidekick/rust/annotate_method.go
haphungw: done

coryan (approved):

PR #6106, opened 2026-05-20
ci(.github/workflows/java): create a postsubmit check to verify library generation

### LINKED ISSUE #5238: java: create a postsubmit check in librarian to verify library generation
We should create a postsubmit check in librarian to verify library generation, similar to https://github.com/googleapis/librarian/blob/main/.github/workflows/golang.yaml

sofisl:
An integration postsubmit check is added to the Java workflow. This job runs on push to the main branch. It sets up Go and Java, checks out googleapis/google-cloud-java, installs the Java tools, and runs `librarian generate --all` to verify that Java library generation completes successfully at HEAD.

Fixes https://github.com/googleapis/librarian/issues/5238


.github/workflows/java.yaml
```diff
@@ -44,3 +45,32 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v6
+      - uses: ./.github/actions/setup-librarian
+      - uses: actions/setup-java@v4
+        with:
+          java-version: "17"
+          distribution: "temurin"
+      - name: Install Java tools
+        run: |
+          ./.github/workflows/scripts/install_java_tools.sh "$HOME/java_tools"
```
julieqiu, line 60: This is outside the scope of this PR, but can we convert this into a Go script? We should avoid using bash when possible: https://github.com/googleapis/librarian/blob/main/CONTRIBUTING.md#language-guidelines

.github/workflows/java.yaml
```diff
@@ -44,3 +45,32 @@ jobs:
           test -x "$(which protoc-gen-java_gapic)"
       - name: Run internal/librarian/java tests and check coverage
         run: go run ./tool/cmd/coverage ./internal/librarian/java
+  integration:
```
julieqiu, line 48: I wonder if this should be its own .github/workflows/ file, since run the same steps across multiple languages

julieqiu (approved):

.github/workflows/java.yaml
sofisl: I had jetski draft a version of this integration workflow:

```
# Copyright 2026 Google LLC
# ... license headers ...
name: Reusable Integration Test

on:
  workflow_call:
    inputs:
      language:
        required: true
        type: string
      repository:
        required: true
        type: string
      setup-java:
        type: boolean
        default: false
      setup-python:
        type: boolean
        default: false
      setup-rust:
        type: boolean
        default: false
      install-taplo:
        type: boolean
        default: false
      librarian-install:
        type: boolean
        default: false
      python-path:
        type: string
        default: ""
      post-generate-command:
        type: string
        default: ""

permissions:
  contents: read
  issues: write

jobs:
  integration:
    runs-on: ubuntu-24.04
    steps:
      - uses: actions/checkout@v6
      - uses: ./.github/actions/setup-librarian

      # Optional Java Setup
      - name: Set up Java
        if: ${{ inputs.setup-java }}
        uses: actions/setup-java@v4
        with:
          java-version: "17"
          distribution: "temurin"
      - name: Install Java tools
        if: ${{ inputs.setup-java }}
        run: |
          ./.github/workflows/scripts/install_java_tools.sh "$HOME/java_tools"
          echo "$HOME/java_tools/bin" >> $GITHUB_PATH

      # Optional Python Setup
      - name: Set up Python
        if: ${{ inputs.setup-python }}
        uses: actions/setup-python@v5
        with:
          python-version: "3.12"
      - name: Install Python dependencies
        if: ${{ inputs.setup-python }}
        run: |
          pip install git+https://github.com/googleapis/synthtool.git

      # Optional Rust Setup
      - name: Set up Rust
        if: ${{ inputs.setup-rust }}
        uses: dtolnay/rust-toolchain@stable
      - name: Install Taplo
        if: ${{ inputs.install-taplo }}
        uses: ./.github/actions/install-taplo

      # Target Repository Checkout & Build
      - name: Checkout target repository
        uses: actions/checkout@v6
        with:
          repository: ${{ inputs.repository }}
          path: repo

      - name: Run librarian install
        if: ${{ inputs.librarian-install }}
        working-directory: repo
        run: librarian install

      - name: Run librarian generate
        working-directory: repo
        env:
          PYTHONPATH: ${{ inputs.python-path != '' && format('{0}/repo/{1}', github.workspace, inputs.python-path) || '' }}
        run: librarian generate --all

      - name: Run post-generate command
        if: ${{ inputs.post-generate-command != '' }}
        working-directory: repo
        run: ${{ inputs.post-generate-command }}

      # Automatically create issue on failure (removes duplicated jobs in the callers)
      - name: Create issue on failure
        if: ${{ failure() && github.event_name == 'push' && github.ref == 'refs/heads/main' }}
        uses: ./.github/actions/create-issue-on-failure
        with:
          title: "${{ inputs.repository }}: librarian generate fails when run at HEAD"
          body: |-
            The ${{ inputs.language }} CI failed on the `main` branch.

            Running `librarian generate --all` on the [${{ inputs.repository }}](<https://github.com/${{ inputs.repository }}>) repository using `librarian` at HEAD failed.

            Note: This integration test only runs as a postsubmit test due to its long execution time.

            Workflow Run: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
```

which we would then call like this:

```
  integration:
    needs: [test]
    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
    uses: ./.github/workflows/reusable-integration.yaml
    with:
      language: Java
      repository: googleapis/google-cloud-java
      setup-java: true
      setup-python: true
      python-path: sdk-platform-java/hermetic_build/library_generation/owlbot
```

IMO, it makes it a bit more difficult to read and/or debug, but it does preserve DRY.

Another option we could pursue potentially is creating a template, like [this](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/main/.github/workflows/utils/ci.yaml.njk).

.github/workflows/java.yaml
```diff
@@ -44,3 +45,40 @@ jobs:
...
+      - name: Install Java tools
+        run: |
+          ./.github/workflows/scripts/install_java_tools.sh "$HOME/java_tools"
+          echo "$HOME/java_tools/bin" >> $GITHUB_PATH
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - name: Install Python dependencies
+        run: |
+          pip install git+https://github.com/googleapis/synthtool.git
```
zhumin8: We need  the [forked synthtool](https://github.com/googleapis/google-cloud-java/tree/main/sdk-platform-java/hermetic_build/library_generation/owlbot/synthtool) in google-cloud-java

.github/workflows/java.yaml
```diff
@@ -44,3 +45,40 @@ jobs:
...
+        uses: actions/checkout@v6
+        with:
+          repository: googleapis/google-cloud-java
+          path: google-cloud-java
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - name: Install Python dependencies
+        run: |
+          pip install ./google-cloud-java/sdk-platform-java/hermetic_build/library_generation/owlbot/synthtool
```
zhumin8: Does this work? I think you might need -e (editable) flag here
```
- name: Install Python dependencies
        run: |
          pip install -e ./sdk-platform-java/hermetic_build/library_generation/owlbot/synthtool
```

The step Run librarian generate is going to look inside that local owlbot folder for modules. If you don't use -e, Python might get confused between the static version copied to site-packages and the local version pointed to by PYTHONPATH. Using -e creates a symlink/pointer from site-packages back to your checked-out repository, ensuring both pip and PYTHONPATH point to the exact same physical files.

sofisl: TIL! Thank you!

zhumin8 (approved):

PR #6108, opened 2026-05-20
feat(swift): support recursive message types with transparent referen…

nolanleastin:
…ce boxing

Introduced support for recursive singular fields and cyclic message dependency graphs (such as self-recursive messages and multi-step loops: A -> B -> C -> A) in the generated Swift client library, preventing compiler infinite struct sizing errors.

Tagged matching fields with `Recursive` and `UnwrappedType` annotations to drive templated Swift generation.

### DIFF
--- a/internal/sidekick/swift/annotate_field.go
+++ b/internal/sidekick/swift/annotate_field.go
@@ -108,7 +108,7 @@ func (c *codec) annotateField(field *api.Field) error {
 	}
 	// Swift value types (structs) cannot contain recursive references directly because their
 	// size must be known at compile time. To break the cycle, we wrap the reference in a box type
-	// when the followin conditions are met:
+	// when the following conditions are met:
 	// 1. field.Recursive: The field is part of a recursive reference cycle.
 	// 2. field.Singular(): Repeated fields ([T]) and Maps ([K: V]) store their elements dynamically
 	//    on the heap, so they do not cause compile-time infinite struct size issues.
--- a/internal/sidekick/swift/annotate_field.go
+++ b/internal/sidekick/swift/annotate_field.go
@@ -15,6 +15,8 @@
 package swift
 
 import (
+	"fmt"
+
 	"github.com/googleapis/librarian/internal/sidekick/api"
 )
 
@@ -45,6 +47,14 @@ type fieldAnnotations struct {
 	//
 	// This is empty for fields that are not part of a oneof group.
 	OneOfPropertyName string
+
+	// Recursive is true if the field is a recursive reference to another message.
+	Recursive bool
+
+	// UnwrappedType is the FieldType of the field without the optional Recursive Wrapper.
+	//
+	// Only set if Recursive is true.
+	UnwrappedType string
 }
 
 func (c *codec) annotateField(field *api.Field) error {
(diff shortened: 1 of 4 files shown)


internal/sidekick/swift/annotate_field.go
```diff
@@ -42,6 +44,14 @@ type fieldAnnotations struct {
...
 	// This is empty for fields that are not part of a oneof group.
 	OneOfPropertyName string
+
+	// Recursive is true if the field is a recursive reference to another message.
+	Recursive bool
+
+	// UnwrappedType is the FieldType of the field without the optional Recursive Wrapper.
+	//
+	// Only set if Recursive is true.
+	UnwrappedType string
```
coryan: How is this different from `BaseFieldType`?

internal/sidekick/swift/templates/common/message.mustache
```diff
@@ -66,7 +67,12 @@ public struct {{Codec.Name}}: Codable, Equatable, {{Codec.Model.WktPackage}}._An
   ) {
     {{#Fields}}
     {{^IsOneOf}}
+    {{#Codec.Recursive}}
+    self.{{Codec.Name}} = {{Codec.Name}}.map { GoogleCloudWkt.Recursive(value: $0) }
```
coryan:
```suggestion
    self.{{Codec.Name}} = {{Codec.Name}}.map { {{Codec.Model.WktPackage}}.Recursive(value: $0) }
```

coryan: Maybe you need some navigation to get to `Model`...

internal/sidekick/swift/annotate_field.go
```diff
@@ -63,6 +73,20 @@ func (c *codec) annotateField(field *api.Field) error {
...
+	// when the following conditions are met:
+	// 1. field.Recursive: The field is part of a recursive reference cycle.
+	// 2. field.Singular(): Repeated fields ([T]) and Maps ([K: V]) store their elements dynamically
+	//    on the heap, so they do not cause compile-time infinite struct size issues.
+	// 3. !field.IsOneOf: Oneof fields are nested inside a Swift enum which handles recursive boxing
+	//    automatically using the native indirect case mechanism.
+	if field.Recursive && field.Singular() && !field.IsOneOf {
+		annotations.Recursive = true
+		annotations.UnwrappedType = fieldType
+		annotations.BaseFieldType = fmt.Sprintf("%s.Recursive<%s>", wellKnownSwiftPackage, baseFieldType)
```
coryan, line 122: I see why we need this... Maybe we can generalize to `InitializerTypeName` or something that works for both?

I am starting to think that the `.with()` approach would eliminate the need for these annotations and special cases too.

coryan (approved):
Please fix the hard-coded `GoogleCloudWkt` value.

The other stuff is not blocking.  It would be nice to have a matching PR in the Swift repo showing how the generated code will look like.

internal/sidekick/swift/annotate_field.go
nolanleastin: Well in this case the BaseFieldType will be something like `GoogleCloudWkt.Recursive<RecursiveType>`

Whereas the `UnwrappedType` would be `RecursiveType`

internal/sidekick/swift/annotate_field.go
nolanleastin: I have renamed it to InitializerType. I think it should work now. 

The way I am thinking of it:

1. InitializerType: user-facing Swift type that appears in the init signature:
`init( ...., child: Node? = nil)`

2. FieldType: the exact type of the Swift Field property e.g.:
`public var child: GoogleCloudWkt.Recursive<Node>?`

3. BaseFieldType: Swift type without extra decorators. Used for the decoding e.g.:
`self.child = try container.decodeIfPresent(GoogleCloudWkt.Recursive<Node>.self, forKey: .child)`

coryan: SG

coryan (approved):

PR #6110, opened 2026-05-21
feat(internal/librarian/nodejs): use pnpm install

quirogas:
## TLDR
This PR refactors our Node.js installer inside librarian. We migrated from our current practice of using the `npm` CLI wrapper to the **pnpm toolchain** utilizing `corepack` to establish absolute package linkage determinism.

By using pnpm and adhering to the lock files installation, we can create a deterministic installation environment. Matching our tool versions to those specified in the lockfiles allows us to match our current generation pipeline output.

---

## Context: Why should we transition to pnpm instead of npm?
1.  **Deterministic Dependency Isolation**: Standard `npm` installations are deterministic when consuming local project files under `package-lock.json`. However, our current practice of performing dynamic global package installations inside librarian (e.g., executing `npm install -g gapic-tools@1.0.5` on host environments) does not respect lockfiles, floating transitives dynamically. Standardizing on `pnpm` resolves this by strictly respecting `pnpm-lock.yaml` across our source-built target tools.
2.  **Infrastructure Alignment**: Moving to `pnpm` directly reflects and aligns our local generation tools with the hermetic sandboxing configurations we use upstream in our current bazel-bot pipeline and postprocessor.
3.  **Monorepo Support & Forward Outlook**: The Node.js platform team has expressed a strategic intent to migrate libraries and generators entirely to `pnpm` in the future due to its first-class workspace support and content-addressable storage design tailored for monorepos.
4.  **Complete npm Decoupling**: To eliminate the runtime footprint of the `npm` binary during tool bootstrapping, we successfully transitioned our installer to Node.js's built-in **`corepack`** wrapper. 
    *   Our installer now automatically creates and maps standard executable bin shims globally using **`corepack enable pnpm`** first.
    *   Our installer activates the target version globally via **`corepack prepare pnpm@7.32.2 --activate`**.
    *   Our installer queries Node's native **`process.execPath`** folder path directly in JavaScript (`node -e "console.log(require('path').dirname(process.execPath))"`) to configure `pnpm`'s `global-bin-dir`. This runs completely independently of standard `npm config` calls.
5.  **Clean Global Links**: We replaced `npm link` with **`pnpm link --global`** to bind the compiled typescript generator to our shared global virtual store.

---

## Why do should we explicitly specify protobufjs?

During local runs of the generator we created against `google-cloud-secretmanager`, the generated `protos.js` file introduced a substantial diff containing recursive depth limits (`long > $util.recursionLimit`) and prototype pollution validations (`keys[i] !== "__proto__"`).

We traced this discrepancy back to a peer dependency resolution mismatch:
*   **The Mismatch**: Upstream compiler shims rely on `protobufjs-cli`, which depends on `protobufjs` as a peer dependency. In floated global registry environments, this peer dependency resolves to the latest `protobufjs@7.6.0`, which automatically injects these recursion and security checks into the generated JavaScript AST.
*   **Does `pnpm` not address this?**:
    *   Normally, when executing project-level installations inside folders with locked files (`pnpm install`), `pnpm`'s virtual store correctly isolates peer dependencies according to the locks. 
    *   However, because `gapic-tools` (which contains the `compileProtos` compilation shim) is installed from the **NPM registry globally** (`pnpm add -g`), published packages do not bundle a lockfile. 
    *   In the absence of a lockfile, `pnpm` dynamically resolves the floated peer range `^7.5.3` under `protobufjs-cli` to the latest minor version (`7.6.0`), triggering the security and depth limit checks.
*   **The Current Pipeline**: However, our current bazel-bot pipeline runfiles are hermetically isolated. During historical compilation, the peer dependency resolved to the only available, older `protobufjs` sub-dependency bundled inside the sandbox (specifically **`7.5.4`** / **`7.5.5`**), generating AST files *without* these checks.
*   **The Aligned Solution**:
    *   To resolve this peer range drift and align with our pipeline output, we added direct peer pins for **`protobufjs-cli@1.2.0`** and **`protobufjs@7.5.5`** inside our tools configuration block in `librarian.yaml`.
    *   This forces the global pnpm virtual store to link the peer dependencies directly to the correct pre-patched versions, resolving the version-mismatch checks completely (leaving only benign monorepo root namespace registry adjustments!).


internal/sidekick/rust/annotate_method_test.go
```diff
@@ -111,6 +111,7 @@ func TestAnnotateDiscoveryAnnotations(t *testing.T) {
 	gotMethod := model.Method(methodID)
 	if gotMethod == nil {
 		t.Fatalf("missing method %s", methodID)
+		return
```
coryan: Why?  `t.Fatalf()` returns immediately:

https://pkg.go.dev/testing#T.Fatalf
https://pkg.go.dev/testing#T.FailNow

internal/librarian/nodejs/generate.go
```diff
@@ -314,6 +314,7 @@ func runPostProcessor(ctx context.Context, cfg *config.Config, library *config.L
 	if err := copyMissingProtos(googleapisDir, outDir); err != nil {
 		return fmt.Errorf("failed to copy missing protos: %w", err)
 	}
+
```
coryan: Is this clarifying anything?

internal/librarian/nodejs/install.go
```diff
@@ -41,9 +41,29 @@ func Install(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("parsing embedded librarian.yaml: %w", err)
 	}
-	for _, tool := range cfg.Tools.NPM {
+	// Ensure pnpm is enabled and shims are linked globally using corepack
+	if err := command.RunStreaming(ctx, "corepack", "enable", "pnpm"); err != nil {
+		return fmt.Errorf("failed to enable pnpm shims: %w", err)
+	}
+
+	// Ensure pnpm version 7.32.2 is prepared and activated globally to enable lockfile-respecting builds
```
coryan: Weird, shouldn't the version be a configuration parameter in `librarian.yaml` ?  Maybe I just don't have the right context.

internal/sidekick/parser/discovery/discovery_test.go
```diff
@@ -175,6 +175,7 @@ func TestDeprecatedField(t *testing.T) {
 	gotMessage := model.Message(id)
 	if gotMessage == nil {
 		t.Fatalf("expected message %s in the API model", id)
+		return
```
coryan: Ditto, please revert these redundant `return` statements.

coryan (dismissed):
Those extra `return` statements after `t.Fatal()` and after `t.Fatalf()` seems wrong, please revert.

internal/librarian/nodejs/install.go
JoeWang1127, line 41: Could you verify all commands are installed before actually running them?

For example, in java package:
```go
for _, cmd := range []string{"java", "mvn", "pip"} {
		if _, err := exec.LookPath(cmd); err != nil {
			return fmt.Errorf("%s is not installed or not in PATH, which is required for Java tool installation: %w", cmd, err)
		}
}
```

internal/librarian/nodejs/install.go
JoeWang1127: +1.

We should list all dependency version in `librarian.yaml`?

internal/librarian/nodejs/install_test.go
```diff
@@ -47,25 +47,27 @@ func TestInstall(t *testing.T) {
...
 esac
-mkdir -p node_modules/.bin
-printf '#!/bin/sh\nmkdir -p build\n' > node_modules/.bin/tsc
-chmod +x node_modules/.bin/tsc
+exit 0
 `
-	if err := os.WriteFile(filepath.Join(bin, "npm"), []byte(npmStub), 0o755); err != nil {
+	corepackStub := `#!/bin/sh
+exit 0
+`
```
JoeWang1127, line 77: nit: remove this empty line.

JoeWang1127 (commented):
Could you revert the changes in sidekick directory? They seems redundant.

JoeWang1127:
Could you fix the unit tests?

internal/librarian/nodejs/install.go
quirogas: Yes, this should definitely be in librarian.yaml (or perhaps a constants file), but librarian.yaml is definitely the right call. Thank you!

internal/sidekick/rust/annotate_method_test.go
quirogas: I have undone these changes. I was seeing linting errors earlier, but they have been resolved in the latest main.

internal/librarian/nodejs/generate.go
quirogas: Not anymore. I have removed the function in favor of using `keep` in the librarian.yaml.

JoeWang1127: nit: remove this empty line.

internal/librarian/nodejs/install.go
```diff
@@ -37,13 +38,25 @@ var librarianYAML []byte
...
+			return fmt.Errorf("%s is not installed or not in PATH, which is required for Node.js tool installation: %w", cmd, err)
+		}
+	}
+
 	cfg, err := yaml.Unmarshal[config.Config](librarianYAML)
 	if err != nil {
 		return fmt.Errorf("parsing embedded librarian.yaml: %w", err)
 	}
-	for _, tool := range cfg.Tools.NPM {
+
```
JoeWang1127: nit: remove this empty line.

internal/librarian/nodejs/install.go
```diff
@@ -52,16 +65,65 @@ func Install(ctx context.Context) error {
...
-		if err := command.RunStreaming(ctx, "npm", "install", "-g", pkg); err != nil {
+		if err := runPNPM(ctx, "", env, "add", "-g", pkg); err != nil {
 			return err
 		}
 	}
 	return nil
 }
 
-func installNPMToolFromSource(ctx context.Context, tool *config.NPMTool) error {
+func getPNPMEnv(ctx context.Context) ([]string, error) {
```
JoeWang1127, line 84: Could you add a comment for this function?

internal/librarian/nodejs/install.go
```diff
@@ -52,16 +65,65 @@ func Install(ctx context.Context) error {
...
+	binOut, err := commandOutput(ctx, "node", "-e", "console.log(require('path').dirname(process.execPath))")
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve node bin directory: %w", err)
+	}
+	globalBin := strings.TrimSpace(binOut)
+
```
JoeWang1127, line 90: nit: remove this empty line.

internal/librarian/nodejs/install.go
```diff
@@ -52,16 +65,65 @@ func Install(ctx context.Context) error {
...
+	cmd.Stdout = &out
+	err := cmd.Run()
+	return out.String(), err
+}
+
+type bytesBuffer struct {
+	strings.Builder
+}
+
+func (b *bytesBuffer) Write(p []byte) (n int, err error) {
```
JoeWang1127: Does this function has any usage? I didn't find any in the codebase.

internal/config/language.go
```diff
@@ -741,6 +741,9 @@ type NodejsPackage struct {
 	// Dependencies maps npm package names to version constraints.
 	Dependencies map[string]string `yaml:"dependencies,omitempty"`
 
+	// PNPMVersion is the desired pnpm package manager version (e.g., "7.32.2").
+	PNPMVersion string `yaml:"pnpm_version,omitempty"`
```
JoeWang1127: How/where is this variable used? I only found it in `mergeNodejs` function, which only assigned the value.

internal/librarian/nodejs/install.go
```diff
@@ -37,13 +38,25 @@ var librarianYAML []byte
 
 // Install installs Node.js tool dependencies.
 func Install(ctx context.Context) error {
+	for _, cmd := range []string{"node", "pnpm"} {
```
JoeWang1127, line 42: Do we need to add `corepack` here?

internal/config/language.go
quirogas: I'm removing it now, as it was used in a previous commit to track the pnpm version before I made it mandatory.

internal/librarian/nodejs/install.go
quirogas: `corepack` comes with node, so checking for one should be sufficient, but  redundancy never hurts -- added :).

JoeWang1127 (approved):

PR #6114, opened 2026-05-21
refactor(internal/librarian): use artifact ID instead of distribution name

### LINKED ISSUE #6048: java: refactor distribution_name_override to artifact_id
In Java configs, we migrated from generation_config.yaml:
- distribution_name_override
- group_id

`distribution_name_override` contains `group_id`. Let's refactor it to just setting `artifact_id` (for overriding the default), so we don't need to deal with case with contradicting info in the two.

This in theory can be done after migrate, putting into migrate milestone to work on if we got bandwidth.

JoeWang1127:
The distribution name override field is removed from the Java configuration module and replaced with a first-class artifact ID field. The logic for deriving Maven coordinates and generating metadata is updated to compute the distribution name directly from the configured group ID and artifact ID.

Fixes #6048


internal/librarian/java/defaults.go
```diff
@@ -161,26 +171,15 @@ func isEmptyJavaModule(j *config.JavaModule) bool {
 }
 
 var (
-	// ErrInvalidDistributionName is returned when a distribution name override
-	// is incorrectly formatted.
-	ErrInvalidDistributionName = fmt.Errorf("invalid distribution name override")
```
sofisl, line 166: wouldn't we need an equivalent error for artifact ID?

sofisl: So I see why we maybe wouldn't error in the same way because we aren't deriving/composing a name. However, since we still have a Validate function, I'd assume we have a potential error path.

sofisl:
also, nit, I think this is more of a fix since the interface is changing

internal/librarian/java/defaults.go
```diff
@@ -17,16 +17,20 @@ package java
...
-	"strings"
 
 	"github.com/googleapis/librarian/internal/config"
 )
 
 const (
-	javaPrefix     = "java-"
-	defaultGroupID = "com.google.cloud"
+	javaPrefix              = "java-"
+	defaultArtifactIDPrefix = "google-cloud-"
```
zhumin8, line 26: nit: this is already defined in https://github.com/googleapis/librarian/blob/88e1baa02a2c639706eccfe1e423bd17f239ecc8/internal/librarian/java/maven_coordinate.go#L27

internal/librarian/java/defaults.go
zhumin8: No. The validation checks "that the distribution name override contains exactly two parts separated by a colon". There is no need of such check now that we construct it from artifact id and group id in `DistributionName`

internal/librarian/java/defaults.go
JoeWang1127: I removed this one since I think putting the default group id and artifact id together makes sense.

zhumin8 (approved):

PR #6125, opened 2026-05-21
feat(tool/cmd/migrate/java): remove redundant keep items from java config

### LINKED ISSUE #6046: java: remove redundant keep items from config matching regex
For files that matches the below regex, they are preserved by default during clean. See
https://github.com/googleapis/librarian/blob/96b31c836b4ce4f02c587494e39af7c7e34090d8/internal/librarian/java/clean.go#L32-L33

Currently keep are migrated from .Owlbot-hermetic.yaml and contain redundant items.

sofisl:
Omit redundant files (such as Version.java and IT*Test.java) as they are already preserved by default. Instead of changing the logic in the keep files, this logic is added to tidy in case someone manually attempts to add a duplicate file in the future.

Fixes https://github.com/googleapis/librarian/issues/6046


internal/librarian/java/defaults.go
```diff
@@ -77,6 +79,46 @@ func Fill(library *config.Library) (*config.Library, error) {
 // Tidy tidies the Java-specific configuration for a library by removing default
 // values.
 func Tidy(library *config.Library) *config.Library {
+	if len(library.Keep) > 0 {
```
JoeWang1127: Could you separate this piece of code to a function?

internal/librarian/java/defaults_test.go
```diff
@@ -360,6 +360,40 @@ func TestTidy(t *testing.T) {
...
+					},
+				},
+				Java: &config.JavaModule{
+					GroupID:    "com.google.cloud",
+					ArtifactID: "google-cloud-vision",
+				},
+				Keep: []string{
+					"google-cloud-vision/src/main/java/com/google/cloud/vision/v1/stub/Version.java",
+					"google-cloud-vision/src/test/java/com/google/cloud/vision/it/ITSystemTest.java",
+					"google-cloud-vision/src/test/resources/placeholder.txt",
```
JoeWang1127, line 380: Why this file is removed from the keep list?

internal/librarian/java/defaults.go
```diff
@@ -77,6 +79,46 @@ func Fill(library *config.Library) (*config.Library, error) {
...
+		}
+		var kept []string
+		for _, keepPath := range library.Keep {
+			keepPathSlash := filepath.ToSlash(keepPath)
+			// 2. Skip files matching standard test/version regexes since they are preserved globally.
+			if isDefaultPreserved(keepPathSlash) {
+				continue
+			}
+			omit := false
+			// 3. Skip manual files in GAPIC src folders since they are preserved by marker-based clean.
```
zhumin8: I would separate out this from this PR. 

This logic is not exactly what we want: it assumes all files listed in keep within GAPIC src folders except the 2 json files are okay to remove. That is not true.
One example is in `aiplatform`,
```
    keep:
      - google-cloud-aiplatform/src/test/java/com/google/cloud/location/MockLocations.java
      - google-cloud-aiplatform/src/test/java/com/google/cloud/location/MockLocationsImpl.java
      - google-cloud-aiplatform/src/test/java/com/google/iam/v1/MockIAMPolicy.java
      - google-cloud-aiplatform/src/test/java/com/google/iam/v1/MockIAMPolicyImpl.java
 ```
These files have the generated marker `@Generated("by gapic-generator-java")` and would otherwise be deleted by clean if not listed in keep.

**Review by zhumin8 — CHANGES_REQUESTED** [review]

File: internal/librarian/java/defaults.go (hunk 3 above)
Comment (sofisl) [reply internal/librarian/java/defaults.go]: spoke offline, will work in a f/u PR for: https://github.com/googleapis/librarian/issues/6163

File: internal/librarian/java/defaults_test.go (hunk 2 above)
Comment (sofisl) [reply internal/librarian/java/defaults_test.go:380]: good call! I think I was confused by another part of the original bug, but I think Min and I discussed offline and I'll add it to a separate PR.

File: internal/librarian/java/defaults.go (hunk 1 above)
Comment (sofisl) [reply internal/librarian/java/defaults.go]: Hm, do you mean just this part? `len(library.Keep) > 0`? I feel like that might be unnecessary, unless you were thinking of another larger part of the code.

Comment (JoeWang1127) [reply internal/librarian/java/defaults.go]: I mean all the code within this `if` statement to a separate function, `tidyKeep`, for example.

File: internal/librarian/java/defaults.go [hunk 4]
```diff
@@ -77,6 +78,22 @@ func Fill(library *config.Library) (*config.Library, error) {
 // Tidy tidies the Java-specific configuration for a library by removing default
 // values.
 func Tidy(library *config.Library) *config.Library {
+	if len(library.Keep) > 0 {
+		var kept []string
```
zhumin8: nit: it's a bit confusing with `kept`, `keepPath` and `library.Keep`
maybe rename `kept` to  something like`filteredPaths` or `filteredKeepPaths` or `remainingPaths`?

internal/librarian/java/defaults.go
zhumin8: plus one to extract helper function

zhumin8 (approved):
left a nit comment

PR #6147, opened 2026-05-22
fix(sidekick/swift): use full return type for external package types

chingor13:
This prevents name conflicts between similarly named types. For example, `GoogleLongrunning` has an `Operation` type which conflicts with a built-in `Operation` type. Without using the fully qualified name, the compiler throws an error.

We will likely need to do something similar for the input types as well.


coryan (approved):
Why is this not a problem for input types of methods too?  I assume mixins consume remote types and they may also conflict?

Don't block on this question, this is an improvement, so merge as-is if it works.

coryan:
> We will likely need to do something similar for the input types as well.

I should have read this before asking. I agree.

Issue #6134, opened 2026-05-22
google-cloud-java: raise PR for migrate -insert-markers


zhumin8:
Update pom.xml with generated markers, these markers are used to mark managed artifacts for pom.xml updates. They are only used when adding new modules, re-generation works fine without these markers. So this can be done independently of #6135.

To do this, run:
install librarian, 
run  `migrate -insert-markers .` 
raise a PR to google-cloud-java repo.

see [#4921](https://github.com/googleapis/librarian/issues/4921) for more details.

zhumin8:
If new library is added by hermetic build after this PR and before we fully use librarian to generate, need to add markers to the newly added library. Therefore, this `migrate -insert-markers .` should be able to skip existing and run repeately.

zhumin8:
CI failing, need to update showcase goldens. 
Additionally, seems that the original post-processor  fix_poms.py  has the bug where it treats comment elements in  <modules>  as "extra modules" and writes them back as real elements, the updated golden file in the workspace now has comment markers converted into actual  <module>  tags.

zhumin8:
**The issue:**
The current code in `fix_pom.py` is not compatible with comments. 
 When processing these pom.xml with comment markers, the output pom.xml it leaves in the workspace contains the illegal XML tags generated by the parsing bug:
    <module> {x-generated-modules-start} </module>
    <module> {x-generated-modules-end} </module>
Leaves broken, non-compilable Maven project.

**Choose minial fix now for `fix_poms.py`:**
- Let it exclude comments during extra module detection, but not preserving them. This means that comment markers won't cause `fix_poms.py` to produce pom.xml with compiling issue. But will ignore the comments in the updated pom.xml files.
- It is not ideal, as it will wipe out comment marker additions when generate is run on each library.
- Why I think it is **okay** for migration purposes:
  - During migrate, we only have limited number of "generate" runs. We will turn off hermetic build workflow after migrated for a week ([go/sdk-librarian-migration-java](http://goto.google.com/sdk-librarian-migration-java))
  - We can rerun `migrate -insert-markers .` and add back any removed comment markers easily.
  - In librarian, these comment markers are only used when adding new modules. So they are okay to be left out temporarily.


Alternative: 
Fix  `fix_poms.py` to also preserve the comments. This would be ~50 lines of changes, plus tests. Extra burden to review.

cc. @blakeli0 let me know if you prefer the alternative?

zhumin8:
[...]

PR #6159, opened 2026-05-26
fix(tool/cmd/migrate): skip inserting markers if already present

zhumin8:
Update the Java migration tool to skip wrapping dependencies/modules with markers if those markers are already present in the target pom.xml. This prevents duplicating markers when the migration tool is run repeatedly on the same repository. This is needed during migration as we might need to rerun this command after new library additions done with hermetic build.

A new test is added to verify that repeated migration runs do not duplicate markers.

For #6134


JoeWang1127 (approved):

JoeWang1127:
We should make sure no manual updates after the initial generation of `librarian.yaml`; otherwise these changes will lose after reruning the migration tool

PR #6161, opened 2026-05-26
feat(internal/librarian/golang): support disabling generator features

JoeWang1127:
A configuration option `disabled_generator_features` is added to the GoAPI struct. This allows disabling specific generator features for individual APIs.

During the GAPIC option build phase in the golang code generator, the enabled generator features for an API are filtered using its disabled generator features before they are passed to the generator. The filtering is performed in an order-preserving, deterministic manner.

Fixes #5255


internal/librarian/library.go
```diff
@@ -92,6 +93,25 @@ func union(a, b []string) []string {
...
+func deleteFromSlice(src, toDelete []string) []string {
+	tdMap := toMap(toDelete)
+	var res []string
+	for _, item := range src {
+		if !tdMap[item] {
+			res = append(res, item)
+		}
+	}
+	return res
+}
```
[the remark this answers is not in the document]

internal/librarian/golang/generate.go
```diff
@@ -367,3 +369,25 @@ func transport(sc *serviceconfig.API) serviceconfig.Transport {
 func isPreview(output string) bool {
 	return strings.Contains(output, "preview/internal")
 }
+
+func deleteFromSlice(src, toDelete []string) []string {
```
noahdietz: Can we replace this custom code with the `slices.DeleteFunc` ([docs](https://pkg.go.dev/slices#DeleteFunc)) and `slices.Contains` ([docs](https://pkg.go.dev/slices#Contains))?

```go
slices.DeleteFunc(src, func(feat string) bool {
  return slices.Contains(toDelete, feat)
})
```

doc/config-schema.md
```diff
@@ -268,6 +268,7 @@ This document describes the schema for the librarian.yaml.
 | :--- | :--- | :--- |
 | `client_package` | string | Is the package name of the generated client. |
 | `diregapic` | bool | Indicates whether generation uses DIREGAPIC (Discovery REST GAPICs). This is typically false. Used for the GCE (compute) client. |
+| `disabled_generator_features` | list of string | Provides a mechanism for disabling generator features at the API level. |
```
zhumin8: Can you specify the behavior if both `enabled_generator_features` and `disabled_generator_features` are specified?

JoeWang1127: Done.

internal/librarian/golang/generate_test.go
```diff
@@ -755,6 +755,65 @@ func TestBuildGAPICOpts(t *testing.T) {
 				"release-level=ga",
 			},
 		},
+		{
+			name:    "disable gen features",
```
zhumin8: What is different between this test case and "a feature is specified in both enabled and disabled list"? The DisabledGeneratorFeatures here is also specified in enabled list.

zhumin8 (approved):

internal/librarian/golang/generate_test.go
JoeWang1127: Good catch, this one is actually overlapped with "a feature is specified in both enabled and disabled list", removed it from the change.

PR #6168, opened 2026-05-27
feat(internal/librarian/golang): support custom sample URIs in README

### LINKED ISSUE #6151: golang: Add configuration option to specify a sample URI for handwritten samples
### Problem
Currently, the generator does not have the capability to point to handwritten samples. If we hand-edit the generated README to include links to custom samples, an in-place regeneration blows away those manual changes. 

For some clients (such as the Maps API clients b/421113149), we need to stop generating snippets and instead provide a link to existing handwritten samples, without those links being overwritten during regeneration.

### Proposed Solution
Provide a configuration option in `librarian.yaml` to specify a custom sample URI. This would allow us to disable automatic snippet generation and reliably point to handwritten samples directly in the generated README.

### Context
This issue was identified when attempting to disable snippet generation for maps clients using a non-existent `omit_snippets` property to avoid overwriting manual changes.

JoeWang1127:
A configuration option sample_uris is added to the API allowlist structure to permit specifying language-specific documentation URIs for code samples.

During README generation for Go client libraries, the generator now resolves the sample URI dynamically from the API configuration. If no custom sample URI is defined for Go, the default Google Cloud Samples link is used.

Fixes #6151

### DIFF
--- a/doc/api-allowlist-schema.md
+++ b/doc/api-allowlist-schema.md
@@ -14,7 +14,8 @@ This document describes the schema for the API Allowlist.
 | `new_issue_uri` | string | Overrides the new issue URI from the service config's publishing section. |
 | `skip_rest_numeric_enums` | list of string | Lists languages that should not pass the rest-numeric-enums flag to the generator. The special value "all" skips it for every language. If empty, all languages use numeric enums. |
 | `open_api` | string | Is the file path to an OpenAPI spec, currently in internal/testdata. This is not an official spec yet and exists only for Rust to validate OpenAPI support. |
-| `release_level` | map[string]string | Is the release level per language. Map key is the language name (e.g., "python", "rust"). Optional. If omitted, the generator default is used.<br><br>TODO(https://github.com/googleapis/librarian/issues/4834): Go uses "alpha", "beta", and "ga" instead of "preview" and "stable". We should standardize release level vocabulary across languages. |
+| `release_level` | map[string]string | Is the release level per language. Map key is the language name (e.g., "python", "rust"). Optional. If omitted, the generator default is used. |
+| `sample_uris` | map[string]string | Is the documentation URI for code samples per language. Map key is the language name (e.g., "go", "python"). Optional. If omitted, a default URI for the language is used. |
 | `short_name` | string | Overrides the API short name from the service config's publishing section. |
 | `service_config` | string | Is the service config file path override. If empty, the service config is discovered in the directory specified by Path. |
 | `service_name` | string | Is a DNS-like logical identifier for the service, such as `calendar.googleapis.com`. |
--- a/internal/librarian/golang/generate.go
+++ b/internal/librarian/golang/generate.go
@@ -386,7 +386,7 @@ func isPreview(output string) bool {
 //
 // The default value is the empty string.
 func sampleURI(sc *serviceconfig.API) string {
-	if sc.SampleURIs == nil {
+	if sc == nil || sc.SampleURIs == nil {
 		return ""
 	}
 	if uri, ok := sc.SampleURIs[config.LanguageGo]; ok {
--- a/internal/librarian/golang/generate.go
+++ b/internal/librarian/golang/generate.go
@@ -74,7 +74,7 @@ func Generate(ctx context.Context, cfg *config.Config, library *config.Library, 
 	}
 
 	var fallbackTitle string
-	var sample string
+	var customSampleURI string
 	for i, api := range library.APIs {
 		goAPI := findGoAPI(library, api.Path)
 		if goAPI == nil {
@@ -97,15 +97,15 @@ func Generate(ctx context.Context, cfg *config.Config, library *config.Library, 
 		if i == 0 {
 			fallbackTitle = sc.Title
 		}
-		// Use the sample URI from the first API found.
-		if sample == "" {
-			sample = sampleURI(sc)
+		// Use the sample URI from the first API that has one defined.
+		if customSampleURI == "" {
+			customSampleURI = sampleURI(sc)
 		}
 		if err := generateRepoMetadata(sc, library, goAPI); err != nil {
 			return fmt.Errorf("failed to generate repo metadata: %w", err)
 		}
 	}
-	if err := generateREADME(library, fallbackTitle, sample, outDir); err != nil {
+	if err := generateREADME(library, fallbackTitle, customSampleURI, outDir); err != nil {
 		return fmt.Errorf("failed to generate README: %w", err)
 	}
 	if err := generateInternalVersionFile(outDir, library.CopyrightYear, library.Version); err != nil {
--- a/internal/librarian/golang/generate.go
+++ b/internal/librarian/golang/generate.go
@@ -328,8 +328,6 @@ func generateREADME(library *config.Library, fallbackTitle, sampleURI, moduleRoo
 	readmePath := filepath.Join(moduleRoot, "README.md")
 	// Skip generating README if it's in the keep list.
 	// Handwritten/veneer libraries should have the top-level README in the keep list.
-	// TODO(https://github.com/googleapis/librarian/issues/4113): investigate the difference between
-	// GAPIC and handwritten libraries.
 	for _, k := range library.Keep {
 		path := filepath.Join(moduleRoot, k)
 		if path == readmePath {
--- a/internal/librarian/golang/generate.go
+++ b/internal/librarian/golang/generate.go
@@ -74,6 +74,7 @@ func Generate(ctx context.Context, cfg *config.Config, library *config.Library, 
 	}
 
 	var fallbackTitle string
+	var sample string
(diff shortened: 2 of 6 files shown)


internal/librarian/golang/generate.go
```diff
@@ -96,12 +97,15 @@ func Generate(ctx context.Context, cfg *config.Config, library *config.Library,
...
+		if sample == "" {
+			sample = sampleURI(sc)
+		}
 		if err := generateRepoMetadata(sc, library, goAPI); err != nil {
 			return fmt.Errorf("failed to generate repo metadata: %w", err)
 		}
-
 	}
-	if err := generateREADME(library, fallbackTitle, outDir); err != nil {
+	if err := generateREADME(library, fallbackTitle, sample, outDir); err != nil {
```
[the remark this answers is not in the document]

internal/librarian/golang/generate.go
```diff
@@ -96,12 +97,15 @@ func Generate(ctx context.Context, cfg *config.Config, library *config.Library,
 		if i == 0 {
 			fallbackTitle = sc.Title
 		}
+		// Use the sample URI from the first API found.
+		if sample == "" {
+			sample = sampleURI(sc)
+		}
```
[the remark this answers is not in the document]

internal/librarian/golang/generate.go
```diff
@@ -74,6 +74,7 @@ func Generate(ctx context.Context, cfg *config.Config, library *config.Library,
 	}
 
 	var fallbackTitle string
+	var sample string
```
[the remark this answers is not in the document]

doc/api-allowlist-schema.md
```diff
@@ -14,7 +14,8 @@ This document describes the schema for the API Allowlist.
...
 | `skip_rest_numeric_enums` | list of string | Lists languages that should not pass the rest-numeric-enums flag to the generator. The special value "all" skips it for every language. If empty, all languages use numeric enums. |
 | `open_api` | string | Is the file path to an OpenAPI spec, currently in internal/testdata. This is not an official spec yet and exists only for Rust to validate OpenAPI support. |
-| `release_level` | map[string]string | Is the release level per language. Map key is the language name (e.g., "python", "rust"). Optional. If omitted, the generator default is used.<br><br>TODO(https://github.com/googleapis/librarian/issues/4834): Go uses "alpha", "beta", and "ga" instead of "preview" and "stable". We should standardize release level vocabulary across languages. |
+| `release_level` | map[string]string | Is the release level per language. Map key is the language name (e.g., "python", "rust"). Optional. If omitted, the generator default is used. |
+| `sample_uris` | map[string]string | Is the documentation URI for code samples per language. Map key is the language name (e.g., "go", "python"). Optional. If omitted, a default URI for the language is used. |
```
noahdietz, line 18: Curious, do we think we need per-language flexibility ?

I went to `https://cloud.google.com/docs/samples?l=go` and I can't really tell the difference from `https://cloud.google.com/docs/samples` _sans_ `?l=go`. That said, it does make a difference in the maps URI used (which has `#go`)

I think it might make sense to start this way and see how things actually differ, then unify if they do not.

WDYT?

internal/librarian/golang/generate.go
```diff
@@ -373,3 +379,16 @@ func transport(sc *serviceconfig.API) serviceconfig.Transport {
...
+//
+// The default value is the empty string.
+func sampleURI(sc *serviceconfig.API) string {
+	if sc == nil || sc.SampleURIs == nil {
+		return ""
+	}
+	if uri, ok := sc.SampleURIs[config.LanguageGo]; ok {
+		return uri
+	}
+	return ""
```
noahdietz, line 395: Should we put the default value here so that we don't need a `if sampleURI == ""` check in `generateReadme`?

doc/api-allowlist-schema.md
JoeWang1127: Sure, I'll make this change.

internal/librarian/golang/generate.go
JoeWang1127: We don't know which API contains the customized sample URI. If we assign a default value here, the lookup will stop at the next API.
```go
if customSampleURI == "" {
   customSampleURI = sampleURI(sc)
}
```

doc/api-allowlist-schema.md
noahdietz: > I think it might make sense to start this way and see how things actually differ, then unify if they do not.

Sorry I wasn't clear, I think your current per-language approach might be fine and we can unify later if we don't need it. But I'm not sure if it is necessary and it would be simpler to have only 1 URI. 

WDYT?

JoeWang1127: Ah, now I understand what you mean.

I thought it was a bit of early for this improvement because we don't have an actual request from Python team. We can refactor it later.

internal/librarian/golang/generate.go
```diff
@@ -341,6 +343,9 @@ func generateREADME(library *config.Library, fallbackTitle string, moduleRoot st
 		// Skip generating README if no title is available.
 		return nil
 	}
+	if sampleURI == "" {
+		sampleURI = "https://cloud.google.com/docs/samples?l=go"
```
noahdietz: Let's put the default in a `const` for easier testing and code readability.

internal/librarian/golang/generate.go
noahdietz: I see good point. I think we should still move the defaulting logic outside of the `generateREADME` logic, to just before we call it. After the API loop, we can see if `customSampleURI` is set, and if not, set to the default, then pass that into the `generateREADME`.

That will make `generateREADME` a bit cleaner in that it just produces what it is given, and the value it uses is decided by its caller.

doc/api-allowlist-schema.md
noahdietz: Cool agreed

internal/librarian/golang/generate.go
JoeWang1127: Done.

internal/librarian/golang/generate_test.go
```diff
@@ -445,6 +446,7 @@ func TestGenerateREADME(t *testing.T) {
 				APIs: []*config.API{{Path: "google/cloud/secretmanager/v1"}},
 			},
 			fallbackTitle: "Secret Manager API",
+			sampleURI:     "https://cloud.google.com/docs/samples?l=go",
```
noahdietz: Let's use the default const

internal/librarian/golang/generate_test.go
```diff
@@ -459,6 +461,7 @@ func TestGenerateREADME(t *testing.T) {
 				TitleOverride: "Custom Title",
 			},
 			fallbackTitle: "Secret Manager API",
+			sampleURI:     "https://cloud.google.com/docs/samples?l=go",
```
noahdietz: same here, let's use the default const

internal/librarian/golang/generate_test.go
```diff
@@ -431,59 +431,81 @@ func TestGenerateLibrary(t *testing.T) {
...
+			library: &config.Library{
+				Name: "secretmanager",
+				APIs: []*config.API{{Path: "google/cloud/secretmanager/v1"}},
+			},
+			fallbackTitle: "Secret Manager API",
+			sampleURI:     "https://cloud.google.com/docs/samples?l=go",
+			wantContains: []string{
+				"Secret Manager API",
+				"cloud.google.com/go/secretmanager",
+				"https://cloud.google.com/docs/samples?l=go",
```
noahdietz: Same, let's use the default const

internal/librarian/golang/generate_test.go
```diff
@@ -431,59 +431,81 @@ func TestGenerateLibrary(t *testing.T) {
...
+				Name:          "secretmanager",
+				APIs:          []*config.API{{Path: "google/cloud/secretmanager/v1"}},
+				TitleOverride: "Custom Title",
+			},
+			fallbackTitle: "Secret Manager API",
+			sampleURI:     "https://cloud.google.com/docs/samples?l=go",
+			wantContains: []string{
+				"Custom Title",
+				"cloud.google.com/go/secretmanager",
+				"https://cloud.google.com/docs/samples?l=go",
```
noahdietz: Same, let's use the default const

noahdietz (approved):
nits, then good to go

internal/librarian/golang/generate_test.go
JoeWang1127: Done.

internal/librarian/golang/generate_test.go
JoeWang1127: Done.

internal/librarian/golang/generate_test.go
JoeWang1127: Done.

internal/librarian/golang/generate_test.go
JoeWang1127: Done.

internal/librarian/golang/generate.go

PR #6174, opened 2026-05-27
chore: add post submit for Java

### LINKED ISSUE #5238: java: create a postsubmit check in librarian to verify library generation

sofisl:
Creates an integration workflow for Java generation.

Fixes #5238

### DIFF
--- a/.github/workflows/java.yaml
+++ b/.github/workflows/java.yaml
@@ -66,6 +66,7 @@ jobs:
           java-version: "17"
           distribution: "temurin"
           cache: "maven"
+     # Explicitly install Maven on this runner (it is a requirement for the workflow).
       - name: Verify Java and Maven installation
         run: |
           java -version
--- a/.github/workflows/java.yaml
+++ b/.github/workflows/java.yaml
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 name: Java
-on: [push, pull_request, merge_group]
+on: [push, pull_request, merge_group, workflow_dispatch]
 permissions:
   contents: read
   issues: write
@@ -30,8 +30,13 @@ jobs:
         with:
           java-version: "17"
           distribution: "temurin"
-      - name: Display Java version
-        run: java -version
+      - name: Verify Java and Maven installation
+        run: |
+          java -version
+          if ! command -v mvn &> /dev/null; then
+            sudo apt-get update && sudo apt-get install -y maven
+          fi
+          mvn -version
       - name: Install Java tools
         run: |
           ./.github/workflows/scripts/install_java_tools.sh "$HOME/java_tools"
@@ -60,6 +65,13 @@ jobs:
         with:
           java-version: "17"
           distribution: "temurin"
+      - name: Verify Java and Maven installation
+        run: |
+          java -version
+          if ! command -v mvn &> /dev/null; then
+            sudo apt-get update && sudo apt-get install -y maven
+          fi
+          mvn -version
       - uses: actions/setup-python@v5
         with:
           python-version: "3.12"
@@ -68,9 +80,6 @@ jobs:
         run: |
           librarian install
           echo "$HOME/java_tools/bin" >> $GITHUB_PATH
-      - name: Install Python dependencies
-        run: |
-          pip install -e ./google-cloud-java/sdk-platform-java/hermetic_build/library_generation
       - name: Run librarian generate
         working-directory: google-cloud-java
         env:
--- a/.github/workflows/java.yaml
+++ b/.github/workflows/java.yaml
@@ -45,3 +45,43 @@ jobs:
           test -x "$(which protoc-gen-java_gapic)"
       - name: Run internal/librarian/java tests and check coverage
         run: go run ./tool/cmd/coverage ./internal/librarian/java
+  integration:
+    runs-on: ubuntu-24.04
+    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+    steps:
+      - uses: actions/checkout@v6
+      - uses: ./.github/actions/setup-librarian
+      - name: Checkout google-cloud-java
+        uses: actions/checkout@v6
+        with:
+          repository: googleapis/google-cloud-java
+          path: google-cloud-java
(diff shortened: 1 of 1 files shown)


.github/workflows/java.yaml
```diff
@@ -45,3 +45,42 @@ jobs:
...
+        with:
+          python-version: "3.12"
+      - name: Run librarian install
+        working-directory: google-cloud-java
+        run: |
+          librarian install
+          echo "$HOME/java_tools/bin" >> $GITHUB_PATH
+      - name: Install Python dependencies
+        run: |
+          pip install -e ./google-cloud-java/sdk-platform-java/hermetic_build/library_generation
```
JoeWang1127: I think this step is included in `librarian install`, am I missing anything?

.github/workflows/java.yaml
```diff
@@ -45,3 +45,42 @@ jobs:
...
+      - name: Run librarian generate
+        working-directory: google-cloud-java
+        env:
+          PYTHONPATH: ${{ github.workspace }}/google-cloud-java/sdk-platform-java/hermetic_build/library_generation/owlbot
+        run: librarian generate --all
+# TODO: uncomment once Java is migrated to Librarian
```
JoeWang1127, line 90: Could you link the issue? How about using the current one?

.github/workflows/java.yaml
zhumin8: `librarian install` installs Python dependencies

.github/workflows/java.yaml
```diff
@@ -45,3 +45,42 @@ jobs:
...
+        with:
+          java-version: "17"
+          distribution: "temurin"
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - name: Run librarian install
+        working-directory: google-cloud-java
+        run: |
+          librarian install
```
zhumin8, line 83: install needs maven installed
https://github.com/googleapis/librarian/blob/ae11239d2acd37e2da6c10c306f808fe596c5467/internal/librarian/java/install.go#L40

.github/workflows/java.yaml
```diff
@@ -30,8 +30,13 @@ jobs:
...
           distribution: "temurin"
-      - name: Display Java version
-        run: java -version
+      - name: Verify Java and Maven installation
+        run: |
+          java -version
+          if ! command -v mvn &> /dev/null; then
+            sudo apt-get update && sudo apt-get install -y maven
+          fi
+          mvn -version
```
JoeWang1127, line 39: Could you just run this command, I think the github runner installs `mvn` by default.

.github/workflows/java.yaml
```diff
@@ -45,3 +50,46 @@ jobs:
...
+      - uses: ./.github/actions/setup-librarian
+      - name: Checkout google-cloud-java
+        uses: actions/checkout@v6
+        with:
+          repository: googleapis/google-cloud-java
+          path: google-cloud-java
+      - uses: actions/setup-java@v4
+        with:
+          java-version: "17"
+          distribution: "temurin"
```
JoeWang1127, line 67: How about using a maven cache to speed up the build: https://github.com/actions/setup-java#caching-maven-dependencies

.github/workflows/java.yaml
sofisl: I did look up online and this seems to be true, but earlier Min pointed out that mvn needed to be installed: https://github.com/googleapis/librarian/pull/6174#discussion_r3313009570. seems safer to do it this way?

sofisl: I also think having this step makes it more explicit if someone is trying to follow it as setup instructions.

sofisl:
FWIW, it has been difficult to test this integration given the issues of "It works on my machine" TM. I think a better way (that I will be doing moving forward) is to test by adding a specific push workflow for my specific branch: ```github.ref == 'refs/heads/addPostSubmit'```

The last integration passed: https://github.com/googleapis/librarian/actions/runs/26533972880/job/78157752279

.github/workflows/java.yaml
sofisl: created a new issue

.github/workflows/java.yaml
zhumin8: Sorry, I might be wrong. But "makes it more explicit if someone is trying to follow it as setup instructions." this makes sense to me. Maybe add a comment?

.github/workflows/java.yaml
```diff
@@ -30,8 +30,13 @@ jobs:
...
+      - name: Verify Java and Maven installation
+        run: |
+          java -version
+          if ! command -v mvn &> /dev/null; then
+            sudo apt-get update && sudo apt-get install -y maven
+          fi
+          mvn -version
       - name: Install Java tools
         run: |
           ./.github/workflows/scripts/install_java_tools.sh "$HOME/java_tools"
```
zhumin8, line 42: I think we can delete this shell script now and replace with librarian install.

Feel free to do as followup.

.github/workflows/java.yaml
```diff
@@ -45,3 +50,49 @@ jobs:
...
+        run: |
+          librarian install
+          echo "$HOME/java_tools/bin" >> $GITHUB_PATH
+      - name: Run librarian generate
+        working-directory: google-cloud-java
+        env:
+          PYTHONPATH: ${{ github.workspace }}/google-cloud-java/sdk-platform-java/hermetic_build/library_generation/owlbot
+        run: librarian generate --all
+# TODO: uncomment once Java is migrated to Librarian
+# https://github.com/googleapis/librarian/issues/6176
```
zhumin8, line 91: nit: Do you mean TODO(https://github.com/googleapis/librarian/issues/6176): ?
ref https://github.com/googleapis/librarian/blob/main/CONTRIBUTING.md#leaving-a-todo

zhumin8 (commented):
Linter is not happy. But otherwise LGTM.

zhumin8 (approved):

PR #6181, opened 2026-05-28
feat(internal/librarian/nodejs): auto-detect and resolve global path helper drift via omit_common_resources

### LINKED ISSUE #5959: nodejs: Strategic Decision on Adopt vs. Patch Legacy Improvements
We need to decide whether to accept `numericEnums = true` and `grpc_service_config.json` improvements during Librarian migration (which will result in a massive API diff on migration day), or patch the legacy pipeline.

Specifically, research if there is a middle ground where we only enable these options selectively for some services and not for others, and investigate how the `BUILD.bazel` configuration files inside individual packages might hold the key to resolving selective enablement.

### Reproduction steps

Examine the generated diff for compute and verify the Timeout and Enum differences:
```bash
# Diffs appear inside client configurations and timeout structures
git diff packages/google-cloud-compute
```

### LINKED ISSUE #6024: nodejs: be selective about passing in google/cloud/common_resources.proto
We currently always pass google/cloud/common_resources.proto into the GAPIC generator, which means it always ends up in `packages/xyz/protos/google/cloud`. 

We should only pass it in when needed - we could probably work out a heuristic for that, and then only configure exceptions to it.

quirogas:
This PR resolves the global path helper methods drift across the Node.js monorepo by introducing and auto-detecting a new **`omit_common_resources`** configuration flag.

Currently librarian's always includes `common_resources.proto` in generation target lists.
However, GCP-independent packages (Google Ads, AdMob, Billing, Quotas) do not depend on it in upstream Bazel rules.
Forcing its inclusion caused them to generate unwanted helper methods, which adds diff to the generation when compared to the expected generation.

### Changes
- Added `omit_common_resources` field under Nodejs config.
- Patched `migrate` in  to parse `BUILD.bazel` dependency graph.
   If the package targets do not depend on `common_resources_proto`, it auto-maps `omit_common_resources: true` under `nodejs:` configuration block inside the generated `librarian.yaml`.
- Updated the Node.js target resolver `resolveNodejsAPI`  to inspect the new configuration option and cleanly omit `commonProtos` from target generation if enabled.

Fixes #5959
Fixes #6024


internal/librarian/nodejs/generate.go
```diff
@@ -128,15 +128,24 @@ func generateAPI(ctx context.Context, api *config.API, library *config.Library,
 func resolveNodejsAPI(library *config.Library, api *config.API) *config.NodejsAPI {
 	res := &config.NodejsAPI{
 		Path:             api.Path,
-		AdditionalProtos: []string{commonProtos},
+		AdditionalProtos: []string{},
 	}
+
+	omitCommon := false
```
jskeet: Possibly just `omitCommon := library.Nodejs && library.Nodejs.OmitCommonResources`?

internal/config/language.go
```diff
@@ -766,6 +766,10 @@ type NodejsPackage struct {
 	// NodejsAPIs is a list of Node.js-specific API configurations.
 	NodejsAPIs []*NodejsAPI `yaml:"nodejs_apis,omitempty"`
 
+	// OmitCommonResources indicates whether to omit the default inclusion of
+	// google/cloud/common_resources.proto.
+	OmitCommonResources bool `yaml:"omit_common_resources,omitempty"`
```
jskeet: I'm not sure whether this should be here or under `NodejsAPI`. I *suspect* it doesn't matter, but it's worth considering.

jskeet (commented):
A couple of comments, but I'm fine with the overall intention. Could do with tests - I think we should be able to write tests for `resolveNodejsAPI` fairly easily.

internal/librarian/nodejs/generate.go
```diff
@@ -128,15 +128,24 @@ func generateAPI(ctx context.Context, api *config.API, library *config.Library,
 func resolveNodejsAPI(library *config.Library, api *config.API) *config.NodejsAPI {
 	res := &config.NodejsAPI{
 		Path:             api.Path,
-		AdditionalProtos: []string{commonProtos},
+		AdditionalProtos: []string{},
```
JoeWang1127: I think this line can be removed.

JoeWang1127 (commented):
Could you add unit tests?

internal/config/language.go
quirogas: Protos are passed on the per package level, so it made more sense in my head to add it to the package instead of the API, but I agree with you—other languages include it in the API struct, so we should follow suit.

internal/librarian/nodejs/generate_test.go
```diff
@@ -1214,6 +1214,65 @@ func TestResolveNodejsAPI(t *testing.T) {
...
+							OmitCommonResources: true,
+						},
+					},
+				},
+			},
+			api: &config.API{Path: "google/api/cloudquotas/v1"},
+			want: &config.NodejsAPI{
+				Path:                "google/api/cloudquotas/v1",
+				OmitCommonResources: true,
+				AdditionalProtos:    nil,
```
JoeWang1127, line 1233: nit: remove this line.

internal/librarian/nodejs/generate_test.go
```diff
@@ -1214,6 +1214,65 @@ func TestResolveNodejsAPI(t *testing.T) {
...
+					},
+				},
+			},
+			api: &config.API{Path: "google/cloud/secretmanager/v1"},
+			want: &config.NodejsAPI{
+				Path:                "google/cloud/secretmanager/v1",
+				OmitCommonResources: true,
+				AdditionalProtos:    []string{"api.proto"},
+			},
+		},
```
JoeWang1127, line 1275: Could you add unit tests to verify these scenarios:

- omit common resource is true, package-level and api-level protos combined
- omit common resource is false, all preserved.
- duplicated protos

tool/cmd/migrate/nodejs.go
JoeWang1127, line 1: Could you add unit tests for changes in this file?

quirogas: I've added the basic test. If we need more comprehensive coverage, I say we address it in a follow-up PR and using this comment to create a tracking issue.

JoeWang1127 (approved):

Issue #6182, opened 2026-05-28
google-cloud-python: Draft a testing strategy


parthea:
**Context and Problem Statement**

A recent update (PR #6171) bumped gapic-generator to version 1.34.0 in Librarian. However, this version could not be used downstream due to a regression discovered in the generated tests. Because gapic-generator lives in our monorepo, this issue could have been detected much sooner. Furthermore, because the regression occurred specifically within the generated unit tests, it raises a fundamental question: Why are we running generated unit tests downstream at all? This incident highlights a need to re-evaluate our overall testing strategy.

**Current Test Coverage**

----

Within gapic-generator, our test suite currently includes:

- Golden APIs: Validates code generation, runs generated unit tests, and performs static analysis for 5–10 golden APIs (including Showcase).

- Integration Tests: Runs system/integration tests for gapic-showcase.

- Fragment Protos: Validates code generation and runs generated unit tests for fragment protos. (Note: We may be able to deprecate these if the test cases can be covered by Showcase).

- Generator Unit Tests: Executes core unit tests for the generator itself.

----

Downstream of gapic-generator, we also run generated unit tests. When these downstream tests fail, it is typically because a specific test case was missing from either our Showcase or fragment protos during generator validation.

-----

**Proposal & Next Steps**

We should aim to increase confidence in gapic-generator releases to the point where downstream testing becomes redundant, without compromising code quality. Shifting our validation entirely upstream represents a significant change in our testing philosophy. As a next step, we should schedule a design review and discussion to align on this new strategy.  Although this specific issue mentions `gapic-generator` for Python, parts of the design may be helpful for other languages that use Librarian and will help to improve efficiency, quality and confidence of GAPIC generator releases.

PR #6183, opened 2026-05-28
feat(internal/librarian/python): update gapic-generator to v1.34.1

parthea:
The version of gapic-generator in the embedded librarian.yaml (used by librarian install) is updated to v1.34.1. This will unblock bigtable generation (b/503326310) which needs the fix from googleapis/google-cloud-python#17260 which is included in v1.34.0. v1.34.0 could not be used due to a regression. The regression was fixed in https://github.com/googleapis/google-cloud-python/pull/17273 (hence why we need gapic-generator v1.34.1 now).

See follow up issue to improve our testing strategy in https://github.com/googleapis/librarian/issues/6182

### DIFF
--- a/internal/librarian/python/librarian.yaml
+++ b/internal/librarian/python/librarian.yaml
@@ -19,7 +19,7 @@ language: python
 tools:
   pip:
     - name: gapic-generator
-      version: "1.34.0"
+      version: "1.34.1"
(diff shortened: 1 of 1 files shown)


chalmerlowe (approved):
LGTM

Issue #5850, opened 2026-05-06
proposal: add $LIBRARIAN_BIN for tool binaries


julieqiu:
We want a place to install binaries that are downloaded and installed via a tarball, such as pandoc and protoc (https://github.com/googleapis/librarian/issues/5818). 

Today, $LIBRARIAN_CACHE (default ~/.cache/librarian/) stores downloaded source tarballs at specific commits. We should introduce a $LIBRARIAN_BIN for tool binaries.

By default, this would live at $LIBRARIAN_CACHE/bin, but $LIBRARIAN_BIN can be set to any location by the user.

The structure would be `$LIBRARIAN_BIN/<tool>/<version>/<extracted tarball tree>`:

```
  $LIBRARIAN_BIN/
  ├── protoc/
  │   ├── 29.0/
  │   │   ├── bin/protoc
  │   │   └── include/google/protobuf/*.proto
  │   └── 30.0/
  │       └── ...
  └── pandoc/
      └── 3.6/
          └── bin/pandoc
```

jskeet:
Presumably we should then have an easy way of code that needs to run a tool saying "Please run tool 'x' based on the config." (Ideally IMO a single-call approach rather than every call site having to resolve the tool from config and then separately run it.)

I'd also propose that we have some value that can be specified in the tools section to mean "just run the version on my path" for ease of local development. (It would be good to be able to specify that on a per-tool basis.)

(These are both corollaries of the main proposal - not trying to broaden the initial scope, but didn't want to lose the thoughts.)

jskeet:
> Presumably we should then have an easy way of code that needs to run a tool saying "Please run tool 'x' based on the config." (Ideally IMO a single-call approach rather than every call site having to resolve the tool from config and then separately run it.)

Alternatively, we could have "please set the path to include the right directories". This may actually be what we need, so that if librarian runs tool X, but tool X runs tool Y, it finds the right version of tool Y...

jskeet:
@julieqiu Are you sufficiently confident in this that it's worth me starting to implement? Or did you already have some code for this?

julieqiu:
I'm actually less convinced by the approach based on the discussion above. If we want to be able to say "just run whatever version happens to be on my path," then it feels like that should just always be the install setup, and this is the wrong abstraction layer.

jskeet:
Will definitely wait before implementing then, but maybe noodle in a doc.

julieqiu:
[...]

PR #6191, opened 2026-05-28
feat(internal/librarian/golang): support custom Go tool installation directory

### LINKED ISSUE #6170: golang: use tools with the right path after running `librarian install`
After running `librarian install` in google-cloud-java, the tools are downloaded to local env. By default, running the tools should be directed to the directory defined with `GOPATH`.

However, this may not be the case with predefined system env, for example, the `PATH` env may define another go directory before `GOPATH` and we may run the wrong version. For example: https://github.com/googleapis/google-cloud-go/issues/14655.

We've add a section in rotation guide to prevent this happen:
```
Run which goimports and ensure it's in your go env GOPATH folder, not /usr/bin/goimports.

If /usr/bin/goimports shows up, configure your PATH environment variable to have $(go env GOPATH)/bin before /usr/bin.
```

We should enforce the tool path programmatically.

JoeWang1127:
A command wrapper `runWithEnv` and helper `mergeEnv` are introduced in the golang package to ensure that executed commands like `protoc` and `goimports` run with the custom installation directory prepended to their `PATH` environment variable.

This ensures that `protoc` can locate the newly installed Go generator plugins and `goimports` can be found by the formatting package without requiring manual shell setup.

Fixes #6170


internal/librarian/golang/install_test.go
```diff
@@ -68,10 +66,55 @@ func TestInstall_Success(t *testing.T) {
...
+			got, err := getInstallDir()
+			if err != nil {
+				t.Fatal(err)
+			}
+			if diff := cmp.Diff(test.want, got); diff != "" {
+				t.Errorf("mismatch (-want +got):\n%s", diff)
+			}
+		})
+	}
+}
```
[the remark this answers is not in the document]
JoeWang1127: Done.

internal/librarian/golang/install.go
```diff
@@ -38,14 +46,36 @@ func Install(ctx context.Context, tools *config.Tools) error {
...
+	installDir, err := getInstallDir()
+	if err != nil {
+		return err
+	}
+	// Go uses the GOBIN environment variable to determine where to install tools,
+	// and it does not allow to specify the installation directory as a command-line argument.
+	// We set it for the duration of the installation.
+	originalGOBIN := os.Getenv(goBinEnvVar)
+	defer os.Setenv(goBinEnvVar, originalGOBIN)
+	env := map[string]string{goBinEnvVar: installDir}
```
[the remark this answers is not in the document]
JoeWang1127: Done.

internal/librarian/golang/install.go
```diff
@@ -18,16 +18,24 @@ import (
...
 
 	"github.com/googleapis/librarian/internal/command"
 	"github.com/googleapis/librarian/internal/config"
 )
 
-// errMissingToolVersion indicates a go tool entry is missing its version.
-var errMissingToolVersion = errors.New("go tool missing version")
+const (
+	goBinEnvVar        = "GOBIN"
+	librarianBinEnvVar = "LIBRARIAN_INSTALL_DIR"
```
suztomo: Can we have the bin directory under LIBRARIAN_INSTALL_DIR ? In Java, we have lib and bin directory under java_tools.

JoeWang1127: Done.

noahdietz (commented):
If we merge this PR, we'd need to change how we use librarian (i.e. ensuring `go_tools` is in our PATH first).

Should we also update the various `command.Run` tool call sites in `internal/librarian/golang` to use `GOBIN=installDir` in this PR?

JoeWang1127:
> If we merge this PR, we'd need to change how we use librarian (i.e. ensuring `go_tools` is in our PATH first).
> 
> Should we also update the various `command.Run` tool call sites in `internal/librarian/golang` to use `GOBIN=installDir` in this PR?

Sure, I'll make one PR to fix the issue.

noahdietz:
@JoeWang1127 @suztomo wdyt about aligning this with https://github.com/googleapis/librarian/issues/5850 ? Can we use LIBRARIAN_BIN as described in the bug (relative to LIBRARIAN_CACHE) and then `/go_tools/bin` and `/java_tools/bin` within that so that all of our tool binaries are in one place?

I'm worried about having multiple places and env vars to configure related things.

JoeWang1127:
> @JoeWang1127 @suztomo wdyt about aligning this with https://github.com/googleapis/librarian/issues/5850 ? Can we use LIBRARIAN_BIN as described in the bug (relative to LIBRARIAN_CACHE) and then /go_tools/bin and /java_tools/bin within that so that all of our tool binaries are in one place?

I'll take a look, thanks for the suggestion.

noahdietz:
I mention this bc I think Java's `java_tools` was the exception, not the rule, in order to deliver the migration. We don't necessarily want all languages to use that, it is just what made sense for our immediate needs.

suztomo:
> LIBRARIAN_BIN

I would like the folder that contains "bin" or "lib. I think this matches the philosophy of "GOPATH" and `--prefix` in [the configure](https://www.gnu.org/prep/standards/html_node/Configuration.html).

I expect that folder replaces `java_tools` folder.

noahdietz:
> Go tools via `LIBRARIAN_BIN` or `LIBRARIAN_CACHE` environment

I think it might be a little trickier than this. I think we need a little bit of an agnostic framework to check & default these env vars.

ATM the [internal/fetch/cache.go](https://github.com/googleapis/librarian/blob/main/internal/fetch/cache.go) handles the `LIBRARIAN_CACHE` loading, but we could put this in an `internal/cache` package, then add `LIBRARIAN_BIN` support relative to that. From there the individual languages can just do `cache.ToolCachePath()` to get the resolved `LIBRARIAN_BIN` (which would default to `LIBRARIAN_CACHE/bin` as described in [the bug](https://github.com/googleapis/librarian/issues/5850)).

@suztomo WDYT about the above? We could then migrate the `java_tools` implementation to use the `internal/cache` package to decide where to install things.

Joe, If you want to roll with something here then come back to consolidate, that would be fine too. Apologies if this is too much back and forth. This is tricky one though.

JoeWang1127:
> I think it might be a little trickier than this. I think we need a little bit of an agnostic framework to check & default these env vars.

@noahdietz I think this is the best way moving forwards, however, I would create a dedicated issue for that and solve the similar issue for Java.

For this pull request, I want to solve the install directory for Go. I can restore the [7ba69be](https://github.com/googleapis/librarian/pull/6191/changes/7ba69be1a856ead46cee8d799953cdd1ccaa8324), what do you think?

noahdietz:
> I would create a dedicated issue for that and solve the similar issue for Java...For this pull request, I want to solve the install directory for Go

SGTM! I'll create the issue

> I can restore the [7ba69be](https://github.com/googleapis/librarian/pull/6191/changes/7ba69be1a856ead46cee8d799953cdd1ccaa8324), what do you think?

SGTM.

Do you think we should update our `internal/golang` `command.Run` call sites to automatically include `go_tools` in the PATH before executing?

JoeWang1127:
> Do you think we should update our internal/golang command.Run call sites to automatically include go_tools in the PATH before executing?

SGTM.

internal/librarian/golang/format_test.go
```diff
@@ -27,6 +27,19 @@ import (
 
 func TestFormat(t *testing.T) {
 	testhelper.RequireCommand(t, "goimports")
+	// goimportsPath, err := exec.LookPath("goimports")
```
noahdietz: Should this be removed or?

internal/librarian/golang/install.go
```diff
@@ -18,16 +18,25 @@ import (
...
 
 	"github.com/googleapis/librarian/internal/command"
 	"github.com/googleapis/librarian/internal/config"
 )
 
-// errMissingToolVersion indicates a go tool entry is missing its version.
-var errMissingToolVersion = errors.New("go tool missing version")
+const (
+	envGoBin        = "GOBIN"
+	envLibrarianDir = "LIBRARIAN_INSTALL_DIR"
```
noahdietz, line 32: Please add a TODO to consolidate based on https://github.com/googleapis/librarian/issues/5850

noahdietz (commented):
I think you are still working on this but thought review was requested so a few small comments and you can "re-request" review when actually ready

internal/librarian/golang/format_test.go
JoeWang1127: I was testing CI, now removed.

internal/librarian/golang/install.go
JoeWang1127: Done.

internal/librarian/golang/command.go
```diff
@@ -0,0 +1,53 @@
...
+	return command.RunInDirWithEnv(ctx, dir, env, cmd, args...)
+}
+
+// mergeEnv merges the given environment with the installation directory.
+func mergeEnv(env map[string]string) (map[string]string, error) {
+	installDir, err := getInstallDir()
+	if err != nil {
+		return nil, err
+	}
+	binDir := filepath.Join(installDir, "bin")
```
noahdietz: Nit: Let's add a constant for `"bin"` and use it everywhere

noahdietz (approved):
Nice!

internal/librarian/golang/command.go
JoeWang1127: Done.

PR #6193, opened 2026-05-29
fix(nodejs): install nvm and pnpm in the Docker nodejs target

jskeet:
The librarian install command requires pnpm to be present, and it is now installed when building a Docker image. We use nvm to give more control over which version of Node is installed.

When installing gapic-generator-typescript in librarian install, the full working directory is provided to pnpm add, so that we end up with a tool called gapic-generator-typescript installed. (Using "pnpm add ." ends up creating a directory called 5 for some reason.)

### DIFF
--- a/.github/workflows/nodejs.yaml
+++ b/.github/workflows/nodejs.yaml
@@ -30,21 +30,16 @@ jobs:
         run: |
           corepack enable pnpm
           corepack prepare pnpm@7.32.2 --activate
-      - uses: actions/setup-python@v6
-        with:
-          python-version: "3.14"
       - name: "Install Node.js dependencies (resolve cache paths)"
         id: tool-paths
         run: |
           echo "npm=$(npm config get prefix)" >> "$GITHUB_OUTPUT"
-          echo "pip=$(python3 -c 'import site; print(site.getsitepackages()[0])')" >> "$GITHUB_OUTPUT"
       - name: "Install Node.js dependencies (restore cache)"
         uses: actions/cache@v5
         id: tools-cache
         with:
           path: |
             ${{ steps.tool-paths.outputs.npm }}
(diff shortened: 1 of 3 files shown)


internal/librarian/nodejs/install.go
```diff
@@ -138,7 +138,11 @@ func installPNPMToolFromSource(ctx context.Context, env []string, tool *config.P
...
 	genDir := filepath.Join(dir, gapicGeneratorSubdir)
 	for _, cmd := range tool.Build {
-		if err := runPNPMBuildCmd(ctx, genDir, env, cmd); err != nil {
+		// We replace $PWD with the working directory, so that we can get the
+		// absolute name of a directory to "pnpm add" - using just "pnpm add ."
+		// ends up creating a directory called "5" unhelpfully.
+		effectiveCmd := strings.ReplaceAll(cmd, "$PWD", genDir)
+		if err := runPNPMBuildCmd(ctx, genDir, env, effectiveCmd); err != nil {
 			return err
 		}
```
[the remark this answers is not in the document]
jskeet: Done.

.github/workflows/nodejs.yaml
```diff
@@ -30,21 +30,16 @@ jobs:
         run: |
           corepack enable pnpm
           corepack prepare pnpm@7.32.2 --activate
-      - uses: actions/setup-python@v6
```
JoeWang1127, line 33: Why setting up python is not required anymore?

jskeet: Because we're not using synthtool any more. It was removed in #6021.

JoeWang1127: That's good to here!

JoeWang1127 (approved):

quirogas (approved):

Issue #5787, opened 2026-05-01
nodejs: Disable snippet_metadata file generation


quirogas:
To completely remove the generation and maintenance of `snippet_metadata_*.json` files across the Node.js client libraries, as these files are no longer actively use (seee [go/snippet-gen-design](http://go/snippet-gen-design)).

### 1. TypeScript Generator (`gapic-generator-typescript`)
The generator is the source of these files. We should stop it from emitting them entirely.
*   **Modify `templater.ts`:**
    *   Remove the `if (outputFilename.match(/snippet_metadata/))` block in `processOneTemplate()`. This block currently bypasses standard template rendering to manually construct the JSON metadata.
    *   Remove the `createSnippetIndexMetadata()` and `createSnippetMetadata()` helper functions.
    *   Remove the `countRegionTagLines()` and `findRegionTag()` helpers if they are not used elsewhere.
*   **Cleanup Templates:**
    *   Delete `templates/esm/typescript_samples/samples/generated/$version/snippet_metadata_$apiNamingProtoPackage.json.njk`.
    *   Delete `templates/cjs/typescript_samples/samples/generated/$version/snippet_metadata_$apiNamingProtoPackage.json.njk`.

### 2. Librarian (`googleapis/librarian`)
Librarian currently contains post-processing logic to update the hardcoded `0.1.0` version in these files.
*   **Modify `internal/librarian/nodejs/generate.go`:**
    *   Remove the call to `updateSnippetMetadataVersion()`.
    *   Delete the `updateSnippetMetadataVersion()` function and its related types (`snippetMetadata`, `clientLibrary`).
*   **Cleanup `v1small` Workarounds:**
    *   The `v1small` skip in `updateSnippetMetadataVersion()` can be removed once the function is deleted.

### 3. Cleanup PR (Monorepo-wide)
Once generation is disabled, we should perform one final sweep of the monorepo to delete all existing `snippet_metadata_*.json` files to avoid having orphaned files.

## Implementation Path
1.  Submit PR to `google-cloud-node` (or `gapic-generator-typescript` repo) to disable generation.
2.  Submit PR to `librarian` to remove the version-updating post-processor.
3.  Execute a monorepo-wide deletion PR (similar to Task 5.3) for all modern UNDERSCORE metadata files.

## Benefits
*   **Reduced Diff Noise:** Eliminates a large source of churn (version bumps and line shifts) in every generation PR.
[...]

PR #6222, opened 2026-05-30
feat(internal/librarian/nodejs): disable snippet metadata generation

quirogas:
Removes the version-updating post-processing logic that was previously used to parse, modify, and format `snippet_metadata_*.json` files. 

Generation of these legacy files has will be retired inside `gapic-generator-typescript, so librarian no longer needs to process or search for them during client generation runs. Removing this logic keeps the generation pipeline clean, lightweight, and prevents any downstream build errors.

For #5787


JoeWang1127 (approved):

JoeWang1127:
> Updates https://github.com/googleapis/librarian/issues/5787

For https://github.com/googleapis/librarian/issues/5787

Issue #5040, opened 2026-04-03
nodejs: move librarian.js post-processing logic to gapic-generator-typescript


julieqiu:
There are currently [13 librarian.js files](https://github.com/search?q=repo%3Agoogleapis%2Fgoogle-cloud-node%20librarian.js&type=code) in google-cloud-node, which apply custom post-processing scripts after the GAPIC generator runs:

```
./packages/google-cloud-sql/librarian.js
./packages/google-cloud-securitycenter/librarian.js
./packages/google-devtools-containeranalysis/librarian.js
./packages/google-storage-control/librarian.js
./packages/google-cloud-speech/librarian.js
./packages/google-cloud-vision/librarian.js
./packages/google-cloud-channel/librarian.js
./packages/google-cloud-tasks/librarian.js
./packages/google-cloud-datacatalog/librarian.js
./packages/google-iam/librarian.js
./packages/google-cloud-secretmanager/librarian.js
./packages/google-cloud-redis/librarian.js
./packages/google-cloud-documentai/librarian.js 
```

These scripts are used to apply custom modifications to the generated code, and from a quick review, seem to do things like:
- Delete specific generated files
- Replace strings with regex
- Rename methods 

Rather than maintaining these custom scripts in the generated repository, we should either move this logic directly into gapic-generator-typescript, so that we can centralize the maintenance burden and eliminate handwritten files from the generated folders.

quirogas:
It might not be necessary to move the librarian.js logic into the postprocessor. https://github.com/googleapis/google-cloud-node/issues/7718 might be the actual issue. Some, if not all, of the code in librarian.js appears to be a temporary fix for this.

jskeet:
I've looked through the librarian.js files, and I *think* only secret-manager uses it for pattern resource compatibility (and I think it's broken there anyway, unless I've missed something). So I think we'll probably need to live with this for now, but definitely revisit after migration.

suztomo:
Current status: we still have https://github.com/googleapis/google-cloud-node/blob/main/packages/google-cloud-sql/librarian.js.

PR #6238, opened 2026-06-01
feat(node): add CompatibilityResources option for NodeJS APIs

jskeet:
If compatibility resources are specified in librarian.yaml, they are passed to the GAPIC generator.

Related GAPIC generator option draft PR:
https://github.com/googleapis/google-cloud-node/pull/8405


jskeet:
(Created as a draft so we can consider if this is the right solution for #5040. Still needs names to be agreed, tests etc.)

PR #6250, opened 2026-06-01
feat(python): move allowed_namespaces to librarian.yaml

### LINKED ISSUE #6145: python: move gapic namespace allowlist to librarian.yaml
We are trying to streamline non-cloud onboarding as per http://go/sdk:expedite-non-cloud and Python onboarding currently rejects `librarian add` calls where the derived namespace is not in an allowlist. This hasn't prevented any issues that I know of, but could also be a source of configuration. We need to investigate this with the aim to remove the extraneous restrictions if we can reasonably derive these values.

noahdietz:
Moves the hard coded python namespace allowlist to `librarian.yaml` under the `default.python` section. Defaults to "allow all" if unset.

Tested in `google-cloud-python` by adding the hard coded list to the `librarian.yaml`, then running several exercises:

- `tidy` to ensure it isn't removed as "unknown" schema
- `add` with an "unallowed" API path --> error as expected
- extend allowlist, then rerun same `add` --> adds as expected
- `add` with an allowlisted API path --> adds as expected

**Important**: We need to add the following stanza to the `google-cloud-python/librarian.yaml` (#6251):
```diff
python:
+    allowed_namespaces:
+      - google.ads
+      - google.apps
+      - google.cloud
+      - google.maps
+      - google.shopping
```

Fixes #6145

### DIFF
--- a/doc/config-schema.md
+++ b/doc/config-schema.md
@@ -368,6 +368,7 @@ This document describes the schema for the librarian.yaml.
 
 | Field | Type | Description |
 | :--- | :--- | :--- |
+| `allowed_namespaces` | list of string | Contains the list of allowed GAPIC namespaces. If empty, all namespaces are allowed. |
 | `common_gapic_paths` | list of string | Contains paths which are generated for any package containing a GAPIC API. These are relative to the package's output directory, and the string "{neutral-source}" is replaced with the path to the version-neutral source code (e.g. "google/cloud/run"). If a library defines its own common_gapic_paths, they will be appended to the defaults. |
 | `library_type` | string | Is the type to emit in .repo-metadata.json. |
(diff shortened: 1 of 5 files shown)


quirogas (approved):
LGTM.

Nit: Change the description to "Updates #6145" and mark it as fixed only once the PR for librarian.yaml is merged into the Python repo.

noahdietz:
> LGTM.
> 
> Nit: Change the description to "Updates #6145" and mark it as fixed only once the PR for librarian.yaml is merged into the Python repo.

I'll do you one better: I'll file a separate bug for the repo! https://github.com/googleapis/librarian/issues/6251

Thanks for the nudge on that.

Issue #6244, opened 2026-06-01
java: google-cloud-java patch version bump breaks librarian version assumption


zhumin8:
(To scope this: this affect README.md generation when a partial release is done. When last release is a patch release, librarian generated README.md has wrong versions.
When regular releases that bumps minor, no such generate issue.)
In readme generation, we assumed released version can be worked out from current version:
https://github.com/googleapis/librarian/blob/f6099e8a93283bbaeecbad3457bbb66abce01340/internal/librarian/java/postprocess.go#L389-L396

This assumption is based on google-cloud-java uses release-please ` "versioning": "always-bump-minor",`.

https://github.com/googleapis/google-cloud-java/blob/dcc2a68b7cfc36ffcfa7e34e3ce91253a4cca074/release-please-config.json#L4

This assumption does not hold true in recent manual partial releases https://github.com/googleapis/google-cloud-java/releases/tag/v1.86.2

As result, seeing generation diffs in recent check for https://github.com/googleapis/google-cloud-java/pull/13319

````
Run git diff --exit-code
diff --git a/java-bigtable/README.md b/java-bigtable/README.md
index 07882cc1fc0..c8a2b142ba7 100644
--- a/java-bigtable/README.md
+++ b/java-bigtable/README.md
@@ -56,13 +56,13 @@ implementation 'com.google.cloud:google-cloud-bigtable'
 If you are using Gradle without BOM, add this to your dependencies:
 
 ```Groovy
-implementation 'com.google.cloud:google-cloud-bigtable:2.78.1'
+implementation 'com.google.cloud:google-cloud-bigtable:2.78.0'
 ```
 
 If you are using SBT, add this to your dependencies:
 
 ```Scala
-libraryDependencies += "com.google.cloud" % "google-cloud-bigtable" % "2.78.1"
+libraryDependencies += "com.google.cloud" % "google-cloud-bigtable" % "2.78.0"
 ```
 
 ## Authentication
@@ -452,7 +452,7 @@ Java is a registered trademark of Oracle and/or its affiliates.
 [javadocs]: https://cloud.google.com/java/docs/reference/google-cloud-bigtable/latest/history
 [stability-image]: https://img.shields.io/badge/stability-stable-green
 [maven-version-image]: https://img.shields.io/maven-central/v/com.google.cloud/google-cloud-bigtable.svg
-[maven-version-link]: https://central.sonatype.com/artifact/com.google.cloud/google-cloud-bigtable/2.78.1
[...]

PR #6252, opened 2026-06-01
fix(librarian/internal/java): explicity list released_version as config

zhumin8:
Introduces an explicit Java config for `released_version` instead of deriving it from `library.version`. Because the derive logic does not work for partial release case that bumps patch versions.
This value is used for generating README.md and sync back to versions.txt for new modules.

In this change, `released_version` needs to be explicitly specified for each library that generated. Will work on a followup to Fill derived version and allow omitting this for most libraries.

For #6244

BEGIN_COMMIT_OVERRIDE
fix(librarian/internal/java): explicitly list released_version as config
END_COMMIT_OVERRIDE

### DIFF
--- a/internal/librarian/java/add.go
+++ b/internal/librarian/java/add.go
@@ -43,6 +43,11 @@ func Add(lib *config.Library) *config.Library {
 	// so we reset it here to avoid redundancy in librarian.yaml.
 	lib.CopyrightYear = ""
 
+	if lib.Java == nil {
+		lib.Java = &config.JavaModule{}
+	}
+	lib.Java.ReleasedVersion = "0.0.0"
+
 	// We use the first API to infer the group ID.
 	// It is unrealistic for a single library to mix cloud and non-cloud APIs.
 	apiPath := lib.APIs[0].Path
@@ -66,9 +71,6 @@ func Add(lib *config.Library) *config.Library {
 }
 
 func setJavaConfig(lib *config.Library, groupID string) *config.Library {
-	if lib.Java == nil {
-		lib.Java = &config.JavaModule{}
-	}
 	lib.Java.ArtifactID = "google-" + lib.Name
 	lib.Java.GroupID = groupID
 	return lib
--- a/internal/librarian/java/add.go
+++ b/internal/librarian/java/add.go
@@ -54,11 +54,11 @@ func Add(lib *config.Library) *config.Library {
 	apiPath := lib.APIs[0].Path
 	switch {
 	case strings.HasPrefix(apiPath, "google/shopping/"):
-		return setJavaConfig(lib, "com.google.shopping")
+		return setNonCloudMavenDefaults(lib, "com.google.shopping")
 	case strings.HasPrefix(apiPath, "google/maps/"):
-		return setJavaConfig(lib, "com.google.maps")
+		return setNonCloudMavenDefaults(lib, "com.google.maps")
 	case strings.HasPrefix(apiPath, "google/ads/"):
-		return setJavaConfig(lib, "com.google.api-ads")
+		return setNonCloudMavenDefaults(lib, "com.google.api-ads")
 	}
 	if !strings.HasPrefix(apiPath, "google/cloud/") {
 		log.Printf(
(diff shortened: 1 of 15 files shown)


noahdietz (commented):
Just some drive-by nits, don't block submission on me :)

internal/librarian/java/add.go
```diff
@@ -66,9 +71,6 @@ func Add(lib *config.Library) *config.Library {
 }
 
 func setJavaConfig(lib *config.Library, groupID string) *config.Library {
-	if lib.Java == nil {
-		lib.Java = &config.JavaModule{}
-	}
```
JoeWang1127, line 71: Why not setting the value here?

internal/librarian/java/defaults_test.go
```diff
@@ -460,6 +495,9 @@ func TestValidate_Error(t *testing.T) {
 	} {
 		t.Run(test.name, func(t *testing.T) {
 			err := Validate(test.lib)
+			if err == nil {
```
JoeWang1127: I don't think you need this check.

internal/librarian/java/postgenerate.go
```diff
@@ -127,16 +124,13 @@ func PostGenerate(ctx context.Context, repoPath string, cfg *config.Config, miss
 	return nil
 }
 
-func deriveVersionLines(missingArtifacts []MissingArtifact) ([]string, error) {
+func deriveVersionLines(missingArtifacts []MissingArtifact) []string {
```
JoeWang1127: The version is not completely derived, could we change the function name?

internal/librarian/java/postgenerate_test.go
```diff
@@ -358,23 +358,20 @@ func TestDeriveVersionLines(t *testing.T) {
...
-	}
-	invalidLibrary := &config.Library{
-		Name:    "invalid",
-		Version: "1.0.0-SNAPSHOT",
+		Java: &config.JavaModule{
+			ReleasedVersion: "1.2.3",
+		},
 	}
 
 	tests := []struct {
```
JoeWang1127: Could you refine this test: https://github.com/googleapis/librarian/blob/main/doc/howwewritego.md#table-driven-tests

internal/librarian/java/postgenerate_test.go
```diff
@@ -386,25 +383,12 @@ func TestDeriveVersionLines(t *testing.T) {
...
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
-			got, err := deriveVersionLines(test.missingArtifacts)
-			if (err != nil) != test.wantErr {
-				t.Errorf("deriveVersionLines() error = %v, wantErr %v", err, test.wantErr)
-				return
-			}
+			got := deriveVersionLines(test.missingArtifacts)
 			if diff := cmp.Diff(test.want, got); diff != "" {
 				t.Errorf("deriveVersionLines() mismatch (-want +got):\n%s", diff)
```
JoeWang1127:
```suggestion
				t.Errorf("mismatch (-want +got):\n%s", diff)
```

internal/librarian/java/add.go
```diff
@@ -43,6 +43,11 @@ func Add(lib *config.Library) *config.Library {
 	// so we reset it here to avoid redundancy in librarian.yaml.
 	lib.CopyrightYear = ""
 
+	if lib.Java == nil {
+		lib.Java = &config.JavaModule{}
+	}
+	lib.Java.ReleasedVersion = "0.0.0"
```
noahdietz: Nit: Can we put this in an informatively named constant with some of the information you included in the PR body?

internal/librarian/java/defaults_test.go
```diff
@@ -460,6 +495,9 @@ func TestValidate_Error(t *testing.T) {
...
+				t.Fatal("Expect error")
+			}
```
noahdietz: Nit: This isn't necessary, `errors.Is` will fail with `error = nil, want ...`

internal/librarian/tidy_test.go
JoeWang1127, line 578: Could you create an issue to create a dedicated test for java? We can put it into post-generate stabilization milestone.

tool/cmd/migrate/java_test.go
```diff
@@ -738,9 +739,9 @@ func TestBuildConfig(t *testing.T) {
 					},
 				},
 			},
-			versions: map[string]string{
-				"google-cloud-java":           "1.79.0",
-				"google-cloud-accessapproval": "2.86.0",
+			versions: map[string]versionInfo{
+				"google-cloud-java":           {Current: "1.79.0"},
```
JoeWang1127: Why `google-cloud-java` doesn't have a released version?

zhumin8: It's not needed, as this is only used by generate. But no hurt to add it.

internal/librarian/java/add.go
zhumin8: `setJavaConfig` is called for non-cloud API path. updated name to clarify.

internal/librarian/tidy_test.go
zhumin8: https://github.com/googleapis/librarian/issues/6272

JoeWang1127 (approved):

PR #6268, opened 2026-06-02
feat(sidekick/swift): map-based pagination

### LINKED ISSUE #6235: sidekick: swift must support map-like pagination
In discovery docs is common to define "aggregated lists", which conform to the pagination API, but return `(key: [list])` pairs of items.

coryan:
Some APIs return "aggregated" lists. A typical example may be listing VM instances grouped by zone.

In this case the `items` field is (in Swift notation) a `[K: T]` map. We can convert this to a vector of tuples (in Swift notation: `[(K, T)]`) and then use normal iteration.

Fixes #6235

### DIFF
--- a/internal/sidekick/swift/annotate_message_test.go
+++ b/internal/sidekick/swift/annotate_message_test.go
@@ -107,7 +107,7 @@ func TestAnnotateMessage(t *testing.T) {
 				},
 				Pagination: &api.PaginationInfo{
 					NextPageToken: &api.Field{Name: "next_page_token", JSONName: "nextPageToken", Typez: api.TypezString},
-					PageableItem:  &api.Field{Name: "pageable_item", JSONName: "pageableItem", Typez: api.TypezString, Codec: &fieldAnnotations{Name: "secretKey", BaseFieldType: "SecretKey"}},
+					PageableItem:  &api.Field{Name: "pageable_item", JSONName: "pageableItem", Typez: api.TypezString, Repeated: true, Codec: &fieldAnnotations{Name: "secretKey", BaseFieldType: "SecretKey"}},
 				},
 			},
 			want: &messageAnnotations{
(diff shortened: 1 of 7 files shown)


internal/sidekick/swift/templates/common/message.mustache
```diff
@@ -170,7 +170,12 @@ public struct {{Codec.Name}}: Codable, Equatable, {{Codec.Model.WktPackage}}._An
   {{#Codec.IsPaginatedResponse}}
 
   public func _getPaginatedItems() -> [{{Codec.PageableItemType}}] {
+    {{^Pagination.PageableItem.Map}}
     return self.{{Codec.PageableItemField}}
+    {{/Pagination.PageableItem.Map}}
+    {{#Pagination.PageableItem.Map}}
+    return self.{{Codec.PageableItemField}}.map { ($0, $1) }
+    {{/Pagination.PageableItem.Map}}
```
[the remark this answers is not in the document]
coryan: Dead wrong: https://godbolt.org/z/83MMzTs57

internal/sidekick/swift/annotate_message_test.go
```diff
@@ -107,7 +107,7 @@ func TestAnnotateMessage(t *testing.T) {
 				},
 				Pagination: &api.PaginationInfo{
 					NextPageToken: &api.Field{Name: "next_page_token", JSONName: "nextPageToken", Typez: api.TypezString},
-					PageableItem:  &api.Field{Name: "pageable_item", JSONName: "pageableItem", Typez: api.TypezString, Codec: &fieldAnnotations{Name: "secretKey", BaseFieldType: "SecretKey"}},
+					PageableItem:  &api.Field{Name: "pageable_item", JSONName: "pageableItem", Typez: api.TypezString, Repeated: true, Codec: &fieldAnnotations{Name: "secretKey", BaseFieldType: "SecretKey"}},
```
chingor13, line 110: non-blocking: We may want to consider helpers for constructing these test fixtures because it seems easy for us to create scenarios that wouldn't be valid.

chingor13 (approved):

internal/sidekick/swift/annotate_message_test.go
coryan: Yeah, @ianthehat has been talking about having better ways to initialize these test models.

PR #6274, opened 2026-06-02
fix(internal/librarian/java): allow omitting ReleasedVersion with fill and tidy

### LINKED ISSUE #6244: java: google-cloud-java patch version bump breaks librarian version assumption

zhumin8:
Bring back `deriveLastReleasedVersion`, so that `released_version` does not needs to be explicitly listed in librarian.yaml when can be derived. Fill `released_version` when omitted and Tidy to remove `released_version`  when same as derived value.
Followup to https://github.com/googleapis/librarian/pull/6252.

Fixes #6244

### DIFF
--- a/internal/librarian/java/defaults.go
+++ b/internal/librarian/java/defaults.go
@@ -172,10 +172,6 @@ var (
 	// ErrOmitCommonResourcesConflict is returned when OmitCommonResources is true
 	// but common_resources.proto is also explicitly listed in AdditionalProtos.
 	ErrOmitCommonResourcesConflict = errors.New("conflict: OmitCommonResources is true but google/cloud/common_resources.proto is explicitly listed in AdditionalProtos")
-	// ErrReleasedVersionRequired is returned when released_version is missing.
-	ErrReleasedVersionRequired = errors.New("released_version is required under java section")
-	// ErrInvalidReleasedVersion is returned when released_version is not a valid semver.
-	ErrInvalidReleasedVersion = errors.New("invalid released_version")
 	// ErrCannotDeriveReleasedVersion is returned when released_version cannot be derived.
 	ErrCannotDeriveReleasedVersion = errors.New("cannot derive released version")
 )
@@ -184,20 +180,14 @@ var (
 // correctly formatted. It ensures that there are no conflicts in common
 // resources configuration.
 func Validate(library *config.Library) error {
-	if !library.SkipGenerate {
-		releasedVersion := ""
-		if library.Java != nil {
-			releasedVersion = library.Java.ReleasedVersion
+	if library.Version != "" {
+		if _, err := semver.Parse(library.Version); err != nil {
+			return fmt.Errorf("library %q: invalid version %q: %w", library.Name, library.Version, err)
 		}
-		if releasedVersion == "" {
-			derived, err := deriveLastReleasedVersion(library.Version)
-			if err != nil {
-				return fmt.Errorf("library %q: %w: %w", library.Name, ErrReleasedVersionRequired, err)
-			}
-			releasedVersion = derived
-		}
-		if _, err := semver.Parse(releasedVersion); err != nil {
-			return fmt.Errorf("library %q: %w: %w", library.Name, ErrInvalidReleasedVersion, err)
+	}
+	if !library.SkipGenerate && library.Java != nil && library.Java.ReleasedVersion != "" {
+		if _, err := semver.Parse(library.Java.ReleasedVersion); err != nil {
+			return fmt.Errorf("library %q: invalid released_version %q: %w", library.Name, library.Java.ReleasedVersion, err)
 		}
 	}
 	for _, api := range library.APIs {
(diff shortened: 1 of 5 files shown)


internal/librarian/java/defaults.go
```diff
@@ -162,17 +176,27 @@ var (
...
+		releasedVersion := ""
+		if library.Java != nil {
+			releasedVersion = library.Java.ReleasedVersion
+		}
+		if releasedVersion == "" {
+			derived, err := deriveLastReleasedVersion(library.Version)
+			if err != nil {
+				return fmt.Errorf("library %q: %w: %w", library.Name, ErrReleasedVersionRequired, err)
+			}
+			releasedVersion = derived
```
JoeWang1127: I think what you verified here is actually library version, right?

internal/librarian/java/defaults_test.go
```diff
@@ -168,6 +169,41 @@ func TestFill(t *testing.T) {
...
+				Output:  "java-secretmanager",
+				Java: &config.JavaModule{
+					ArtifactID:      "google-cloud-secretmanager",
+					GroupID:         "com.google.cloud",
+					ReleasedVersion: "1.1.0",
+				},
+			},
+		},
+		{
+			name: "do not fill released version if skip generate",
```
JoeWang1127, line 190: Could you add a test to verify a released version is not override by Fill?

internal/librarian/java/defaults.go
zhumin8: ah, good catch. Validate should not derive, Fill that that. At this point, value isn't filled, so validate any version that is explicitly listed in config instead..

internal/librarian/java/defaults.go
```diff
@@ -158,22 +172,22 @@ var (
...
 )
 
 // Validate checks that the Java-specific configuration for a library is
 // correctly formatted. It ensures that there are no conflicts in common
 // resources configuration.
 func Validate(library *config.Library) error {
-	if !library.SkipGenerate {
-		if library.Java == nil || library.Java.ReleasedVersion == "" {
-			return fmt.Errorf("library %q: %w", library.Name, ErrReleasedVersionRequired)
+	if library.Version != "" {
```
JoeWang1127, line 183: Do we have a scenario that the library version is empty?

JoeWang1127 (approved):

internal/librarian/java/defaults.go
zhumin8: I don't think so, should we validate at language agnostic level as required field? I can do a followup for this.

zhumin8: Ceated https://github.com/googleapis/librarian/issues/6307 to track.

PR #6275, opened 2026-06-02
feat(sidekick/rust): add `gcp.resource.destination.id` and fix incorrect `gcp.longrunning.done` status in lro traces

haphungw:
Record `gcp.resource.destination.id` directly onto Wait spans by extracting it ambiently from the `LroRecorder` context.

Also fix a bug in the generated poller loop where `gcp.longrunning.done` was incorrectly hardcoded to `false`. The completion status is now accurately parsed and recorded directly from the operation body in the final polling iteration.

### DIFF
--- a/internal/sidekick/rust/templates/crate/src/tracing.rs.mustache
+++ b/internal/sidekick/rust/templates/crate/src/tracing.rs.mustache
@@ -85,9 +85,11 @@ where T: super::stub::{{Codec.Name}} + std::fmt::Debug + Send + Sync {
         {
             if let Some(recorder) = google_cloud_lro::LroRecorder::current() {
                 if let Some(attempt) = recorder.attempt_count() {
-                    _span.record("gcp.longrunning.poll_attempt_count", attempt);
+                    _span.record("gcp.longrunning.poll_attempt_count", attempt as i64);
                 }
-                _span.record("gcp.longrunning.done", false);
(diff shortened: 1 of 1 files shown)


internal/sidekick/rust/templates/crate/src/tracing.rs.mustache
```diff
@@ -83,20 +83,25 @@ where T: super::stub::{{Codec.Name}} + std::fmt::Debug + Send + Sync {
         {{#Codec.IsLroPoller}}
         #[cfg(google_cloud_unstable_tracing)]
         {
-            if let Ok(attempt) = google_cloud_lro::POLL_ATTEMPT_COUNT.try_with(|c| *c) {
-                _span.record("gcp.longrunning.poll_attempt_count", attempt);
-                _span.record("gcp.longrunning.done", false);
+            if let Some(recorder) = google_cloud_lro::LroRecorder::current() {
```
westarle: consider putting all this in a macro like we did for client signals.

westarle (approved):

internal/sidekick/rust/templates/crate/src/tracing.rs.mustache
haphungw: macro @ https://github.com/googleapis/google-cloud-rust/pull/5838

PR #6290, opened 2026-06-03
chore(internal/fetch): export CacheDir function

JoeWang1127:
The cacheDir function in internal/fetch is exported as CacheDir to make it accessible to other packages within the module.

For #5850


noahdietz:
I think it would be cleaner to move it to its own package `internal/cache` rather than export it from `internal/fetch`.

`internal/cache` provides authoritative APIs for interacting with env vars, constructing target directory paths, etc.

`internal/fetch` defers to `internal/cache` for where to put the things it downloads.

Thoughts?

JoeWang1127:
> I think it would be cleaner to move it to its own package internal/cache rather than export it from internal/fetch.

Makes sense.

noahdietz (approved):

PR #6291, opened 2026-06-03
feat(internal/librarian/java): remove redundant keep items in librarian.yaml

### LINKED ISSUE #6163: java: remove redundant keep items in gapic client module
For manual files inside **gapic client module**, they are preserved by default, don't need to be explicit listed after https://github.com/googleapis/librarian/pull/5861.

Manual files are files that does not have any of these 2 markers:
https://github.com/googleapis/librarian/blob/b5716f6507bc1366f48d34492ea61f25918a6d9c/internal/librarian/java/clean.go#L33-L34

So if a file listed in `keep`, is inside gapic client module, and is a manual file, then it can be omitted from keep list.

Note: if a file is listed in `keep`, but 
- it has generate markers
- Or, it is outside of the client module
Then it is still necessary and should not be removed.

sofisl:
This PR removes redundant items in the `keep` property in a librarian.yaml if it is inside gapic client module, and is a manual file. This PR also introduces a small refactor: since the Java clean.go also makes a determination whether a file is manual or not, this PR extracts that logic, creates a helper function, and then reuses it for our purposes as well.

Fixes https://github.com/googleapis/librarian/issues/6163


internal/librarian/java/defaults.go
```diff
@@ -154,28 +157,54 @@ func tidyKeep(keep []string) []string {
 	return filteredKeepPaths
 }
 
+// isInGAPICModule checks if the given keepPath belongs to a generated GAPIC module of the library.
+func isInGAPICModule(library *config.Library, keepPath string) bool {
```
sofisl, line 173: @zhumin8 this is the only part of the logic I really wasn't sure about. I couldn't find any other examples in the Java code about where we make this determination.

internal/librarian/tidy.go
```diff
@@ -85,24 +85,24 @@ func RunTidyOnConfig(ctx context.Context, repoDir string, cfg *config.Config) er
...
-	if cfg.Libraries, err = tidyLibraries(cfg); err != nil {
+	if cfg.Libraries, err = tidyLibraries(repoDir, cfg); err != nil {
 		return err
 	}
 	cfg = tidyConfig(cfg)
 	return yaml.Write(filepath.Join(repoDir, config.LibrarianYAML), formatConfig(cfg))
 }
 
-func tidyLibraries(cfg *config.Config) ([]*config.Library, error) {
+func tidyLibraries(repoDir string, cfg *config.Config) ([]*config.Library, error) {
```
sofisl: @zhumin8 I also don't love changing this signature for all languages, but here are the options as I see it:

1. Change this signature and make all the corresponding languages change as well (as in this PR)
2. Change only the Java implementation signature, basically making `tidyLanguageConfig` into a switch statement where we call the Java implementation with `repoDir`.
3. An interesting suggestion from Jetski: 
>We can add a runtime-only field to the config.Library struct that is ignored during YAML serialization (yaml:"-"). This allows us to attach the runtime repository path to the library object before passing it to the tidiers, keeping the signatures untouched.
```
type Library struct {
	// ... existing fields ...
	// RepoDir is the repository root directory on the local machine.
	// This is set at runtime and is not serialized to YAML.
	RepoDir string `yaml:"-"`
}
```
I do think something like repoDir fits well here, but I also think it may be confusing if it only appears during runtime (especially for debugging purposes).

Also, this conversation could all be for nothing if it turns out we don't need the path to figure out if it's a gapic module. LMK what you think!

internal/librarian/java/defaults.go
zhumin8: see if you can reuse `cleanPatterns` instead?

sofisl: that simplifies things a. lot - thanks!

internal/librarian/java/clean.go
```diff
@@ -173,16 +176,32 @@ func isDefaultPreserved(path string) bool {
 	return itTestRegexp.MatchString(path) || versionRegexp.MatchString(path)
 }
 
+// isManualFile checks if the file at the given path is manually maintained.
+func isManualFile(path string, isDir bool, name string) (bool, error) {
```
zhumin8: Passing  `isDir` into a function named `isManualFile` sound odd. Consider removing the  `isDir`  param from  `isManualFile`  by checking early in the  defaults.go  caller.

sofisl:
@zhumin8 I tested against google-cloud-java and you're right, there was a bug! fixed in https://github.com/googleapis/librarian/pull/6291/commits/3782058c3ba657e4d567625a5c5e46f2c83ff9db

internal/librarian/java/defaults.go
```diff
@@ -89,6 +89,9 @@ func Fill(library *config.Library) (*config.Library, error) {
 // Tidy tidies the Java-specific configuration for a library by removing default
 // values.
 func Tidy(library *config.Library) (*config.Library, error) {
+	if _, err := Fill(library); err != nil {
```
zhumin8, line 92: Can you explain to me why we need to Fill when running Tidy?

sofisl: Yes! I had some help from Jetski debugging:

```
The issue was that librarian tidy was running on the raw configuration loaded from librarian.yaml without populating language-specific defaults first.

For bigquerystorage, the group_id and artifact_id are not explicitly defined in librarian.yaml (they use defaults). Because tidy didn't fill these defaults:

The Java coordinate derivation fell back to empty values.
cleanPatterns could not resolve the GAPIC module name (which should have been google-cloud-bigquerystorage).
isInGAPICModule returned false for all keep entries because it couldn't match them against the empty module name.
As a result, tidyKeep did not recognize these entries as redundant and left them untouched.
```

(this was specifically to identify why bigquerystorage's `keeps` weren't being removed)

zhumin8: SG. Do you mind attach config changes for `bigquerystorage` running Tidy with this change?

sofisl: All `keep` items are removed. I did some spot checking and this is correct.

zhumin8 (approved):
Quick note: we expect no generation diff due to this change. @sofisl can you update the pseudo version in google-cloud-java to verify after merging this?

PR #6293, opened 2026-06-04
feat(nodejs): support client_documentation and client_documentation_override

quirogas:
This PR addresses comments on the metadata generation for Node.js libraries, ensuring we support necessary overrides while strictly maintaining the established common patterns across all languages.

Instead of introducing global overrides or changing the shared generator behavior (which would cause churn in other languages), this PR isolates Node-specific requirements entirely within the Node.js generator.

#### Why do we need `client_documentation_override`?
Node.js package reference documentation URLs default to `https://cloud.google.com/nodejs/docs/reference/<pkgSuffix>/latest`. However, there are three common scenarios where this default is incorrect, requiring an explicit override in `librarian.yaml`:
1. **Legacy `googleapis.dev` Hosting**: Many established packages still host their official reference docs on the legacy `googleapis.dev` domain.
   * *Example*: `google-cloud-speech` ➔ `https://googleapis.dev/nodejs/speech/latest`
2. **Non-Cloud / Alternative Brand Scopes**: Packages belonging to other Google product areas (like Google Maps, Google Assistant/AI, or Google Ads) point to developer-specific portals.
   * *Example*: `google-ai-generativelanguage` ➔ `https://developers.generativeai.google/guide/palm_api_overview#node.js`.
3. **Package Name Suffix Discrepancies**:
   * **Unified Documentation**: All BeyondCorp sub-packages (like `google-cloud-beyondcorp-appconnections`) share a single unified documentation page: `.../reference/beyondcorp/latest`.
   * **Product vs. Package Name**: The package `google-cloud-configdelivery` generates documentation for *Infrastructure Manager*, so it must point to `.../reference/infra-manager/latest`.

> NOTE: A follow up [PR](https://github.com/googleapis/librarian/pull/6310) will update migrate to auto populate the fields correctly for `librarian.yaml`.

###  Before (What we were seeing)
The default generator stripped the `/docs` suffix from product documentation, and populated a broken issue tracker template link:
```json
{
    "api_id": "secretmanager.googleapis.com",
    "distribution_name": "@google-cloud/secret-manager",
    "client_documentation": "https://cloud.google.com/nodejs/docs/reference/secret-manager/latest",
    "issue_tracker": "https://issuetracker.google.com/issues/new?component=784854", 
    "product_documentation": "https://cloud.google.com/secret-manager/", 
    "library_type": "GAPIC_AUTO"
}

```

###  After (Using overrides)
Configuring overrides in `librarian.yaml` now produces the correct, working paths:
```json
{
    "api_id": "secretmanager.googleapis.com",
    "distribution_name": "@google-cloud/secret-manager",
    "client_documentation": "https://cloud.google.com/nodejs/docs/reference/secret-manager/latest",
    "issue_tracker": "https://issuetracker.google.com/issues/new?component=784854&template=1380926", 
    "product_documentation": "https://cloud.google.com/secret-manager/docs", 
    "library_type": "GAPIC_AUTO"
}
```


noahdietz:
> o address this, we have added three configuration fields to Node.js packages in `librarian.yaml`:
> 
> * `client_documentation_override` (also auto-inferred for `@google-cloud/` packages)
> * `issue_tracker_override`
> * `product_documentation_override`

I think we have the bottom two already in sdk.yaml: https://github.com/googleapis/librarian/blob/9552dcdce156e4b4f24ab638eff01bcf69ce17d2/internal/serviceconfig/api.go#L64-L77

We should prefer those instead of adding more related configs.

> * `client_documentation_override` (also auto-inferred for `@google-cloud/` packages)

Do other languages populate this and if so how / do they need an override ?

suztomo:
> The default generator stripped the /docs suffix from product documentation, and populated a broken issue tracker template link

I think we should fix the source ("the default generator ").

JoeWang1127:
> I think we should fix the source ("the default generator ").

I think if the fix in librarian doesn't take much of effort, we can fix it in librarian and file a bug for post migration milestone.

quirogas:
> 
> > * `client_documentation_override` (also auto-inferred for `@google-cloud/` packages)
> 
> Do other languages populate this and if so how / do they need an override ?

yes, [Python]and [Java] currently populate this field, but only for their configuration files. I've reverted my implementation and favored theirs, but I think this is worth talking.

I'll add this as a topic for the sync to see if it's worth exposing this field globally, or keeping it local to every language.


Yes, [Python](https://github.com/googleapis/librarian/blob/d123ec9cac761646dcbaa0234776fec1d4454f25/internal/config/language.go#L361)  and [Java](https://github.com/googleapis/librarian/blob/d123ec9cac761646dcbaa0234776fec1d4454f25/internal/config/language.go#L488) currently populate this field, but only for their configuration files. I have reverted my implementation in favor of theirs, but I believe this is worth discussing.

I will add this to our sync agenda to determine whether we should expose this field globally or keep it local to each language.

internal/librarian/nodejs/generate_test.go
```diff
@@ -1016,6 +1016,43 @@ func TestWriteRepoMetadata(t *testing.T) {
...
 	want.Language = cfg.Language
 	want.Repo = cfg.Repo
+	want.ClientDocumentation = "https://cloud.google.com/nodejs/docs/reference/secretmanager/latest"
+	want.ProductDocumentation = "https://cloud.google.com/secret-manager/docs"
+	if diff := cmp.Diff(want, got); diff != "" {
+		t.Errorf("mismatch (-want +got):\n%s", diff)
+	}
+}
+
+func TestWriteRepoMetadata_Override(t *testing.T) {
```
JoeWang1127: Could you combine this test with `TestWriteRepoMetadata` to a table driven test as these two tests need similar setup.

internal/librarian/nodejs/generate.go
```diff
@@ -497,7 +497,35 @@ func writeRepoMetadata(cfg *config.Config, library *config.Library, googleapisDi
 	metadata.DistributionName = DerivePackageName(library)
 	metadata.DefaultVersion = filepath.Base(library.APIs[0].Path)
 	metadata.LibraryType = repometadata.GAPICAutoLibraryType
-	return metadata.Write(outDir)
+
+	if strings.HasPrefix(metadata.DistributionName, "@google-cloud/") {
+		pkgSuffix := strings.TrimPrefix(metadata.DistributionName, "@google-cloud/")
+		metadata.ClientDocumentation = fmt.Sprintf("https://cloud.google.com/nodejs/docs/reference/%s/latest", pkgSuffix)
+	}
+
```
JoeWang1127, line 508: nit: remove this empty line.

internal/librarian/nodejs/generate.go
```diff
@@ -497,7 +497,35 @@ func writeRepoMetadata(cfg *config.Config, library *config.Library, googleapisDi
...
+	if library.Nodejs != nil && library.Nodejs.ClientDocumentationOverride != "" {
+		metadata.ClientDocumentation = library.Nodejs.ClientDocumentationOverride
+	}
+
+	if strings.HasPrefix(metadata.ProductDocumentation, "https://cloud.google.com/") {
```
JoeWang1127, line 513: I think this piece of code is redundant if we control how to generate the metadata.json.

Could you add a todo for clean up after the migration?

internal/librarian/nodejs/generate.go
```diff
@@ -497,7 +497,35 @@ func writeRepoMetadata(cfg *config.Config, library *config.Library, googleapisDi
...
+			metadata.ProductDocumentation = strings.TrimSuffix(metadata.ProductDocumentation, "/") + "/docs"
+		}
+	}
+
+	if err := metadata.Write(outDir); err != nil {
+		return err
+	}
+
+	// Go's json.MarshalIndent escapes HTML characters by default, but we want a
+	// literal ampersand in the .repo-metadata.json.
```
JoeWang1127, line 524: Why do we need a literal ampersand?

quirogas: This preserves the current .repo-metadata.json format, but we should remove it once we have retired the file.

noahdietz:
> Yes, [Python](https://github.com/googleapis/librarian/blob/d123ec9cac761646dcbaa0234776fec1d4454f25/internal/config/language.go#L361) and [Java](https://github.com/googleapis/librarian/blob/d123ec9cac761646dcbaa0234776fec1d4454f25/internal/config/language.go#L488) currently populate this field, but only for their configuration files. I have reverted my implementation in favor of theirs, but I believe this is worth discussing.
> 
> I will add this to our sync agenda to determine whether we should expose this field globally or keep it local to each language.

Looking at Java & Python, the values are language specific. So my answer is likely "Yes, but likely not as part of an sdk.yaml, rather a `config.Library` property, not `config.Library.{Language}`". Most languages need it, so why make it a language specific property and have N of them. We can do this as a follow up stabilization/cleanup item. Feel free to make a bug from this comment.

suztomo:
I don't see a problem in using the new location. Shall we use the new location? Example: https://docs.cloud.google.com/nodejs/docs/reference/speech/latest

> Hosting: Many established packages still host their official reference docs on the legacy googleapis.dev domain.
Example: google-cloud-speech ➔ https://googleapis.dev/nodejs/speech/latest

internal/librarian/nodejs/generate.go
JoeWang1127: If that's the case, could you add a todo and resolve this comment?

internal/librarian/nodejs/generate_test.go
```diff
@@ -996,28 +996,69 @@ func TestWriteRepoMetadata(t *testing.T) {
...
+		},
 	}
-	want := sample.RepoMetadata()
-	want.DistributionName = "@google-cloud/secretmanager"
-	want.Language = cfg.Language
-	want.Repo = cfg.Repo
-	if diff := cmp.Diff(want, got); diff != "" {
-		t.Errorf("mismatch (-want +got):\n%s", diff)
+
+	for _, tc := range tests {
```
JoeWang1127: Could you use this format to write table driven test: https://github.com/googleapis/librarian/blob/main/doc/howwewritego.md#table-driven-tests

internal/librarian/nodejs/generate.go
quirogas: https://github.com/googleapis/librarian/pull/6293/changes#diff-2d7149693b48b049787a2fe3ab3ceff2ba3c453194f7c5b1d75b26737b617f36R488

internal/librarian/nodejs/generate_test.go
```diff
@@ -996,28 +996,67 @@ func TestWriteRepoMetadata(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	outDir := t.TempDir()
+
```
JoeWang1127, line 999: nit: remove this empty line.

internal/librarian/nodejs/generate_test.go
```diff
@@ -996,28 +996,67 @@ func TestWriteRepoMetadata(t *testing.T) {
...
-	if err != nil {
-		t.Fatal(err)
-	}
-	want := sample.RepoMetadata()
-	want.DistributionName = "@google-cloud/secretmanager"
-	want.Language = cfg.Language
-	want.Repo = cfg.Repo
-	if diff := cmp.Diff(want, got); diff != "" {
-		t.Errorf("mismatch (-want +got):\n%s", diff)
+
```
JoeWang1127, line 1004: nit: remove this empty line.

JoeWang1127 (approved):

PR #6309, opened 2026-06-04
feat(sidekick/swift): generate `with` helper

### LINKED ISSUE #6303: sidekick: swift snippets and the argument order
The generated code for `google/cloud/storageinsights/v1` fails to compile with this error:

```
/Users/coryan/google-cloud-swift/generated/google-cloud-storageinsights-v1/Snippets/StorageInsights_UpdateReportConfig.swift:34:7: error: argument 'updateMask' must precede argument 'reportConfig'
32 |         name: "projects/\(projectId)/locations/\(locationId)/reportConfigs/\(reportConfigId)",
33 |       ),
34 |       updateMask: GoogleCloudWkt.FieldMask(paths: ["field.path1", "field.path2"]),
   |       `- error: argument 'updateMask' must precede argument 'reportConfig'
35 |     )
36 |   )
```

The are probably two solutions:

- Make the mustache template more complicated to respect the argument order
  - The code would be harder to grok too.
- Generate (and use) a `with(_ config: (inout Self) -> Void) -> Self` helper instead of a member-by-member initializer (see #6201)

coryan:
Generate a `with()` helper for each message. This provides a much nicer experience to customers, as they do not need to worry about field ordering.

It also simplifies the code generation for us.

Fixes #6303

### DIFF
--- a/internal/sidekick/swift/annotate_message.go
+++ b/internal/sidekick/swift/annotate_message.go
@@ -38,6 +38,9 @@ type messageAnnotations struct {
 	PageableItemType    string
 	DependsOn           map[string]*Dependency
 
+	// The name of a field to use in message examples.
+	SampleField string
+
 	// GatedBy is the list of package traits that enables this message.
 	//
 	// Empty unless the package is configured with `per_service_traits` enabled.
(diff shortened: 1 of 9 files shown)


AlexGhiondea (approved):
This only enables changing a single field at one time, right?

coryan:
> This only enables changing a single field at one time, right?

No, you can change many fields at a time:

```swift
let foo = Foo().with {
  $0.field0 = 42
  $0.field1 = "Hello"
}
```

Issue #6178, opened 2026-05-27
java: fix header for grafeas in generate


JoeWang1127:
The header of java files in grafeas are different from the rest of files:
```
 * Copyright 2026 The Grafeas Authors. All rights reserved.
```

We should support this in java generation, possibly adding a library config.

PR #6311, opened 2026-06-04
feat(internal/librarian/java): support alternate license header files

sofisl:
Added support for custom external alternate license header files during Java code regeneration.
When `alternate_headers` is configured with a relative file path in the library's `java` configuration block, Librarian reads that file and appends its contents as the license header instead of the standard header. A temporary, backwards-compatible migration check ensures monolithic libraries retain their existing skip behavior until their config is updated.

There will be two f/u PRs: the first in google-cloud-java to add the property and the header. The second in librarian to remove the monolithic block since (I think) it only applies to grafeas, and just rely on the alternate headers path.

For https://github.com/googleapis/librarian/issues/6178


internal/librarian/java/postprocess.go
```diff
@@ -113,7 +113,7 @@ func postProcessAPI(ctx context.Context, p postProcessParams) error {
 			return fmt.Errorf("failed to unzip %s: %w", srcjarPath, err)
 		}
 	}
-	if err := addHeadersIfRequired(p, []string{gRPCDir, protoDir}); err != nil {
+	if err := addHeaders(p, []string{gRPCDir, protoDir}); err != nil {
```
sofisl: renaming this since we'll always be adding headers

internal/librarian/java/postprocess.go
```diff
@@ -160,9 +160,11 @@ func addHeadersIfRequired(p postProcessParams, dirs []string) error {
 
 // addMissingHeaders prepends the license header to all Java files in the given directory
 // if they don't already have one.
-func addMissingHeaders(dir string) error {
-	year := time.Now().Year()
-	licenseText := buildLicenseText(year)
+func addMissingHeaders(p postProcessParams, dir string) error {
```
JoeWang1127:
```suggestion
func addMissingHeaders(params postProcessParams, dir string) error {
```

internal/librarian/java/postprocess.go
```diff
@@ -174,10 +176,26 @@ func addMissingHeaders(dir string) error {
 		if license.HasHeader(content) {
 			return nil
 		}
-		return os.WriteFile(path, append([]byte(licenseText), content...), 0644)
+		return os.WriteFile(path, append(headerText, content...), 0644)
 	})
 }
 
+func getLicenseText(p postProcessParams) ([]byte, error) {
+	if p.library != nil && p.library.Java.AlternateHeaders != "" {
```
JoeWang1127: Is it possible that `p.library.Java == nil`?

internal/librarian/java/postprocess.go
```diff
@@ -174,10 +176,26 @@ func addMissingHeaders(dir string) error {
 		if license.HasHeader(content) {
 			return nil
 		}
-		return os.WriteFile(path, append([]byte(licenseText), content...), 0644)
+		return os.WriteFile(path, append(headerText, content...), 0644)
 	})
 }
 
+func getLicenseText(p postProcessParams) ([]byte, error) {
```
JoeWang1127: Could you add a comment for this function?

internal/librarian/java/postprocess.go
```diff
@@ -174,10 +176,26 @@ func addMissingHeaders(dir string) error {
...
+	if p.library != nil && p.library.Java.AlternateHeaders != "" {
+		headerPath := filepath.Join(p.outDir, p.library.Java.AlternateHeaders)
+		b, err := os.ReadFile(headerPath)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read alternate header file %s: %w", headerPath, err)
+		}
+		if len(b) > 0 && b[len(b)-1] != '\n' {
```
JoeWang1127: Does this step add a blank line at the end of the header?

Could you check whether the formatting step can do this and remove this step?

internal/librarian/java/postprocess_test.go
```diff
@@ -432,10 +432,12 @@ func TestPostProcessAPI_SkipHeaders(t *testing.T) {
...
 		monolithic bool
-		wantHeader bool
+		altHeader  string
+		wantHeader string
 	}{
-		{"default - adds header", false, true},
-		{"monolithic - skips header", true, false},
+		{"default adds header", false, "", "/*\n * Copyright"},
+		{"monolithic skips header", true, "", "package"},
+		{"alternate adds alt header", false, "/* Alternate */\n", "/* Alternate */"},
```
JoeWang1127: Could we create a new test to verify alternative headers? Looks like the testing logic is different than the original test.

internal/librarian/java/postprocess_test.go
```diff
@@ -816,6 +825,19 @@ func TestAddMissingHeaders(t *testing.T) {
 			filename: "test.txt",
 			content:  "some text",
 		},
+		{
+			name: "alternate header",
```
JoeWang1127: Same as the previous commend, could you create a new test for alternative headers?

internal/librarian/java/postprocess_test.go
```diff
@@ -841,6 +876,22 @@ func TestAddMissingHeaders(t *testing.T) {
...
+	tmpDir := t.TempDir()
+	p := postProcessParams{
+		outDir: tmpDir,
+		library: &config.Library{
+			Java: &config.JavaModule{
+				AlternateHeaders: "missing-header.txt",
+			},
+		},
+	}
+	if err := addMissingHeaders(p, tmpDir); err == nil {
```
JoeWang1127: Could you check the error type: https://github.com/googleapis/librarian/blob/main/doc/howwewritego.md#verify-with-errorsis-and-errorsas

internal/librarian/java/postprocess.go
sofisl: It does not add a blank line so it is necessary!

internal/librarian/java/postprocess_test.go
sofisl: I would maybe argue against it, I think what we'd be testing there is an implementation strategy rather than something we want to confirm. In other words, we want to assert that a different header is added; whether that's done through the config, or a command line argument, or whatever, is a detail we can change later and maybe don't want to assert. However I'm happy to add a test if you feel strongly, just my thoughts!

internal/librarian/java/postprocess_test.go
sofisl: see response above!

internal/librarian/java/postprocess_test.go
```diff
@@ -449,24 +451,30 @@ func TestPostProcessAPI_SkipHeaders(t *testing.T) {
...
-			p := postProcessParams{
+			params := postProcessParams{
 				outDir:  outdir,
 				apiBase: apiBase,
 				library: &config.Library{Java: &config.JavaModule{}},
 				javaAPI: &config.JavaAPI{Monolithic: test.monolithic},
 			}
-			// We only care about the skip logic before restructure/cleanup.
-			if err := addHeadersIfRequired(p, []string{gRPCDir}); err != nil {
+			if test.altHeader != "" {
```
JoeWang1127: I think we should try to reduce the logic in tests.

This setup only requires if the alternate header is not empty, I think it's better to create a separate test, rather than modifying the existing tests.

What do you think.

internal/librarian/java/postprocess.go
```diff
@@ -174,16 +182,35 @@ func addMissingHeaders(dir string) error {
...
 	})
 }
 
-func copyFiles(p postProcessParams) error {
-	if p.javaAPI == nil || len(p.javaAPI.CopyFiles) == 0 {
+// If a library has an alternate header file, this function will read the contents of
+// the alternate_header property (a filepath). Otherwise it will grab the default license
+// header.
+func getLicenseText(params postProcessParams) ([]byte, error) {
+	if params.library != nil && params.library.Java != nil && params.library.Java.AlternateHeaders != "" {
```
JoeWang1127: nit: I would put the simpler logic (set year and return) inside the if statement.

internal/librarian/java/postprocess.go
```diff
@@ -174,16 +182,35 @@ func addMissingHeaders(dir string) error {
...
+		headerPath := filepath.Join(params.outDir, params.library.Java.AlternateHeaders)
+		b, err := os.ReadFile(headerPath)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read alternate header file %s: %w", headerPath, err)
+		}
+		if len(b) > 0 && b[len(b)-1] != '\n' {
```
JoeWang1127: Could you add a comment?

internal/librarian/java/postprocess_test.go
sofisl: ok sounds good, separated out the tests to simplify the logic!

internal/librarian/java/postprocess.go
```diff
@@ -174,16 +182,36 @@ func addMissingHeaders(dir string) error {
...
 			return nil
 		}
-		return os.WriteFile(path, append([]byte(licenseText), content...), 0644)
+		return os.WriteFile(path, append(headerText, content...), 0644)
 	})
 }
 
-func copyFiles(p postProcessParams) error {
-	if p.javaAPI == nil || len(p.javaAPI.CopyFiles) == 0 {
+// If a library has an alternate header file, this function will read the contents of
```
JoeWang1127: The function comment should start with `getLicenseText`

internal/librarian/java/postprocess_test.go
```diff
@@ -841,6 +904,23 @@ func TestAddMissingHeaders(t *testing.T) {
 	}
 }
 
+func TestAddMissingHeaders_AlternateHeadersError(t *testing.T) {
```
JoeWang1127:
```suggestion
func TestAddMissingHeaders_AlternateHeaders_Error(t *testing.T) {
```

internal/librarian/java/postprocess_test.go
```diff
@@ -816,6 +853,19 @@ func TestAddMissingHeaders(t *testing.T) {
 			filename: "test.txt",
 			content:  "some text",
 		},
+		{
+			name: "alternate header",
```
JoeWang1127: Is this test case different from `TestPostProcessAPI_AlternateHeaders`?

sofisl: you're right, can be safely removed

JoeWang1127: Did you remove this test? I still see it in the pull request.

JoeWang1127 (approved):

internal/librarian/java/postprocess_test.go
sofisl: sounds good, think there was a mixup but I cleaned it in [b33a88dc9624ebc9bbd61ed0f2e68565dcb50d8e](https://github.com/googleapis/librarian/pull/6311/changes/b33a88dc9624ebc9bbd61ed0f2e68565dcb50d8e)

PR #6315, opened 2026-06-05
feat(internal/cache): add `BinDirectory` and `LIBRARIAN_BIN` override

### LINKED ISSUE #6199: librarian: consolidate cache logic in dedicated pkg
> Go tools via `LIBRARIAN_BIN` or `LIBRARIAN_CACHE` environment

I think it might be a little trickier than this. I think we need a little bit of an agnostic framework to check & default these env vars.

ATM the [internal/fetch/cache.go](https://github.com/googleapis/librarian/blob/main/internal/fetch/cache.go) handles the `LIBRARIAN_CACHE` loading, but we could put this in an `internal/cache` package, then add `LIBRARIAN_BIN` support relative to that. From there the individual languages can just do `cache.ToolCachePath()` to get the resolved `LIBRARIAN_BIN` (which would default to `LIBRARIAN_CACHE/bin` as described in [the bug](https://github.com/googleapis/librarian/issues/5850)).

We could then migrate the `java_tools` implementation to use the `internal/cache` package to decide where to install things.

_Originally posted by @noahdietz in https://github.com/googleapis/librarian/issues/6191#issuecomment-4568957515_

JoeWang1127:
A `BinDirectory` function is added to internal/cache to resolve the directory where librarian bin files are stored.

It checks the new `LIBRARIAN_BIN` environment variable for a custom directory path, falling back to the bin/ subdirectory under the main cache directory if not specified.

For #5850
Fixes #6199

### DIFF
--- a/internal/cache/cache.go
+++ b/internal/cache/cache.go
@@ -22,6 +22,7 @@ import (
 )
 
 const (
+	// EnvLibrarianBin is the environment variable used to override the default bin directory.
 	EnvLibrarianBin = "LIBRARIAN_BIN"
 	// EnvLibrarianCache is the environment variable used to override the default cache directory.
 	EnvLibrarianCache = "LIBRARIAN_CACHE"
--- a/internal/cache/cache.go
+++ b/internal/cache/cache.go
@@ -21,8 +21,11 @@ import (
 	"path/filepath"
 )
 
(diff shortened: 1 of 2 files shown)


noahdietz (approved):
sweet

PR #6321, opened 2026-06-05
fix(internal/fetch): add support for symlink extraction

### LINKED ISSUE #6313: fetch: symlinks (and other unhandled types) silently ignored in tarballs
I've just spent ages tracking down a problem in the Node generator (#6312), because "librarian install" uses "fetch" to fetch and expand a tarball... and a symlink is being ignored

Two problems here:

- We should create symlinks like `tar xzf` would
- If we come across an entry in a tarball that we don't understand, we should fail rather than silently ignoring it

(There may be some entries we *want* to ignore, but we should do that explicitly.)

JoeWang1127:
The `extractTarball` function has been refactored to support extracting symlink headers (`tar.TypeSymlink`) via `os.Symlink`.

It now also explicitly rejects unsupported file types (like block devices) by returning a new `errUnsupportedFileType` error instead of silently skipping them.

Unit tests have been added to verify successful symlink extraction (TestExtractTarball_Symlink) and correct error handling of unsupported files. The existing `TestExtractTarball_PathError` has been fixed to resolve compilation errors and to correctly assert `*fs.PathError` using errors.As.

Fixes #6313


internal/fetch/fetch.go
```diff
@@ -447,6 +447,15 @@ func extractTarball(tarballPath, destDir string) error {
 				return err
 			}
 			out.Close()
+		case tar.TypeSymlink:
+			if err := os.MkdirAll(filepath.Dir(target), 0755); err != nil {
+				return err
+			}
+			if err := os.Symlink(hdr.Linkname, target); err != nil {
+				return err
+			}
```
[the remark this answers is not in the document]
JoeWang1127: Done.

internal/fetch/fetch_test.go
```diff
@@ -617,49 +618,70 @@ func createTestTarball(t *testing.T, topLevelDir string, files map[string]string
...
+func TestExtractTarball_Error(t *testing.T) {
 	for _, test := range []struct {
 		name        string
 		tarballPath func(t *testing.T) string // Function to create the test file
 		dest        func(t *testing.T) string
-		wantErr     bool
+		wantErr     error
 	}{
 		{
-			name: "tarball does not exist",
```
jskeet, line 628: Has this test case been deleted? Did it need to be?

internal/fetch/fetch.go
```diff
@@ -447,6 +449,26 @@ func extractTarball(tarballPath, destDir string) error {
...
+		case tar.TypeSymlink:
+			// Ensure the symlink target does not escape the destination directory.
+			linkTarget := hdr.Linkname
+			if filepath.IsAbs(linkTarget) {
+				return fmt.Errorf("%w: %s", errAbsSymlinks, linkTarget)
+			}
+			var resolvedTarget string
+			resolvedTarget = filepath.Join(filepath.Dir(target), linkTarget)
+			relLink, err := filepath.Rel(destDir, resolvedTarget)
+			if err != nil || strings.HasPrefix(relLink, "..") {
```
jskeet: Does this handle cases of (say) "foo/../../bar"? (We probably don't need to worry *too* much.)

jskeet (approved):
Mostly fine (thanks!) - just two questions.

internal/fetch/fetch.go
JoeWang1127: Ah, I'll update the test to fail if the file path has `..`, rather than the prefix.

internal/fetch/fetch_test.go
JoeWang1127: No, it moved to `TestExtractTarball_PathError`.

internal/fetch/fetch.go
JoeWang1127: Done.

internal/fetch/fetch.go
```diff
@@ -458,7 +458,7 @@ func extractTarball(tarballPath, destDir string) error {
 			var resolvedTarget string
 			resolvedTarget = filepath.Join(filepath.Dir(target), linkTarget)
 			relLink, err := filepath.Rel(destDir, resolvedTarget)
-			if err != nil || strings.HasPrefix(relLink, "..") {
+			if err != nil || strings.Contains(relLink, "..") {
```
jskeet, line 461: I think this will fail for valid filenames like "a..b".

I hadn't spotted that you're checking against `relLink`... it *might* have been okay before. I'm not sure how much canonicalization is going on here.

jskeet (approved):
I think this is probably okay, even though it's overly restrictive - if we get any tarballs which fail due to having a file like "a..b" (as a symlink target) we can address it at that point.

internal/fetch/fetch.go
JoeWang1127: We will address this if we have a use case.

PR #6325, opened 2026-06-05
feat(internal/librarian/java): use com.google.cloud group id for CLOUD organization APIs

### LINKED ISSUE #6267: java: add new api to use com.google.cloud group id if organization cloud in service config
If service config specifies `organization: CLOUD` in `publishing` section (e.g., [backstory](
https://github.com/googleapis/googleapis/blob/0ff3c706a5e0964c6430529c406208f2ff57cf65/backstory/backstory.yaml#L24)), it should be reasonable to assign `com.google.cloud` as group id.

Currently `librarian add` for Java does not guess group id for API path outside of `google/cloud/` dir, see:
https://github.com/googleapis/librarian/blob/e8092db48f44ead0e718ea664b5d68d9e4a0d51d/internal/librarian/java/add.go#L57-L64

cc. @blakeli0

sofisl:
When onboarding a new API in Java, we check the service config to see if the publishing organization is configured as CLOUD ([example](https://github.com/googleapis/googleapis/blob/0ff3c706a5e0964c6430529c406208f2ff57cf65/backstory/backstory.yaml#L24)). If so, we treat it as a Google Cloud API and use standard Cloud maven coordinates (group ID com.google.cloud) instead of prompting for custom coordinate configuration or warning about unrecognized non-cloud API path.

Fixes https://github.com/googleapis/librarian/issues/6267

### DIFF
--- a/internal/librarian/add.go
+++ b/internal/librarian/add.go
@@ -91,11 +91,15 @@ A typical librarian workflow for adding a new client library is:
 }
 
 func runAdd(ctx context.Context, cfg *config.Config, api string) error {
-	name, cfg, err := addLibrary(cfg, api)
+	sources, err := LoadSources(ctx, cfg.Sources)
 	if err != nil {
 		return err
 	}
-	cfg, err = resolveDependencies(ctx, cfg, name)
+	name, cfg, err := addLibrary(cfg, api, sources.Googleapis)
+	if err != nil {
+		return err
+	}
+	cfg, err = resolveDependencies(ctx, cfg, name, sources)
 	if err != nil {
 		return err
 	}
@@ -109,16 +113,16 @@ func runAdd(ctx context.Context, cfg *config.Config, api string) error {
 	return RunTidyOnConfig(ctx, ".", cfg)
 }
 
-func resolveDependencies(ctx context.Context, cfg *config.Config, name string) (*config.Config, error) {
+func resolveDependencies(ctx context.Context, cfg *config.Config, name string, sources *sources.Sources) (*config.Config, error) {
 	switch cfg.Language {
 	case config.LanguageJava:
-		lib, sources, err := setupResolve(ctx, cfg, name)
+		lib, err := FindLibrary(cfg, name)
 		if err != nil {
 			return nil, err
 		}
 		return java.ResolveMixinDependencies(cfg, lib, sources)
 	case config.LanguageRust:
-		lib, sources, err := setupResolve(ctx, cfg, name)
+		lib, err := FindLibrary(cfg, name)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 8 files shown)


internal/librarian/java/add_test.go
```diff
@@ -135,9 +135,29 @@ func TestAdd(t *testing.T) {
...
+			},
+			want: &config.Library{
+				Name: "iam",
+				APIs: []*config.API{
+					{Path: "google/iam/v3"},
+				},
+				Version:       defaultVersion,
+				CopyrightYear: "",
+				Java: &config.JavaModule{
+					ReleasedVersion: defaultReleasedVersion,
```
sofisl, line 154: we don't need to assert cloud.google.com groupID because this is the default; if it is empty, it means it will be defaulted to this in java.Fill.

internal/librarian/add.go
```diff
@@ -91,11 +91,15 @@ A typical librarian workflow for adding a new client library is:
 }
 
 func runAdd(ctx context.Context, cfg *config.Config, api string) error {
-	name, cfg, err := addLibrary(cfg, api)
+	sources, err := LoadSources(ctx, cfg.Sources)
```
sofisl, line 94: small refactor of `setupResolve` so we're only calling it once in this new method. Here, we LoadSources beforehand and pass it to `resolveDependencies` instead of calling it twice during `resolveDependencies`.

internal/serviceconfig/serviceconfig.go
```diff
@@ -195,6 +196,9 @@ func populateFromServiceConfig(api *API, cfg *Service) *API {
 		if api.ShortName == "" {
 			api.ShortName = publishing.GetApiShortName()
 		}
+		if api.Organization == "" && publishing.GetOrganization() != annotations.ClientLibraryOrganization_CLIENT_LIBRARY_ORGANIZATION_UNSPECIFIED {
+			api.Organization = publishing.GetOrganization().String()
```
sofisl, line 200: are there any other conditions here that would make us not want to set the api.Organization property?

internal/serviceconfig/api.go
```diff
@@ -103,6 +103,10 @@ type API struct {
 	// If empty, the service config is discovered in the directory specified by Path.
 	ServiceConfig string `yaml:"service_config,omitempty"`
 
+	// Organization is the organization from the service config's publishing section.
+	// This is populated from the service config during Find.
+	Organization string `yaml:"organization,omitempty"`
```
zhumin8, line 108: I don't think we want to write this back to sdk.yaml. To keep this populated at runtime only:
```suggestion
	Organization string `yaml:"-"`
```

zhumin8 (commented):
@noahdietz considering you proposal in [go/sdk:expedite-non-cloud](http://goto.google.com/sdk:expedite-non-cloud), is it worth to do this change in the meantime?  
This is inspired by the recent new API [backstory](https://github.com/googleapis/googleapis/tree/aee6df1596e23ecb4d186a8ebf7fe7a840a54ed3/backstory), it saves one step after running `librarian add backstory` that it does not block with a fake group id, instead let `librarian generate` derive group id as `com.google.cloud` and artifact id as `google-cloud-backstory`.

That said, I suppose this is pretty rare, and this just saves one manual step, in return risk it being unnoticed. How confident are we on `organization: CLOUD` from publish section?

noahdietz:
> That said, I suppose this is pretty rare, and this just saves one manual step, in return risk it being unnoticed. How confident are we on `organization: CLOUD` from publish section?

See my reply in https://github.com/googleapis/librarian/issues/6267#issuecomment-4636332689, in particular the referenced comment thread with Blake. I don't think we should derive from `organization: CLOUD` just yet, I don't think it is 100% trust worthy, and Java team still wants to weigh in on group IDs.

Agreed that this backstory case was an odd one, but not one we should be optimizing for.

zhumin8:
Thanks @noahdietz for weighing in and confirming my hesitation. Let's do nothing and keep the manual for these cases for now.

PR #6333, opened 2026-06-08
fix(nodejs): remove google/cloud/common_resources.proto after generation

### LINKED ISSUE #6024: nodejs: be selective about passing in google/cloud/common_resources.proto

jskeet:
Deletes protos/google/cloud/common_resources.proto in the post-processor for Node. We don't want the file to ever exist in the protos directory, but we do need to specify it during generation in many cases.

Fixes #6024

### DIFF
--- a/internal/librarian/nodejs/generate.go
+++ b/internal/librarian/nodejs/generate.go
@@ -38,7 +38,7 @@ import (
 )
 
 const (
-	commonProtos = "google/cloud/common_resources.proto"
+	cloudCommonResourcesProto = "google/cloud/common_resources.proto"
 )
 
 // IsMixedLibrary reports whether the library has handwritten code wrapping
@@ -149,7 +149,7 @@ func resolveNodejsAPI(library *config.Library, api *config.API) *config.NodejsAP
 
 	var protos []string
 	if !omitCommon {
-		protos = append(protos, commonProtos)
+		protos = append(protos, cloudCommonResourcesProto)
 	}
 
 	if library.Nodejs == nil {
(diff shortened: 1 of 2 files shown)


JoeWang1127:
I added a comment to the issue since I think fixing google-cloud-node would be a better approach.

However, we can use this as a temporarily solution.

internal/librarian/nodejs/generate.go
```diff
@@ -394,6 +394,12 @@ func runPostProcessor(ctx context.Context, cfg *config.Config, library *config.L
 		return fmt.Errorf("failed to remove redundant linter files: %w", err)
 	}
 
+	// Remove google/cloud/common_resources.proto from the protos directory.
+	// We don't need it in the repo (it isn't in googleapis-gen) and we don't
+	// want it to be in the diff.
+	if err := os.Remove(filepath.Join(outDir, "protos", cloudCommonResourcesProto)); err != nil && !errors.Is(err, fs.ErrNotExist) {
```
JoeWang1127, line 400: Could you add a todo to remove this piece of code after common proto is removed from google-cloud-node?

JoeWang1127 (approved):

internal/librarian/nodejs/generate.go
jskeet: No, because we want to keep it to prevent new instances. (We may want to stop copying protos entirely, but that's a different matter.)

quirogas (approved):

PR #6337, opened 2026-06-08
fix(librarian): revert pipeline state for failed library releases via deep copy

### LINKED ISSUE #6299: google-cloud-dotnet: google-cloud-sdk-librarian-dotnet-robot does not revert changes in state file for failed integration tests run before the release
For example: Spanner and BigQuery integration tests failed recently when attempting a release. The state was still updated as if the release had happened, so we had to roll it back (e.g. https://github.com/googleapis/google-cloud-dotnet/pull/14990)

JoeWang1127:
When creating a release PR, if a library fails to prepare, build, or test during the loop, its working directory changes are reverted on disk, but its in-memory state modifications in `state.pipelineState` were not reverted.

This caused subsequent successful library release commits in the same loop to serialize
and commit the incorrect version updates for the previously failed library.

Fix this by using `proto.Clone` to deep-copy the original library state before modifying it in the loop. If any of the release steps fail, we restore the original library state in the `libraries` slice and clean the working tree, ensuring that no stale version updates are written to `pipeline-state.json`.

Fixes #6299

### DIFF
--- a/internal/librarian/createreleasepr.go
+++ b/internal/librarian/createreleasepr.go
@@ -214,7 +214,7 @@ func generateReleaseCommitForEachLibrary(ctx context.Context, state *commandStat
 
 	pr := new(PullRequestContent)
 
-	for _, library := range libraries {
+	for i, library := range libraries {
 		// If we've specified a single library to release, skip all the others.
 		if cfg.LibraryID != "" && library.Id != cfg.LibraryID {
 			continue
@@ -258,10 +258,7 @@ func generateReleaseCommitForEachLibrary(ctx context.Context, state *commandStat
 		}
 
 		// Save the original values of the library state fields, so we can restore them if the release fails.
-		origVersion := library.CurrentVersion
-		origNextVersion := library.NextVersion
-		origLastReleasedCommit := library.LastReleasedCommit
-		origReleaseTimestamp := library.ReleaseTimestamp
+		originalLib := library.deepCopy()
 		// Update the pipeline state to record what we're releasing and when, and to clear the next version field.
 		// Performing this before anything else means that container code can use the pipeline state for the steps
 		// below, if it doesn't want/need to store the version separately.
@@ -275,16 +272,18 @@ func generateReleaseCommitForEachLibrary(ctx context.Context, state *commandStat
 
 		if err := cc.PrepareLibraryRelease(ctx, cfg, languageRepo.Dir, inputDirectory, library.Id, releaseVersion); err != nil {
 			addErrorToPullRequest(pr, library.Id, err, "preparing library release")
+			libraries[i] = originalLib
 			// Clean up any changes before starting the next iteration.
-			if err := rollbackLibraryState(state, library, origVersion, origNextVersion, origLastReleasedCommit, origReleaseTimestamp); err != nil {
+			if err := languageRepo.CleanWorkingTree(); err != nil {
 				return nil, err
 			}
 			continue
 		}
 		if err := cc.BuildLibrary(ctx, cfg, languageRepo.Dir, library.Id); err != nil {
 			addErrorToPullRequest(pr, library.Id, err, "building/testing library")
+			libraries[i] = originalLib
 			// Clean up any changes before starting the next iteration.
-			if err := rollbackLibraryState(state, library, origVersion, origNextVersion, origLastReleasedCommit, origReleaseTimestamp); err != nil {
+			if err := languageRepo.CleanWorkingTree(); err != nil {
 				return nil, err
 			}
 			continue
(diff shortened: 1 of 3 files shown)


internal/librarian/createreleasepr.go
```diff
@@ -438,3 +443,14 @@ func isReleaseWorthy(messages []*CommitMessage, libraryId string) bool {
...
+
+func rollbackLibraryState(state *commandState, library *statepb.LibraryState, origVersion, origNextVersion, origLastReleasedCommit string, origReleaseTimestamp *timestamppb.Timestamp) error {
+	library.CurrentVersion = origVersion
+	library.NextVersion = origNextVersion
+	library.LastReleasedCommit = origLastReleasedCommit
+	library.ReleaseTimestamp = origReleaseTimestamp
+	if err := savePipelineState(state); err != nil {
+		return err
+	}
+	return state.languageRepo.CleanWorkingTree()
```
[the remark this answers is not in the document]
JoeWang1127: Done.

internal/librarian/createreleasepr.go
```diff
@@ -438,3 +443,16 @@ func isReleaseWorthy(messages []*CommitMessage, libraryId string) bool {
...
 	return false
 }
+
+func rollbackLibraryState(state *commandState, library *statepb.LibraryState, origVersion, origNextVersion, origLastReleasedCommit string, origReleaseTimestamp *timestamppb.Timestamp) (err error) {
+	library.CurrentVersion = origVersion
+	library.NextVersion = origNextVersion
+	library.LastReleasedCommit = origLastReleasedCommit
+	library.ReleaseTimestamp = origReleaseTimestamp
+	defer func() {
+		if cleanErr := state.languageRepo.CleanWorkingTree(); cleanErr != nil {
```
jskeet: Why are we deferring this rather than just doing it immediately?

internal/librarian/createreleasepr.go
```diff
@@ -438,3 +443,16 @@ func isReleaseWorthy(messages []*CommitMessage, libraryId string) bool {
...
+			err = cleanErr
+		}
+	}()
+	return savePipelineState(state)
```
jskeet: We don't need to save the pipeline state at this point. We haven't saved the "bad" version of the state - the bug only exists because we were changing the state in memory, so *next* time the state was saved, we'd end up with bad state.

internal/librarian/createreleasepr.go
```diff
@@ -257,6 +257,11 @@ func generateReleaseCommitForEachLibrary(ctx context.Context, state *commandStat
 			return nil, err
 		}
 
+		// Save the original values of the library state fields, so we can restore them if the release fails.
+		origVersion := library.CurrentVersion
```
jskeet: It irks me to have these four separate variables. If we were expecting this code to last for longer, I'd think harder about how to clone the whole `LibraryState` here and pass the clone in, but this will definitely work.

internal/librarian/createreleasepr.go
JoeWang1127: Do you mean we only need to reset library configurations (line 448 - 451) and clean working tree?

jskeet: Yes. Cleaning the working tree will reset the file to the previous state already. Basically the files were already correct at the end of each round here - but because the state in memory had the change, the *next* iteration of the loop saved the broken data.

internal/librarian/createreleasepr.go
JoeWang1127: Done.

internal/librarian/createreleasepr.go
JoeWang1127: Ah, understood. Updated the code.

internal/librarian/createreleasepr.go
JoeWang1127: Removed this function.

internal/librarian/createreleasepr.go
```diff
@@ -438,3 +445,18 @@ func isReleaseWorthy(messages []*CommitMessage, libraryId string) bool {
...
 }
+
+func deepCopyLibrary(lib *statepb.LibraryState) *statepb.LibraryState {
+	return &statepb.LibraryState{
+		Id:                        lib.Id,
+		CurrentVersion:            lib.CurrentVersion,
+		NextVersion:               lib.NextVersion,
+		GenerationAutomationLevel: lib.GenerationAutomationLevel,
+		ReleaseAutomationLevel:    lib.ReleaseAutomationLevel,
+		ReleaseTimestamp:          proto.Clone(lib.ReleaseTimestamp).(*timestamppb.Timestamp),
```
jskeet: If `proto.Clone()` performs a deep copy already, can we not use that for LibraryState?

So `originalLib := &proto.Clone(*library).(statepb.LibraryState)` I'm guessing - the syntax might be slightly off, but basically I don't *think* we should need this code at all.

JoeWang1127: Done.

jskeet (approved):
Nice :)

PR #6346, opened 2026-06-09
feat(librarian): sync to release-please in add command

### LINKED ISSUE #5029: golang: remove functions to update state.yaml
Remove functions to update state.yaml after fully migrating off legacylibrarian.

codyoss:
Update release-please bulk configuration files when onboarding Go and Python libraries via librarian add, replacing the obsolete .librarian/state.yaml synchronization.

Synchronization is skipped if the bulk configuration files are missing. Legacy state.yaml logic is completely removed.

Example runs of branches with release-please config(ignore the reordering as I did not have keys sorted in the initial generation of the config files}:
- https://github.com/codyoss/google-cloud-python/pull/3
- https://github.com/codyoss/google-cloud-go/pull/9

Fixes: #5029
Internal Bug: b/519647333


internal/librarian/release_please.go
```diff
@@ -0,0 +1,272 @@
...
+	bulkManifestFile            = ".release-please-bulk-manifest.json"
+	bulkConfigFile              = "release-please-bulk-config.json"
+	defaultReleasePleaseVersion = "0.0.0"
+)
+
+func hasBulkReleasePleaseConfigs(dir string) bool {
+	_, errM := os.Stat(filepath.Join(dir, bulkManifestFile))
+	_, errC := os.Stat(filepath.Join(dir, bulkConfigFile))
+	return errM == nil && errC == nil
+}
```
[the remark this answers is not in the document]
codyoss: Done

internal/librarian/add.go
```diff
@@ -91,7 +89,11 @@ A typical librarian workflow for adding a new client library is:
 }
 
 func runAdd(ctx context.Context, cfg *config.Config, api string) error {
-	name, cfg, err := addLibrary(cfg, api)
+	cleanAPI := strings.TrimSuffix(strings.TrimSpace(api), "/")
+	if cleanAPI == "" || cleanAPI == "preview" || cleanAPI == "preview/" {
```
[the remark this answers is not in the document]
codyoss: Done.

codyoss:
Note there is a limitation here that python does not allow using librarian add if there are custom options configured in the librarian yaml. I think this is a good use case where instead of failing to add when there are options we should duplicate the options for the new api and log a warning for the person running the tooling to validate they are correct for the new api. In my quick audit this should work most of the time without requiring modifications.

internal/librarian/python/add.go
```diff
@@ -134,3 +142,39 @@ func versionless(apiPath string) string {
...
+	for _, api := range lib.APIs {
+		protoPackage := strings.ReplaceAll(api.Path, "/", ".")
+		version := serviceconfig.ExtractVersion(api.Path)
+
+		var versionlessPath string
+		if version != "" {
+			versionlessPath = path.Dir(api.Path)
+		} else {
+			versionlessPath = api.Path
+		}
```
noahdietz: Nit: do we need the `else` if we just default to `versionlessPath = api.Path` to start?

```suggestion
		versionlessPath := api.Path
		if version != "" {
			versionlessPath = path.Dir(api.Path)
		}
```

Not sure which reads better but I tend to try to reduce conditional blocks. Probably preference. Take your pick

internal/librarian/python/add.go
```diff
@@ -134,3 +142,39 @@ func versionless(apiPath string) string {
...
+		if !addedVersionless[versionlessPath] {
+			addedVersionless[versionlessPath] = true
+			extraFiles = append(extraFiles, versionlessPath+"/gapic_version.py")
+		}
+
+		if version != "" {
+			extraFiles = append(extraFiles, versionlessPath+"_"+version+"/gapic_version.py")
+		}
+
+		snippetMetadata := map[string]any{
```
noahdietz, line 168: Nit: Is there a release-please link we could drop in here as a reference, because this looks wonky to someone that isn't familiar with release-please.

internal/librarian/add.go
```diff
@@ -91,6 +89,9 @@ A typical librarian workflow for adding a new client library is:
 }
 
 func runAdd(ctx context.Context, cfg *config.Config, api string) error {
+	if api == "" || api == "preview" {
```
noahdietz: Not quite following why `api == "preview"` is special cased? Just because its in googleapis and doesn't have any protos in it?

I think the `api == ""` is covered by the above CLI code "errWrongAPICount" ?

internal/librarian/release_please.go
noahdietz, line 1: Nit: please name it `release_please.go` and `release_please_test.go` for consistency with casing in code being PascalCase i.e. ReleasePlease

internal/librarian/releaseplease.go
```diff
@@ -0,0 +1,258 @@
...
+		}
+	default:
+		return nil
+	}
+
+	if err := syncPackageToReleasePlease(manifest, packages, pkgPath, lib.Version, component, extraFiles); err != nil {
+		return err
+	}
+	if lib.Preview != nil {
+		if err := syncPackageToReleasePlease(manifest, packages, previewPath, lib.Preview.Version, component+"-preview", previewExtraFiles); err != nil {
```
noahdietz: The component name should not have the `-preview` suffix - it will impact the tag format otherwise and we need the tag (at least for Go) to be identical.

noahdietz (commented):
I did a first pass but need to look more closely at the release please merging code

internal/librarian/python/add.go
codyoss: I like it, done.

internal/librarian/python/add.go
codyoss: Sure, added.

internal/librarian/add.go
codyoss: You are right, I will just remove this block. Thanks.

internal/librarian/release_please.go
codyoss: done.

internal/librarian/releaseplease.go
codyoss: ack, done.

codyoss:
I will note with the merging code I purposefully did not create structs for marshaling the data. It make the code a little harder to read but allows us to be more decoupled from changes to the release-please config schema. Because we surgically edit specific fields we just marshal back out other data we don't touch. This ensures if release-please adds a new field in the future we automatically handle it without code changes.

internal/librarian/release_please.go
```diff
@@ -0,0 +1,258 @@
...
+		if lib.Preview != nil {
+			previewPath = golang.ReleasePleasePreviewPkgPrefix + lib.Name
+		}
+	case config.LanguagePython:
+		pkgPath = python.ReleasePleasePkgPrefix + lib.Name
+		extraFiles = python.ReleasePleaseExtraFiles(lib)
+		if lib.Preview != nil {
+			previewPath = python.ReleasePleasePreviewPkgPrefix + lib.Name
+			previewExtraFiles = python.ReleasePleaseExtraFiles(lib.Preview)
+		}
```
noahdietz: I don't think we should include previews in the bulk config for now.

Two issues:
1. The release-please version bumping for previews in release please does not follow the designed semantics (stable-relative bumping). 
2. We don't want Go to appear in the bulk release PR bc it won't be released from main

In my testing, and in my proposed config layout ([for Go](https://docs.google.com/document/d/1NQtgQlRKch5bqFYmbOKuiEUx2dONvd8Ld_QITDSf0eM/edit?pli=1&tab=t.byathahmsz03#heading=h.wb8byd1bhdm)), the release-please configs are separated out into `-preview` variants that we only interact with through the CLI.

So I think you can delete all of the Preview related code for now? If we want to implement stable-relative version bumps in release-please, we could at least add Python back in to bulk, and if we could disable creating the github release on main for Go preview packages only but still keep them in the version bump PR on main, then we could also onboard Go to bulk operations.

internal/librarian/release_please.go
```diff
@@ -0,0 +1,258 @@
...
+		} else {
+			if ef.isMap && !existingEF.isMap {
+				seen[ef.path] = ef
+			} else if ef.isMap && existingEF.isMap {
+				if !reflect.DeepEqual(ef.rawMap, existingEF.rawMap) {
+					return fmt.Errorf("conflicting configurations for extra-file %q:\nexisting: %v\nnew: %v", ef.path, existingEF.rawMap, ef.rawMap)
+				}
+			}
+		}
+	}
```
noahdietz, line 234: Consider returning early and de-nesting this to make it more readable
```suggestion
		if !exists {
			seen[ef.path] = ef
			return nil
		}
		if ef.isMap && !existingEF.isMap {
			seen[ef.path] = ef
			return nil
		}
		if ef.isMap && existingEF.isMap && !reflect.DeepEqual(ef.rawMap, existingEF.rawMap) {
			return fmt.Errorf("conflicting configurations for extra-file %q:\nexisting: %v\nnew: %v", ef.path, existingEF.rawMap, ef.rawMap)
		}
```

internal/librarian/release_please.go
codyoss: Seems reasonable had meant to ask you about this.

internal/librarian/release_please.go
codyoss: Took the spirt of this, good idea!

internal/librarian/golang/add.go
noahdietz, line 1: Can probably just revert changes to this file?

internal/librarian/release_please.go
```diff
@@ -0,0 +1,243 @@
...
+
+	switch cfg.Language {
+	case config.LanguageGo:
+		pkgPath = lib.Name
+	case config.LanguagePython:
+		pkgPath = python.ReleasePleasePkgPrefix + lib.Name
+		extraFiles = python.ReleasePleaseExtraFiles(lib)
+	default:
+		return nil
+	}
```
noahdietz: Do we need this anymore?

Might be simpler to default to `pkgPath := lib.Name` then `if cfg.Language == config.LanguagePython` do the special stuff.

noahdietz (approved):
Just nits left. Pretty slick, but that merging code hurts my brain lol

internal/librarian/golang/add.go
codyoss: done.

internal/librarian/release_please.go
codyoss: agree, done.

Issue #6357, opened 2026-06-10
nodejs: consider DefaultVersion requirements


jskeet:
We should either restructure our Node libraries to remove the "default version" aspect entirely, or require every library that contains generated code to explicitly state the default version, and set it during onboarding.
Relying on ordering of API paths is fragile.

For initial migration, we can just set it for all existing libraries so we know it's correct.

suztomo:
Here is the current state of `default_version` support in Bazel and the NodeJS generation pipeline:

### 1. Bazel Rule Definition (`nodejs_gapic_combined_pkg`)
The `default_version` is specified in the `nodejs_gapic_combined_pkg` rule in the API's top-level `BUILD.bazel` (e.g., [google/cloud/vision/BUILD.bazel](https://github.com/googleapis/googleapis/blob/master/google/cloud/vision/BUILD.bazel#L46)).

*(Note: The default version location was moved from `.repo-metadata.json` to `BUILD.bazel` in 2025 by Sofia in cl/808815559).*

The NodeJS Bazel rules are defined in the `google-cloud-node` monorepo (which hosts the generator):
*   **Rule Definition:** [`typescript_gapic_combined_pkg` in `core/generator/gapic-generator-typescript/rules_typescript_gapic/typescript_gapic_combined_pkg.bzl`](https://github.com/googleapis/google-cloud-node/blob/main/core/generator/gapic-generator-typescript/rules_typescript_gapic/typescript_gapic_combined_pkg.bzl#L72)
*   **Action Script:** It passes `default_version` to [`core/generator/gapic-generator-typescript/rules_typescript_gapic/combine_script.sh`](https://github.com/googleapis/google-cloud-node/blob/main/core/generator/gapic-generator-typescript/rules_typescript_gapic/combine_script.sh#L84), which runs the post-processor:
    ```bash
    $PROCESS_LIBRARIES combine-library --source-path $LIBRARY_DIR --default-version "$DEFAULT_VERSION" ...
    ```

### 2. Node.js Processing CLI (`gapic-node-processing`)
The post-processor CLI (also in `google-cloud-node` repo) uses the `default_version` to combine versioned libraries:
*   **CLI Option:** It accepts `--default-version` in [`core/packages/gapic-node-processing/src/commands/generate-combined-library.ts`](https://github.com/googleapis/google-cloud-node/blob/main/core/packages/gapic-node-processing/src/commands/generate-combined-library.ts#L47).
[...]

PR #6358, opened 2026-06-10
feat(nodejs): add a DefaultVersion field to NodeJSPackage

jskeet:
Adds a DefaultVersion field to the Node config, which is then written to .repo-metadata.json, overriding the default of "the first API path variant". The combine-libraries part of the post-processor is also provided with the default version to create the appropriate index.ts.

Like Python, our Node packages have the concept of a "default version", which is the API variant which is imported implicitly.

See #6357 for further work we'll need to consider post-migration.

### DIFF
--- a/internal/librarian/nodejs/generate.go
+++ b/internal/librarian/nodejs/generate.go
@@ -286,6 +286,7 @@ func runPostProcessor(ctx context.Context, cfg *config.Config, library *config.L
 		"combine-library",
 		"--source-path", stagingDir,
 		"--destination-path", outDir,
+		"--default-version", resolveDefaultVersion(library),
 	}
 	if library.Nodejs != nil && library.Nodejs.ESM {
 		combineArgs = append(combineArgs, "--is-esm")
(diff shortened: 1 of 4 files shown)


internal/librarian/nodejs/generate.go
```diff
@@ -668,6 +669,15 @@ func DefaultOutput(name, defaultOutput string) string {
...
 }
 
+// deriveDefaultVersion returns the default API version (v1, v1beta etc) for
+// a library
+func deriveDefaultVersion(library *config.Library) string {
+	if library.Nodejs != nil && library.Nodejs.DefaultVersion != "" {
+		return library.Nodejs.DefaultVersion
+	}
+	return filepath.Base(library.APIs[0].Path)
+}
```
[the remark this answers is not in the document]
jskeet: Done.

jskeet:
One interesting point to note - this doesn't make index.ts correct in all cases. For example, 

https://github.com/googleapis/google-cloud-node/blob/main/packages/google-cloud-maintenance-api/src/index.ts uses `v1` as the default, but https://github.com/googleapis/google-cloud-node/blob/main/packages/google-cloud-maintenance-api/.repo-metadata.json shows `v1beta`

See https://github.com/googleapis/google-cloud-node/commit/f0649df5793f5f647ddbd10a403457b66169de71 though...

internal/librarian/nodejs/generate.go
```diff
@@ -668,6 +669,20 @@ func DefaultOutput(name, defaultOutput string) string {
 	return filepath.Join(defaultOutput, name)
 }
 
+// resolveDefaultVersion returns the default API version (v1, v1beta etc) for
+// a library, using the Node-specific override if present, or the path of the
+// first API otherwise. If the library has no override and no APIs, an empty
+// string is returned.
```
quirogas, line 675:
```suggestion
// string is returned.
// TODO(https://github.com/googleapis/librarian/issues/6357): determine whether we should keep or remove this function
```

quirogas (approved):

internal/librarian/nodejs/generate.go
jskeet: Resolving for now, merging, then will unresolve and create this in another PR

PR #6361, opened 2026-06-10
feat(sidekick/swift): non-string maps

### LINKED ISSUE #5808: sidekick: swift supports maps with non-string keys
The generated code needs to support maps where the keys are not strings.  I am not sure if the "Right Thing":tm: is to generate a custom encoder / decoder for these maps (possibly) or we need to change our ProtoJSONDecoder to magically convert maps indexed by the possible key types (`bool`, `int32`, `int64`, `uint32` and `uint64`) to their representation in ProtoJSON.

coryan:
Protobuf can define maps where the key is not a string. These require custom serialization and deserialization. It may be possible to reduce the size of the generated code, but I cannot find a way at the moment.

Fixes #5808

### DIFF
--- a/internal/sidekick/swift/annotate_message.go
+++ b/internal/sidekick/swift/annotate_message.go
@@ -144,6 +144,10 @@ func (c *codec) annotateMessage(message *api.Message, model *modelAnnotations) e
 				annotations.DependsOn[dep.Name] = dep
 			}
 			if field.Map && !fieldCodec.IsStringKeyed() {
+				// In ProtoJSON map fields with non-string keys need to be
+				// serialized as JSON objects with key fields. In the generated
+				// Swift code, that requires a custom implementation of the
+				// `Decodable` and `Encodable` protocol.
 				annotations.CustomSerialization = true
 			}
 		}
--- a/internal/sidekick/swift/annotate_message.go
+++ b/internal/sidekick/swift/annotate_message.go
@@ -143,6 +143,9 @@ func (c *codec) annotateMessage(message *api.Message, model *modelAnnotations) e
 				}
 				annotations.DependsOn[dep.Name] = dep
 			}
+			if field.Map && !fieldCodec.IsStringKeyed() {
+				annotations.CustomSerialization = true
+			}
 		}
 	}
 
--- a/internal/sidekick/swift/templates/common/message.mustache
+++ b/internal/sidekick/swift/templates/common/message.mustache
@@ -84,6 +84,7 @@ public struct {{Codec.Name}}: Codable, Equatable, {{Codec.Model.WktPackage}}._An
     let container = try decoder.container(keyedBy: CodingKeys.self)
     {{#Fields}}
     {{^IsOneOf}}
+    {{#Singular}}
     {{#Optional}}
     self.{{Codec.Name}} = try container.decodeIfPresent({{{Codec.BaseFieldType}}}.self, forKey: .{{Codec.Name}})
     {{/Optional}}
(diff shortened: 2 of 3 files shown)


internal/sidekick/swift/templates/common/message.mustache
```diff
@@ -119,7 +142,24 @@ public struct {{Codec.Name}}: Codable, Equatable, {{Codec.Model.WktPackage}}._An
...
     {{^IsOneOf}}
+    {{#Map}}
+    {{#Codec.IsStringKeyed}}
+    {{! string keys have easy encoding }}
+    try container.encode(self.{{Codec.Name}}, forKey: .{{Codec.Name}})
+    {{/Codec.IsStringKeyed}}
+    {{^Codec.IsStringKeyed}}
+    do {
+      let stringKeyed = Dictionary(
+        uniqueKeysWithValues: self.{{Codec.Name}}.lazy.map { (Swift.String($0), $1) }
```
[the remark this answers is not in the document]
coryan: wrong.

internal/sidekick/swift/templates/common/message.mustache
```diff
@@ -95,6 +96,28 @@ public struct {{Codec.Name}}: Codable, Equatable, {{Codec.Model.WktPackage}}._An
...
+    self.{{Codec.Name}} = try container.decode({{{Codec.FieldType}}}.self, forKey: .{{Codec.Name}})
+    {{/Repeated}}
+    {{#Map}}
+    {{#Codec.IsStringKeyed}}
+    self.{{Codec.Name}} = try container.decode({{{Codec.FieldType}}}.self, forKey: .{{Codec.Name}})
+    {{/Codec.IsStringKeyed}}
+    {{^Codec.IsStringKeyed}}
+    self.{{Codec.Name}} = try { () throws in
+      let stringKeyed = try container.decode([Swift.String : {{{Codec.ValueType}}}].self, forKey: .{{Codec.Name}})
+      let tuples = try stringKeyed.lazy.map { (key, value) throws -> ({{{Codec.KeyType}}}, {{{Codec.ValueType}}}) in
```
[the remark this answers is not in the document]
coryan: Wrong, see `#262` in the internal repository.

internal/sidekick/swift/annotate_message.go
```diff
@@ -143,6 +143,9 @@ func (c *codec) annotateMessage(message *api.Message, model *modelAnnotations) e
 				}
 				annotations.DependsOn[dep.Name] = dep
 			}
+			if field.Map && !fieldCodec.IsStringKeyed() {
```
chingor13, line 146: nit: can we add a comment why we need custom serialization set. above it kinda makes sense because conditional is that the field name and JSON name don't match.

internal/sidekick/swift/templates/common/message.mustache
```diff
@@ -95,6 +96,28 @@ public struct {{Codec.Name}}: Codable, Equatable, {{Codec.Model.WktPackage}}._An
...
+        guard let newKey = {{{Codec.KeyType}}}(key) else {
+          throw DecodingError.typeMismatch(
+            {{{Codec.KeyType}}}.self, DecodingError.Context(codingPath: decoder.codingPath, debugDescription: "Incorrect type for map key '{{Codec.Name}}'"))
+        }
+        return (newKey, value)
+      }
+      return Dictionary(uniqueKeysWithValues: tuples)
+    }()
+    {{/Codec.IsStringKeyed}}
+    {{/Map}}
```
chingor13, line 120: non-blocking: for the future, these conditionals are getting deep and a little hard to read. Not sure on the best solution though

chingor13 (approved):

nolanleastin (approved):

internal/sidekick/swift/annotate_message.go
coryan: Done

internal/sidekick/swift/templates/common/message.mustache
coryan: Ack.  It may be time to refactor some of this to partials.

PR #6363, opened 2026-06-10
feat(internal/librarian): enable structured logging with slog

### LINKED ISSUE #6338: proposal: allow logging in librarian
Today in howwewritego.md we say:

```
Do not use `slog` in this codebase. A function should either succeed or report
failure. Logs that note progress or debug state feel useful when you write them,
but they accumulate into noise that everyone learns to ignore. Return errors.


// Good
func parseReleaseLevel(path string) (string, error) {
	// ...
	return "", fmt.Errorf("parse release level in %s: %w", path, err)
}

// Bad
func parseReleaseLevel(path string) string {
	// ...
	slog.Warn("failed to parse release level", "path", path, "error", err)
	return ""
}


For progress output, `internal/command` provides a `Verbose` flag. When enabled
by an application (e.g., via `-v`), all executed commands are printed.
```

The current rule, "return errors, don't log them", correctly targets a bad practice: swallowing errors. However, banning slog entirely misses the mark by conflating control flow (handling failures) with observability (understanding application state). Librarian is a CLI application, not a pure library (which should stick to returning errors). Relying solely on returned errors and a command-execution flag (-v) creates blind spots. We should strictly enforce returning errors, but allow slog to track what the CLI is actually doing before it fails.

Why we should have logging:
- provide visibility for long running tasks
- make the code easier to debug and reproduce errors
- if we ever automate calling the CLI it would be good to add transparency to what the CLI is doing and for those logs to appear somewhere
- `-v` only helps for exec commands whereas a lot of the code in librarian does not necessary benefit from enabling this flag 
- it may be appropriate to log a decision that was made in the code for some defaulting that may later require human review. (logging a WARN)

I am not suggesting we need to go backfill a bunch of logging this minute, but that we lift this ban and allow logging going forward.

codyoss:
Allow and configure structured logging (slog) in Librarian, changing direction from the previous policy of forbidding it. Configure a default text logger to stderr that defaults to warning level and switches to debug level when the verbose flag is set.
Additionally, document logging guidelines in doc/howwewritego.md.

Fixes: #6338


internal/librarian/librarian.go
```diff
@@ -21,6 +21,9 @@ import (
 	"errors"
 	"fmt"
 
+	"log/slog"
+	"os"
```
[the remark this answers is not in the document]
codyoss: Done.

doc/howwewritego.md
```diff
@@ -403,27 +403,22 @@ func TestTransform(t *testing.T) {
...
-	slog.Warn("failed to parse release level", "path", path, "error", err)
-	return ""
-}
-```
+- **Default**: Logs are filtered to `slog.LevelWarn` and above.
+- **Verbose**: If the verbose flag is set (e.g., `librarian -v` or `librarian --verbose`), the level is set to `slog.LevelDebug`, displaying all logs.
+
+### Best Practices
+
+1. **Prefer returning errors**: Deeply nested functions should prefer returning errors rather than logging and returning nil/empty values. Logging should be reserved for cross-cutting execution concerns, warnings where execution can proceed, or debugging info.
```
noahdietz: I'd make the second sentence its own list item

doc/howwewritego.md
```diff
@@ -403,27 +403,22 @@ func TestTransform(t *testing.T) {
...
+2. **Provide structured context**: Always pass structured key-value pairs to log messages:
```
noahdietz: Can we also add something about being consistent in the keys we use?

I don't mean defining each one, but let's make sure that when we start using a key like`"api"` it is used consistently. Hopefully AIs will pick up on this and keep us honest.

internal/librarian/verbose_test.go
```diff
@@ -42,6 +45,15 @@ func TestVerboseFlag(t *testing.T) {
 			if command.Verbose != test.wantVerbose {
 				t.Errorf("command.Verbose = %t, want %t", command.Verbose, test.wantVerbose)
 			}
+
+			// Verify slog level configuration.
+			ctx := t.Context()
+			if got := slog.Default().Enabled(ctx, slog.LevelDebug); got != test.wantVerbose {
+				t.Errorf("slog.Default().Enabled(Debug) = %t, want %t", got, test.wantVerbose)
```
JoeWang1127: use cmp.Diff: https://github.com/googleapis/librarian/blob/main/doc/howwewritego.md#use-cmpdiff-for-comparisons

doc/howwewritego.md
codyoss: Done.

doc/howwewritego.md
codyoss: Yes. Once we start purposefully start adding more observability into librarian we can create loggers with wrapped context so all log lines have these keys built in without needing to add them every time. But I think things like that needs to come with time.

internal/librarian/verbose_test.go
codyoss: Done.

But I do disagree with needing cmp for simple value comparisons. It is great for structs or complex data but seems like overkill for simple value comparisons. I think that is what the guide is getting at when it is saying use cmp instead of deepequals which could be used for deeply nested data.

noahdietz (approved):

Issue #5748, opened 2026-04-30
sidekick: swift, discovery docs, and bytes serialization


coryan:
Services using discovery doc use a different base64 encoding vs. the normal ProtoJSON encoding:

<img width="1998" height="107" alt="Image" src="https://github.com/user-attachments/assets/fbaafd5f-1807-438b-93da-f648e4fdbef8" />

https://developers.google.com/discovery/v1/type-format

We need to handle the encoding properly.

PR #6389, opened 2026-06-11
refactor(sidekick/swift): `annotateField()` return

coryan:
Return the annotations from `annotateField()` so we can use them in `annotateMessage()` directly without a downcast.

Towards #5748


nolanleastin (approved):
lgtm

Issue #6388, opened 2026-06-11
antique-librarian: the unit tests failed in antique-librarian branch


JoeWang1127:
See the error log: https://github.com/googleapis/librarian/actions/runs/27350196204/job/80809642163

We should fix the CI before migrating antique-librarian in case we need to make changes.

JoeWang1127:
Resolved via https://github.com/googleapis/librarian/pull/6421

PR #6418, opened 2026-06-13
chore: update dependencies

JoeWang1127:
Update dependencies.

For https://github.com/googleapis/librarian/issues/6388


jskeet (commented):
Could we separate the dependency changes from the changes to avoid Builder.WriteString? It's unclear whether the latter changes are needed after updating dependencies.

JoeWang1127:
> Could we separate the dependency changes from the changes to avoid Builder.WriteString?

I can try to do it.

JoeWang1127:
The CI failed with:
```
--- FAIL: TestGolangCILint (13.47s)
    all_test.go:100: /opt/hostedtoolcache/go/1.26.4/x64/bin/go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@latest run: exit status 1
Error:         internal/docker/environment.go:110:4: QF1012: Use fmt.Fprintf(...) instead of WriteString(fmt.Sprintf(...)) (staticcheck)
        			builder.WriteString(fmt.Sprintf("%s=%s\n", variable.Name, value))
        			^
Error:         internal/docker/environment.go:113:4: QF1012: Use fmt.Fprintf(...) instead of WriteString(fmt.Sprintf(...)) (staticcheck)
        			builder.WriteString(fmt.Sprintf("# No value for %s\n", variable.Name))
        			^
Error:         internal/librarian/mergereleasepr.go:463:3: QF1012: Use fmt.Fprintf(...) instead of WriteString(fmt.Sprintf(...)) (staticcheck)
        		builder.WriteString(fmt.Sprintf("%s: %s\n", suspectRelease.LibraryID, suspectRelease.Reason))
        		^
        3 issues:
        * staticcheck: 3
        exit status 1
```

@jskeet Do you think we can merge this in without fixing the CI errors? We can make this pull request only updates dependencies.

I will create another pull request to fix this.

jskeet:
Okay, I hadn't realised this was fixing an error that would happen regardless of the dependency updates. In that case, this is fine with all the changes together.

jskeet (approved):

Issue #6298, opened 2026-06-04
java: create unified methodOperations function supporting delete, duplicate, and deprecate actions in postprocessing package


yangyzs:
Implement the unified `methodOperations` post-processing function in Go to support deleting, duplicating, and deprecating Java methods.

See [go/sdk-librarian-mod-gen](https://docs.google.com/document/d/1a4R-tNZCbGN5TUTMRR6bG-CCDob1x0R7YDzmh5-9oC4/edit?resourcekey=0-XeSTzmtjPWsC6mmkpr0hMQ&tab=t.0#heading=h.lg3orppkodpi) for more details.

yangyzs:
Completed.

PR #6436, opened 2026-06-15
feat(postprocessing): implement Java method deletion

yangyzs:
This change adds DeleteMethod to internal/postprocessing as part of issue #6298. It uses regex pre-cleaning and brace tracking to safely delete Java method bodies.

For #6298

### DIFF
--- a/internal/postprocessing/delete.go
+++ b/internal/postprocessing/delete.go
@@ -21,6 +21,8 @@ import (
 	"os"
 	"regexp"
 	"strings"
+
+	"github.com/googleapis/librarian/internal/config"
 )
 
 var (
@@ -36,8 +38,13 @@ var (
 
 // DeleteMethod deletes a method from a Java file.
 // It handles brace counting to remove the entire method body.
+//
+// Note: funcName must be the full method declaration line (including modifiers
+// and return type, e.g., "public static void foo()") to avoid matching a substring
+// of another method name (e.g., "foo()" matching inside "afoo()").
+// TODO: Enforce word boundaries during matching as a safety net against accidental substring matches.
 func DeleteMethod(path, funcName, language string) error {
-	if language != "java" {
+	if language != config.LanguageJava {
 		return fmt.Errorf("%w: %s", errUnsupportedLanguage, language)
 	}
 	if !strings.Contains(funcName, "(") || !strings.Contains(funcName, ")") {
@@ -62,13 +69,11 @@ func DeleteMethod(path, funcName, language string) error {
 	}
 	finalStart, finalEnd := adjustDeleteBounds(data, idx, closeBraceIdx+1)
 	newContent := append(data[:finalStart], data[finalEnd:]...)
-	if err := os.WriteFile(path, newContent, 0644); err != nil {
-		return fmt.Errorf("writing updated file: %w", err)
-	}
-	return nil
+	return os.WriteFile(path, newContent, 0644)
 }
 
-// cleanJavaCode replaces comments and strings with equal-length spaces to keep offsets.
+// cleanJavaCode replaces comments, strings, and char literals with spaces.
+// This prevents brace count mismatch while keeping the original offsets.
 func cleanJavaCode(content []byte) []byte {
 	return javaCleanRegex.ReplaceAllFunc(content, func(match []byte) []byte {
 		return bytes.Repeat([]byte(" "), len(match))
@@ -77,11 +82,10 @@ func cleanJavaCode(content []byte) []byte {
 
 // findOpeningBrace returns the index of the first '{' after start, or error if ';' is found first.
 func findOpeningBrace(cleaned []byte, start int) (int, error) {
-	for i := start; i < len(cleaned); i++ {
-		c := cleaned[i]
+	for i, c := range cleaned[start:] {
 		switch c {
 		case '{':
-			return i, nil
+			return start + i, nil
 		case ';':
 			return -1, errOpeningBraceNotFound
 		}
@@ -92,30 +96,49 @@ func findOpeningBrace(cleaned []byte, start int) (int, error) {
 // findClosingBrace returns the index of the matching closing brace.
 func findClosingBrace(cleaned []byte, openBraceIdx int) (int, error) {
 	braceCount := 1
-	for i := openBraceIdx + 1; i < len(cleaned); i++ {
-		c := cleaned[i]
+	start := openBraceIdx + 1
+	for i, c := range cleaned[start:] {
 		switch c {
 		case '{':
 			braceCount++
 		case '}':
 			braceCount--
 			if braceCount == 0 {
-				return i, nil
+				return start + i, nil
 			}
 		}
 	}
 	return -1, errClosingBraceNotFound
 }
 
-// adjustDeleteBounds expands the delete range to remove entire lines if the method is on its own lines.
+// adjustDeleteBounds adjusts the delete range [start, end) to include the entire line(s)
+// if the method is the only content on those lines.
+//
+// Examples:
+// - For a block method on its own lines:
+//
+//	class T {
+//	    void m() {
+//	    }
+//	}
+//
+//	If deleting "void m() {\n    }", it returns bounds for "    void m() {\n    }\n"
+//	to remove the leading indentation and trailing newline.
+//
+// - For an inline method:
+//
+//	class T { void m() {} void o() {} }
+//
+//	If deleting "void m() {}", it returns the exact bounds of "void m() {}"
+//	to preserve "void o() {}".
 func adjustDeleteBounds(content []byte, start, end int) (int, int) {
 	startLineIdx := 0
-	if lastNL := bytes.LastIndexByte(content[:start], '\n'); lastNL != -1 {
-		startLineIdx = lastNL + 1
+	if lastNl := bytes.LastIndexByte(content[:start], '\n'); lastNl != -1 {
+		startLineIdx = lastNl + 1
 	}
 	endLineIdx := len(content)
-	if nextNL := bytes.IndexByte(content[end:], '\n'); nextNL != -1 {
-		endLineIdx = end + nextNL
+	if nextNl := bytes.IndexByte(content[end:], '\n'); nextNl != -1 {
+		endLineIdx = end + nextNl
 	}
 	if len(bytes.TrimSpace(content[startLineIdx:start])) == 0 && len(bytes.TrimSpace(content[end:endLineIdx])) == 0 {
 		finalEnd := endLineIdx
--- a/internal/postprocessing/delete.go
+++ b/internal/postprocessing/delete.go
@@ -15,6 +15,7 @@
 package postprocessing
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"os"
@@ -46,9 +47,8 @@ func DeleteMethod(path, funcName, language string) error {
 	if err != nil {
 		return fmt.Errorf("reading file: %w", err)
 	}
-	content := string(data)
-	cleaned := cleanJavaCode(content)
-	idx := strings.Index(cleaned, funcName)
+	cleaned := cleanJavaCode(data)
+	idx := bytes.Index(cleaned, []byte(funcName))
 	if idx == -1 {
 		return fmt.Errorf("%w %q in %s", errMethodNotFound, funcName, path)
 	}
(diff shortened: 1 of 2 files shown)


internal/postprocessing/delete.go
```diff
@@ -0,0 +1,128 @@
...
+	errClosingBraceNotFound = errors.New("closing brace not found")
+	errInvalidSignature     = errors.New("invalid method signature")
+	// javaCleanRegex matches Java text blocks, double-quoted strings, char literals,
+	// block comments, and line comments.
+	javaCleanRegex = regexp.MustCompile(`"""[\s\S]*?"""|"(?:[^"\\]|\\.)*"|'(?:[^'\\]|\\.)*'|/\*[\s\S]*?\*/|//.*`)
+)
+
+// DeleteMethod deletes a method from a Java file.
+// It handles brace counting to remove the entire method body.
+func DeleteMethod(path, funcName, language string) error {
```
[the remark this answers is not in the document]
yangyzs: The legacy `remove_method` in `owlbot.py` also used exact string matching and left Javadocs/annotations dangling. You can see here: [java-bigquerydatatransfer/owlbot.py#L24-L29](https://github.com/googleapis/google-cloud-java/blob/main/java-bigquerydatatransfer/owlbot.py#L24-L29).

Also, in the new design, adding deprecation annotations happens *after* deletions, so they don't conflict. Leftover annotations/Javadocs can also be cleaned up using `replace` / `replace_regex` rules if needed.

So I don't feel it is necessary for now, but please let me know if you see any issues.

internal/postprocessing/delete.go
```diff
@@ -0,0 +1,128 @@
...
+
+package postprocessing
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"regexp"
+	"strings"
+)
```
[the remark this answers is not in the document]
yangyzs: Done. Switched to `[]byte` to avoid string copying and unnecessary memory allocations.

internal/postprocessing/delete.go
```diff
@@ -0,0 +1,128 @@
...
+	errInvalidSignature     = errors.New("invalid method signature")
+	// javaCleanRegex matches Java text blocks, double-quoted strings, char literals,
+	// block comments, and line comments.
+	javaCleanRegex = regexp.MustCompile(`"""[\s\S]*?"""|"(?:[^"\\]|\\.)*"|'(?:[^'\\]|\\.)*'|/\*[\s\S]*?\*/|//.*`)
+)
+
+// DeleteMethod deletes a method from a Java file.
+// It handles brace counting to remove the entire method body.
+func DeleteMethod(path, funcName, language string) error {
+	if language != "java" {
```
JoeWang1127: Use `config.LanguageJava`.

internal/postprocessing/delete.go
```diff
@@ -0,0 +1,128 @@
...
+	if err != nil {
+		return fmt.Errorf("deleting method %s: %w", funcName, err)
+	}
+	closeBraceIdx, err := findClosingBrace(cleaned, openBraceIdx)
+	if err != nil {
+		return fmt.Errorf("deleting method %s: %w", funcName, err)
+	}
+	finalStart, finalEnd := adjustDeleteBounds(data, idx, closeBraceIdx+1)
+	newContent := append(data[:finalStart], data[finalEnd:]...)
+	if err := os.WriteFile(path, newContent, 0644); err != nil {
```
JoeWang1127: How about `return os.WriteFile(path, newContent, 0644)`

internal/postprocessing/delete.go
```diff
@@ -0,0 +1,128 @@
...
+		return fmt.Errorf("writing updated file: %w", err)
+	}
+	return nil
+}
+
+// cleanJavaCode replaces comments and strings with equal-length spaces to keep offsets.
+func cleanJavaCode(content []byte) []byte {
```
JoeWang1127, line 78: Why do we need this function?

internal/postprocessing/delete.go
```diff
@@ -0,0 +1,128 @@
...
+	return javaCleanRegex.ReplaceAllFunc(content, func(match []byte) []byte {
+		return bytes.Repeat([]byte(" "), len(match))
+	})
+}
+
+// findOpeningBrace returns the index of the first '{' after start, or error if ';' is found first.
+func findOpeningBrace(cleaned []byte, start int) (int, error) {
+	for i := start; i < len(cleaned); i++ {
```
JoeWang1127: Could we use `for i, c := range cleaned`?

internal/postprocessing/delete.go
```diff
@@ -0,0 +1,128 @@
...
+			if braceCount == 0 {
+				return i, nil
+			}
+		}
+	}
+	return -1, errClosingBraceNotFound
+}
+
+// adjustDeleteBounds expands the delete range to remove entire lines if the method is on its own lines.
+func adjustDeleteBounds(content []byte, start, end int) (int, int) {
```
JoeWang1127, line 135: I think this function is a bit difficult to understand, could you add some examples in the comments? Also, could you add some unit tests?

internal/postprocessing/delete_test.go
```diff
@@ -0,0 +1,278 @@
...
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestDeleteMethod(t *testing.T) {
+	for _, test := range []struct {
+		name     string
+		content  string
+		funcName string
+		language string
```
JoeWang1127: We currently only support `java`, I think its reasonable to set the value to `java` in the function call.

internal/postprocessing/delete_test.go
```diff
@@ -0,0 +1,278 @@
...
+class Test {
+    public void myMethod() {} }`,
+			funcName: "public void myMethod()",
+			language: "java",
+			want: `package com.example;
+class Test {
+     }`,
+		},
+	} {
+		t.Run(test.name, func(t *testing.T) {
```
JoeWang1127, line 203: How about using parallel testing?

internal/postprocessing/delete_test.go
```diff
@@ -0,0 +1,278 @@
...
+			funcName: "public void myMethod()",
+			language: "java",
+			want: `package com.example;
+class Test {
+    public void otherMethod() {
+    }
+}`,
+		},
+		{
+			name: "braces in strings",
```
JoeWang1127, line 74: This test is interesting, if we only print `{`, the test case would fail, right?

internal/postprocessing/delete_test.go
```diff
@@ -0,0 +1,278 @@
...
+			language: "java",
+			wantErr:  errOpeningBraceNotFound,
+		},
+		{
+			name:     "invalid signature",
+			content:  "class T { void m() {} }",
+			funcName: "void m",
+			language: "java",
+			wantErr:  errInvalidSignature,
+		},
```
JoeWang1127, line 259: Could you add a test to verify `errClosingBraceNotFound`?

internal/postprocessing/delete_test.go
```diff
@@ -0,0 +1,278 @@
...
+func TestDeleteMethod_Error(t *testing.T) {
+	for _, test := range []struct {
+		name     string
+		content  string
+		funcName string
+		language string
+		wantErr  error
+	}{
+		{
+			name:     "function not found",
```
JoeWang1127, line 233: Could you write a test to verify `errMethodNotFound` with a function whose name is a substring of the target function name.

For example, the file contains a function, `foo`, but we want to remove function `foobar`.

JoeWang1127:
Two things in our meeting:
- Write a regex to verify the function exists in the file, return an error if no match or multiple matches. We expect full function name, i.e., `<access modifiers> <static> <return type> <function name> (<function parameter>)`
- Write more unit tests to verify the regex can find the right function to delete.

We could create follow up pull requests to do these things.

internal/postprocessing/delete.go
```diff
@@ -0,0 +1,128 @@
...
+	closeBraceIdx, err := findClosingBrace(cleaned, openBraceIdx)
+	if err != nil {
+		return fmt.Errorf("deleting method %s: %w", funcName, err)
+	}
+	finalStart, finalEnd := adjustDeleteBounds(content, idx, closeBraceIdx+1)
+	newContent := content[:finalStart] + content[finalEnd:]
+	if err := os.WriteFile(path, []byte(newContent), 0644); err != nil {
+		return fmt.Errorf("writing updated file: %w", err)
+	}
+	return nil
```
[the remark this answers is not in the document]
yangyzs: Done.

internal/postprocessing/delete.go
yangyzs: Done.

internal/postprocessing/delete.go
yangyzs: Done.

internal/postprocessing/delete.go
yangyzs: To ignore braces inside comments and strings during brace matching, preventing count mismatches from unbalanced braces (like `// {` or `" { "`). Added detailed explanation in the comment.

internal/postprocessing/delete.go
yangyzs: Done.

internal/postprocessing/delete.go
yangyzs: Done. Added examples and comments.

internal/postprocessing/delete_test.go
yangyzs: Done.

internal/postprocessing/delete_test.go
yangyzs: Done.

internal/postprocessing/delete_test.go
yangyzs: It won't fail because `cleanJavaCode` clears these braces. Added explanation for this test.

internal/postprocessing/delete_test.go
yangyzs: Added TODO to address this in future PRs.

internal/postprocessing/delete_test.go
yangyzs: Added.

internal/postprocessing/delete.go
```diff
@@ -0,0 +1,151 @@
...
+	javaCleanRegex = regexp.MustCompile(`"""[\s\S]*?"""|"(?:[^"\\]|\\.)*"|'(?:[^'\\]|\\.)*'|/\*[\s\S]*?\*/|//.*`)
+)
+
+// DeleteMethod deletes a method from a Java file.
+// It handles brace counting to remove the entire method body.
+//
+// Note: funcName must be the full method declaration line (including modifiers
+// and return type, e.g., "public static void foo()") to avoid matching a substring
+// of another method name (e.g., "foo()" matching inside "afoo()").
+// TODO: Enforce word boundaries during matching as a safety net against accidental substring matches.
```
JoeWang1127: Add an issue link in TODO: https://github.com/googleapis/librarian/blob/main/CONTRIBUTING.md#leaving-a-todo

You can use the current issue.

internal/postprocessing/delete_test.go
```diff
@@ -0,0 +1,326 @@
...
+			content:     "class T { void m() {} void o() {} }",
+			startSub:    "void m()",
+			endSub:      "}",
+			wantDeleted: "void m() {}",
+		},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			content := []byte(test.content)
+			start := strings.Index(test.content, test.startSub)
+			if start == -1 {
```
JoeWang1127: Does this line always returns `false`? How about removing this check?

internal/postprocessing/delete_test.go
```diff
@@ -0,0 +1,326 @@
...
+				t.Fatalf("startSub %q not found in content", test.startSub)
+			}
+			endRelative := strings.Index(test.content[start:], test.endSub)
+			if endRelative == -1 {
```
JoeWang1127: Same as above, could we remove this check?

internal/postprocessing/delete_test.go
yangyzs: Yes. Good suggestion. removed.

internal/postprocessing/delete_test.go
yangyzs: Removed.

internal/postprocessing/delete.go
yangyzs: Added.

JoeWang1127 (approved):

PR #6440, opened 2026-06-16
refactor(nodejs): extract functions from runPostProcessor

jskeet:
Two "chunky" functions are extracted from runPostProcessor:

- movePackageFromStaging performs all the required actions to move code from owl-bot-staging into the package directory
- generateReadme performs replacements on the README.md file

Although generateReadme is currently not huge, it will be expanded to perform the "initial generation" step that includes the release level and populates the list of samples.

### DIFF
--- a/internal/librarian/nodejs/generate.go
+++ b/internal/librarian/nodejs/generate.go
@@ -260,43 +260,9 @@ func buildGeneratorArgs(api *config.API, library *config.Library, googleapisDir,
 // runPostProcessor combines versioned API outputs from owl-bot-staging/ into
 // the output directory using gapic-node-processing, then compiles protos.
 func runPostProcessor(ctx context.Context, cfg *config.Config, library *config.Library, googleapisDir, repoRoot, outDir string) error {
-	// combine-library wipes the destination directory before writing generated
-	// files (src/, protos/). Save the keep files it would delete, then restore
-	// them afterward.
-	backupDir, err := os.MkdirTemp(filepath.Dir(outDir), "librarian-backup-*")
-	if err != nil {
-		return fmt.Errorf("failed to create backup dir: %w", err)
-	}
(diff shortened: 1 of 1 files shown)


jskeet:
Note: there are no test changes because this is a pure refactor. However, it gives us the *opportunity* to write more fine-grained tests for the extracted functions, without setting up everything needed for `runPostProcessor` in every situation.

jskeet:
See #6441 for the later intended change to `generateReadme`.

internal/librarian/nodejs/generate.go
```diff
@@ -339,7 +305,70 @@ func runPostProcessor(ctx context.Context, cfg *config.Config, library *config.L
...
+	if err != nil {
+		return fmt.Errorf("failed to create backup dir: %w", err)
+	}
+	defer os.RemoveAll(backupDir)
+	// Backup keep files.
+	if err := moveKeep(library.Keep, outDir, backupDir); err != nil {
+		return err
+	}
 
+	stagingDir := filepath.Join(repoRoot, "owl-bot-staging", library.Name)
```
quirogas, line 340: Nit: change the staging directory name. Now that we've removed owl-bot we can also change the name to something more meaningful. Maybe librarian-stagin?

quirogas (approved):

internal/librarian/nodejs/generate.go
jskeet: Definitely not doing that during a purely-refactoring PR - but I'd also suggest not doing that just yet anyway. Basically at the moment we're trying to replicate the existing process, and it's easiest if that includes using the existing directory names. (It's possible that some of the post-processing code *assumes* this directory structure - it does for Python.) We should absolutely try to move away from mentioning owl-bot eventually, but I don't think just yet.

PR #6441, opened 2026-06-16
feat(nodejs): generate README.md files with release level and samples

jskeet:
Calls gapic-node-processing generate-readme with "initial generation" (even when it's not actual initial generation) in order to create a suitable README with a list of samples and a release level disclaimer. With this in place, we can stop preserving the old README.md files.


jskeet:
~~The first commit of this is from #6440. This is a draft (the tests fail) to show the purpose of the refactoring in #6440.~~

(That PR is now merged, and I've rebased this PR.)

Issue #5171, opened 2026-04-07
java: define a JavaDefault instead of JavaModule


zhumin8:
Java currently reuses JavaModule for default, there is not much overlap. Should define a JavaDefault with `LibrariesBomVersion` and any other field needed.

https://github.com/googleapis/librarian/blob/37b702b99ab2c51f1d44d6693ccd4d4b1ac23d28/internal/config/config.go#L200

Issue #5152, opened 2026-04-07
java: tidy should validate libraries bom version is present in default


zhumin8:
This is intentional, libraries bom version should be same across libraries, library override is ignored.
Should add a comment to the config, and validate its presence in tidy. Filing a followup

_Originally posted by @zhumin8 in https://github.com/googleapis/librarian/pull/5136#discussion_r3042474293_

PR #6447, opened 2026-06-16
refactor(internal/config): define a JavaDefault configuration instead of JavaModule

sofisl:
The Java default configuration previously reused the JavaModule struct, which included many library-specific fields (e.g. artifact_id, group_id) that are invalid in a global default block. Conversely, the libraries_bom_version field belongs
solely at the global level and should not be overridden per-library.

This change introduces a dedicated JavaDefault struct containing libraries_bom_version, updates Default.Java to use this type, and removes libraries_bom_version from JavaModule.

The only item I could see that would be relevant would be bom version (from reading https://github.com/googleapis/librarian/issues/5152)

For https://github.com/googleapis/librarian/issues/5171


internal/config/language.go
```diff
@@ -510,9 +516,6 @@ type JavaModule struct {
 	// to be overridden.
 	IssueTrackerOverride string `yaml:"issue_tracker_override,omitempty"`
 
-	// LibrariesBOMVersion is the version of the libraries-bom to use for Java.
```
JoeWang1127, line 513: Is this a breaking change for Java? I think we can add a default struct but keep this one as-is. After updating librarian.yaml in google-cloud-java, we can remove this.

sofisl: good catch. I'll do this in a separate PR and make the transition in a few PRs. Thanks!

JoeWang1127 (approved):

Issue #5930, opened 2026-05-08
sidekick: swift clients support method signatures


coryan:
We want to generate overloads for the method signatures

PR #6451, opened 2026-06-16
feat(sidekick): parse method signatures

coryan:
In some languages (and soon in Swift) we want to generate multiple overloads per-RPC, as some RPCs can be invoked with one or two string fields. An overload can request a couple of strings, or maybe a few strings and a body message, with is a better DX than the full request as input.

For protobuf-based APIs, the `method_signature` annotation defines such overloads.

For discovery-based APIs, the path parameters define an obvious overload.

Towards #5930

### DIFF
--- a/internal/sidekick/api/method_signature.go
+++ b/internal/sidekick/api/method_signature.go
@@ -1,0 +1,34 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package api
+
+// MethodSignature defines an alternative signature for the method.
+//
+// Some methods include annotations to generate additional overloads for the
+// corresponding generated code. That is, while most generators emit a single
+// function for each RPC, these annotations may define additional overloads for
+// the same RPC.
+//
+// These overloads select a subset of the fields in the request message, and the
+// generator emits a function for each list.
+type MethodSignature struct {
+	// Fields define the list of fields from the request message included in
+	// this signature.
+	Fields []*Field
+
+	// Names define the list of field names from the request message included in
+	// this signature.
+	Names []string
+}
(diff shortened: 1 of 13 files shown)


internal/sidekick/parser/protobuf_method_signature.go
```diff
@@ -0,0 +1,43 @@
...
+		return signatures, nil
+	}
+	ext, ok := proto.GetExtension(m.GetOptions(), extensionId).([]string)
+	if !ok {
+		return nil, fmt.Errorf("invalid signature type for %v", m.Name)
+	}
+	for _, signature := range ext {
+		names := strings.Split(signature, ",")
+		signatures = append(signatures, &api.MethodSignature{Names: names})
+	}
```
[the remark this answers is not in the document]
coryan: TIL... and eeeewww.

internal/sidekick/api/method_signature.go
```diff
@@ -0,0 +1,40 @@
...
+// These overloads select a subset of the fields in the request message, and the
+// generator emits a function for each list.
+type MethodSignature struct {
+	// Fields define the list of fields from the request message included in
+	// this signature.
+	Fields []*Field
+
+	// Names define the list of field names from the request message included in
+	// this signature.
+	Names []string
```
chingor13: Are `Fields` and `Names` mutually exclusive? Are we tracking both incase we have Field which has more data?

internal/sidekick/api/method_signature.go
```diff
@@ -0,0 +1,40 @@
...
+
+	// Method cross-references the method containing this signature.
+	//
+	// This is useful in mustache templates, as the template can access the
+	// method and the method's annotations from within a signature.
+	Method *Method
```
chingor13, line 42: This seems like a circular reference -- are we concerned at all?

internal/sidekick/api/method_signature.go
coryan: No, `Fields` cannot be initialized at first, because when the `Names` are captured, the `Messages` and `Fields` may not exist yet.

This is a know problem in sidekick, the initial data structures capture cross-references by name, and only later the data structure gains all the pointers.  @ianthehat has ideas to make this better.

internal/sidekick/api/method_signature.go
coryan: No, why would you be?

chingor13: refcounts and freeing up the memory but I'm not a golang person so maybe that's not an issue

chingor13 (approved):

internal/sidekick/api/method_signature.go
coryan: Ack. Go is garbage collected, cycles are fine.

Issue #6442, opened 2026-06-16
nodejs: regenerate README.md with release level and samples


jskeet:
Currently we have README.md in a keep list, which means that even though we apply *some* substitutions to the generated file from .readme-partials.yaml, we then overwrite it with a backup version. The readme we get before then doesn't have the note based on the release level (preview/stable) and also doesn't include a list of samples.

We should:

- Stop having README.md in the default keep list, instead using the regenerated version
- Apply release level and sample substitutions

PR #6441 (based on #6440) does this, but has no tests.

jskeet:
(Assigned to @JoeWang1127 as per offline discussion with @quirogas)

JoeWang1127:
In progress.

Move it to post generation stabilization milestone b/c it is not a blocking issue.

PR #6454, opened 2026-06-16
feat(internal/librarian/nodejs): extract sample metadata for node readme

JoeWang1127:
Functionality is added to scan and collect metadata from generated Node.js sample files during the post-processing phase.

This metadata is gathered using directory traversal and is intended to be used for dynamically generating sample listings in library README files.

Implement helper function first and create main function in follow ups.

For #6442

### DIFF
--- a/internal/librarian/nodejs/generate_readme.go
+++ b/internal/librarian/nodejs/generate_readme.go
@@ -16,6 +16,7 @@ package nodejs
 
 import (
 	"errors"
+	"fmt"
 	"io/fs"
 	"os"
 	"path/filepath"
@@ -49,7 +50,7 @@ func findSampleMetadata(output string) ([]sampleMetadata, error) {
 		}
 		metadata = append(metadata, sampleMetadata{
 			name:     extractSampleName(d.Name()),
-			filePath: filepath.Join(repoURLPrefix, path),
+			filePath: fmt.Sprintf("%s/%s", repoURLPrefix, path),
 		})
 		return nil
 	})
--- a/internal/librarian/nodejs/generate_readme.go
+++ b/internal/librarian/nodejs/generate_readme.go
@@ -49,7 +49,7 @@ func findSampleMetadata(output string) ([]sampleMetadata, error) {
 			return nil
 		}
 		metadata = append(metadata, sampleMetadata{
-			name:     extractSampleName(d.Name()),
+			name: extractSampleName(d.Name()),
 			// use string concat and filepath.ToSlash because we need a url format
 			// path to use in markdown file.
 			filePath: fmt.Sprintf("%s/%s", repoURLPrefix, filepath.ToSlash(path)),
--- a/internal/librarian/nodejs/generate_readme.go
+++ b/internal/librarian/nodejs/generate_readme.go
@@ -23,7 +23,7 @@ import (
 	"strings"
 )
 
-const repoURLPrefix = "https://github.com/googleapis/google-cloud-node/blob/main/packages"
+const repoURLPrefix = "https://github.com/googleapis/google-cloud-node/blob/main"
 
 var (
 	errorFindSampleMetadata = errors.New("error finding sample metadata")
--- a/internal/librarian/nodejs/generate_readme.go
+++ b/internal/librarian/nodejs/generate_readme.go
@@ -50,7 +50,7 @@ func findSampleMetadata(output string) ([]sampleMetadata, error) {
 		}
 		metadata = append(metadata, sampleMetadata{
 			name:     extractSampleName(d.Name()),
-			filePath: fmt.Sprintf("%s/%s", repoURLPrefix, path),
+			filePath: fmt.Sprintf("%s/%s", repoURLPrefix, filepath.ToSlash(path)),
 		})
 		return nil
 	})
--- a/internal/librarian/nodejs/generate_readme.go
+++ b/internal/librarian/nodejs/generate_readme.go
@@ -1,0 +1,69 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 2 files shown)


internal/librarian/nodejs/generate_readme.go
```diff
@@ -0,0 +1,69 @@
...
+		if err != nil {
+			return err
+		}
+		if d.IsDir() || filepath.Ext(path) != ".js" {
+			return nil
+		}
+		metadata = append(metadata, sampleMetadata{
+			name:     extractSampleName(d.Name()),
+			filePath: filepath.Join(repoURLPrefix, path),
+		})
```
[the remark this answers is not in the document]
JoeWang1127: The file path is part of a GitHub URL, so we don't need to use forward slash.

internal/librarian/nodejs/generate_readme.go
```diff
@@ -0,0 +1,69 @@
...
+	err := filepath.WalkDir(samplesPath, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if d.IsDir() || filepath.Ext(path) != ".js" {
+			return nil
+		}
+		metadata = append(metadata, sampleMetadata{
+			name:     extractSampleName(d.Name()),
+			filePath: filepath.Join(repoURLPrefix, path),
```
quirogas: Does filepath.Join won't work on URLs? Even if it does, It might be brittle since it is OS dependent. We should use a URL builder instead.

internal/librarian/nodejs/generate_readme.go
```diff
@@ -0,0 +1,69 @@
...
+		})
+		return nil
+	})
+	if err != nil {
+		return nil, err
```
quirogas: Nit, add context for this error.

internal/librarian/nodejs/generate_readme.go
quirogas: Or we could just concatenate the string we generate to the URL, since we know the exact URL

quirogas (changes requested):

internal/librarian/nodejs/generate_readme_test.go
```diff
@@ -0,0 +1,128 @@
...
+				generatedDir := filepath.Join(dir, "samples", "generated")
+				return []sampleMetadata{
+					{
+						name:     "nested sample",
+						filePath: filepath.Join(repoURLPrefix, filepath.Join(generatedDir, "sub/v1.nested_sample.js")),
+					},
+					{
+						name:     "do something",
+						filePath: filepath.Join(repoURLPrefix, filepath.Join(generatedDir, "v2.do_something.js")),
+					},
```
quirogas, line 91: We should include explicit, static versions of the expected results here to help us identify potential issues. ([go/tott/746](https://goto.google.com/tott/746))

internal/librarian/nodejs/generate_readme.go
JoeWang1127: Makes sense, I'll use string concatenation.

JoeWang1127: Done.

internal/librarian/nodejs/generate_readme.go
JoeWang1127: Done.

internal/librarian/nodejs/generate_readme_test.go
JoeWang1127: I have to use a val here to replace the library output. I updated the test to use string concat, what do you think?

internal/librarian/nodejs/generate_readme_test.go
```diff
@@ -0,0 +1,128 @@
...
+			},
+			want: func(dir string) []sampleMetadata {
+				return []sampleMetadata{
+					{
+						name:     "nested sample",
+						filePath: fmt.Sprintf("%s/%s/samples/generated/sub/v1.nested_sample.js", repoURLPrefix, dir),
+					},
+					{
+						name:     "do something",
+						filePath: fmt.Sprintf("%s/%s/samples/generated/v2.do_something.js", repoURLPrefix, dir),
```
quirogas: Why can’t we assert on this string instead?
```suggestion
						filePath: "https://github.com/googleapis/google-cloud-
     	  node/blob/main/packages/my-package/samples/generated/v2.do_something.
          js",
```

internal/librarian/nodejs/generate_readme_test.go
quirogas: I still don't think I understand why we have to generate the path instead of knowing what the path should look like. Perhaps you could clarify:

for the test Above are we expecting something like `https://github.com/googleapis/google-cloud-node/blob/main/packages/my-package/samples/generated/sub/v1.nested_sample.js`

or are we expecting something like `"https://github.com/googleapis/google-cloud-node/blob/main/packages/<TMP_DIR>/packages/my-package/samples/generated/sub/v1.nested_sample.js"`?

quirogas (approved):
I'm fine if we merge the Pr and address the comments in a follow up change if needed. Code and logic looks good :)

internal/librarian/nodejs/generate_readme_test.go
JoeWang1127: Done.

JoeWang1127:
> I still don't think I understand why we have to generate the path instead of knowing what the path should look like

This depends on the output in `findSampleMetadata(output string)`. I expect this is the library output, e.g., `packages/google-storage-secretmanager`, then the sample file should be `https://github.com/googleapis/google-cloud-node/blob/main/packages/google-storage-secretmanager/samples/generated/...`

quirogas:
> > I still don't think I understand why we have to generate the path instead of knowing what the path should look like
> 
> This depends on the output in `findSampleMetadata(output string)`. I expect this is the library output, e.g., `packages/google-storage-secretmanager`, then the sample file should be `https://github.com/googleapis/google-cloud-node/blob/main/packages/google-storage-secretmanager/samples/generated/...`

In that case, I believe the current code contains a bug that the tests aren't catching because they use a dynamically built path matching the one generated by the function, including the temp directory. We should address this in a follow-up PR.

PR #6461, opened 2026-06-16
feat(sidekick): parse method signatures

### LINKED ISSUE #6457: sidekick: method signature parsing fails for some APIs
> ```
> ibrarian: generate library "google-cloud-billing-v1" (rust): cannot find field  in method signature for method .google.cloud.billing.v1.CloudCatalog.ListServices
> ```
> 
> caused by:
> 
> https://github.com/googleapis/googleapis/blob/7af3f2c7b8927cffb548fd2cf09b04c6371437ee/google/cloud/billing/v1/cloud_catalog.proto#L48 

 _Originally posted by @coryan in [#6455](https://github.com/googleapis/librarian/issues/6455#issuecomment-4724184816)_

coryan:
In some languages (and soon in Swift) we want to generate multiple overloads per-RPC, as some RPCs can be invoked with one or two string fields. An overload can request a couple of strings, or maybe a few strings and a body message, with is a better DX than the full request as input.

For protobuf-based APIs, the `method_signature` annotation defines such overloads.

For discovery-based APIs, the path parameters define an obvious overload.

----

Towards #5930, fixes #6457.  Is not exactly the same as #6451, it adds some checks for empty method signatures, skips method signatures with sub-message fields, and adds test cases for them.  Manually tested on `google-cloud-rust`.

### DIFF
--- a/internal/sidekick/api/method_signature.go
+++ b/internal/sidekick/api/method_signature.go
@@ -1,4 +1,4 @@
-// Copyright 2024 Google LLC
+// Copyright 2026 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/internal/sidekick/api/method_signature.go
+++ b/internal/sidekick/api/method_signature.go
@@ -1,0 +1,43 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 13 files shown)


internal/sidekick/api/method_signature.go
```diff
@@ -0,0 +1,43 @@
+// Copyright 2024 Google LLC
```
chingor13:
```suggestion
// Copyright 2026 Google LLC
```

Missed this on the first review

chingor13 (approved):

internal/sidekick/api/method_signature.go
coryan: Thanks, fixed.

Issue #6453, opened 2026-06-16
nodejs: support mixing configuration per api client version.


quirogas:
Currently, we do not apply mixins per API client version, which results in the mixin being used for both versions.

For example, in google-cloud-functions, the proto file definitions handle IAM mixins differently: v1 has the IAM RPC defined within the proto file itself, while v2 does not. This means we should add the mixin for v2 without impacting v1.

To support this, we should explore encoding this information in librarian.yaml, allowing us to include or exclude mixins for specific API versions as necessary.  

The service yaml for v2 DOES include IAM: https://github.com/googleapis/googleapis/blob/master/google/cloud/functions/v2/cloudfunctions_v2.yaml#L9

This could be genuine issue. If I understand the generator correctly, a service yaml like this should trigger the inclusion of these lines.

_Originally posted by @pearigee in https://github.com/googleapis/google-cloud-node/pull/8600#discussion_r3406376498_

PR #6462, opened 2026-06-17
feat(nodejs): support per-API version mixin configuration

quirogas:
Currently, `librarian.yaml` only supports defining generator mixins at the package level under `library.nodejs.mixins`. When a client contains multiple versions (such as `v1` and `v2`), this singular package-level configuration forces all API client versions to inherit the exact same generator mixin flags.

This global constraint blocks or complicates onboarding of client libraries that have diverging proto schemas implementations for each mixin:

1. **`google-cloud-functions`**: The stable `v1` client defines IAM policy methods directly within its own service proto (`functions.proto`). However, `v2`, `v2alpha`, and `v2beta` do not define IAM RPCs in their protos and rely entirely on the `google.iam.v1.IAMPolicy` mixin. If configured globally at the package level, `v1` encounters duplicate/conflicting method generation errors. Without it, `v2` variants lose their IAM methods.

2. **`google-cloud-dialogflow-cx`**: Stable `v3` and preview `v3beta1` require distinct mixin inclusions (such as `google.longrunning.Operations` vs `google.cloud.location.Locations`). A singular global flag introduces massive unwanted diffs across surfaces.

This PR introduces unified, API-level mixin configuration exclusively under `apis[].nodejs.mixins` (aligning Node.js structure with Go and Java API overrides) and allows multi-version packages to configure distinct generator mixins per version (or explicitly set `mixins: none` to disable all mixin inclusions on specific API surfaces).

For #6453

---
## Follow-up Work

1. **Monorepo Configuration Adoption (`google-cloud-node`)**:
   - Once this Librarian CLI release is published, submit a pull request to `googleapis/google-cloud-node` updating its top-level `librarian.yaml`. This will migrate existing client libraries (`google-cloud-functions`, `google-cloud-dialogflow-cx`, `google-cloud-datacatalog`, `google-cloud-dataplex`, `google-cloud-retail`) to use the new granular `apis[].nodejs.mixins` syntax and enable automated client generation across those packages.

2. **Generator Engine Investigation (`gapic-generator-typescript`)**:
   - Investigate why `gapic-generator-typescript` requires explicit CLI `--mixins="..."` flags to activate mixins in certain multi-version libraries (e.g. `google-cloud-functions`), whereas standard single-version libraries automatically infer mixins directly from their service config YAML files (`--service-yaml`). Explore streamlining the generator's internal selective GAPIC filters to eliminate the need for manual CLI overrides where possible.


internal/config/language.go
```diff
@@ -789,6 +786,10 @@ type NodejsAPI struct {
 	// This is typically false. Used for the GCE (compute) client.
 	DIREGAPIC bool `yaml:"diregapic,omitempty"`
 
+	// Mixins controls mixin behavior for this API (e.g., "none" to disable).
```
JoeWang1127, line 789: Why do we need this change?

internal/librarian/nodejs/generate_test.go
```diff
@@ -234,15 +234,19 @@ func TestBuildGeneratorArgs(t *testing.T) {
...
+				},
+			},
 			library: &config.Library{
 				Name: "google-cloud-translate",
 				Nodejs: &config.NodejsPackage{
 					BundleConfig:          "google/cloud/translate/v3/translate_gapic.yaml",
 					ExtraProtocParameters: []string{"auto-populate-field-oauth-scope"},
 					HandwrittenLayer:      true,
 					MainService:           "translate",
-					Mixins:                "none",
```
JoeWang1127, line 245: Why this line is removed?

quirogas: Because mixin in no longer present in the NodejsPackage.

internal/config/language.go
quirogas: In order to support having mixing per version of a library.

JoeWang1127 (approved):

PR #6470, opened 2026-06-17
fix(internal/librarian/nodejs):  path leak during generte_readme

quirogas:
When `librarian` post-processes a generated client library (e.g., `packages/google-cloud-vision`), `library.Output` is resolved to an absolute path for working directory safety. When `filepath.WalkDir` navigates this directory, it passes absolute disk file paths (`/workspace/google-cloud-node/packages/...`) to the callback.

When we appended this walk path to `repoURLPrefix`, we generated Markdown hyperlinks that leaked local continuous integration working folder paths:
```text
https://github.com/googleapis/google-cloud-node/blob/main/packages//workspace/google-cloud-node/packages/google-cloud-vision/samples/generated/v1/foo.js
```

> NOTE: The previous unit tests failed to catch this defect because the test computed expected URLs at runtime using the exact same dynamic path manipulation logic as the implementation.

### DIFF
--- a/internal/librarian/nodejs/generate_readme.go
+++ b/internal/librarian/nodejs/generate_readme.go
@@ -36,6 +36,7 @@ type sampleMetadata struct {
 }
 
 func findSampleMetadata(output string) ([]sampleMetadata, error) {
+	output = filepath.Clean(output)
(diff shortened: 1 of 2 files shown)


JoeWang1127 (approved):
Thanks.

PR #6477, opened 2026-06-17
chore: remove surfer

noahdietz:
Removes `surfer` codebase. AFAIK there is currently no plan to continue developing this. We should remove it to avoid bearing the potential maintenance cost, test execution time, extra repo size, etc.

Everything will be parked on a branch for safe keeping: https://github.com/googleapis/librarian/tree/surfer

If we want to revive it, this PR can be reverted and any diffs in the librarian/sidekick APIs it used can be rectified as part of the cost of reviving the project.

Argument for keeping it is that we *may* want to use it in the future and it is easier to make incremental changes to the code as its dependencies (librarian / sidekick) in the repo change. The problem with that is the cost of keeping surfer up to date falls on the folks developing librarian/sidekick. While there hasn't been much friction to date, we'd rather not wait to find out.

Piggybacks on https://github.com/googleapis/librarian/pull/6474 removing `gcloud`.


noahdietz:
@coryan @jameslynnwu I've spoken with both of you separately, and James mentioned discussing with @ianthehat .

See rationale in PR description.

coryan (approved):

jameslynnwu (approved):
Let's create a tag or branch to make it easy to bring back.

noahdietz:
> Let's create a tag or branch to make it easy to bring back.

Great idea: https://github.com/googleapis/librarian/tree/surfer

PR #6481, opened 2026-06-17
feat(internal/librarian/java): support alternate_headers for monolithc libraries

### LINKED ISSUE #6178: java: fix header for grafeas in generate

sofisl:
Final F/U PR described in https://github.com/googleapis/librarian/pull/6311. This PR "flips the switch" so that if a library has alternate_headers, we should see it in the library. Otherwise keeps other monolithic libraries stable.

Fixes https://github.com/googleapis/librarian/issues/6178

### DIFF
--- a/internal/librarian/java/postprocess_test.go
+++ b/internal/librarian/java/postprocess_test.go
@@ -505,6 +505,20 @@ func TestPostProcessAPI_AlternateHeaders(t *testing.T) {
 	if !bytes.HasPrefix(got, []byte(wantHeader)) {
 		t.Errorf("mismatch got = %q, want %q", got, wantHeader)
 	}
+	if err := os.WriteFile(grpcFile, []byte("package com.test;"), 0644); err != nil {
+		t.Fatal(err)
+	}
+	params.javaAPI.Monolithic = true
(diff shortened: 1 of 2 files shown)


internal/librarian/java/postprocess_test.go
```diff
@@ -507,6 +507,43 @@ func TestPostProcessAPI_AlternateHeaders(t *testing.T) {
 	}
 }
 
+func TestPostProcessAPI_AlternateHeaders_Monolithic(t *testing.T) {
```
JoeWang1127: I think you already tested the scenario of alternative header, could you verify the `addHeaders` return early if no alternative header is provided in librarian.yaml?

sofisl: this is more specific to the monolithic being true... but yes, will do! I'll replace it.

JoeWang1127 (approved):
I think the test needs more refactor but it's fine for now.

PR #6491, opened 2026-06-18
chore: update starting tag

codyoss:
The tag referenced before was not a tag that existed on this branch, so at release-please defaulted to searching further back in history just to try to gather all the context.

Release-As: 0.1.2


jskeet:
Does this mean it's going to try to create a tag of v0.1.1, which *does* exist but isn't a commit on this branch?

jskeet:
(Is that tag actually useful? Would it be worth getting admin access in order to delete that tag? I don't normally like reusing tag names, but in this case it might be less awkward.)

codyoss:
Good point, I will put in an override so we release as 0.1.2 -- I think this is why I started at 0.1.1 when I made the config 👀

codyoss:
> Does this mean it's going to try to create a tag of v0.1.1, which _does_ exist but isn't a commit on this branch?

Go is very tag sensitive so we don't want to change where they are pointing.

codyoss:
Okay when merged this should now go to 0.1.2 because I included `Release-As` in the commit

JoeWang1127 (approved):

PR #6502, opened 2026-06-18
test(.github/workflows/java): cache maven and librarian tools to speed up test execution

### LINKED ISSUE #6482: java: librarian tests take longer than 5 minutes
The `Java / tests` workflow is regularly cancelled for exceeding the 5 min timeout, even on `main`.

Image

We need to look at speeding these tests update: https://github.com/googleapis/librarian/blob/4a38581c80d85d3cd901ef1710a4fc48dd72164b/.github/workflows/java.yaml#L23-L58

cc @zhumin8

sofisl:
Optimizes the `Java / tests` workflow in `java.yaml` to prevent timeouts by introducing caching for built dependencies and tools:

- Enable Maven dependency caching (`cache: "maven"`) in `setup-java`.
- Set up a Python virtual environment (`venv`) to isolate and cache pip-installed dependencies like `synthtool`.
- Use `actions/cache` to cache both the compiled Java tools (`~/.cache/librarian`) and the Python virtual environment (`venv`).
- Make the Java/Maven verification and `librarian install` steps conditional on a cache miss, skipping them entirely when tools are restored from cache.

Future test idea speedup: to do `sparse-checkout`, we could change `buildLocalMavenProject` to extract the parent directory of the local project (e.g., sdk-platform-java) and run the mvn command from inside that directory, so that we don't need to check out the whole repo and mvn doesn't have to validate the whole repo when running.

Fixes #6482

### DIFF
--- a/.github/workflows/java.yaml
+++ b/.github/workflows/java.yaml
@@ -61,12 +61,13 @@ jobs:
             sudo apt-get update && sudo apt-get install -y maven
           fi
           mvn -version
+      - name: Add Java tools to PATH
+        run: echo "$HOME/.cache/librarian/bin/java_tools/bin" >> $GITHUB_PATH
       - name: Run librarian install
         if: steps.cache-tools.outputs.cache-hit != 'true'
         working-directory: google-cloud-java
         run: |
           librarian install
-          echo "$HOME/.cache/librarian/bin/java_tools/bin" >> $GITHUB_PATH
       - name: Debug - List Java tools directory
         run: |
           echo "=== 1. CHECKING PATH ENVIRONMENT VARIABLE ==="
--- a/.github/workflows/java.yaml
+++ b/.github/workflows/java.yaml
@@ -47,7 +47,7 @@ jobs:
           path: |
             ~/.cache/librarian
             venv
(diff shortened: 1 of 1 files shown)


JoeWang1127:
The unit test still needs more than 5 min to [run](https://github.com/googleapis/librarian/actions/runs/27779092073/job/82198897355?pr=6502), will it be faster after merging this pull request?

sofisl:
It's certainly right on the edge (I'm seeing it at 5m 27 sec), but it's also green (i.e., it's not being canceled).

We could merge this, since at least tests are passing and this caches the dependencies; and in a future issue (if we see this pop up again) we could do the more involved idea I have in the PR body (would require a refactor of `buildLocalMavenProject`)

JoeWang1127:
> It's certainly right on the edge (I'm seeing it at 5m 27 sec), but it's also green (i.e., it's not being canceled).

This is b/c we increased the timeout to 10 min: https://github.com/googleapis/librarian/pull/6483

sofisl:
I reran the job because I wanted to see if the 5 min was only because it wasn't using the cache. Then I ran into a bug in which the format dependency wasn't being properly cached. Now that the bug is fixed, the tests are running in 3 min.

JoeWang1127 (approved):

JoeWang1127:
Could you remove the TODO added in https://github.com/googleapis/librarian/pull/6483?

sofisl:
And lowered it back to 5!

PR #6503, opened 2026-06-18
feat(sidekick/rust): add condition to include `google-cloud-lro` as dependency

haphungw:
Include the `google-cloud-lro` dependency in a generated crate's `Cargo.toml` if any of its methods are configured as LRO pollers (`IsLroPoller`).

### DIFF
--- a/internal/sidekick/rust/codec.go
+++ b/internal/sidekick/rust/codec.go
@@ -385,7 +385,7 @@ func resolveUsedPackages(model *api.API, extraPackages []*packagez) {
 		// not save us any computations.
 
 		for _, m := range s.Methods {
-			if m.OperationInfo != nil || m.DiscoveryLro != nil {
+			if m.OperationInfo != nil || m.DiscoveryLro != nil || m.IsLroPoller {
 				hasLROs = true
 			}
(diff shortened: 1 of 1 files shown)


internal/sidekick/rust/codec.go
```diff
@@ -385,7 +385,7 @@ func resolveUsedPackages(model *api.API, extraPackages []*packagez) {
 		// not save us any computations.
 
 		for _, m := range s.Methods {
-			if m.OperationInfo != nil || m.DiscoveryLro != nil {
+			if m.OperationInfo != nil || m.DiscoveryLro != nil || m.IsLroPoller {
```
dbolduc: nit: 
```suggestion
			if m.OperationInfo != nil || m.IsLroPoller {
```

looks like `m.DiscoveryLro` implies `m.IsLroPoller`

https://github.com/googleapis/librarian/blob/520fe9d501df0f31227289cfa821da67f1006ceb/internal/sidekick/rust/annotate_method.go#L365-L368

dbolduc (approved):

internal/sidekick/rust/codec.go
haphungw: done!

PR #6506, opened 2026-06-18
feat(sidekick/swift): qualified names for requests

### LINKED ISSUE #6493: sidekick: swift support nested request messages
For discovery-based APIs sidekick injects requests messages into the data model. The discovery doc does not define specific messages for each request.

These request types are named as `${MethodName}Request` which obviously could (and does) create clashes because multiple services can have the name method name (e.g. `Foo.list()` and `Bar.list()`). Sidekick generates the requests in the scope of a (also synthetic) message named after the service.

Currently the Swift codec generates methods that reference the unqualified name of the message, for example:

https://github.com/googleapis/librarian/blob/27a72beb520feb5d7c04701e967a3454f1367b39/internal/sidekick/swift/templates/common/method_signature/request.mustache#L16

That expands to something like `request: FooRequest` which would not work if `FooRequest` is in some namespace, we need the fully qualified name, to generate something like `request: Container.FooRequest`.

coryan:
In discovery-based services the requests will be nested messages. We need to generate code that fully qualifies the request type, like we do for fields. This requires new annotations for the messages.

Fixes #6493

### DIFF
--- a/internal/sidekick/swift/annotate_message.go
+++ b/internal/sidekick/swift/annotate_message.go
@@ -58,8 +58,8 @@ type messageAnnotations struct {
 	// imported with the mixin, e.g. `import GoogleCloudIamV1` imports the IAM
 	// mixin request types.
 	//
-	// For discovery-based APIs the request are synthetic and generated within a
-	// scope. They need to be fully qualified.
+	// For discovery-based APIs, the request are synthetic and generated within
+	// a scope. They need to be fully qualified.
 	ParameterTypeName string
 }
 
--- a/internal/sidekick/swift/annotate_message.go
+++ b/internal/sidekick/swift/annotate_message.go
@@ -51,6 +51,16 @@ type messageAnnotations struct {
 	// service that needs them is enabled. Messages that do not map to any
 	// service use " && ".
 	GatedOp string
+
+	// The message type name when it appears as a method parameter name.
+	//
+	// Most of the time the request types are in the package namespace, or are
+	// imported with the mixin, e.g. `import GoogleCloudIamV1` imports the IAM
+	// mixin request types.
+	//
+	// For discovery-based APIs the request are synthetic and generated within a
+	// scope. They need to be fully qualifed.
+	ParameterTypeName string
 }
 
 // MessageImports returns the list of dependencies for this message.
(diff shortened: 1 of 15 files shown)


internal/sidekick/swift/templates/common/service.swift.mustache
```diff
@@ -202,7 +202,7 @@ extension {{Codec.Name}} {
   {{#Signatures}}
 
   public func {{> /templates/common/method_signature/overload}} {
-    let request = {{Method.InputType.Codec.Name}}().with {
+    let request = {{Method.InputType.Codec.ParameterTypeName}}().with {
```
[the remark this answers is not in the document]
coryan: Fixed.

internal/sidekick/swift/annotate_message.go
```diff
@@ -51,6 +51,16 @@ type messageAnnotations struct {
...
 	// service use " && ".
 	GatedOp string
+
+	// The message type name when it appears as a method parameter name.
+	//
+	// Most of the time the request types are in the package namespace, or are
+	// imported with the mixin, e.g. `import GoogleCloudIamV1` imports the IAM
+	// mixin request types.
+	//
+	// For discovery-based APIs the request are synthetic and generated within a
```
nolanleastin: nit: hard to read.

Suggestion: "For discover-based APIs, the request types are synthetic ..."

internal/sidekick/swift/annotate_method_test.go
```diff
@@ -148,7 +148,6 @@ func TestAnnotateMethod(t *testing.T) {
 			service := &api.Service{
 				Name:    "TestService",
 				ID:      ".test.TestService",
-				Package: "test",
```
nolanleastin, line 151: Why is this removed?

nolanleastin (approved):

internal/sidekick/swift/annotate_message.go
coryan: Done.

internal/sidekick/swift/annotate_method_test.go
coryan: Preserving this was more disruptive as it requires other things to change. Fixed.

PR #6511, opened 2026-06-19
fix(internal/librarian/java/defaults.go): keep autogenerated files and non-Java files if explicitly listed in librarian.yaml

### LINKED ISSUE #6510: java: remove redundant keep files removes too many files
https://github.com/googleapis/librarian/issues/6163 caused a regression that removed autogenerated-marked, yet explicitly stated files:

```
(.venv) sofialeon@sofialeon:~/google-cloud-java$ git status
On branch pseudoVersion
Your branch is up to date with 'origin/pseudoVersion'.

Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        deleted:    java-aiplatform/google-cloud-aiplatform/src/test/java/com/google/cloud/location/MockLocations.java
        deleted:    java-aiplatform/google-cloud-aiplatform/src/test/java/com/google/cloud/location/MockLocationsImpl.java
        deleted:    java-aiplatform/google-cloud-aiplatform/src/test/java/com/google/iam/v1/MockIAMPolicy.java
        deleted:    java-aiplatform/google-cloud-aiplatform/src/test/java/com/google/iam/v1/MockIAMPolicyImpl.java
        deleted:    java-bigquerystorage/google-cloud-bigquerystorage/src/main/resources/META-INF/native-image/com.google.cloud/google-cloud-bigquerystorage/reflect-config.json
        deleted:    java-firestore/google-cloud-firestore/src/main/resources/META-INF/native-image/com.google.cloud/google-cloud-firestore/reflect-config.json
        deleted:    java-firestore/google-cloud-firestore/src/test/resources/META-INF/native-image/reflect-config.json
        deleted:    java-spanner/google-cloud-spanner/src/main/resources/META-INF/native-image/com.google.cloud.spanner/reflect-config.json
        modified:   librarian.yaml
```

In order to fix, we need to save generated non-Java files and autogenerated files that are explicitly marked as keep.

This was tested by upgrading the version of Librarian to the pseudo version in Java, running `librarian tidy`, and then `generate --all`

sofisl:
This PR fixes a regression introduced in #6291 because it erroneously removed files that were not redundant (where redundant means "automatically saved or generated by the current logic so does not need to be put into the `keep` list). Instead, if a file is explicitly listed in the `keep` list, and it is autogenerated, it should be kept. It also incorrectly removed non-Java files. So the new overall logic is:

- If it is NOT in a gapic module, then it is NOT redundant
- If it's outside of the gapic module's src file, then it would automatically get cleaned anyways, so it IS redundant
- If it's a file that should always be present (otherwise, in `generateIfMissing`), then it will be always be generated so it is redundant
- If it's a generated but non-Java file, it IS redundant
- If it does NOT have an autogenerated marker, it IS redundant. Conversely if it has an autogenerated marker and is in the keep list, it is NOT redundant.

This PR institutes the logic above. I have confirmed it works by running generate on the problem libraries listed in https://github.com/googleapis/librarian/issues/6510. However: this logic feels extremely fragile. It's really hard to wrap my head around it, and makes for really confusing code. I *might* just advocate for reverting #6291 and not implementing this feature at all. There are so many corner cases that it makes it difficult to confidently say which files should be there or not; and I'm sure this will break someone in the future (why was my explicitly listed file removed all of a sudden?)

Fixes https://github.com/googleapis/librarian/issues/6510


sofisl:
@zhumin8 @JoeWang1127 FYI, I tested #6921 and found a regression in #6510. This PR fixes the regression (I confirmed by running `generate` on the APIs listed in the issue), but I also opened this https://github.com/googleapis/librarian/pull/6512. Since I'll be OOO on Monday I opened both so you can decide to take whatever path you prefer.

Issue #6537, opened 2026-06-23
nodejs: "librarian tidy" shows diff


suztomo:
google-cloud-node has diff after running `librarian tidy`:

```
(venv) suztomo@suztomo:~/librarian-2026/google-cloud-node$ V=$(go run github.com/googleapis/librarian/cmd/librarian@latest config get version)
(venv) suztomo@suztomo:~/librarian-2026/google-cloud-node$ go run github.com/googleapis/librarian/cmd/librarian@${V} tidy
(venv) suztomo@suztomo:~/librarian-2026/google-cloud-node$ git diff
diff --git a/librarian.yaml b/librarian.yaml
index b0b2a2ba18..e9c8c0fd32 100644
--- a/librarian.yaml
+++ b/librarian.yaml
@@ -611,11 +611,7 @@ libraries:
     version: 5.2.2
     apis:
       - path: google/cloud/datacatalog/v1
-        nodejs:
-          mixins: none
       - path: google/cloud/datacatalog/v1beta1
-        nodejs:
-          mixins: none
     copyright_year: "2026"
     nodejs:
       default_version: v1
@@ -665,8 +661,6 @@ libraries:
     version: 5.6.0
     apis:
       - path: google/cloud/dataplex/v1
-        nodejs:
-          mixins: google.cloud.location.Locations;google.longrunning.Operations
     copyright_year: "2026"
     nodejs:
       default_version: v1
@@ -729,11 +723,7 @@ libraries:
     version: 5.8.2
     apis:
       - path: google/cloud/dialogflow/cx/v3
-        nodejs:
-          mixins: google.longrunning.Operations;google.cloud.location.Locations
       - path: google/cloud/dialogflow/cx/v3beta1
-        nodejs:
-          mixins: google.longrunning.Operations;google.cloud.location.Locations
     copyright_year: "2026"
     skip_generate: true
```

suztomo:
The main branch of Librarian already preserves the fields.

```
suztomo@suztomo:~/librarian-2026/librarian$ git log -1 --oneline
1373628c (HEAD -> main, upstream/main, upstream/HEAD) feat(nodejs): bump pnpm version to 11.7.0 and support v11+ global bin layout (#6494)
suztomo@suztomo:~/librarian-2026/librarian$ go build  -o /tmp/librarian ./cmd/librarian
suztomo@suztomo:~/librarian-2026/librarian$ cd ../google-cloud-node
[...]

PR #6539, opened 2026-06-24
fix(internal/librarian/nodejs): preserve non-empty api-level configs

quirogas:
When normalizing library configurations with librarian tidy, explicit nodejs configurations declared on individual APIs are omitted.

This modifies nodejs normalization to iterate over all APIs in a library and preserve non-empty api-level nodejs blocks

### DIFF
--- a/internal/librarian/nodejs/tidy.go
+++ b/internal/librarian/nodejs/tidy.go
@@ -21,15 +21,25 @@ import (
 
 // Tidy tidies configuration for a library.
 func Tidy(lib *config.Library) (*config.Library, error) {
-	if lib.Nodejs == nil {
-		return lib, nil
+	for _, api := range lib.APIs {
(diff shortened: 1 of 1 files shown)


internal/librarian/nodejs/tidy.go
```diff
@@ -21,15 +21,25 @@ import (
 
 // Tidy tidies configuration for a library.
 func Tidy(lib *config.Library) (*config.Library, error) {
-	if lib.Nodejs == nil {
```
suztomo, line 24: In the old code, I don't see any logic that would remove the `api.nodejs` fields.

I observe the main branch of Librarian already preserves the fields https://github.com/googleapis/librarian/issues/6537#issuecomment-4789547841.

quirogas:
Closing as this has already been addressed in v0.22.0

Issue #6513, opened 2026-06-19
java: move known group id mapping into config default section


zhumin8:
When adding new libraries, java currently have these hardcoded known group ids other than com.google.cloud:
https://github.com/googleapis/librarian/blob/24b170e3aa64fb584dd2b9514f8498b082de4006/internal/librarian/java/add.go#L55-L62

We should convert these into config in the defaults for java:
https://github.com/googleapis/librarian/blob/24b170e3aa64fb584dd2b9514f8498b082de4006/internal/config/config.go#L221
Preserve the mapping from API path to relevant group id.

This approach has benefits that:
- Supported group ids are visible from config
- When adding new group ids in the future, this is configurable and no changes to librarian needed

PR #6554, opened 2026-06-27
feat(internal/librarian): populate Java Maven coordinates from defaults

JoeWang1127:
The library configuration loader is updated to automatically populate Java Maven coordinates (GroupID and ArtifactID) based on API path prefix mappings defined in default configuration.

When a library's Java module configuration is missing or its GroupID is empty, the loader scans the library's API paths against the default api_path_to_group_id map to resolve the appropriate GroupID and sets the ArtifactID using the standard "google-<library>" naming convention.

Add default mappings in follow up PRs.

For #6513


internal/librarian/library.go
```diff
@@ -92,6 +95,29 @@ func union(a, b []string) []string {
...
+				lib.Java.GroupID = groupID
+				break
+			}
+		}
+		if lib.Java.GroupID != "" {
+			break
+		}
+	}
+	return lib
+}
```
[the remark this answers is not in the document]
JoeWang1127: 1. I don't think we need a fixed order.
2. Done.
3. Done.
4. I don't think this is relevant.

internal/librarian/library.go
```diff
@@ -92,6 +95,29 @@ func union(a, b []string) []string {
 	return res
 }
 
+// fillJava populates empty Java-specific fields in lib from the provided default.
+func fillJava(lib *config.Library, d *config.Default) *config.Library {
+	if lib.Java == nil {
+		lib.Java = &config.JavaModule{}
+	}
+	if lib.Java.ArtifactID != "" || lib.Java.GroupID != "" || d.Java.APIPathToGroupID == nil {
```
zhumin8: If user has `lib.Java.ArtifactID` explicitly set but leaves out `lib.Java.GroupID`, this logic exits too early and does not populates group id from defaults.

internal/librarian/library.go
```diff
@@ -92,6 +95,29 @@ func union(a, b []string) []string {
...
+		return lib
+	}
+	for _, api := range lib.APIs {
+		for apiPrefix, groupID := range d.Java.APIPathToGroupID {
+			if strings.HasPrefix(api.Path, apiPrefix+"/") {
```
zhumin8: Not sure we ever gonna have this case of exact match, but better be safe here:
`path == prefix || strings.HasPrefix(path, prefix+"/")`

internal/config/language.go
```diff
@@ -460,6 +460,8 @@ type DartPackage struct {
 type JavaDefault struct {
 	// LibrariesBOMVersion is the version of the libraries-bom to use for Java.
 	LibrariesBOMVersion string `yaml:"libraries_bom_version,omitempty"`
+	// APIPathToGroupID is a map of API paths to group IDs for Java.
+	APIPathToGroupID map[string]string `yaml:"api_path_to_group_id,omitempty"`
```
zhumin8: How about naming this "CustomGroupIDs"? It tells the user: "use this to define exceptions to the default Group ID."
And give more context in the comment, e.g. 
```
        // CustomGroupIDs maps API path prefixes (e.g., "google/shopping") to their
        // corresponding Maven Group IDs (e.g., "com.google.shopping").
        // Use this to override the default "com.google.cloud" Group ID for
        // specific API paths (e.g., maps, ads, shopping).
```

JoeWang1127: Done.

internal/librarian/library.go

internal/librarian/library.go
JoeWang1127: The artifact id and group id, if empty, will be filled with default value in https://github.com/googleapis/librarian/blob/feat/group-id-mapping/internal/librarian/java/defaults.go#L45

zhumin8: Right, that fills cloud defaults. 
I am saying, what if say we are processing API path `google/shopping/a/b/c`, with an explicit for example `lib.Java.ArtifactID=toy`. We expect this process to set `lib.Java.GroupID` to "com.google.shopping". But with this logic it exists early and leaves GroupID to default (later filled with com.google.cloud).

JoeWang1127: Updated the logic to return early only if artifact id and group id are set. PTAL.

zhumin8: do you need to go to below steps if lib.Java.GroupID  is explicitly set but lib.Java.ArtifactID empty? I suppose it will get defaults filled in later anyway?

We only look into the custom group id mapping if the group id in librarian.yaml is empty.

internal/librarian/library_test.go
```diff
@@ -233,6 +233,217 @@ func TestFillDefaults(t *testing.T) {
...
+				APIs: []*config.API{
+					{Path: "google/shopping/type"},
+				},
+				Java: &config.JavaModule{
+					GroupID: "com.google.api.grpc",
+				},
+			},
+		},
+		{
+			name: "do not fill group id if api path contains library name as prefix",
```
zhumin8: this test case name is misleading

zhumin8 (approved):

internal/librarian/library_test.go
JoeWang1127: Renamed.

internal/librarian/library.go
```diff
@@ -47,6 +47,9 @@ func fillDefaults(lib *config.Library, d *config.Default) *config.Library {
 	if d.Go != nil {
 		return fillGo(lib, d)
 	}
+	if d.Java != nil {
```
ianthehat: While you are following the pattern of the surrounding code, the surrounding code is wrong.
At the lowest level, a sequence of mutually exclusive if statements like this should be a switch instead.
At a higher level, why do we have a bunch of fields that seem to be mutually exclusive, do we actually verify that they are somewhere, if not does the order in which they are checked matter, why does Java get higher priority than rust... the fact that all these questions exist implies a fundamental design fault with another API somewhere (probably whatever config.Default is or whatever builds it).
At first glance it looks like there should be a single field in that thing, of an interface that supports a Fill method, rather than what seems like weird type switch and method dispatch block in this code.

internal/librarian/library.go
```diff
@@ -97,6 +100,28 @@ func union(a, b []string) []string {
 	return res
 }
 
+// fillJava populates empty Java-specific fields in lib from the provided default.
+func fillJava(lib *config.Library, d *config.Default) *config.Library {
+	if lib.Java == nil {
+		lib.Java = &config.JavaModule{}
+	}
+	if lib.Java.GroupID != "" || d.Java.CustomGroupIDs == nil {
+		return lib
```
ianthehat: This seems like a footgun
This method reads like one that is allowed to set multiple fields to their default values, but this return makes the assumption that GroupID is the only field that this will ever be true for.
If I was adding a new default later one I would add it at the bottom of the function and be very confused why it was not always being set...

internal/librarian/library.go
```diff
@@ -97,6 +100,28 @@ func union(a, b []string) []string {
...
+	}
+	for _, api := range lib.APIs {
+		for apiPrefix, groupID := range d.Java.CustomGroupIDs {
+			if api.Path == apiPrefix || strings.HasPrefix(api.Path, apiPrefix+"/") {
+				lib.Java.GroupID = groupID
+				break
+			}
+		}
+		if lib.Java.GroupID != "" {
+			break
```
ianthehat: Adding an outer break that depends on an inner break triggering is a code smell, it almost always implies some code that should be extracted to a function and the breaks should be returns instead.

JoeWang1127: Done, replaced the inner `break` to `return`.

internal/librarian/library.go
JoeWang1127: Created a helper function to fill group id.

This can make the function easily extend to fill other fields for java library.

internal/librarian/library.go
JoeWang1127: As a first step, I've refactor `fillDefaults` to use a switch.

I'll create follow up PRs to refactor `config.Default` struct.

internal/librarian/library.go
```diff
@@ -97,6 +97,32 @@ func union(a, b []string) []string {
...
+	if lib.Java == nil {
+		lib.Java = &config.JavaModule{}
+	}
+	return fillGroupIDIfEmpty(lib, d)
+}
+
+// fillGroupIDIfEmpty sets the Java group ID on lib if one is not already configured.
+// It matches the library's API paths against the custom group ID prefixes in default
+// and assigns the first matching group ID.
+func fillGroupIDIfEmpty(lib *config.Library, d *config.Default) *config.Library {
```
ianthehat: No point returning the lib, it is always the same as the one passed in, don't make the reader guess wether there is a good reason for that.

ianthehat (commented):
One very minor comment, cl is good apart from that.

internal/librarian/library.go

PR #6557, opened 2026-06-27
chore(internal/librarian): refactor unit tests

JoeWang1127:
For unit tests of `applyDefaults`, split it into success and error cases.

See https://github.com/googleapis/librarian/blob/main/doc/howwewritego.md#separate-error-tests


internal/librarian/library.go
```diff
@@ -288,7 +293,7 @@ func applyDefaults(language string, lib *config.Library, defaults *config.Defaul
 	}
 	if lib.Output == "" {
 		if isMixedLibrary(language, lib) {
-			return nil, fmt.Errorf("library %q requires an explicit output path", lib.Name)
+			return nil, fmt.Errorf("%s: %w", lib.Name, errNoExplicitOutput)
```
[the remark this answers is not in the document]

internal/librarian/library.go
```diff
@@ -288,7 +293,7 @@ func applyDefaults(language string, lib *config.Library, defaults *config.Defaul
 	}
 	if lib.Output == "" {
 		if isMixedLibrary(language, lib) {
-			return nil, fmt.Errorf("library %q requires an explicit output path", lib.Name)
+			return nil, fmt.Errorf("%w: %s", errNoExplicitOutput, lib.Name)
```
codyoss: By convention when wrapping Go error the wrapped error should be put at the end of the string, even though it is allowed to go anywhere by the tooling. errors should be formatted "context: underlying-error"

codyoss (approved):

PR #6564, opened 2026-06-29
feat(internal/librarian/java): Add JSpecify dep to proto module template

lqiu96:
This changes adds the JSpecify dep to Java proto module template.


JoeWang1127 (approved):
Could you add a description about this change?

Issue #6515, opened 2026-06-19
java: generate README.md using template


JoeWang1127:
Currently, the README.md [template](https://github.com/googleapis/google-cloud-java/blob/main/sdk-platform-java/hermetic_build/library_generation/owlbot/templates/java_library/README.md) lives in google-cloud-java.

We should create a template and render README in librarian. This is part of the effort to eliminate the sythtool dependency.

See go/sdk-librarian-mod-gen for more details.

PR #6574, opened 2026-06-29
feat(internal/librarian/java): add README sample extraction helpers

yangyzs:
Add helper functions collectSampleFiles and parseCodeSample to scan samples/src/main/java and extract sample metadata. Add codeSample struct and unit tests.

For #6515


internal/librarian/java/readme.go
```diff
@@ -38,6 +39,62 @@ var (
...
 
+// codeSample represents a discovered Java code sample along with its derived title.
+type codeSample struct {
+	Title string
+	File  string
+}
+
+// collectSampleFiles recursively scans dir/samples for Java production files.
+func collectSampleFiles(dir string) ([]string, error) {
+	samplesDir := filepath.Join(dir, "samples")
```
JoeWang1127, line 50: Could you check the existence of this directory, so that you can avoid check it inside the `WalkDir` function.

internal/librarian/java/readme.go
```diff
@@ -38,6 +39,62 @@ var (
...
+	var files []string
+	err := filepath.WalkDir(samplesDir, func(path string, d os.DirEntry, err error) error {
+		if err != nil {
+			if errors.Is(err, fs.ErrNotExist) {
+				return nil
+			}
+			return err
+		}
+		// Skip directories and non-regular files.
```
JoeWang1127: Could you remove this comment as it doesn't add much of value. In addition, does it check diectory?

internal/librarian/java/readme.go
```diff
@@ -38,6 +39,62 @@ var (
...
+		if !d.Type().IsRegular() {
+			return nil
+		}
+		// Get relative path of file and filter for only Java files in src/main/java/ production trees.
```
JoeWang1127: Could you remove this comment as it doesn't add much of value.

internal/librarian/java/readme.go
```diff
@@ -38,6 +39,62 @@ var (
...
+}
+
+// parseCodeSample reads a Java sample file and constructs a codeSample struct with its title and relative path.
+func parseCodeSample(dir, file string) (codeSample, error) {
+	// Derive default title by stripping extension and converting CamelCase to space-separated words.
+	base := strings.TrimSuffix(filepath.Base(file), ".java")
+	title := decamelize(base)
+	titleOverride, err := extractTitle(filepath.Join(dir, file))
+	if err != nil {
+		return codeSample{}, fmt.Errorf("failed to extract title from %s: %w", file, err)
```
JoeWang1127:
```suggestion
		return nil, fmt.Errorf("failed to extract title from %s: %w", file, err)
```

internal/librarian/java/readme.go
yangyzs: Done.

internal/librarian/java/readme.go
yangyzs: Removed. Yes, it checks and filters out directories so only files are collected (if that is what you're asking).

internal/librarian/java/readme.go
yangyzs: Done.

internal/librarian/java/readme.go
yangyzs: Done.

JoeWang1127 (approved):

PR #6576, opened 2026-06-29
feat(internal/librarian): add debug command with env subcommand

### LINKED ISSUE #6374: proposal: create librarian env command
With more of our toolchain being configurable from the environment (`LIBRARIAN_CACHE` and `LIBRARIAN_BIN`), but still having defaults, it could be useful to have a command for printing out the `librarian` interpretation of the environment it is run in. This could also print out the various language tool installation locations.

This is probably most useful in debugging scenarios or whenever we end up building some automation.

```sh
$ librarian env

LIBRARIAN_CACHE=/absolute/path/to/.cache
LIBRARIAN_BIN=/absolute/path/to/.cache/bin

Language-specific tool installation directories:
  golang: /absolute/path/to/.cache/bin/golang_tools
  java: /absolute/path/to/.cache/bin/java_tools
  ...
```

hj690:
A new `debug` command with an `env` subcommand is added to the CLI to print the librarian environment. This includes resolved paths for `LIBRARIAN_CACHE`, `LIBRARIAN_BIN`, and language-specific tool installation directories. 
To support accessing the tool installation directories from the new command, the internal `getInstallDir` functions in the golang and java packages are exported as `InstallDir`.

__Local Testing__

run `go run ./cmd/librarian debug env`, result:
```
LIBRARIAN_CACHE=/usr/local/google/home/hongdaj/.cache/librarian
LIBRARIAN_BIN=/usr/local/google/home/hongdaj/.cache/librarian/bin

Language-specific tool installation directories:
  golang: /usr/local/google/home/hongdaj/.cache/librarian/bin/go_tools
  java: /usr/local/google/home/hongdaj/.cache/librarian/bin/java_tools
```

Fixes #6374


internal/librarian/debug_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"fmt"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+func TestRunEnv(t *testing.T) {
+	cacheDir := t.TempDir()
+	binDir := t.TempDir()
+
```
JoeWang1127: nit: remove this empty line.

internal/librarian/debug_test.go
```diff
@@ -0,0 +1,77 @@
...
+	t.Setenv("LIBRARIAN_CACHE", cacheDir)
+	t.Setenv("LIBRARIAN_BIN", binDir)
+
```
JoeWang1127: nit: remove this empty line.

internal/librarian/debug_test.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	got := buf.String()
+	wants := []string{
+		fmt.Sprintf("LIBRARIAN_CACHE=%s", cacheDir),
+		fmt.Sprintf("LIBRARIAN_BIN=%s", binDir),
+		fmt.Sprintf("golang: %s", filepath.Join(binDir, "go_tools")),
+		fmt.Sprintf("java: %s", filepath.Join(binDir, "java_tools")),
+	}
+
```
JoeWang1127: nit: remove this empty line.

internal/librarian/debug_test.go
```diff
@@ -0,0 +1,77 @@
...
+	for _, want := range wants {
+		if !strings.Contains(got, want) {
```
JoeWang1127, line 42: Are the strings not the same?

internal/librarian/debug_test.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	got := buf.String()
+	wants := []string{
+		"LIBRARIAN_CACHE=<error:",
+		"LIBRARIAN_BIN=<error:",
+		"golang: <error:",
+		"java: <error:",
+	}
+
```
JoeWang1127: nit: remove this empty line.

internal/librarian/debug_test.go
```diff
@@ -0,0 +1,77 @@
...
+	t.Setenv("XDG_CACHE_HOME", "")
+
+	var buf bytes.Buffer
+	if err := runEnv(&buf); err != nil {
+		t.Fatal(err)
+	}
+
+	got := buf.String()
+	wants := []string{
+		"LIBRARIAN_CACHE=<error:",
```
JoeWang1127, line 60: Is it possible to specify the error string?

JoeWang1127 (commented):
Could you create an issue for this feature?

We should document the rational for this feature in the issue.

internal/librarian/debug_test.go
hj690: This is the "failure" condition, i.e. if the output string doesn't include the expected path, then we fail the test. I intentionally avoided direct string comparison here so that we don't get flaky test from output format update etc.

internal/librarian/debug_test.go
hj690: It's possible, although my intention is to de-couple this test behavior from exact error message produced from `cache` module, since we care more about "error is printed" rather than "what error is printed" here, this would make test a bit less brittle from error message update for example. WDYT?

internal/librarian/debug.go
```diff
@@ -0,0 +1,75 @@
...
+	cacheDir := dirOrErr(cache.Directory())
+	buildDir := dirOrErr(cache.BinDirectory())
+	goToolsDir := dirOrErr(golang.InstallDir())
+	javaToolsDir := dirOrErr(java.InstallDir())
+	fmt.Fprintf(w, "LIBRARIAN_CACHE=%s\n", cacheDir)
+	fmt.Fprintf(w, "LIBRARIAN_BIN=%s\n", buildDir)
+	fmt.Fprintf(w, "\nLanguage-specific tool installation directories:\n")
+	fmt.Fprintf(w, "  golang: %s\n", goToolsDir)
+	fmt.Fprintf(w, "  java: %s\n", javaToolsDir)
+	return nil
```
JoeWang1127: Do we need this return since we never return an error.

JoeWang1127 (commented):
Sorry I have one more comment about the return type.

internal/librarian/debug.go
hj690: Good catch! Originally this return statement was just to make it so I can directly return the `runEnv` result in `Action` function above. But looking further, I noticed the writer error is indeed something I could catch and return here, I made the change.

JoeWang1127 (approved):

Issue #6527, opened 2026-06-22
java: move logic that updates versions.txt from `generate` to `add`


zhumin8:
Now that other languages are updating release please configs in `librarian add` step (https://github.com/googleapis/librarian/pull/6526), we should move steps in java that updates versions.txt from `generate` to `add` too. 
This is used in Java when onboarding new API paths (adding to existing library or adding as new library)

Current logic identifies modules to add line for with https://github.com/googleapis/librarian/blob/6ee3e9554a129098bfd2befcaa6c0e243a055250/internal/librarian/generate.go#L236
Then appends the lines in post-generate step:
https://github.com/googleapis/librarian/blob/6ee3e9554a129098bfd2befcaa6c0e243a055250/internal/librarian/java/postgenerate.go#L104-L108

It is relatively standalone logic and makes sense to move to `add` for consistency as it should have all info needed.

PR #6580, opened 2026-06-29
refactor(java): simplify transport lookup

noahdietz:
Simplifies the `lookupTransport` implementation, and by proxy, the `IdentifyMissingModules` implementation by introducing new `serviceconfig.FindTransport` API. This new API attempts to find the `API` entry in `sdk.yaml` by API path and then provide the `Transport` for said `API` entry. It does not need to read anything from `sources`, which the implementation previously required by using `serviceconfig.Find` (loads the entire API config from sources). `API.Transport` is entirely defined in librarian either by default or `sdk.yaml` entry.

Also refactors some common lookup code into an unexported `findAPI` method.

Towards #6527


internal/serviceconfig/serviceconfig.go
```diff
@@ -76,6 +76,24 @@ func Read(serviceConfigPath string) (*Service, error) {
...
+	}
+	var result *API
+	for _, api := range APIs {
+		if api.Path == path || api.OpenAPI == path || api.Discovery == path {
+			r := api
+			result = &r
+			break
+		}
+	}
+	return validateAPI(path, language, result)
```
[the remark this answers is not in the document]
noahdietz: Done

internal/serviceconfig/api_test.go
```diff
@@ -376,3 +376,57 @@ func TestRepoMetadataTransport(t *testing.T) {
 		})
 	}
 }
+
+func TestFindTransport(t *testing.T) {
```
JoeWang1127, line 387: Could you separate the error test to a different test: https://github.com/googleapis/librarian/blob/main/doc/howwewritego.md#separate-error-tests

noahdietz: Sure thing, done!

internal/serviceconfig/api_test.go
```diff
@@ -382,3 +382,68 @@ func TestRepoMetadataTransport(t *testing.T) {
...
+	}{
+		{
+			name:     "matching path but not allowed language",
+			path:     "google/ads/admanager/v1",
+			language: config.LanguageGo,
+		},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			_, err := FindTransport(test.path, test.language)
+			if err == nil {
```
JoeWang1127: Could you check the error here: https://github.com/googleapis/librarian/blob/main/doc/howwewritego.md#verify-with-errorsis-and-errorsas

noahdietz: Had to add a new error to actually test this. PTAL

JoeWang1127 (approved):

JoeWang1127:
I thought all APIs are allowed now, am I missing anything?

noahdietz:
> I thought all APIs are allowed now, am I missing anything?

Yes, they are.

Either an API has an entry in sdk.yaml that contains some overrides e.g. a language-specific `transport` override OR it doesn't have an entry and we use default values (which the `api.Transport()` helper does already).

Issue #6558, opened 2026-06-27
all: add `protoc` in tool section


JoeWang1127:
We should add `Protoc` to Tools section in librarian.yaml so that we can install protobuf compiler using `librarian install`.

`protoc` should be installed into a versioned directory within bin directory managed by Librarian.

JoeWang1127:
Reopen because we haven't add protoc section in librarian.yaml.

PR #6583, opened 2026-06-30
feat(internal/librarian): add `protoc` installation support

JoeWang1127:
Add the ability to install the protoc compiler from GitHub releases.

The install function downloads the protoc zip archive for the host platform, verifies its checksum, and extracts it to the librarian bin directory.

For #6558


internal/librarian/protoc.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func filter(path string) (string, bool) {
+	cleanPath := filepath.Clean(path)
+	if cleanPath == filepath.Join("bin", "protoc") ||
+		strings.HasPrefix(cleanPath, "include"+string(filepath.Separator)) {
+		return cleanPath, true
+	}
+	return "", false
+}
```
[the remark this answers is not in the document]
JoeWang1127: We extract all files, this is not relevant any more.

internal/librarian/protoc.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"github.com/googleapis/librarian/internal/cache"
+	"github.com/googleapis/librarian/internal/config"
+	"github.com/googleapis/librarian/internal/fetch"
+)
+
+// downloadURL returns the download URL for the protoc binary for the given version.
+var downloadURL = func(version string) string {
+	return fmt.Sprintf("https://github.com/protocolbuffers/protobuf/releases/download/v%s/protoc-%s-linux-x86_64.zip", version, version)
+}
```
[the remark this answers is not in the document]
JoeWang1127: Done.

internal/librarian/protoc_test.go
```diff
@@ -0,0 +1,177 @@
...
+		},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			got := downloadURL(test.version)
+			if diff := cmp.Diff(test.want, got); diff != "" {
+				t.Errorf("mismatch (-want +got):\n%s", diff)
+			}
+		})
+	}
+}
```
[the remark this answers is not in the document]
JoeWang1127: Done.

internal/librarian/protoc.go
```diff
@@ -0,0 +1,75 @@
...
+		"linux":  "linux",
+	}
+	archMap = map[string]string{
+		"arm64": "aarch_64",
+		"amd64": "x86_64",
+	}
+)
+
+// install installs the protoc tool.
+func install(ctx context.Context, protoc *config.Protoc) error {
```
noahdietz: This function name is a bit too ambiguous to be part of the broader `librarian` package. We should name it `installProtoc` or move this `protoc.go` file to a different subpackage.

This comment applies to most of the identifiers in this file.

I think the former is easier and I don't have a good idea of where it would go for the latter.

internal/librarian/protoc.go
```diff
@@ -0,0 +1,75 @@
...
+
+	"github.com/googleapis/librarian/internal/cache"
+	"github.com/googleapis/librarian/internal/config"
+	"github.com/googleapis/librarian/internal/fetch"
+	"github.com/googleapis/librarian/internal/filesystem"
+)
+
+var (
+	// downloadURL returns the download URL for the protoc binary for the given version and platform.
+	downloadURL = func(version string) string {
```
noahdietz: Maybe let's make the base (`https://github.com`) a global var that we can replace with the test server, and make this an actual function that we can test coverage for.

That would be cleaner code all around.

internal/librarian/protoc.go
```diff
@@ -0,0 +1,75 @@
...
+		return fmt.Sprintf("https://github.com/protocolbuffers/protobuf/releases/download/v%s/protoc-%s-%s.zip", version, version, platformSuffix())
```
noahdietz: embedding execution of  `platformSuffix` may make this a little harder to test on various platforms. Maybe we should have a `platform string` parameter that tests can just provide, rather than have the potential point of variance introduced by running this on different dev / CI machines.

internal/librarian/protoc.go
```diff
@@ -0,0 +1,75 @@
...
+// installDir returns the directory where the protoc binary should be installed.
+func installDir(version string) (string, error) {
+	binDir, err := cache.BinDirectory()
+	if err != nil {
+		return "", err
+	}
+	return filepath.Join(binDir, "protoc", fmt.Sprintf("v%s", version)), nil
+}
+
+func platformSuffix() string {
```
noahdietz: Nit: please add a comment, all the other funcs have them :)

noahdietz (changes requested):

noahdietz:
Stoked about this! I'm guessing a follow up will introduce the code changes to language packages to use the librarian installed protoc ?

internal/librarian/protoc.go
JoeWang1127: Good point, I actually forgot this file is part of `librarian` package.

internal/librarian/protoc.go
JoeWang1127: Done.

internal/librarian/protoc.go
JoeWang1127: Done.

internal/librarian/protoc.go
JoeWang1127: Done.

JoeWang1127:
> Stoked about this! I'm guessing a follow up will introduce the code changes to language packages to use the librarian installed protoc ?

Yes.

We choose to do it now b/c Java wants to release a LTS version, which requires different version of protoc.

internal/librarian/protoc.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+// downloadURL returns the download URL for the protoc binary for the given version, OS, and arch.
+func downloadURL(version, os, arch string) string {
+	suffix := platformSuffix(os, arch)
+	return fmt.Sprintf("%s/protocolbuffers/protobuf/releases/download/v%s/protoc-%s-%s.zip", githubURLBase, version, version, suffix)
+}
+
+// installDir returns the directory where the protoc binary should be installed.
+func installDir(version string) (string, error) {
```
noahdietz: I think you reverted the "fix func name" commit accidentally bc I still see `installDir` and such.

JoeWang1127: Thanks for catching this, updated to `protocInstallDir`.

internal/librarian/protoc.go
```diff
@@ -0,0 +1,72 @@
...
+		"darwin": "osx",
+		"linux":  "linux",
+	}
+	archMap = map[string]string{
+		"arm64": "aarch_64",
+		"amd64": "x86_64",
+	}
+)
+
+func downloadAndExtract(ctx context.Context, url, dir, sha256 string) error {
```
noahdietz: This name is still kind of generic for the `librarian` package - maybe just `installProtoc`

internal/librarian/protoc.go
```diff
@@ -0,0 +1,72 @@
...
+	tarball := filepath.Join(dir, "protoc.zip")
+	if err := fetch.Download(ctx, tarball, url, sha256); err != nil {
+		return err
+	}
+	defer os.Remove(tarball)
+	return filesystem.Unzip(ctx, tarball, dir)
+}
+
+// downloadURL returns the download URL for the protoc binary for the given version, OS, and arch.
+func downloadURL(version, os, arch string) string {
```
noahdietz: This name is still kind of generic for the `librarian` package - maybe just `protocDownloadURL`

noahdietz (approved):
Just two naming nits, but approving for merge once addressed.

internal/librarian/protoc.go
JoeWang1127: Done.

internal/librarian/protoc.go
JoeWang1127: Done.

PR #6584, opened 2026-06-30
refactor(internal/config): remove deprecated libraries_bom_version from JavaModule

### LINKED ISSUE #5171: java: define a JavaDefault instead of JavaModule

sofisl:
Finalizes the migration to JavaDefault by removing the deprecated libraries_bom_version field from the JavaModule struct, its merging logic, and its library-level fallback check in the generator. This is backwards compatibile because it is only defined once at the global default.java level in librarian.yaml in google-cloud-java. Follow up to https://github.com/googleapis/librarian/pull/6447.

Fixes https://github.com/googleapis/librarian/issues/5171


internal/librarian/java/repometadata_test.go
```diff
@@ -112,7 +112,7 @@ func TestDeriveRepoMetadata_Overrides(t *testing.T) {
 				APIDescription:       "Custom description",
 				ClientDocumentation:  "https://custom.client.docs",
 				ReleaseLevel:         s.ReleaseLevel,
-				Transport:            "both",
+				Transport:            "grpc+rest",
```
g-husam, line 115: does this break anything? same for other Transport field changes

g-husam (approved):
LGTM but might want someone more familiar to review

internal/serviceconfig/api.go
```diff
@@ -167,24 +167,10 @@ func (api *API) ReleaseLevel(language, version string) string {
...
-		}
-		switch transport {
-		case GRPC:
-			return "grpc"
-		case Rest:
-			return "http"
-		default:
-			return "both"
-		}
+	if language == config.LanguageJava && library != nil && library.Java != nil && library.Java.TransportOverride != "" {
```
JoeWang1127, line 172: Could you restore this change, looks like this is merged with https://github.com/googleapis/librarian/pull/6582

sofisl: you're right, good catch. Thanks!

internal/librarian/java/repometadata_test.go
sofisl: this should have been in https://github.com/googleapis/librarian/pull/6582, cleaned it up.

JoeWang1127 (approved):

PR #6585, opened 2026-06-30
chore(internal/librarian): add extractZip helper for protoc installation

JoeWang1127:
Add an extractZip helper function to the internal/librarian package to extract precompiled `protoc` ZIP archives to a destination directory.

This helper extracts only the protoc binary (inside the bin folder) and the protocol buffer header files (inside the include folder), ignoring other auxiliary files in the archive.

For #6558

### DIFF
--- a/internal/librarian/protoc.go
+++ b/internal/librarian/protoc.go
@@ -1,0 +1,75 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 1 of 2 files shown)


internal/librarian/protoc.go
```diff
@@ -0,0 +1,69 @@
...
+		}
+		defer rc.Close()
+		out, err := os.OpenFile(target, os.O_CREATE|os.O_RDWR|os.O_TRUNC, file.Mode())
+		if err != nil {
+			return err
+		}
+		defer out.Close()
+		if _, err := io.Copy(out, rc); err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
JoeWang1127: It should be fine as we verify the checksum before extracting.

internal/librarian/protoc.go
```diff
@@ -0,0 +1,71 @@
...
+			continue
+		}
+		if err := unzipFile(file, destDir, cleanPath); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func unzipFile(file *zip.File, destDir, cleanPath string) error {
```
zhumin8, line 46: Can we simplify and just use `unzip` command?
https://github.com/googleapis/librarian/blob/027103b817777f101d72ad7df3af2bf7255e7b6a/internal/filesystem/filesystem.go#L82

JoeWang1127: Looks like this helper function is not needed.

PR #6589, opened 2026-06-30
docs(adr): create decision record guide and template

g-husam:
View the rendered readme guide here: https://github.com/googleapis/librarian/blob/docs/adr-template/doc/decision-records/README.md

### DIFF
--- a/doc/decision-records/000000-decision-record-template.md
+++ b/doc/decision-records/000000-decision-record-template.md
@@ -9,6 +9,7 @@ The title should typically be formatted as "Choose X", and the file name should 
 ## Status
 
 What is the status, such as proposed, accepted, rejected, deprecated, superseded, etc.?
+When merged, there should be at least two rows, one for the proposed date, and one for the resultant state (such as Accepted or Rejected).
 
 * 2025-06-25: Proposed
 * 2025-07-01: Accepted
--- a/doc/decision-records/000000-decision-record-template.md
+++ b/doc/decision-records/000000-decision-record-template.md
@@ -1,6 +1,6 @@
 _This template is based on Michael Nygard's, available [here](https://github.com/architecture-decision-record/architecture-decision-record/blob/main/locales/en/templates/decision-record-template-by-michael-nygard/index.md)_.
 
-In each ADR file (named as `#####-choose-x.md`), write these sections:
+In each ADR file (named as `######-choose-x.md`), write these sections:
 
 # Title
 
--- a/doc/decision-records/README.md
+++ b/doc/decision-records/README.md
@@ -12,6 +12,8 @@ By capturing how and why a decision was made, it can be re-assessed much more ea
 Decision records are typically for documenting when a particular choice must be made among several options, whether it is a particular technology or a practice for the team to follow.
 They are not well-suited for larger designs that may involve a variety of intertwined design choices.
 
+The process is intended to be lightweight, and more focused on content and record-keeping with context than formality.
+
 Some examples of good candidates:
 
 1. Choosing a test framework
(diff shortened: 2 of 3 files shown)


doc/decision-records/README.md
```diff
@@ -0,0 +1,42 @@
...
+1. Choosing a vendor for capability X
+1. Choosing a code review methodology or workflow (if specific, if a more involved SDLC, may be better fit for a design doc)
+
+Some examples of unsuitable candidates:
+
+1. Designing a build system
+1. Designing an API
+1. Code generation implementation
+
+When unsure, ask a friend, or err on the side of writing a quick one by using Gemini to help structure your thoughts and context using the template linked below.
```
codyoss, line 34: Should we tell gemini about these files/dir and when to write/update one in `GEMINI.md`?

hj690 (approved):
💯

doc/decision-records/000000-decision-record-template.md
```diff
@@ -0,0 +1,27 @@
...
+
+The title should typically be formatted as "Choose X", and the file name should be based on this.
+
+## Status
+
+What is the status, such as proposed, accepted, rejected, deprecated, superseded, etc.?
+
+* 2025-06-25: Proposed
+* 2025-07-01: Accepted
+* 2025-12-15: Superseded
```
noahdietz, line 16: I kind of wish we just had `Current` and `Superseded` - I feel like we shouldn't submit the PR in `Proposed` state bc its not clear if its authoritative or not ("well it was checked in, but its not approved so...?"), and when we'd update it to `Accepted` or remove it altogether.

Basically I feel like any PR introducing a new one is effectively the `Proposed` state, and if it is checked in it is at first `Current`, until it is deprecated/superseded, at which point it is `Superseded`.

Thoughts?

g-husam: yeah good point, it should be in one of the (semi) terminal states, like accepted, rejected (still very valuable!!), superseded, etc.

g-husam: although I guess proposed captures when it was first created/started discussion

doc/decision-records/README.md
g-husam: yeah we can tell it about it in GEMINI.md, to reference when appropriate, I'll add (tho historically Gemini is pretty poor at following those instructions IME lol). Maybe not to write one, as I'd leave it to the human to decide that since its usually the human knowing when to write one and driving that decision.

g-husam: actually I'm not sure when this context is really needed, as its more historical reference, so might be better to just leave it out and let developers add to context when deemed necessary, to avoid polluting it?

doc/decision-records/000000-decision-record-template.md
noahdietz: Cool, I like the adjustments, thank you!

> although I guess proposed captures when it was first created/started discussion

The "Proposed date" to capture timeline is great, good idea.

doc/decision-records/README.md
noahdietz: Agreed that it might not always be relevant to loop gemini in. Example of when it would be: our design pattern for new CLI commands. Example of when it wouldn't be: why we chose a specific database.

Whether or not this readme has a breadcrumb to remind authors of looping in gemini, personally idc.

noahdietz (approved):

quirogas (approved):

doc/decision-records/README.md
g-husam: I'll leave it out of AGENTS.md for now, as I think having control of your agent's context is best, and AGENTS.md-equivalent files is often consuming context without as much benefit as explicitly provided instructions. It may be due to the lost-in-the-middle problem, as these files actually end up somewhere in the middle after the system prompt and other stuff. (refs: https://atlan.com/know/llm/lost-in-the-middle-problem/, https://arxiv.org/abs/2307.03172)

Issue #5271, opened 2026-04-09
sidekick: Swift supports custom LROs


coryan:
Some APIs (compute is what I have in mind, but container too) have custom LRO-type things. 

It may be possible to provide LRO helpers for them, if we write a small amount of hand-crafted code. I think in Rust that was less than 100 lines.

coryan:
This is done.

PR #6604, opened 2026-07-01
feat(librarian/swift): use discovery config

coryan:
Use the discovery configuration to generate `getOperation()` mixins in discovery-based APIs.

Towards #5271 

Verified that this has no effect on Rust and Swift.

### DIFF
--- a/internal/sidekick/api/xref.go
+++ b/internal/sidekick/api/xref.go
@@ -361,11 +361,11 @@ func sortOneOfFieldForExamples(f1, f2 *Field) int {
 }
 
 func enrichMethodSamples(m *Method) {
-	m.IsSimple = m.Pagination == nil &&
-		!m.ClientSideStreaming && !m.ServerSideStreaming &&
-		m.OperationInfo == nil && m.DiscoveryLro == nil
-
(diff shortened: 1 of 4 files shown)


internal/sidekick/api/xref.go
```diff
@@ -361,11 +361,11 @@ func sortOneOfFieldForExamples(f1, f2 *Field) int {
 }
 
 func enrichMethodSamples(m *Method) {
-	m.IsSimple = m.Pagination == nil &&
-		!m.ClientSideStreaming && !m.ServerSideStreaming &&
-		m.OperationInfo == nil && m.DiscoveryLro == nil
-
-	m.IsLRO = m.OperationInfo != nil
+	// Methods with AIP-151 LRO annotations *OR* discovery LRO annotations are LROs.
+	m.IsLRO = m.OperationInfo != nil || m.DiscoveryLro != nil
```
westarle, line 365: does this need a test change?

westarle (approved):

internal/sidekick/api/xref.go
coryan: Done.

PR #6624, opened 2026-07-04
refactor(internal/librarian): dispatch `fillDefaults` by language instead of type switch

JoeWang1127:
Replace the type switch in `fillDefaults` that inferred the language from which `Default` field was non-nil with a `languageDefaultFillers` map keyed by the explicit language string from `Config.Language`.

This eliminates the implicit priority ordering between languages (e.g. Java before Rust) and the fragile dispatch based on data rather than the known language identifier.

Each language-specific fill function now guards against its default field being nil, since the type switch no longer prevents calls when the language-specific default is absent.

Motivated by https://github.com/googleapis/librarian/pull/6554#discussion_r3502066428

### DIFF
--- a/internal/librarian/library.go
+++ b/internal/librarian/library.go
@@ -33,8 +33,19 @@ var (
 	errNoExplicitOutput = errors.New("library requires an explicit output path")
 )
 
+// languageDefaultFillers maps a language to a function that fills
+// language-specific library fields from the provided defaults.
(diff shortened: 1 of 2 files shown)


JoeWang1127:
Talked with @ianthehat, this change doesn't solve the design issue of `config.Config`: it ties the public API to a serialize form so they have to change in sync.

Issue #6509, opened 2026-06-18
nodejs: move tools install list to google-cloud-node/librarian.yaml


noahdietz:
As described in #5158, we need to move the "hardcoded" list of tools to install from `internal/librarian/nodejs/librarian.yaml` to the `google-cloud-node/librarian.yaml`: https://github.com/googleapis/librarian/blob/2bdb3b5f262bad05ce2a569828f06b9445ab78bd/internal/librarian/nodejs/librarian.yaml#L15-L22

This will impact our Docker image building workflow for Node.js: https://github.com/googleapis/librarian/blob/2bdb3b5f262bad05ce2a569828f06b9445ab78bd/cmd/librarian/Dockerfile#L287-L288

cc @hj690

noahdietz:
cc @quirogas

suztomo:
> This will impact our Docker image building workflow for Node.js

## Java

google-cloud-java employed the design: https://github.com/googleapis/google-cloud-java/blob/20999d390e3d3d7853f70623eefcf6d0c1e8cb2a/librarian.yaml#L24

Java does not use Docker.

## Go

google-cloud-go has employed the design: https://github.com/googleapis/google-cloud-go/blob/ad72e9e46158494e0043f1a29dd322c66ad06086/librarian.yaml#L21

Go does not use Docker.

## Python

google-cloud-python has **not** employed the design: https://github.com/googleapis/google-cloud-python/blob/8248d8e2f891e68d875d417fe64d7d9de4703b62/librarian.yaml#L20

Python uses Docker.


Unless we solve the problem of relying on the language mono repo in Docker build, moving the tools section means removing Docker from the operation. Maybe next step is to ensuring we can run the NodeJS-related Librarian operations without Docker.

## builddockerimages

`go run github.com/googleapis/librarian/tool/cmd/builddockerimages@latest --language nodejs --version=${V}`

This checks out librarian and runs Docker for https://github.com/googleapis/librarian/blob/main/cmd/librarian/Dockerfile.

suztomo:
@JoeWang1127 Am I right that if we implement this ticket, then NodeJS operation will look like Go and Java. (As we haven't implemented this yet, NodeJS operation looks like Python now.)

JoeWang1127:
> @JoeWang1127 Am I right that if we implement this ticket, then NodeJS operation will look like Go and Java. (As we haven't implemented this yet, NodeJS operation looks like Python now.)

[...]

PR #6646, opened 2026-07-06
build(docker): remove redundant PNPM_CONFIG_DANGEROUSLY_ALLOW_ALL_BUILDS inline env vars

suztomo:
Remove redundant `PNPM_CONFIG_DANGEROUSLY_ALLOW_ALL_BUILDS=true` inline variables in `cmd/librarian/Dockerfile` since it is already set as a global `ENV` variable. 

Additionally explicitly defines `PNPM_HOME=/root/.nvm/versions/node/current` to fix pnpm global path validation failures.

### Why `PNPM_HOME=/root/.nvm/versions/node/current` was chosen
This matches the logic in our Node.js installation helper where `PNPM_HOME` is set to the parent of the resolved `node` bin directory (so that `PNPM_HOME/bin` points directly to the `node` bin directory which is already in `PATH`):
https://github.com/googleapis/librarian/blob/7569927/internal/librarian/nodejs/install.go#L91-L94

Without `PNPM_HOME`, the build failed with the following error during `pnpm add -g`:
```
[ERROR] The configured global bin directory "/root/.local/share/pnpm/bin" is not in PATH
Run "pnpm setup" to update your shell configuration.
```

### Installation in `/opt` and Linking
To prevent broken symlinks and missing runtime dependencies (specifically `google-gax`, which is a `devDependency` in `package.json` but required at startup to resolve proto paths), the generator is extracted and built in a permanent directory: `/opt/gapic-generator-typescript`. 

We then run `pnpm add -g "$PWD"` inside that directory. This links the global command to `/opt/...` and preserves its `node_modules` (including `google-gax` installed during build time), mirroring the behavior of the Go `librarian install` helper.

This also resolves the cosmetic version resolution issue where `yargs` was showing the version of `nvm` (`0.40.4`) instead of the generator because it was resolving the parent directories of the global bin. It now correctly resolves the generator's `package.json` version (`4.12.1`).

### Verification

1. **Docker Image Build**:
   ```bash
   docker build --target nodejs -t librarian:nodejs -f cmd/librarian/Dockerfile .
   ```
   *Result*: Built successfully (no cache).

2. **Installed Binaries Verification**:
   We verify that the installed tools are executable and correctly configured in `PATH` (exiting with 0):

   ```bash
   docker run --rm --entrypoint /bin/bash librarian:nodejs -c "gapic-generator-typescript --version && gapic-node-processing --version && which pbjs"
   ```
   *Output*:
   ```
   4.12.1
   0.1.8
   /root/.nvm/versions/node/v22.23.1/bin/pbjs
   ```

For https://github.com/googleapis/librarian/issues/6509

### DIFF
--- a/cmd/librarian/Dockerfile
+++ b/cmd/librarian/Dockerfile
@@ -332,7 +332,9 @@ RUN pnpm add -g \
     pnpm install && \
     ./node_modules/.bin/tsc && \
     cp -a templates protos build/ && \
-    pnpm add -g "$PWD" && \
+    node -e "const fs = require('fs'); const f = 'package.json'; const p = JSON.parse(fs.readFileSync(f)); p.dependencies['google-gax'] = p.devDependencies['google-gax']; delete p.devDependencies['google-gax']; fs.writeFileSync(f, JSON.stringify(p, null, 2))" && \
+    pnpm pack && \
+    pnpm add -g ./*.tgz && \
     rm -rf /tmp/generator /tmp/generator.tar.gz
 
 # ============== Java Stage ==============
--- a/cmd/librarian/Dockerfile
+++ b/cmd/librarian/Dockerfile
@@ -317,7 +317,7 @@ ENV PNPM_CONFIG_STORE_DIR=/root/.nvm/versions/node/current/pnpm-store
 ENV PNPM_CONFIG_DANGEROUSLY_ALLOW_ALL_BUILDS=true
 
 # Install Node generation tools
-RUN PNPM_CONFIG_DANGEROUSLY_ALLOW_ALL_BUILDS=true pnpm add -g \
+RUN pnpm add -g \
         gapic-node-processing@${NODEJS_GAPIC_NODE_PROCESSING_VERSION} \
         gapic-tools@${NODEJS_GAPIC_TOOLS_VERSION} \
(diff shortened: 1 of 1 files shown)


cmd/librarian/Dockerfile
```diff
@@ -311,28 +311,33 @@ RUN chmod a+x /root
...
+    # source directory would break the global installation with MODULE_NOT_FOUND.
+    mkdir -p /opt/gapic-generator-typescript && \
+    tar -xzf /tmp/generator.tar.gz -C /opt/gapic-generator-typescript --strip-components=4 google-cloud-node-${NODEJS_GAPIC_GENERATOR_VERSION}/core/generator/gapic-generator-typescript && \
+    rm /tmp/generator.tar.gz && \
+    cd /opt/gapic-generator-typescript && \
     pnpm install && \
     ./node_modules/.bin/tsc && \
     cp -a templates protos build/ && \
-    PNPM_CONFIG_DANGEROUSLY_ALLOW_ALL_BUILDS=true pnpm add -g "$PWD" && \
-    rm -rf /tmp/generator /tmp/generator.tar.gz
```
suztomo, line 335: This "rm" command has moved to line 335.

quirogas (approved):

Issue #6516, opened 2026-06-19
java: rewrite Java post-processing logic to use functions in postprocessing package


JoeWang1127:
Update [postProcessAPI](https://github.com/googleapis/librarian/blob/main/internal/librarian/java/postprocess.go#L111) to apply postprocessing logic based on `postprocess.yaml`, if exists.

See go/sdk-librarian-mod-gen for more details.

PR #6651, opened 2026-07-06
feat(internal/config): add declarative postprocessing config schema

yangyzs:
Add declarative configuration structs for postprocessing operations in config.go and auto-generate updated schema tables in config-schema.md.

For #6516

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -335,6 +335,78 @@ type Library struct {
 
 	// Swift contains Swift-specific library configuration.
 	Swift *SwiftPackage `yaml:"swift,omitempty"`
+
+	// Postprocess contains post-processing operations executed after code generation.
+	Postprocess *Postprocess `yaml:"postprocess,omitempty"`
+}
+
+// Postprocess represents post-processing configuration options integrated into librarian.yaml.
+type Postprocess struct {
+	// Replace contains literal string replacement rules.
+	Replace []ReplaceConfig `yaml:"replace,omitempty"`
+
+	// ReplaceRegex contains regular expression replacement rules.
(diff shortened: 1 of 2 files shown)


internal/config/config.go
```diff
@@ -335,6 +335,78 @@ type Library struct {
 
 	// Swift contains Swift-specific library configuration.
 	Swift *SwiftPackage `yaml:"swift,omitempty"`
+
+	// Postprocess contains post-processing operations executed after code generation.
+	Postprocess *Postprocess `yaml:"postprocess,omitempty"`
```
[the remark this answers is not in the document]
yangyzs: Make sense. Done.

internal/config/config.go
```diff
@@ -337,6 +340,75 @@ type Library struct {
...
+	RemoveFile []string `yaml:"remove_file,omitempty"`
+
+	// MethodOperations contains method-level operations (`delete`, `duplicate`, `deprecate`).
+	MethodOperations []MethodOperation `yaml:"method_operations,omitempty"`
+}
+
+// MethodOperation represents a method-level operation like delete, duplicate, or deprecate.
+type MethodOperation struct {
+	// Path specifies the relative file path to modify.
+	Path string `yaml:"path"`
```
JoeWang1127, line 364: Do we intentional not set `omitempty`?

yangyzs: Yes. these fields (`path`, `action`, `func_name`, etc.) are mandatory for an operation to be valid and should never be empty, so `omitempty` is not added. Only optional fields use it.

JoeWang1127 (approved):

PR #6653, opened 2026-07-06
feat(java): append versions.txt on add

### LINKED ISSUE #6527: java: move logic that updates versions.txt from `generate` to `add`

noahdietz:
Moves the extension of the Java monorepo `versions.txt` file out of `librarian generate` and into `librarian add`. When a completely new API is onboarded with `librarian add` all relevant artifacts are appended to `versions.txt`. When a new API path for an existing library entry is onboarded, only the necessary artifacts (those that do not already appear in `versions.txt`) are appended to `versions.txt`.

This eliminates one of the post-`generate` steps implemented in librarian today. It also moves what is conceptually a one-time onboarding step into the purpose-built onboarding command. It does add some complexity to the Java `add` implementation as it now needs to do the following:
1. fill out the library entry with Java defaults
2. derive the artifact IDs
3. read `versions.txt`, then append only new artifacts
4. tidy the new library entry
5. return

Example commit of adding a completely new API: https://github.com/noahdietz/google-cloud-java/commit/2efbac63314a434cbd5393f2ac3ea2c39b87604b

Example commit of adding a new API path to an existing library: https://github.com/noahdietz/google-cloud-java/commit/85e200be74ec1c42c63e833993b67029a2247a97

Fixes #6527


internal/librarian/java/add.go
```diff
@@ -35,40 +40,201 @@ const (
...
+		if line == "" || strings.HasPrefix(line, "#") {
+			continue
+		}
+		parts := strings.Split(line, ":")
+		if len(parts) > 0 {
+			modules[parts[0]] = true
+		}
+	}
+	return modules, nil
+}
```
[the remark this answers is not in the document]
noahdietz: Done

internal/librarian/java/add.go
```diff
@@ -35,40 +40,201 @@ const (
...
+		versions = append(versions, fmt.Sprintf("%s:%s:%s", id, lib.Java.ReleasedVersion, lib.Version))
+	}
+	if err := appendVersions(versions); err != nil {
+		return nil, err
+	}
+
+	lib, err = Tidy(lib)
+	if err != nil {
+		return nil, err
+	}
```
noahdietz, line 106: This `Fill` + `Tidy` pattern worries me a bit, because if `Tidy` doesn't remove the default that `Fill` injects, we are explicitly including default config in an individual entry when we don't need to be and which will exclude it from updates to said default given that it will be considered "set". The alternative of making a local copy is hard due to the depth of nested pointer structs and slices.

Any other ideas?

noahdietz: Proposing a fix to `java.Tidy` to remove default `released_version` in https://github.com/googleapis/librarian/pull/6654 fwiw

internal/librarian/java/add.go
```diff
@@ -35,40 +40,201 @@ const (
...
 	// Java generation defaults to the system year for license headers,
 	// so we reset it here to avoid redundancy in librarian.yaml.
 	lib.CopyrightYear = ""
 
 	if lib.Java == nil {
 		lib.Java = &config.JavaModule{}
 	}
-	lib.Java.ReleasedVersion = defaultReleasedVersion
+	if lib.Java.ReleasedVersion == "" {
```
zhumin8: For existing libraries, this value can be omitted meaning can be derived from lib.Version.

internal/librarian/java/add.go
```diff
@@ -35,40 +40,201 @@ const (
 	defaultVersion         = "0.1.0-SNAPSHOT"
 	defaultReleasedVersion = "0.0.0"
 	fakeGroupID            = "please-configure-java-group-id"
+	versionsFileName       = "versions.txt"
 )
 
-// Add initializes a new Java library with default values.
-func Add(lib *config.Library) *config.Library {
-	lib.Version = defaultVersion
+// Add initializes a new Java library with default values and registers its modules in versions.txt.
```
zhumin8: > a new Java library

I believe you are also using this to add new API path to existing library

internal/librarian/java/add.go
noahdietz: Yep, great catch. Fixed by checking if we are adding a new api to an existing library, updated relevant test to use non-`defaultReleaseVersion` to ensure the correct versions were used, and produced [another example](https://github.com/noahdietz/google-cloud-java/commit/e5de48aa70ffad241ec6807d00905a071223e0eb) with a diff library confirming that `released_version` isn't set to `0.0.0`, and the versions used in `versions.txt` are relative to the existing versions.

internal/librarian/java/add.go
noahdietz: Thanks, fixed.

internal/librarian/java/add.go
noahdietz: > Proposing a fix to `java.Tidy` to remove default `released_version` in #6654 fwiw

Reverted and fixed elsewhere.

Seems like this _might_ be OK?

internal/librarian/java/add.go
noahdietz: PTAL and resolve if my response is satisfactory.

internal/librarian/java/add.go
```diff
@@ -35,40 +40,203 @@ const (
...
-	lib.Java.ReleasedVersion = defaultReleasedVersion
+	if lib.Java.ReleasedVersion == "" && addedAPI == nil {
+		lib.Java.ReleasedVersion = defaultReleasedVersion
+	}
 
 	// We use the first API to infer the group ID.
 	// It is unrealistic for a single library to mix cloud and non-cloud APIs.
 	apiPath := lib.APIs[0].Path
 	switch {
 	case strings.HasPrefix(apiPath, "google/shopping/"):
```
zhumin8, line 68: this can be done separately, but after https://github.com/googleapis/librarian/pull/6554, this logic setting group id for non cloud based on api path is redundant.

noahdietz: The PR and bug are marked closed - are they? 

I suppose its a matter of "who wins" - if Joe refactors this logic before this PR is submitted, I will have to update my PR. If I submit this sooner, then Joe just does the refactor as normal.

internal/librarian/java/add.go
zhumin8: I feel it's okay.
> which will exclude it from updates to said default given

Good call out. I'd argue we should fix `Tidy` in such cases.

noahdietz: > I'd argue we should fix `Tidy` in such cases.

Agreed. I think `Tidy` should remove individual config values that match the `default` for that language. We can find out if there are gaps :)

internal/librarian/java/add.go
```diff
@@ -35,40 +40,203 @@ const (
...
+	}
+
+	return lib, nil
+}
+
+func deriveAddedArtifactIDs(lib *config.Library, addedAPI *config.API) ([]string, error) {
+	libCoord := deriveLibraryCoordinates(lib)
+	var artifacts []string
+
+	addAPIArtifacts := func(api *config.API) error {
```
zhumin8: Feels like this is duplicating logic in`discoverModules` in internal/librarian/java/pom.go, can you perhaps extract a helper function `expectedAPIModules`  to share?

internal/librarian/java/add.go
```diff
@@ -35,40 +40,203 @@ const (
...
+			if shouldGenerateGAPIC(javaAPI) || shouldGenerateResourceNames(javaAPI) {
+				shouldGenerateClient = true
+			}
+			if err := addAPIArtifacts(api); err != nil {
+				return nil, err
+			}
+		}
+		if shouldGenerateClient {
+			artifacts = append(artifacts, libCoord.GAPIC.ArtifactID)
+		}
```
zhumin8: this logic for " initialize a new library, register all of its modules" sounds very similar to https://github.com/googleapis/librarian/blob/939ae6f16e82cadcdd09aa111d7f71a0e81062da/internal/librarian/java/pom.go#L127

Can we refactor and reuse same logic?

noahdietz: Sure thing, done!

internal/librarian/java/add.go
noahdietz: Yep! Refactored! PTAL!

internal/librarian/java/add.go
zhumin8: cc. @JoeWang1127 so you guys can race.

internal/librarian/java/add.go
```diff
@@ -110,55 +110,57 @@ func Add(lib *config.Library, addedAPI *config.API) (*config.Library, error) {
...
-			}
-			if err := addAPIArtifacts(api); err != nil {
-				return nil, err
+		// Reorder modules to match versions.txt expectation: Parent, BOM, APIs, Client.
+		var parent *expectedModule
+		var bom *expectedModule
+		var client *expectedModule
+		var apis []expectedModule
+
+		for _, m := range modules {
```
zhumin8: Do instead `for i := range modules {`.
 `m` is a copy of the data, leading to all pointers targeting a single, shared loop variable address.

noahdietz: This was fixed in Go 1.22 https://go.dev/blog/loopvar-preview#the-fix. The variable is initialized per iteration, not per loop, so we can take the variable without copying.

noahdietz: Also I think the tests would fail if it hadn't been fixed bc we'd only add one artifact (the last one to be set by the loop) to the versions.txt when we should be adding multiple.

internal/librarian/java/add.go
JoeWang1127: This is not needed anymore as the non-cloud group id can be derived from the custom group id mapping.

This feature is complete and the map is added to google-cloud-java: https://github.com/googleapis/google-cloud-java/blob/main/librarian.yaml#L55

internal/librarian/java/add.go
```diff
@@ -35,40 +40,205 @@ const (
...
+	}
+
+	return lib, nil
+}
+
+func deriveAddedArtifactIDs(lib *config.Library, addedAPI *config.API) ([]string, error) {
+	libCoord := deriveLibraryCoordinates(lib)
+	var modules []expectedModule
+
+	if addedAPI != nil {
```
JoeWang1127, line 115: Can we make this very long if-else a helper function?

internal/librarian/java/add.go
```diff
@@ -35,40 +40,205 @@ const (
...
+				filtered = append(filtered, art)
+			}
+		}
+		artifacts = filtered
+	}
+
+	return artifacts, nil
+}
+
+func readExistingModules(path string) (map[string]bool, error) {
```
JoeWang1127, line 191: Could you add a comment to this function, e.g., what is the keep of the returned map?

internal/librarian/java/add.go
```diff
@@ -35,40 +40,205 @@ const (
...
+	if len(lines) == 0 {
+		return nil
+	}
+	existing, err := os.ReadFile(path)
+	if err != nil {
+		return err
+	}
+	var buf bytes.Buffer
+	buf.Write(existing)
+	// Ensure the file ends with a newline before appending.
```
JoeWang1127: Could you add a comment about the reason?

internal/librarian/java/add_test.go
```diff
@@ -129,22 +154,100 @@ func TestAdd(t *testing.T) {
...
+				t.Fatal(err)
+			}
+			got, err := Add(test.lib, nil)
+			if err != nil {
+				t.Fatalf("Add() error = %v", err)
+			}
 			if diff := cmp.Diff(test.want, got); diff != "" {
 				t.Errorf("mismatch (-want +got):\n%s", diff)
 			}
+			content, err := os.ReadFile(versionsFileName)
```
JoeWang1127, line 247: nit: could we create a new test to verify `versions.txt`? I feel this test does to many things and the verification logic is complex.

internal/librarian/java/add_test.go
```diff
@@ -167,3 +270,72 @@ func TestDefaultLibraryName(t *testing.T) {
...
+	}
+}
+
+func TestAppendVersions_Error(t *testing.T) {
+	tmpDir := t.TempDir()
+	t.Chdir(tmpDir)
+	if err := os.Mkdir(versionsFileName, 0755); err != nil {
+		t.Fatal(err)
+	}
+	if err := appendVersions([]string{"line"}); err == nil {
```
JoeWang1127, line 403: Could you check the error?

internal/librarian/java/add.go
noahdietz: I don't think this is feature complete if [this code we are reviewing in `add` still exists](https://github.com/googleapis/librarian/blob/c90b861346145613eedfdead95e24081f13d9247/internal/librarian/java/add.go#L52-L70) - the logic was not actually switched over to the new `defaults` based solution if this `add` logic (which I'm only modifying) still exists, right? The application of defaults is only done on Generate, not Add. So as it stands on `HEAD` - Java `add` is doing one thing while Java `generate` is doing another thing.

I don't want to change this in the same PR as it is a different feature than what this PR is addressing.

It's also worth noting that we now have `java.Fill` (called by `librarian.fillLibraryDefaults`) and `librarian.fillJava` (called by `librarian.fillDefaults`) - this is confusing why do we have two different language oriented fill functions?

internal/librarian/java/add_test.go
noahdietz: This code is just moved from a different file, it was not newly authored. There is no sentinel error from these helpers to check for. Happy to add them but in a separate PR if that's OK.

internal/librarian/java/add_test.go
noahdietz: good idea. done

internal/librarian/java/add.go
noahdietz: I just moved this code so I don't know the original reason, but I've added a guess based on the function comment.

internal/librarian/java/add.go
noahdietz: good catch, done

internal/librarian/java/add.go
noahdietz: Yeah good call, this was messy. I kept the if/else in this method, but moved the massive `else` body into a new function. The branching of new api path vs. new library in this logic I think is important to keep here.

internal/librarian/java/add.go
noahdietz: I will need https://github.com/googleapis/librarian/pull/6754 and we will need to change the `Add` signature to include `cfg *config.Config` to access the `default.java` OR `librarian.Add` logic needs to pre-`fillDefaults`. This logic is a bit tangled. 

Not sure if I should merge this as is and follow up or block this PR until that refactoring is done.

JoeWang1127: OK, looks like we need more refactor which is outside the scope of this pull request.

I'm OK to make it simple for now.

JoeWang1127 (approved):

internal/librarian/java/add.go
noahdietz: Tracking via https://github.com/googleapis/librarian/issues/6755

internal/librarian/java/add_test.go
noahdietz: Tracking with https://github.com/googleapis/librarian/issues/6756

PR #6654, opened 2026-07-06
fix(java): tidy default released version

noahdietz:
Tidy the `java.released_version` if it was set to the default of `0.0.0`.

[Noticed this being serialized](https://github.com/noahdietz/google-cloud-java/commit/85e200be74ec1c42c63e833993b67029a2247a97) while testing a diff change.

### DIFF
--- a/internal/librarian/java/defaults.go
+++ b/internal/librarian/java/defaults.go
@@ -240,6 +240,10 @@ func tidyReleasedVersion(library *config.Library) {
 	if library.Java.ReleasedVersion == "" {
 		return
 	}
+	if library.Java.ReleasedVersion == defaultReleasedVersion {
+		library.Java.ReleasedVersion = ""
+		return
+	}
 	if !isSnapshot(library.Version) {
 		return
 	}
--- a/internal/librarian/java/defaults_test.go
+++ b/internal/librarian/java/defaults_test.go
@@ -468,6 +468,18 @@ func TestTidy(t *testing.T) {
 			},
 		},
 		{
+			name: "tidy released version if default",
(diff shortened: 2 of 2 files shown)


zhumin8:
I think this better be part of `add`. 
Based on [the example](https://github.com/noahdietz/google-cloud-java/commit/85e200be74ec1c42c63e833993b67029a2247a97), IIUC you are adding a new API path (google/cloud/kms/v2) to an existing library. 
So I expect `librarian add google/cloud/kms/v2` not changing released_version of the library. Note that the original config has `released_version` omitted because it can be derived from `version: 2.97.0-SNAPSHOT`,  it should be`released_version: 2.96.0` . `released_version: 0.0.0` is not a right config.
Removing this via `tidy` is saying this is a valid config, only has redundant info.

noahdietz:
> I think this better be part of `add`. Based on [the example](https://github.com/noahdietz/google-cloud-java/commit/85e200be74ec1c42c63e833993b67029a2247a97), IIUC you are adding a new API path (google/cloud/kms/v2) to an existing library. So I expect `librarian add google/cloud/kms/v2` not changing released_version of the library. Note that the original config has `released_version` omitted because it can be derived from `version: 2.97.0-SNAPSHOT`, it should be`released_version: 2.96.0` . `released_version: 0.0.0` is not a right config. Removing this via `tidy` is saying this is a valid config, only has redundant info.

Thanks and [your comment](https://github.com/googleapis/librarian/pull/6653#discussion_r3537541049) on the `versions.txt` PR clarified my confusion. Good catch. Closing this PR.

PR #6655, opened 2026-07-06
feat(librarian): config get libraries apiPath

### LINKED ISSUE #6475: proposal: Introduce `config find` to find library name for API Paths
### Motivation

When onboarding a new API either by human today or potentially with script, we need to know the name of the library associated with a specific API path (e.g., mapping `google/cloud/asset/v1` to `asset`).
```
librarian add "${API_PATH}"
# need to read the librarian.yaml addition, find out the name added
librarian generate <name>
```

Currently, the mapping and name derivation logic is internal to [add.go](https://github.com/googleapis/librarian/blob/main/internal/librarian/add.go), and language specific.
Exposing this via a CLI command enables clean orchestration:

```bash
# 1. Add config to librarian.yaml
librarian add google/cloud/newapi/v1

# 2. Find the library name
LIBRARY_NAME=$(librarian config find google/cloud/newapi/v1)

# 3. Generate the client library using the name
librarian generate "$LIBRARY_NAME"
```

### Proposed Command

`librarian config find <api_path>`

### Examples

```bash
# Output: asset
librarian config find google/cloud/asset/v1
```

### Implementation Details

1. Define a new command `find` under the `config` command group in
   [config.go](https://github.com/googleapis/librarian/blob/main/internal/librarian/config.go).
2. The command action loads `librarian.yaml`.
3. Checks if the API path is already associated with a library in the configuration.
5. Prints the library name to stdout.

Some corner cases to consider: API path that are associated with more than one library.

noahdietz:
Adds `libraries` config path support for `librarian config get` that, when followed by an API path, will attempt to find the library entry in `librarian.yaml` and return the library name for that entry, or an error if it doesn't exist.

Fixes #6475


cmd/librarian/doc.go
```diff
@@ -42,7 +42,7 @@ Usage:
 
 Usage:
 
-	librarian config get [path]
+	librarian config get [path] [value]
```
zhumin8, line 45: Can you add a bit more documenation about this new usage?

internal/librarian/config.go
```diff
@@ -90,3 +101,10 @@ func runConfigSet(path, value string) error {
 	}
 	return yaml.Write(config.LibrarianYAML, updated)
 }
+
+func libraryName(cfg *config.Config, apiPath string) (string, error) {
+	if library := findExistingLibraryForNewAPI(cfg, apiPath); library != nil {
```
zhumin8: Maybe worth renaming `findExistingLibraryForNewAPI` to `findExistingLibraryForAPI`?  This usage is not passing in a new API anymore

noahdietz: Sure! Done.

cmd/librarian/doc.go
noahdietz: Yeah! Added a `Description` with use cases

sofisl (approved):

JoeWang1127 (approved):

PR #6613, opened 2026-07-01
feat(internal/librarian/java): pre-validate java libraries bom version in config

sofisl:
This change modifies librarian to pre-validate that the LibrariesBOMVersion is present for Java libraries during the configuration parsing and tidying phase (`librarian tidy`). Previously, this check was performed deep in the
generation phase, which could result in a later, less predictable failure.

Fixes https://github.com/googleapis/librarian/issues/5152

PR #6674, opened 2026-07-07
feat(internal/librarian/java): pre-validate java libraries bom version in config

### LINKED ISSUE #5152: java: tidy should validate libraries bom version is present in default

sofisl:
A BOM version is now required and validated upfront in the config, and findBOMVersion is removed.

This reapplies the changes from pull request [6613](https://github.com/googleapis/librarian/pull/6613) (exactly the same) and adds a fix for the undefined findBOMVersion error in internal/librarian/java/readme.go by accessing the config value directly (removing the call to `findBOMVersion`, which we deleted).

Fixes https://github.com/googleapis/librarian/issues/5152


zhumin8:
Do you mind clarify what is the re-applied code and what is the fix?

sofisl:
> Do you mind clarify what is the re-applied code and what is the fix?

Sure - it's in the PR body but basically it's removing the call to `findBOMVersion` in `internal/librarian/java/readme.go` (which we deleted). See failure: https://github.com/googleapis/librarian/actions/runs/28871929651/job/85636736068

internal/librarian/tidy_test.go
```diff
@@ -87,6 +87,13 @@ func TestValidateLibraries(t *testing.T) {
 				Language:  test.language,
 				Libraries: test.libraries,
 			}
+			if test.language == config.LanguageJava {
```
JoeWang1127, line 90: Just an FYI, I think this test needs refactor as the verification logic is too complex, it's fine to do it in a separate pull request.

internal/librarian/java/pom.go
```diff
@@ -523,22 +523,21 @@ func writePOM(pomPath, templateName string, data any) (err error) {
...
 	}
-	return "", errMonorepoVersion
+	return "", errNotFound
+}
+
+func findMonorepoVersion(cfg *config.Config) (string, error) {
+	return findLibraryVersion(cfg, rootLibrary, errMonorepoVersion)
 }
 
 // TODO(https://github.com/googleapis/librarian/issues/6411):
```
JoeWang1127: Can we remove this TODO?

internal/librarian/java/pom.go
```diff
@@ -523,22 +523,21 @@ func writePOM(pomPath, templateName string, data any) (err error) {
 	return nil
 }
 
-func findMonorepoVersion(cfg *config.Config) (string, error) {
+func findLibraryVersion(cfg *config.Config, name string, errNotFound error) (string, error) {
```
JoeWang1127: Why do we need to pass an error as an input to the function?

sofisl: It's so that we can reuse it in other functions (`findParentPOMVersion`, `findMonoRepoVersion`) and they'll return the correct errors. Instead I'll just wrap it individually within each function.

internal/librarian/tidy_test.go
sofisl: I see what you're saying. I'll create an issue!

internal/librarian/java/default_test.go
```diff
@@ -682,6 +698,60 @@ func TestDeriveLastReleasedVersion_Error(t *testing.T) {
 	}
 }
 
+func TestValidate_Config(t *testing.T) {
+	for _, test := range []struct {
+		name string
+		def  *config.Default
+	}{
+		{
+			name: "valid bom version",
```
JoeWang1127, line 707: nit: we don't have to use a table driven test if there's only one test case.

internal/librarian/java/defaults_test.go
```diff
@@ -682,6 +698,60 @@ func TestDeriveLastReleasedVersion_Error(t *testing.T) {
 	}
 }
 
+func TestValidate_Config(t *testing.T) {
```
JoeWang1127:
```suggestion
func TestValidate(t *testing.T) {
```

internal/librarian/java/defaults_test.go
```diff
@@ -682,6 +698,60 @@ func TestDeriveLastReleasedVersion_Error(t *testing.T) {
...
+		t.Run(test.name, func(t *testing.T) {
+			cfg := &config.Config{Default: test.def}
+			if err := Validate(cfg); err != nil {
+				t.Fatal(err)
+			}
+		})
+	}
+}
+
+func TestValidate_ConfigError(t *testing.T) {
```
JoeWang1127:
```suggestion
func TestValidate_Error(t *testing.T) {
```

internal/librarian/tidy.go
```diff
@@ -179,6 +176,9 @@ func validateLibraries(cfg *config.Config) error {
 			errs = append(errs, fmt.Errorf("%w: %s (appears %d times)", errDuplicateAPIPath, path, count))
 		}
 	}
+	if err := validateLanguageConfig(cfg); err != nil {
```
JoeWang1127, line 179: The function name is `validateLibraries` but we are validating `config.Config`.

I think this part needs further refactor, could you create a todo?

internal/librarian/tidy_test.go
```diff
@@ -757,7 +767,7 @@ func TestTidy_UnusedSections(t *testing.T) {
 		{
 			name: "maven preserved",
 			cfg: &config.Config{
-				Language: config.LanguageJava,
+				Language: config.LanguageRust,
```
JoeWang1127: Why do we change this line?

internal/librarian/tidy.go
sofisl: I don't think this needs a refactor, if anything it needs a renaming of the top-level function. This logic was introduced here: xhttps://github.com/googleapis/librarian/pull/3103. It seems like validateLibraries only ever really touched the config. In any case I don't think this deserves a refactor.

JoeWang1127: I mean to refactor `validateLibraries`, not `validateLanguageConfig`.

internal/librarian/java/defaults_test.go
sofisl: I disagree with the naming but I did rename!

internal/librarian/tidy_test.go
sofisl: GREAT catch! This was laziness from jetski that I totally missed. Thank you!

JoeWang1127 (approved):

Issue #4490, opened 2026-03-10
config: move RequiresBilling to sdk.yaml


zhumin8:
RequiresBilling should not be a language specific properly, should move to sdk.yaml eventually.

Currently used for .repo-metadata.json, but not populated by internal/repometadata/repometadata.go.

In Java, it is an optional field in generation_config.yaml for hermetic_build (e.g. [here](https://github.com/googleapis/google-cloud-java/blob/e6bf14901e90b7cb252b11cfe3e941d1df9a193c/generation_config.yaml#L36)). So if we have this value in sdk.yaml, Java might still use some override values to reduce diffs for initial migration.

jskeet:
Moving to the Java post-stabilization milestone, as we don't use this in Python.

zhumin8:
Need to investigate where it is used downstream for java, check README and client doc generation.

zhumin8:
They are used in README.md template.
https://github.com/googleapis/google-cloud-java/blob/86d35e4caa364ce02766cb560d73d3ce6affc62d/sdk-platform-java/hermetic_build/library_generation/owlbot/templates/java_library/README.md?plain=1#L115
and
https://github.com/googleapis/google-cloud-java/blob/86d35e4caa364ce02766cb560d73d3ce6affc62d/sdk-platform-java/hermetic_build/library_generation/owlbot/templates/java_library/README.md?plain=1#L259

@JoeWang1127 @blakeli0 This info is migrated from generated_config.yaml, do you know the source of it? Looks like we currently have 29 not requiring billing. How useful is this info? Should we migrate it to sdk.yaml?

PR #6679, opened 2026-07-07
feat(internal/serviceconfig): add RequiresBilling to library config

sofisl:
This change adds RequiresBilling to api.go to allow specifying whether an API requires billing centrally in sdk.yaml. This allows all language generators to access this metadata instead of relying on language-specific overrides if needed. Specifically at least Java and Node use this setting.

For https://github.com/googleapis/librarian/issues/4490


internal/config/config.go
```diff
@@ -283,6 +283,9 @@ type Library struct {
 	// CopyrightYear is the copyright year for the library.
 	CopyrightYear string `yaml:"copyright_year,omitempty"`
 
+	// RequiresBilling indicates whether the API requires billing.
+	RequiresBilling *bool `yaml:"requires_billing,omitempty"`
```
noahdietz: This is being added to the `Library` struct when I think you want this added to the internal/serviceconfig/api.go struct? At least that is where we've traditionally put API-level overrides for repometadata stuff

sofisl: yes, thank you!

internal/librarian/java/repometadata.go
```diff
@@ -106,7 +106,19 @@ func deriveRepoMetadata(cfg *config.Config, library *config.Library, sourceDir s
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to find api config: %w", err)
+	}
+
+	if apiCfg.RequiresBilling != nil {
+		metadata.RequiresBilling = *apiCfg.RequiresBilling
+	} else if library.Java != nil {
+		metadata.RequiresBilling = !library.Java.BillingNotRequired
+	} else {
+		metadata.RequiresBilling = true
```
zhumin8: move this logic to internal/repometadata/repometadata.go

zhumin8:
nit: in description, this is for #4490 (the child issue)

internal/serviceconfig/api.go
```diff
@@ -90,6 +90,9 @@ type API struct {
 	// Optional. If omitted, the generator default is used.
 	ReleaseLevels map[string]string `yaml:"release_level,omitempty"`
 
+	// RequiresBilling indicates whether the API requires billing.
```
noahdietz, line 93: Nit: What is the more common case / default? Is the exceptional scenario where the API _doesn't_ require billing? Or is it exception for an API to required billing? I would assume the former.

It would be good to document that in the comments imo.

internal/serviceconfig/api_test.go
```diff
@@ -450,3 +451,31 @@ func TestFindTransport_Error(t *testing.T) {
 		})
 	}
 }
+
+func TestParseRequiresBilling(t *testing.T) {
```
noahdietz, line 455: Can we refactor this into a table driven test? Each test case would have a single entry sdk.yaml content, and a `want *bool` to compare with using `cmp.Equal` (which handles primitive pointers)

internal/librarian/python/generate_test.go
```diff
@@ -1207,6 +1208,7 @@ func TestCreateRepoMetadata(t *testing.T) {
 				ReleaseLevel:         "stable",
 				Language:             config.LanguagePython,
 				Repo:                 "googleapis/google-cloud-python",
+				RequiresBilling:      true,
```
zhumin8: Are these intended?

internal/librarian/rust/generate_test.go
```diff
@@ -734,6 +735,7 @@ func TestCreateRepoMetadata(t *testing.T) {
 				APIID:               "showcase.googleapis.com",
 				APIShortname:        "showcase",
 				ReleaseLevel:        "preview",
+				RequiresBilling:     true,
```
zhumin8: Is this intended change for rust?

internal/repometadata/repometadata.go
```diff
@@ -143,6 +146,16 @@ func fromAPI(cfg *config.Config, api *serviceconfig.API, library *config.Library
...
 		}
 	}
+
+	var requiresBilling bool
+	if api.RequiresBilling != nil {
+		requiresBilling = *api.RequiresBilling
+	} else if library.Java != nil {
+		requiresBilling = !library.Java.BillingNotRequired
+	} else {
+		requiresBilling = true
```
zhumin8: If only Java and Node uses this field, we should only populate if language is one of these?

internal/repometadata/repometadata.go
```diff
@@ -143,6 +146,16 @@ func fromAPI(cfg *config.Config, api *serviceconfig.API, library *config.Library
 			recommendedPackage = library.Java.RecommendedPackage
 		}
 	}
+
+	var requiresBilling bool
+	if api.RequiresBilling != nil {
+		requiresBilling = *api.RequiresBilling
+	} else if library.Java != nil {
+		requiresBilling = !library.Java.BillingNotRequired
```
zhumin8: Is this intended for backward compatible to google-could-java/librarian.yaml as we change this? If so, can you add a TODO item to remove once done?

internal/librarian/python/generate_test.go
sofisl: These are intended. I assumed, since we are adding this resources to the shared repo metadata, we'd want to add it to all languages. Also, seems harmless, since it's just an extra piece of metadata, and perhaps more confusing to "silence" it for other languages?

internal/repometadata/repometadata.go
sofisl: Hm, would we want to totally eliminate the possibility to override? I'd assume we'll essentially nullify it but do you also want to remove the logic eventually?

sofisl:
closing this in favor of https://github.com/googleapis/librarian/pull/6737

PR #6697, opened 2026-07-07
feat(swift): add protobuf generation support

nolanleastin:
This change introduces support for generating raw Swift Protobuf and gRPC-Swift stub files directly using protoc, bypassing the standard GAPIC client generation when requested. This is done by adding a new "swift-protobuf" template option to the Swift module configuration.

Produces the following: https://depot.code.corp.goog/experimental/google-cloud-swift/pull/300 

Issue: https://github.com/googleapis/google-cloud-swift/issues/452

### DIFF
--- a/internal/config/swift.go
+++ b/internal/config/swift.go
@@ -109,7 +109,7 @@ type SwiftModule struct {
 	// APIPath is the proto path to generate from (e.g., "google/storage/v2").
 	APIPath string `yaml:"api_path"`
 
-	// ModuleType is the type of module to generate (e.g., "swift-protobuf", "convert-swift", or empty for standard GAPIC).
+	// ModuleType is the type of module to generate (e.g., "swift-protobuf", "convert-swift", or empty/"default" for standard GAPIC).
 	ModuleType string `yaml:"module_type,omitempty"`
 
 	// IncludeList is a subset of proto files under the target API path to include.
--- a/internal/config/swift.go
+++ b/internal/config/swift.go
@@ -113,6 +113,8 @@ type SwiftModule struct {
 	ModuleType string `yaml:"module_type,omitempty"`
 
 	// IncludeList is a subset of proto files under the target API path to include.
+	// This is typically reserved for special cases (e.g., resolving dependency issues or
+	// compiling subsets of shared proto namespaces) and should not be used for standard libraries.
 	IncludeList []string `yaml:"include_list,omitempty"`
 
 	// ModulePath is the module import path or target containing stubs (used by convert-swift).
--- a/internal/config/swift.go
+++ b/internal/config/swift.go
@@ -109,8 +109,8 @@ type SwiftModule struct {
 	// APIPath is the proto path to generate from (e.g., "google/storage/v2").
 	APIPath string `yaml:"api_path"`
 
-	// Template is the generation template to use (e.g., "swift-protobuf", "convert-swift", or empty for standard GAPIC).
-	Template string `yaml:"template,omitempty"`
+	// ModuleType is the type of module to generate (e.g., "swift-protobuf", "convert-swift", or empty for standard GAPIC).
+	ModuleType string `yaml:"module_type,omitempty"`
 
 	// IncludeList is a subset of proto files under the target API path to include.
 	IncludeList []string `yaml:"include_list,omitempty"`
--- a/internal/config/swift.go
+++ b/internal/config/swift.go
@@ -113,8 +113,10 @@ type SwiftModule struct {
 	ModuleType string `yaml:"module_type,omitempty"`
 
 	// IncludeList is a subset of proto files under the target API path to include.
-	// This is typically reserved for special cases (e.g., resolving dependency issues or
-	// compiling subsets of shared proto namespaces) and should not be used for standard libraries.
+	// This is typically reserved for special cases to avoid generating unused/dead code.
+	// For example, in Storage we need Protobuf gencode for a subset of the protos in the google/type
+	// directory. This code is private to the package (google-cloud-storage in Rust, GoogleCloudStorage
+	// in Swift). All other files in google/type would be dead code.
 	IncludeList []string `yaml:"include_list,omitempty"`
 
 	// ModulePath is the module import path or target containing stubs (used by convert-swift).
--- a/internal/config/swift.go
+++ b/internal/config/swift.go
@@ -108,6 +108,15 @@ type SwiftModule struct {
 
 	// APIPath is the proto path to generate from (e.g., "google/storage/v2").
 	APIPath string `yaml:"api_path"`
+
+	// Template is the generation template to use (e.g., "swift-protobuf", "convert-swift", or empty for standard GAPIC).
+	Template string `yaml:"template,omitempty"`
+
+	// IncludeList is a subset of proto files under the target API path to include.
+	IncludeList []string `yaml:"include_list,omitempty"`
+
+	// ModulePath is the module import path or target containing stubs (used by convert-swift).
+	ModulePath string `yaml:"module_path,omitempty"`
 }
 
 // SwiftDiscovery contains discovery-specific configuration for LRO polling.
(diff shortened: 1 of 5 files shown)


internal/config/swift.go
```diff
@@ -108,6 +108,15 @@ type SwiftModule struct {
 
 	// APIPath is the proto path to generate from (e.g., "google/storage/v2").
 	APIPath string `yaml:"api_path"`
+
+	// Template is the generation template to use (e.g., "swift-protobuf", "convert-swift", or empty for standard GAPIC).
+	Template string `yaml:"template,omitempty"`
+
+	// IncludeList is a subset of proto files under the target API path to include.
+	IncludeList []string `yaml:"include_list,omitempty"`
```
noahdietz, line 120: When do we want to only generate a subset of protos in a package?

Would it be better to invert this to `ExcludeList` if it's only one or two files to avoid? Or do we want to avoid a new file being added to googleapis that is automatically picked up if not added to the `ExcludeList`?

noahdietz (commented):
Reviewing mostly from a Librarian config perspective.

noahdietz:
adding @coryan to reviewers from Swift side

internal/config/swift.go
coryan: > When do we want to only generate a subset of protos in a package?

https://github.com/googleapis/google-cloud-rust/blob/b91c00ea47501039c60815259a3bee92aaf07dbc/librarian.yaml#L1477-L1484

In Storage we need Protobuf gencode for a subset of the protos in the `google/type` directory.  This code is private to the package (google-cloud-storage in Rust, GoogleCloudStorage in Swift).  All other files would be dead code.

> 
> Would it be better to invert this to `ExcludeList` if it's only one or two files to avoid? Or do we want to avoid a new file being added to googleapis that is automatically picked up if not added to the `ExcludeList`?

The latter, we want to avoid new protos adding dead code unexpectedtly.

internal/config/swift.go
```diff
@@ -108,6 +108,15 @@ type SwiftModule struct {
 
 	// APIPath is the proto path to generate from (e.g., "google/storage/v2").
 	APIPath string `yaml:"api_path"`
+
+	// Template is the generation template to use (e.g., "swift-protobuf", "convert-swift", or empty for standard GAPIC).
```
coryan: Is this used?  If so, I would expect to see new files in the `.../swift/templates` directory?

internal/librarian/swift/generate_module.go
```diff
@@ -25,16 +30,84 @@ import (
 
 func generateModule(ctx context.Context, library *config.Library, src *sources.Sources) error {
 	for _, module := range library.Swift.Modules {
-		modelConfig := moduleToModelConfig(library, module, src)
-		model, err := parser.CreateModel(modelConfig)
+		switch module.Template {
+		case "swift-protobuf":
```
coryan, line 30: no action required: I wonder if this should be a different "language" vs. a different template.

internal/config/swift.go
noahdietz: Thank you, makes sense. Would it be reasonable to add some red tape to the documentation of this field that it is typically for special cases? Just a thought.

internal/librarian/swift/generate_module.go
```diff
@@ -25,16 +30,84 @@ import (
 
 func generateModule(ctx context.Context, library *config.Library, src *sources.Sources) error {
 	for _, module := range library.Swift.Modules {
-		modelConfig := moduleToModelConfig(library, module, src)
-		model, err := parser.CreateModel(modelConfig)
+		switch module.Template {
```
coryan: nit-ish: Seems like "template" is the wrong name, in Rust the "template" parameter referred to a sub-directory of the `.../rust/templates/` directory, which made some sense. Here it seems to be a completely different thing, more akin to a different language (as is the `.../sidekick/rust/` and `.../sidekick/rust_prost` directories).

internal/librarian/swift/generate_module.go
```diff
@@ -25,16 +30,84 @@ import (
...
-			return err
+		for _, entry := range entries {
+			if !entry.IsDir() && strings.HasSuffix(entry.Name(), ".proto") {
+				protoFiles = append(protoFiles, filepath.Join(apiPathAbs, entry.Name()))
+			}
 		}
 	}
-	return nil
+
+	if len(protoFiles) == 0 {
```
coryan: That seems like it should be an error, the configuration is wrong and it should not fail silently.

internal/librarian/swift/generate_module.go
```diff
@@ -25,16 +30,84 @@ import (
...
+				return err
+			}
+		default:
+			return fmt.Errorf("unknown template %q", module.Template)
+		}
+	}
+	return nil
+}
+
+func compileProtobufs(ctx context.Context, library *config.Library, module *config.SwiftModule, src *sources.Sources) error {
```
coryan: This probably belongs in a separate file, does not look like part of the "generate_module" things.

coryan (changes requested):
There is something awkward about a "template" parameter that does not control what template subdirectory to use.  Does this belong in a separate "language"?

internal/librarian/swift/generate_module.go
```diff
@@ -25,16 +30,84 @@ import (
...
+		}
+	}
+
+	for _, r := range sourceConfig.ActiveRoots {
+		addImport(sourceConfig.Root(r))
+	}
+	addImport(src.Googleapis)
+	addImport(src.ProtobufSrc)
+	addImport(src.Showcase)
+	addImport(src.Conformance)
```
noahdietz: Are these necessary if we are always setting `roots` in the librarian.yaml as necessary and it defaults to `googleapis` otherwise, which would be imported via the above `for` loop over `ActiveRoots`?

[Rust showcase](https://github.com/googleapis/google-cloud-rust/blob/b91c00ea47501039c60815259a3bee92aaf07dbc/librarian.yaml#L1351-L1353)

[Rust conformance](https://github.com/googleapis/google-cloud-rust/blob/b91c00ea47501039c60815259a3bee92aaf07dbc/librarian.yaml#L1689-L1694)

[Rust wkt](https://github.com/googleapis/google-cloud-rust/blob/b91c00ea47501039c60815259a3bee92aaf07dbc/librarian.yaml#L1656-L1662)

internal/librarian/swift/generate_module.go
```diff
@@ -25,18 +26,31 @@ import (
...
-		if err := sidekickswift.Generate(ctx, model, module.Output, modelConfig, library.Swift); err != nil {
-			return err
+		switch module.ModuleType {
+		case "swift-protobuf":
+			if err := compileProtobufs(ctx, library, module, src); err != nil {
+				return err
+			}
+		case "convert-swift":
+			return fmt.Errorf("module type %q is not yet supported", module.ModuleType)
+		case "":
```
coryan: optional: Should this have an explicit name too?  Maybe as in:


```suggestion
		case "", "default":
```

coryan (approved):

internal/sidekick/parser/discovery/discovery_test.go
```diff
@@ -175,6 +175,7 @@ func TestDeprecatedField(t *testing.T) {
 	gotMessage := model.Message(id)
 	if gotMessage == nil {
 		t.Fatalf("expected message %s in the API model", id)
+		return
```
coryan: AFAIK these `return` things after `t.Fatalf()` are not needed and are not idiomatic. Can you remove them?

coryan (changes requested):

coryan (approved):

internal/config/swift.go
```diff
@@ -108,6 +108,19 @@ type SwiftModule struct {
 
 	// APIPath is the proto path to generate from (e.g., "google/storage/v2").
 	APIPath string `yaml:"api_path"`
+
+	// ModuleType is the type of module to generate (e.g., "swift-protobuf", "convert-swift", or empty for standard GAPIC).
```
noahdietz: Nit: You've got `default` as a supported value synonymous with empty string. May be worth documenting.

noahdietz (approved):

PR #6731, opened 2026-07-09
feat(internal/librarian/java): add ApplyMoveActionsToLibrary helper and unit tests

yangyzs:
Add ApplyMoveActionsToLibrary helper function to relocate generated files from temporary source paths to target library directories while applying file preservation logic.

For #6516


internal/librarian/java/postprocess_test.go
```diff
@@ -1139,3 +1140,130 @@ func TestCreateOrVerifyOwlbotPy_Error(t *testing.T) {
...
+	destDir := filepath.Join(dir, "dest")
+	writeFiles(t, srcDir, map[string]string{"mismatch/file.txt": "content"})
+	writeFiles(t, destDir, map[string]string{"mismatch": "content"})
+	actions := []moveAction{
+		{src: srcDir, dest: destDir, description: "test files"},
+	}
+	err := ApplyMoveActionsToLibrary(actions, destDir, nil)
+	if !errors.Is(err, syscall.ENOTDIR) {
+		t.Errorf("ApplyMoveActionsToLibrary() error = %v, wantErr %v", err, syscall.ENOTDIR)
+	}
```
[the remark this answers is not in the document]
yangyzs: We only use Linux system for now.

internal/librarian/java/postprocess_test.go
```diff
@@ -23,6 +23,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"syscall"
 	"time"
```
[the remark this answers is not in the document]
yangyzs: See below.

internal/librarian/java/postprocess_test.go
```diff
@@ -1139,3 +1140,130 @@ func TestCreateOrVerifyOwlbotPy_Error(t *testing.T) {
...
+			}
+			got := readDirFiles(t, destDir)
+			if diff := cmp.Diff(test.wantFiles, got); diff != "" {
+				t.Errorf("mismatch (-want +got):\n%s", diff)
+			}
+		})
+	}
+}
+
+func TestApplyMoveActionsToLibrary_NonExistentSource(t *testing.T) {
```
JoeWang1127, line 1244: Should we return an error in this case?

internal/librarian/java/postprocess_test.go
```diff
@@ -1139,3 +1140,130 @@ func TestCreateOrVerifyOwlbotPy_Error(t *testing.T) {
...
+			t.Fatal(err)
+		}
+	}
+}
+
+func readDirFiles(t *testing.T, dir string) map[string]string {
+	t.Helper()
+	got := make(map[string]string)
+	err := filepath.WalkDir(dir, func(path string, d os.DirEntry, err error) error {
+		if err != nil || d.IsDir() {
```
JoeWang1127: Why returning an error if d is a directory?

internal/librarian/java/postprocess.go
```diff
@@ -535,3 +535,22 @@ func createOrVerifyOwlbotPy(outDir string) (err error) {
...
+			if err := os.MkdirAll(action.dest, 0755); err != nil {
+				return fmt.Errorf("failed to create directory %s: %w", action.dest, err)
+			}
+			err := filesystem.MoveAndMergeWithKeep(action.src, action.dest, destRoot, func(rel string) bool {
+				return shouldPreserve(rel, keepSet)
+			})
+			if err != nil {
+				return fmt.Errorf("failed to move %s: %w", action.description, err)
+			}
+		}
```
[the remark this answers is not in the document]
yangyzs: Great catch. changed.

internal/librarian/java/postprocess_test.go
yangyzs: In this case the `err` is `nil`, so it actually returns nil. 
But it's confusing, so we separated the checks for clarity.

internal/librarian/java/postprocess_test.go
yangyzs: Good point. I updated to return fs.ErrNotExist, but [TestGenerateAPI_NoTools](https://github.com/googleapis/librarian/blob/main/internal/librarian/java/generate_test.go#L446-L448) (which mocks runProtoc to do nothing) now fails during post-processing because the expected output folders do not exist on disk. The legacy logic silently skipped missing directories, which is likely why the test originally passed.

Do we want to update the test setups? Let me know what you think.

JoeWang1127: I think it's fine to keep it as-is for now.

Could you add a todo to modify this test after removing `owlbot.py`?

JoeWang1127 (approved):

internal/librarian/java/postprocess_test.go
yangyzs: Sure. Added a TODO with the issue.

Issue #5157, opened 2026-04-07
php: add support for librarian install


julieqiu:
Add support for `librarian install dotnet` and update CI jobs to use the command. See https://github.com/googleapis/librarian/issues/4848 for context.

See example implementations:

Python:
- https://github.com/googleapis/librarian/commit/9b32279fb48372e5da84bd3928ac1fe283decc3d
- https://github.com/googleapis/librarian/commit/25f1161899790dc97421d65c33c410d1ce93534d

NodeJS:
- https://github.com/googleapis/librarian/commit/83bdc5a1f2a694994f2e920a78992c0e1dc4b5f3

JoeWang1127:
Already supported

PR #6736, opened 2026-07-10
feat(tool/cmd/migrate/php): scaffold composer tools for php

sofisl:
This change modifies the migration command to dynamically scaffold the tools section with gapic-generator-php when migrating a PHP codebase. It also introduces the Composer struct that we'll use later during install.

For https://github.com/googleapis/librarian/issues/5157


internal/config/config.go
```diff
@@ -122,6 +125,24 @@ type CargoTool struct {
...
+	Name string `yaml:"name"`
+
+	// Version is the version to install.
+	Version string `yaml:"version"`
+
+	// Package is the URL or path of the package to install.
+	Package string `yaml:"package,omitempty"`
+
+	// Checksum is the SHA256 checksum of the package.
+	Checksum string `yaml:"checksum,omitempty"`
```
JoeWang1127: nit: we use `sha256` in other structs.

JoeWang1127 (approved):

PR #6737, opened 2026-07-10
feat(internal/librarian/java): populate RequiresBilling from API config in README template

### LINKED ISSUE #4490: config: move RequiresBilling to sdk.yaml

sofisl:
This PR populates the RequiresBilling field in the Java README directly from the API config rather than requiring it to be output into .repo-metadata.json. It completely removes the RequiresBilling field from Java's .repo-metadata.json serialization, which means we don't have to repopulate it.

This PR favors removing the field instead of repopulating it according to the discussion in https://chat.google.com/room/AAQA3zZCFy4/1upTQLCsD4E

Fixes https://github.com/googleapis/librarian/issues/4490


internal/librarian/java/readme.go
```diff
@@ -129,6 +131,18 @@ func renderREADME(params libraryPostProcessParams, keepSet map[string]bool) erro
 	if err != nil {
 		return fmt.Errorf("failed to extract samples: %w", err)
 	}
+
+	apiRequiresBilling := false
+	if len(params.library.APIs) > 0 {
+		api, err := serviceconfig.Find(params.primaryDir, params.library.APIs[0].Path, params.cfg.Language)
+		if err == nil && api.RequiresBilling != nil {
```
JoeWang1127: Could you separate the two conditions? Why do we need to continue if `err != nil`?

internal/librarian/java/readme.go
```diff
@@ -129,6 +131,18 @@ func renderREADME(params libraryPostProcessParams, keepSet map[string]bool) erro
...
+			apiRequiresBilling = *api.RequiresBilling
+		} else {
+			// TODO: Remove this logic once backward compatibility for google-cloud-java/librarian.yaml is no longer needed.
```
JoeWang1127: Could you link a issue for this todo?

internal/librarian/java/readme.go
sofisl: good call out. TIL Go would prefer to handle errors explicitly, separating out the business logic. Updated.

internal/librarian/java/readme.go
```diff
@@ -129,6 +131,22 @@ func renderREADME(params libraryPostProcessParams, keepSet map[string]bool) erro
...
+	apiRequiresBilling := false
+	if len(params.library.APIs) > 0 {
+		api, err := serviceconfig.Find(params.primaryDir, params.library.APIs[0].Path, params.cfg.Language)
+		if err != nil {
+			return fmt.Errorf("failed to find api config for %s: %w", params.library.APIs[0].Path, err)
+		}
+		if api.RequiresBilling != nil {
+			apiRequiresBilling = *api.RequiresBilling
+		} else {
+			// TODO: Remove this logic once backward compatibility for google-cloud-java/librarian.yaml is no longer needed.
```
JoeWang1127:
```suggestion
			// TODO(// https://github.com/googleapis/librarian/issues/6285): Remove this logic once backward compatibility for google-cloud-java/librarian.yaml is no longer needed.
```

internal/librarian/java/readme.go
```diff
@@ -129,6 +131,22 @@ func renderREADME(params libraryPostProcessParams, keepSet map[string]bool) erro
...
+			// https://github.com/googleapis/librarian/issues/6285
```
JoeWang1127: You can remove this line.

JoeWang1127 (approved):
One more comment on the TODO, then it's good to go.

Issue #6628, opened 2026-07-06
tool/cmd/migrate: support Php


JoeWang1127:
Ensure that a valid librarian.yaml can be created for Php.

zhumin8:
Closing the parent ticket, as the basics is populating with migrate script. 
Will continue to add/fix as we develop generate, or work on individual tickets prefixes with "tool/cmd/migrate".

PR #6739, opened 2026-07-10
fix(tool/cmd/migrate): populate API paths from .OwlBot.yaml during migrate for php

zhumin8:
Updates the PHP migration tool to parse `.OwlBot.yaml` (if present) and extract correct API paths from `deep-copy-regex` sources.
PHP library names (e.g., `Ces`, `AccessApproval`) do not reliably map to proto paths in `googleapis`. Need to explicitly list API path in librarian.yaml.

For https://github.com/googleapis/librarian/issues/6628

### DIFF
--- a/tool/cmd/migrate/php_test.go
+++ b/tool/cmd/migrate/php_test.go
@@ -91,3 +91,129 @@ func TestRunPHPMigration(t *testing.T) {
 		t.Errorf("mismatch (-want +got):\n%s", diff)
 	}
 }
+
+func TestExtractAPIPath(t *testing.T) {
+	tests := []struct {
+		name     string
+		source   string
+		wantPath string
+		wantOk   bool
+	}{
+		{
+			name:     "versioned api",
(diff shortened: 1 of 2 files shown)


tool/cmd/migrate/php_test.go
```diff
@@ -91,3 +91,127 @@ func TestRunPHPMigration(t *testing.T) {
...
+				}
+				return path
+			},
+		},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			dir := t.TempDir()
+			path := test.setupFile(dir)
+			_, err := extractAPIsFromOwlBot(path)
+			if err == nil {
```
JoeWang1127, line 212: We should check the error but maybe it's find here since the migration tool will only be used once.

JoeWang1127 (approved):

tool/cmd/migrate/php_test.go
zhumin8: I'll leave it as there is only one error case now.

PR #6742, opened 2026-07-10
refactor(internal/librarian/java): align default output logic with other languages

zhumin8:
Align Java's default output resolution with the common librarian flow, to rely on the common default-application and tidy logic, deduplicating code and removing Java-specific logic from the tidy and fill steps. Also if a global `default: output` is ever configured, Java libraries will correctly inherit it as a prefix.

Specifically, define and register `java.DefaultOutput`, remove `java.Fill` and  `java.Tidy` logic for output and delegate them to the common flows.

This is a pure refactoring and does not introduce any behavior changes or differences in `google-cloud-java` configurations or generation.


zhumin8:
note: I came across this while working on php. Went straight to this PR without an issue.

JoeWang1127 (approved):
Thanks.

Issue #6753, opened 2026-07-10
librarian: consolidate/reorganize language fill logic


noahdietz:
We have two layers of filling:
1. from the `default.{language}` level
2. derived at the library level

The first is applied by internal `librarian` APIs here: https://github.com/googleapis/librarian/blob/c90b861346145613eedfdead95e24081f13d9247/internal/librarian/library.go#L36-L52

The second applied by an internal `librarian` API that calls language-specific APIs here: https://github.com/googleapis/librarian/blob/c90b861346145613eedfdead95e24081f13d9247/internal/librarian/library.go#L361-L370

They are called in sequence here: https://github.com/googleapis/librarian/blob/c90b861346145613eedfdead95e24081f13d9247/internal/librarian/library.go#L334

This is rather confusing and error prone.

The fact that they are separate, for one, is not ideal. The fact that they must be called in sequence is also not ideal. The idea that pkg `librarian` owns `fillJava`, but pkg `java` owns `Fill` is also confusing.

We should refactor the filling and defaulting code to:
1. keep like-logic aligned and easy to use (not separated)
2. keep language specific logic in the language package, and language agnostic logic in librarian

PR #6754, opened 2026-07-10
refactor: move fillJava to java.FillDefaultJava

noahdietz:
Refactors Java-related filling and defaulting code into the Java package. No actual logic is changed, just moved things around to address separation of concerns and code colocation.

We still need to figure out how to consolidate `Fill` and `FillDefaultJava`.

Towards #6753


internal/librarian/java/default.go
JoeWang1127, line 1: I feel the name of this file should be `default.go`: https://github.com/googleapis/librarian/blob/main/doc/howwewritego.md#collection-names

internal/librarian/java/defaults.go
```diff
@@ -84,6 +85,35 @@ func Fill(library *config.Library) (*config.Library, error) {
 	return library, nil
 }
 
+// FillDefaultJava populates empty Java-specific fields in lib from the
+// [config.Default][], specifically from [config.JavaDefault][].
```
JoeWang1127: Why there are `[]` in the comment?

JoeWang1127 (approved):

internal/librarian/java/defaults.go
noahdietz: Oops, habit from protobuf markdown. We don't need the trialing `[]`, but bracketing the Go identifiers in the docs makes them linkable both in the IDE and in GoDoc.

Fixed the trailing `[]`

internal/librarian/java/default.go
noahdietz: Done!

PR #6757, opened 2026-07-10
feat(internal/librarian/java): add RestructureToLibrary helper and unit tests

yangyzs:
Add RestructureToLibrary helper function to reorganize and relocate generated Java source files (GAPIC, proto, gRPC, and samples) from intermediate directories into target library directories. It also handles conflicting file cleanup and copying of public proto files.

For #6516

### DIFF
--- a/internal/librarian/java/postprocess.go
+++ b/internal/librarian/java/postprocess.go
@@ -569,3 +569,83 @@ func ToKeepSet(keep []string) map[string]bool {
 	}
 	return keepSet
 }
+
+// RestructureToLibrary moves all generated source code to the library root directories.
+// It also removes conflicting files, and copies public proto files to the library.
+func RestructureToLibrary(params postProcessParams, destRoot string, keepSet map[string]bool) error {
+	tempProtoSrcDir := params.protoDir()
+	isCommonProtos := params.library != nil && params.library.Name == commonProtosLibrary
+	if !isCommonProtos {
+		if err := removeConflictingFiles(tempProtoSrcDir); err != nil {
+			return err
+		}
+	}
+	if err := moveSourcesToLibrary(params, destRoot, keepSet); err != nil {
+		return err
+	}
+	if err := copyProtoFilesToLibrary(params, destRoot); err != nil {
+		return err
+	}
+	return nil
+}
+
+// moveSourcesToLibrary relocates the generated Java source files (GAPIC, proto,
+// gRPC, resource names, and samples) to their repository destinations.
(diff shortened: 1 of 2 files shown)


internal/librarian/java/postprocess.go
```diff
@@ -569,3 +569,83 @@ func ToKeepSet(keep []string) map[string]bool {
 	}
 	return keepSet
 }
+
+// RestructureToLibrary moves all generated source code to the library root directories.
+// It also removes conflicting files, and copies public proto files to the library.
+func RestructureToLibrary(params postProcessParams, destRoot string, keepSet map[string]bool) error {
+	tempProtoSrcDir := params.protoDir()
+	isCommonProtos := params.library != nil && params.library.Name == commonProtosLibrary
```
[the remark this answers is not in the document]
yangyzs: Good Catch. Changed.

internal/librarian/java/postprocess_test.go
```diff
@@ -1260,6 +1260,162 @@ func TestApplyMoveActionsToLibrary_NonExistentSource(t *testing.T) {
...
+			gotFiles := readDirFiles(t, destDir)
+			if diff := cmp.Diff(tc.wantFiles, gotFiles); diff != "" {
+				t.Errorf("mismatch (-want +got):\n%s", diff)
+			}
+		})
+	}
+}
+
+// TestRestructureToLibrary_OverwritesExistingFiles verifies that existing files in the destination are overwritten.
+func TestRestructureToLibrary_OverwritesExistingFiles(t *testing.T) {
```
JoeWang1127, line 1351: How does this function behave if a file is removed from the src? Does the file exist after the move?

For example, say 2 files, `src/A.java`, `src/B.java`, are generated and moved to `dst/`. Then, at the regeneration time, only `src/A.java` is generated. In this case, will `dst/B.java` still exist after the move?

yangyzs: No, it won't exist. 

During regeneration, the pipeline runs a **Clean** phase first, which deletes all previously generated files (like `B.java`) from `dst/`. Because `B.java` is not regenerated, `MoveAndMergeWithKeep` inside silently skips it, so it remains deleted. 

The only exception is if `B.java` is added to the `keep` config: in that case, both the Clean and Move phases will skip it, preserving it in `dst/`.

This test wants to make sure even if the Clean phase fails, the Move phase still correctly overwrites existing files in `dst/`.

JoeWang1127 (approved):

PR #6766, opened 2026-07-11
fix(librarian/rust): detect inconsistent repos

coryan:
In Rust, to generate a new library the library needs to be inserted into the top-level `Cargo.toml` file *and* the new directory must be created. If only the directory exists, or only the entry in `Cargo.toml` exist, then the commands to update the state will fail with strange errors.

With this change, the user gets an explicit error and they can take action. I believe it would be unsafe to automatically repair the problem, but maybe with more experience we can figure out some default action to take.

### DIFF
--- a/internal/librarian/rust/crate_exists.go
+++ b/internal/librarian/rust/crate_exists.go
@@ -48,6 +48,8 @@ func crateExists(output string) (bool, error) {
 	if err != nil {
 		return false, err
 	}
+	// The production version of `Cargo.toml` is always formatted with `taplo
+	// fmt`. That guarantees the entries will follow this strict format.
 	hasCargoEntry := bytes.Contains(contents, fmt.Appendf(nil, "\n  \"%s\",\n", output))
 	if !hasCargoEntry && hasDirectory {
 		return false, fmt.Errorf("inconsistent repository state, crate missing in Cargo.toml, but the directory exists: %s", output)
--- a/internal/librarian/rust/crate_exists.go
+++ b/internal/librarian/rust/crate_exists.go
@@ -1,0 +1,59 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package rust
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+)
+
+// crateExists determines if a crate (the Rust mapping for "library") exists.
+//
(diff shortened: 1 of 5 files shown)


internal/librarian/rust/crate_exists.go
```diff
@@ -0,0 +1,59 @@
...
+			return false, fmt.Errorf("cannot access output directory %q: %w", output, err)
+		}
+	}
+	// The librarian commands are always executed at the top-level directory of
+	// the monorepo.
+	contents, err := os.ReadFile("Cargo.toml")
+	if err != nil {
+		return false, err
+	}
+	hasCargoEntry := bytes.Contains(contents, fmt.Appendf(nil, "\n  \"%s\",\n", output))
```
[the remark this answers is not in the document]
coryan: nonsense, as we use `taplo fmt`.

internal/librarian/rust/crate_exists.go
```diff
@@ -0,0 +1,61 @@
...
+	// the monorepo.
+	contents, err := os.ReadFile("Cargo.toml")
+	if err != nil {
+		return false, err
+	}
+	// The production version of `Cargo.toml` is always formatted with `taplo
+	// fmt`. That guarantees the entries will follow this strict format.
+	hasCargoEntry := bytes.Contains(contents, fmt.Appendf(nil, "\n  \"%s\",\n", output))
+	if !hasCargoEntry && hasDirectory {
+		return false, fmt.Errorf("inconsistent repository state, crate missing in Cargo.toml, but the directory exists: %s", output)
```
JoeWang1127, line 55: Could you create an error variable so it can be checked in the test?

internal/librarian/rust/crate_exists_test.go
```diff
@@ -0,0 +1,126 @@
...
+[workspace]
+resolver = "3"
+
+members = [
+ "src/auth",
+%s
+]
+`
+
+func TestCrateExists(t *testing.T) {
```
JoeWang1127, line 33: Could you separate the error cases: https://github.com/googleapis/librarian/blob/main/doc/howwewritego.md#separate-error-tests

internal/librarian/rust/crate_exists_test.go
```diff
@@ -0,0 +1,126 @@
...
+	t.Chdir(d)
+	// Create part of the path, but make it unreadable.
+	if err := os.MkdirAll("src", 0000); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile("Cargo.toml", []byte("# test only"), 0644); err != nil {
+		t.Fatal(err)
+	}
+	if got, err := crateExists("src/generated/unreachable"); err == nil {
+		t.Errorf("expected an error from crateExists(), got=%v", got)
```
JoeWang1127, line 125: Could you check the error?

internal/librarian/rust/crate_exists.go
coryan: No, I don't care what the error is, just that it returns an error.

internal/librarian/rust/crate_exists_test.go
coryan: I don't think we should. The error specifics are not interesting.

internal/librarian/rust/crate_exists_test.go
coryan: Done.

JoeWang1127 (approved):

PR #6767, opened 2026-07-11
feat(internal/librarian/golang): support protoc tool configuration

### LINKED ISSUE #6769: golang: run `protoc` from managed bin directory

JoeWang1127:
Update `internal/librarian/golang` to support executing a managed protoc installation during Go client generation.

A runProtoc helper function is introduced to execute the configured protoc tool binary if `tools.protoc` is defined in configuration, or fall back to system protoc from PATH when unconfigured.

Fixes https://github.com/googleapis/librarian/issues/6769

### DIFF
--- a/internal/librarian/golang/command.go
+++ b/internal/librarian/golang/command.go
@@ -29,16 +29,12 @@ const envPath = "PATH"
 
 // runProtoc runs the protoc command with the given environment variable and arguments.
 func runProtoc(ctx context.Context, pc *config.Protoc, env map[string]string, args ...string) error {
-	if pc == nil {
-		// Fallback to global protoc if no tool configured.
-		return runWithEnv(ctx, env, "protoc", args...)
-	}
 	// Ensure that the toolchain environment variables are set before calling protoc.
 	env, err := mergeEnv(env)
 	if err != nil {
 		return err
 	}
-	return protoc.Run(ctx, env, pc, args...)
+	return protoc.RunOrSystem(ctx, env, pc, args...)
 }
 
 // runWithEnv runs a command with the given environment.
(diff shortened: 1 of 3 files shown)


internal/librarian/golang/command.go
```diff
@@ -21,10 +21,26 @@ import (
...
 )
 
 const envPath = "PATH"
 
+// runProtoc runs the protoc command with the given environment variable and arguments.
+func runProtoc(ctx context.Context, pc *config.Protoc, env map[string]string, args ...string) error {
+	if pc == nil {
+		// Fallback to global protoc if no tool configured.
+		return runWithEnv(ctx, env, "protoc", args...)
+	}
```
zhumin8: use `protoc.RunOrSystem` to avoid duplicating this logic.

internal/librarian/golang/command.go
zhumin8, line 60: `buildCmd` from the core command package already merges PATHs. This is a bit redundant?
https://github.com/googleapis/librarian/blob/3086d632d01845d9e16f4701300d738406590a40/internal/command/command.go#L119-L120

We can just return the map and let command handle adding the PATH: 
`	res := map[string]string{envPath: toolsBinDir}`

internal/librarian/golang/command.go
JoeWang1127: Done.

internal/librarian/golang/command.go
JoeWang1127: This seems like redundant. I'll refactor it in follow ups.

Tracked via https://github.com/googleapis/librarian/issues/6785

zhumin8 (approved):

PR #6772, opened 2026-07-13
chore: exploring idea of Librarian reading .release-please-manifest.json

suztomo:
## Description

This draft PR explores resolving library release versions dynamically from `.release-please-manifest.json` during `librarian generate`.

### Context & Motivation
Currently, `librarian.yaml` contains explicit `version:` tags for each library, while Release Please ALSO maintains package versions in `.release-please-manifest.json`. During releases, Release Please updates both manifest files and `librarian.yaml` via `LibrarianYamlUpdater`, causing extra commits and state duplication.

### Proposed Solution
1. **Single Source of Truth**: Allow `.release-please-manifest.json` to serve as the single source of truth for library versions.
2. **Dynamic Resolution**: During `librarian generate`, Librarian reads `.release-please-manifest.json` if present at the workspace root and resolves each library's version by matching its output directory path or package name.
3. **Fallback & Compatibility**: If `.release-please-manifest.json` is not present or does not contain a given package, Librarian falls back to explicit versions declared in `librarian.yaml`.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -25,6 +25,9 @@ const (
 	// LibrarianYAML is the filename for the librarian configuration file.
 	LibrarianYAML = "librarian.yaml"
 
+	// ReleasePleaseManifest is the filename for the release-please manifest file.
+	ReleasePleaseManifest = ".release-please-manifest.json"
+
 	// RemoteUpstream is the default git remote name.
 	RemoteUpstream = "upstream"
 )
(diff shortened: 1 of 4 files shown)


suztomo:
This feature in Librarian is simple.

However, I changed my mind. I'm now thinking Release Please editing librarian.yaml is fine and that's actually the key feature of Release Please. For example, although Maven has capability to manage the versions in pom.xml (mvn versions:set), we let Release Please to edit the XML files. It's the same for Release Please + Librarian integration.

PR #6782, opened 2026-07-13
feat(tool/cmd/migrate): support union versions in PHP OwlBot configs

### LINKED ISSUE #6779: tool/cmd/migrate(php): explicitly list each version variant API path
In the current librarian.yaml populated, I see
```
  - name: SecretManager
      version: 2.3.0
      apis:
        - path: google/cloud/secretmanager
```

  We must explicitly list each version variant (e.g.  google/cloud/secretmanager/v1  and  google/cloud/secretmanager/v1beta2 ) instead of a single unversioned path ( google/cloud/secretmanager )

zhumin8:
Updates the PHP migration tool to support version union groups in `.OwlBot.yaml` (e.g. `(v1|v1beta2)`). Ensures individual versioned API paths are included in config for compound libraries.

Fixes #6779

### DIFF
--- a/tool/cmd/migrate/php.go
+++ b/tool/cmd/migrate/php.go
@@ -80,8 +80,14 @@ type deepCopyRegexSpec struct {
 	Dest   string `yaml:"dest"`
 }
 
+// extractAPIPaths extracts target API paths from an OwlBot source matcher pattern.
+// It supports both unversioned paths and versioned paths, including union matchers
+// (e.g. "(v1|v1beta2)") which are expanded into separate versioned paths.
+// Returns nil if the pattern is invalid.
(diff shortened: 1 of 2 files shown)


tool/cmd/migrate/php.go
```diff
@@ -80,14 +80,20 @@ type deepCopyRegexSpec struct {
 	Dest   string `yaml:"dest"`
 }
 
-func extractAPIPath(source string) (string, bool) {
+func extractAPIPaths(source string) []string {
```
JoeWang1127, line 87: Could you add a comment about what are the matching result?

JoeWang1127 (approved):

Issue #6791, opened 2026-07-13
tool/cmd/migrate(php): add protoc to tools section


zhumin8:
protoc install is language agnostic and should be ready to use.
Add a protoc config to the librarian.yaml for php

PR #6793, opened 2026-07-13
feat(too/cmd/migrate): specify protoc version in tools section

zhumin8:
Add protoc version to tool section in librarian.yaml produced. Version 31.0 comes from googleapis WORKSPACE [here](https://github.com/googleapis/googleapis/blob/2c4dd791cc44104f1fdc49b4cbdadb160772b11e/WORKSPACE#L181).

For #6791

### DIFF
--- a/tool/cmd/migrate/php.go
+++ b/tool/cmd/migrate/php.go
@@ -53,6 +53,9 @@ func runPHPMigration(ctx context.Context, repoPath string) error {
 					Build:   []string{"composer install"},
 				},
 			},
+			Protoc: &config.Protoc{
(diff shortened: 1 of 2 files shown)


suztomo:
As in other languages, we upgrade the protoc version from 31 to 33.2 unless there's a technical blocker.

zhumin8:
> As in other languages, we upgrade the protoc version from 31 to 33.2 unless there's a technical blocker.

Ack. Will create a followup to upgrade. I want to eliminate any potential diff created by protoc version at the current stage.

sofisl (approved):

Issue #6743, opened 2026-07-10
php(generate): add mixins to generate


zhumin8:
From local generate test on Ces, currently location mixins are missed.

Find source config for mixins (likely BUILD.bazel), migrate them into librarian.yaml as needed.
In generate, add protos to generate mixins.

zhumin8:
Took one example `Ces`, where generating with naive librarian code is missing location mixin.

### 1. How the old Bazel pipeline gets this info:

  In the  googleapis  directory under BUILD.bazel:
```
    proto_library_with_info(
        name = "ces_proto_with_info",
        deps = [
            ":ces_proto",
            "//google/cloud:common_resources_proto",
            "//google/cloud/location:location_proto",  # <--- Explicit dependency
        ],
    )
```
  • Because  location_proto  is listed under  deps  inside  proto_library_with_info , Bazel's internal rules resolve this dependency and pass  
  google/cloud/location/location.proto  directly as an input target file to the  protoc  command (not just as an import directory  -I ).
  • When the  gapic-generator-php  plugin runs:
      1. It reads the service YAML configuration (ces_v1.yaml) which specifies  google.cloud.location.Locations  under  apis .
      2. Because  location.proto  was passed directly as a target file to compile, the generator parses the  Locations  service.
      3. It matches  Locations  as an active mixin service, and generates the  getLocation  /  listLocations  mixin client code inside the output service clients (like AgentServiceClient.php ).


  ### 3. How we can resolve this in Librarian:

  To fix this, we should introduce support for  additional_protos  in the PHP configuration block of Librarian, mirroring how it is handled in Java.

  1. Config Definition: Update language.go to support  additional_protos  under  PHPAPI .
  2. Proto Gathering: Update generate.go to append these configured additional protos to  targetProtos .

  Once implemented, we would configure  librarian.yaml  like this:

      - name: Ces
        apis:
          - path: google/cloud/ces/v1
            php:
              additional_protos:
                - google/cloud/location/location.proto
Then in `add` command, we should also populate these additional_protos similar to how java do.

In currently onboarded languages, this is same pattern used in Java and Node. 
Python do not need this because the generator ( gapic-generator-python ) uses Python's multiple inheritance and delegates mixin calls at runtime to shared helper classes inside  google-api-core.
Go's client design guidelines do not use class-level mixin inheritance. Instead,common helper services like Locations and Operations are treated as completely separate clients (e.g.  cloud.google.com/go/location/apiv1 )

[...]

PR #6801, opened 2026-07-13
feat(internal/config): add additional_protos configuration support for php

zhumin8:
Add support for `additional_protos` in PHP package and API-level configuration structures. This enables specifying mixin proto dependencies (e.g., location.proto, iam_policy.proto) to compile them directly as target protos for client code generation.

For #6743

### DIFF
--- a/internal/librarian/library_test.go
+++ b/internal/librarian/library_test.go
@@ -1566,3 +1566,39 @@ func TestMergeRust(t *testing.T) {
 		})
 	}
 }
+
+func TestMergePHP(t *testing.T) {
+	for _, test := range []struct {
+		name string
+		dst  *config.PHPPackage
+		src  *config.PHPPackage
+		want *config.PHPPackage
+	}{
+		{
+			name: "nil src returns dst",
+			dst:  &config.PHPPackage{AdditionalProtos: []string{"a.proto"}},
+			src:  nil,
+			want: &config.PHPPackage{AdditionalProtos: []string{"a.proto"}},
+		},
+		{
+			name: "nil dst returns src",
+			dst:  nil,
+			src:  &config.PHPPackage{AdditionalProtos: []string{"b.proto"}},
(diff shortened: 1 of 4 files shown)


internal/librarian/library_test.go
```diff
@@ -1566,3 +1566,38 @@ func TestMergeRust(t *testing.T) {
...
+			want: &config.PHPPackage{AdditionalProtos: []string{"a.proto"}},
+		},
+		{
+			name: "nil dst returns src",
+			dst:  nil,
+			src:  &config.PHPPackage{AdditionalProtos: []string{"b.proto"}},
+			want: &config.PHPPackage{AdditionalProtos: []string{"b.proto"}},
+		},
+		{
+			name: "merges all fields",
```
JoeWang1127, line 1590: Could you add a test case to verify additional protos in the merging result has no duplicates?

JoeWang1127 (approved):

JoeWang1127:
I think the test should be fine if you merge main branch.

internal/librarian/library_test.go
zhumin8: It looks like that the "merging" in  ResolvePreview  is primarily designed as an override mechanism (preview replaces stable values), rather than a combining mechanism (union). So I don't think this is a good place to dedup.
However, src.AdditionalProtos  itself might contain duplicates and I will create a followup to Tidy to validate/dedup.

In the meantime, it looks like the merge functions for each language is a bit inconsistent and I am not super clear about what needs to be merged and what not. I will create a separate issue for clarifying and perhaps refactor for each language.

zhumin8: Filed https://github.com/googleapis/librarian/issues/6809 for ResolvePreview
And added note to https://github.com/googleapis/librarian/issues/6743#issuecomment-4970438762

PR #6810, opened 2026-07-14
feat(internal/librarian/php): integrate additional_protos into client generation

zhumin8:
Gather additional protos from package-level and API-level configurations, deduplicate them, and append them as target compile files to the protoc command.
This allows generator plugins to resolve and generate code for mixed-in APIs.

Tested locally with `librarian generate Ces`, looks like mixin (location) is generated as expected.

For #6743

### DIFF
--- a/internal/librarian/php/generate.go
+++ b/internal/librarian/php/generate.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"slices"
 	"sort"
 	"strings"
 
@@ -108,7 +109,9 @@ func generateAPI(ctx context.Context, params *generateAPIParams) error {
 		return err
 	}
 	opts := gapicOpts(params.api, apiMetadata, grpcConfigPath)
-	targetProtos, err := gatherTargetProtos(googleapisDir, params.api.Path)
+	additionalProtos := resolveAdditionalProtos(params.library, params.api)
+
+	targetProtos, err := gatherTargetProtos(googleapisDir, params.api.Path, additionalProtos)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


internal/librarian/php/generate.go
```diff
@@ -139,6 +142,9 @@ func gatherTargetProtos(googleapisDir, apiPath string) ([]string, error) {
 	if _, err := os.Stat(commonResources); err == nil {
 		targetProtos = append(targetProtos, commonResources)
 	}
+	for _, p := range additionalProtos {
+		targetProtos = append(targetProtos, filepath.Join(googleapisDir, filepath.FromSlash(p)))
+	}
```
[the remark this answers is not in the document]
zhumin8: In practice, there will be no duplicates. It any, it is a config issue and we should deal with it at validate config stage (tidy)

internal/librarian/php/generate.go
```diff
@@ -108,7 +109,9 @@ func generateAPI(ctx context.Context, params *generateAPIParams) error {
 		return err
 	}
 	opts := gapicOpts(params.api, apiMetadata, grpcConfigPath)
-	targetProtos, err := gatherTargetProtos(googleapisDir, params.api.Path)
+	additionalProtos := resolveAdditionalProtos(params.library, params.api)
+
```
JoeWang1127: nit: remove this empty line.

internal/librarian/php/generate.go
```diff
@@ -223,3 +229,15 @@ func gapicOpts(api *config.API, apiMetadata *serviceconfig.API, grpcConfigPath s
 func DefaultOutput(name, defaultOutput string) string {
 	return filepath.Join(defaultOutput, name)
 }
+
+func resolveAdditionalProtos(library *config.Library, api *config.API) []string {
```
JoeWang1127, line 232: Could you add a comment for this function?

internal/librarian/php/generate_test.go
```diff
@@ -181,11 +181,12 @@ func TestGapicOpts(t *testing.T) {
...
-		setupFiles []string
-		apiPath    string
-		wantProtos []string
-		wantErr    bool
+		name             string
+		setupFiles       []string
+		apiPath          string
+		additionalProtos []string
+		wantProtos       []string
+		wantErr          bool
```
JoeWang1127: Could you separate the test case: https://github.com/googleapis/librarian/blob/main/doc/howwewritego.md#separate-error-tests

It's fine to do it in a separate PR.

JoeWang1127 (approved):

internal/librarian/php/generate.go
zhumin8: I am going to refactor this in a followup to remove the package level config, will defer to add comment then (if not clear after simplifying)
See https://github.com/googleapis/librarian/issues/6743#issuecomment-4971042056

internal/librarian/php/generate_test.go
zhumin8: updated.

PR #6814, opened 2026-07-14
feat(tool/cmd/migrate): parse BUILD.bazel and populate AdditionalProtos for php

zhumin8:
Update findPHPLibraries in the PHP migration tool to parse BUILD.bazel files for each API. Extract Locations and IAM mixin dependencies from the deps list of proto_library_with_info targets, mapping them to their proto files, and write them to the generated librarian.yaml config as PHPAPI additional_protos.

See librarian.yaml produced [here](https://paste.googleplex.com/6029171531841536).

For #6743


tool/cmd/migrate/php.go
```diff
@@ -170,6 +191,75 @@ func findPHPLibraries(repoPath string) ([]*config.Library, error) {
...
+				// load this to dedicated config
+				if strings.Contains(dep, "common_resources_proto") {
+					continue
+				}
+				// Ignore LROs since PHP does not compile LRO methods as mixins.
+				if strings.HasPrefix(dep, "//google/longrunning:") {
+					continue
+				}
+				// Ignore policy_proto as it only defines structs; the IAMPolicy service is in iam_policy_proto.
+				if dep == "//google/iam/v1:policy_proto" {
```
JoeWang1127, line 222: Do you need these conditions when the `dep` only added to `additionalProtos` by the mapping key?

zhumin8: Technically no. But if we are not explicit with these conditions, they would fall in the `else` block and print a warning log.
I want to be explicit about what we are skipping deliberately

JoeWang1127 (approved):

PR #6820, opened 2026-07-14
feat(sidekick/swift): generate telemetry headers

### LINKED ISSUE #5940: sidekick: swift clients should include the x-goog-api-client header
The generated code should populate a constant with the:

- The client library language (swift)
- The client library type: gapic by default, but allow overrides to `gccl == veneer`.
- The client library version.
- The gax version.

At runtime that information is used to create the header and merged with the information from auth, which includes:

- The auth type (if known, auth returns it in the same header).
- The auth library version (ditto).

coryan:
Use the gax helper to initialize a `x-goog-api-client` header value (only once) and then use this value in each request. This requires propagating the package version information through several layers.

Fixes googleapis/librarian#5940


internal/librarian/swift/generate.go
```diff
@@ -90,6 +90,7 @@ func libraryToModelConfig(library *config.Library, apiCfg *config.API, src *sour
 		specFormat = library.SpecificationFormat
 	}
 
+	releaseLevel := svcConfig.ReleaseLevel(config.LanguageRust, library.Version)
```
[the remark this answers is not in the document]
coryan: Done

internal/sidekick/swift/templates/common/stub.swift.mustache
```diff
@@ -92,6 +92,7 @@ extension Clients {
       {{/Codec.QueryParams}}
       var req = try await self.inner.Request(path: path, query: query)
       req.httpMethod = "{{Codec.HTTPMethod}}"
+      req.setValue(Clients.clientHeader, forHTTPHeaderField: "X-Goog-Api-Client")
```
chingor13, line 95: nit: would it make sense to configure this once for the GoogleCloudGax.HTTPClient rather than need to generate it on every single RPC method stub?

chingor13 (approved):

internal/sidekick/swift/templates/common/stub.swift.mustache
coryan: I think you are right. I am going to merge as-is and refactor later, we need changes in gax first.

Tracked in https://github.com/googleapis/librarian/issues/6859

PR #6840, opened 2026-07-15
feat(swift): format all module output directories

nolanleastin:
Previously, the format runner was hardcoded to only run `swift-format` on `library.Output`, leaving secondary modules (like conversions generated under unformatted.

for https://github.com/googleapis/google-cloud-swift/issues/452

### DIFF
--- a/internal/librarian/swift/generate.go
+++ b/internal/librarian/swift/generate.go
@@ -53,14 +53,14 @@ func Generate(ctx context.Context, cfg *config.Config, library *config.Library, 
 func Format(ctx context.Context, library *config.Library) error {
 	// Collect unique output directories to format to avoid running the formatter
 	// multiple times on the same directory (e.g. library.Output and module.Output mapping to same path).
-	dirs := map[string]bool{}
+	dirs := map[string]struct{}{}
 	if library.Output != "" {
-		dirs[library.Output] = true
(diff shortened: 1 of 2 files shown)


internal/librarian/swift/generate.go
```diff
@@ -51,7 +51,29 @@ func Generate(ctx context.Context, cfg *config.Config, library *config.Library,
 
 // Format formats a generated Swift library.
 func Format(ctx context.Context, library *config.Library) error {
-	return command.Run(ctx, "swift-format", "format", "--in-place", "--recursive", library.Output)
+	// Collect unique output directories to format to avoid running the formatter
+	// multiple times on the same directory (e.g. library.Output and module.Output mapping to same path).
+	dirs := map[string]bool{}
```
coryan: nit: I am told the cool kids use `map[string]struct{}{}` ... an empty struct is to a zero-sized thing (making the map more compact and therefore more cache friendly). A `bool` always takes room.

coryan (approved):

internal/librarian/swift/generate.go
nolanleastin: 1. Sadly, since I didn't use it, I am not a cool kid. 
2. Given we have established (1), do I have to use it?
3. Also: if I use it will I become a cool kid? 

in more serious: fixed.

Issue #6630, opened 2026-07-06
php: add install command


JoeWang1127:
Add install support for Php.

Tools should be installed into subdirectory within Librarian Bin directory, see [code](https://github.com/googleapis/librarian/blob/ac43e52b3a539e9ad574680fcc9ce88ab51d1728/internal/cache/cache.go#L49) for how to find this directory. A similar implementation in nodejs package is https://github.com/googleapis/librarian/pull/6680.

Also, `librarian install` should work in the root directory of google-cloud-php repository.

JoeWang1127:
One way to proceed with this issue is to use `bazel` command to find the protoc command to generate a client, for example:
`bazelisk build --subcommands //google/cloud/dataform/v1:google-cloud-dataform-v1-php`

This command will output all commands in order to build a php library tarball.

In addition, synthtool is likely one of the tools since google-cloud-php has `owlbot.py`.

You can reference [code](https://github.com/googleapis/librarian/blob/a0930f4b0f52383ee4b19f23ae666ae4017f40f6/internal/librarian/python/librarian.yaml#L32) in python package on how to install synthtool.

JoeWang1127:
`protoc` is installed as a step before invoking language-specific install step: [code](https://github.com/googleapis/librarian/blob/main/internal/librarian/librarian.go#L108), so you don't need to implement it in `php` package.

sofisl:
I think all the sub-issues are done, is this ok to call done @zhumin8 and @JoeWang1127 ?

sofisl:
I'm going to close - feel free to reopen if it's not true!

PR #6799, opened 2026-07-13
feat(internal/librarian/php): install tools using composer

sofisl:
Adds the ability to iterate over tools.Composer from librarian.yaml for dynamic PHP tool installation. Also adds a pip installation that we'll need for the post-processor (synthtool). Retains the original installGenerator function as a fallback for workflows without librarian.yaml.

For #6630

PR #6843, opened 2026-07-15
fix(internal/librarian/php): update instructions and recursively create folder structure

sofisl:
Adds instructions to install php-bcmat (a plugin to PHP to give it the ability to perform math operations on numbers of any size) which is explicitly missing in Debian/Ubuntu envs. And recursively creates the folders necessary for the bash script in the installed dep directory.

For #6630


sofisl:
Blocked by https://github.com/googleapis/librarian/issues/6846

JoeWang1127:
> After https://github.com/googleapis/librarian/pull/6799, I wanted to attempt to generate a single library. I found two small bugs when generating a library, which were that php-bcmath was missing (a plugin to PHP to give it the ability to perform math operations on numbers of any size) which is explicitly missing in Debian/Ubuntu envs. And, a small bug to recursively create the folders necessary for the bash script in the installed dep directory.

Could you write what is the reason for this pull request?

internal/librarian/php/install.go
```diff
@@ -165,6 +165,9 @@ func generatorDir(ctx context.Context) (string, error) {
 func createBinWrapper(wrapperName, destPath, binDir string) error {
 	wrapperPath := filepath.Join(binDir, wrapperName)
 	content := fmt.Sprintf("#!/bin/sh\nexec %q \"$@\"\n", destPath)
+	if err := os.MkdirAll(filepath.Dir(wrapperPath), 0755); err != nil {
```
zhumin8, line 168: Can you add/modify test to cover this case? A nested wrapperName should do?

internal/librarian/php/DEVELOPMENT.md
```diff
@@ -10,7 +10,7 @@ system `PATH`:
 *   **Go** (to build the Librarian CLI)
 *   **PHP** and **Composer** (required to install and run the PHP generator
     plugin)
-    *   On Debian/Ubuntu (gLinux): `sudo apt-get install php-cli composer`
+    *   On Debian/Ubuntu (gLinux): `sudo apt-get install php-cli php-bcmath composer`
```
zhumin8, line 13: Thanks for adding.
Interesting. I see my env already has this installed. Wonder that is done differently

sofisl: aren you working in glinux? if so it should be an extra step. I think macos and windows have it natively

zhumin8: I am on glinux. (It doesn't hurt to explicitly install anyway)
Can you add a brief note to why it is needed?

internal/librarian/php/install_test.go
```diff
@@ -192,6 +194,46 @@ func TestInstall_Error(t *testing.T) {
...
+				t.Fatal(err)
+			}
+			wrapperPath := filepath.Join(binDir, test.wrapperName)
+			b, err := os.ReadFile(wrapperPath)
+			if err != nil {
+				t.Fatal(err)
+			}
+			want := fmt.Sprintf("#!/bin/sh\nexec %q \"$@\"\n", destPath)
+			if diff := cmp.Diff(want, string(b)); diff != "" {
+				t.Errorf("wrapper content mismatch (-want +got):\n%s", diff)
```
zhumin8, line 224: nit: Follow https://github.com/googleapis/librarian/blob/main/doc/howwewritego.md#use-cmpdiff-for-comparisons
```suggestion
				t.Errorf("mismatch (-want +got):\n%s", diff)
```

zhumin8:
> > After #6799, I wanted to attempt to generate a single library. I found two small bugs when generating a library, which were that php-bcmath was missing (a plugin to PHP to give it the ability to perform math operations on numbers of any size) which is explicitly missing in Debian/Ubuntu envs. And, a small bug to recursively create the folders necessary for the bash script in the installed dep directory.
> 
> Could you write what is the reason for this pull request?

Can you please address this and reword description per https://github.com/googleapis/librarian/blob/main/CONTRIBUTING.md#commit-messages?

zhumin8 (approved):

Issue #6638, opened 2026-07-06
nodejs: generation diff as of July 6th


suztomo:
```
V=$(go run github.com/googleapis/librarian/cmd/librarian@latest config get version)
go run github.com/googleapis/librarian/cmd/librarian@${V} tidy
```

There's no diff after the tidy command. Good.

For now (until https://github.com/googleapis/librarian/issues/6509), we use the Docker-based generation:

```
go run github.com/googleapis/librarian/tool/cmd/builddockerimages@latest --language nodejs --version=${V}
docker run -u $(id -u):$(id -g) -v .:/repo -v ~/.cache:/.cache -w /repo docker.io/library/librarian-nodejs:${V}  generate -v --all
```

Pull request creation failed:

```
suztomo@suztomo:~/librarian-2026/google-cloud-node$ git add . && git commit -m "chore: generation diff as of July 6th"
...
suztomo@suztomo:~/librarian-2026/google-cloud-node$ gh pr create --draft
? Where should we push the 'generation-diff-july-6th' branch? suztomo/google-cloud-node

Creating draft pull request for suztomo:generation-diff-july-6th into main in googleapis/google-cloud-node

? Title (required) chore: generation diff as of July 6th
? Choose a template PULL_REQUEST_TEMPLATE.md
? Body <Received>
? What's next? Submit as draft
remote: warning: File packages/google-cloud-aiplatform/protos/protos.js is 60.58 MB; this is larger than GitHub's recommended maximum file size of 50.00 MB        
remote: error: Trace: 5d11ea147b4f5e0c2de65ae57df0821cd99c10947e820974ff17ef788628b340        
remote: error: See https://gh.io/lfs for more information.        
remote: error: File packages/google-cloud-compute/protos/protos.js is 102.10 MB; this exceeds GitHub's file size limit of 100.00 MB        
remote: error: GH001: Large files detected. You may want to try Git Large File Storage - https://git-lfs.github.com.        
To https://github.com/suztomo/google-cloud-node.git
 ! [remote rejected]       HEAD -> generation-diff-july-6th (pre-receive hook declined)
error: failed to push some refs to 'https://github.com/suztomo/google-cloud-node.git'
failed to run git: exit status 1
```

https://github.com/googleapis/google-cloud-node/blob/main/packages/google-cloud-compute/protos/protos.js (as of today) has 98 MB.

Changing the push destination to the upstream (https://github.com/googleapis/google-cloud-node.git) didn't help:

```
[...]

Issue #4815, opened 2026-03-25
proposal: download protoc binary based on sources before librarian generation


JoeWang1127:
In an ideal world, librarian should generate all SDKs using the same version of protoc. This is not the case today, e.g., Rust is using v31.0 and Go is using v25.7.

This version diverge makes the on-call rotation toilsome. For example, on-call needs to download one version of protoc, generate one SDK and download another version before generating another SDK.

We should download protoc binary and cache it, the version can be specified via tag in sources section in librarian.yaml.

In this way, the protoc binary is downloaded or reused from cache when running `librarian generate` with the right version. No need to download it manually.

JoeWang1127:
Since Java uses the highest protoc version (v33.2) so far, we can update the issues or open new issues for protoc v33.2 compatibility.

noahdietz:
Now that we have `Tools` and `librarian install`, should we just make `protoc` a configurable tool to declare in librarian.yaml and installable via `librarian install`? 

See https://github.com/googleapis/librarian/issues/5282#issuecomment-4245842455

noahdietz:
> We probably do need a model where different languages can use different protoc versions, but that’s something we should handle after migrating Go generate and before onboarding the next language. Let's use this issue track that plan.

We are now in that position, and with Java onboarding approaching we need to look at this again. Java maintains multiple branches for LTS and those branches could differ in the tools they use, including protoc. Furthermore, I don't think the Librarian _team_ should be on the hook for managing the version of protobuf used in a language - like other `tools` with versions configured in `librarian.yaml` the language team can update them at will. If we enable the same CUJ for `protoc`, we do not need all of the `{lang}: determine protoc vX.Y compatibility` bugs we've filed - it just becomes config.

I think #5850 is the technical proposal to work through. I propose that we close this bug if we have consensus on the requirement of supporting varying versions of protoc across languages and focus on the other bug as a technical proposal to achieve that.

Thoughts?

zhumin8:
Plus one to `protoc` as config in `librarian.yaml`, it makes things clear especially in future Java LTS branches.

On the technical part, I think we got a similar issue to solve in java tool install:  https://github.com/googleapis/librarian/pull/6089#issuecomment-4501540910

noahdietz:
Closing this as "requirement accepted", moving convo on technical discussion to #5850

PR #6847, opened 2026-07-16
feat(internal/librarian/nodejs): install pnpm during librarian install

### LINKED ISSUE #6846: node: compatibility issue with pnpm v11.7.0, which strictly enforces that the PNPM_HOME directory is exactly present in the PATH
There is a bug in the main branch of nodejs which went undetected for a month due to the
following sequence of events:
1. June 15: GitHub Actions updated their runners to use pnpm v11.7.0.
2. The librarian CI aggressively caches the nodejs_tools installation, so
   the CI runners kept restoring the old cache (which used an older
   pnpm) and skipped the installation step entirely.
3. July 14: PR #6805 changed the cache key calculation, effectively
   invalidating the cache.
4. July 15: The next PR opened experienced a cache miss and was forced
   to run 'librarian install nodejs' from scratch. The newly downloaded
   pnpm v11.7.0 was finally executed, surfacing this strict PATH
   validation error.

this is blocking #6843 and maybe others.

quirogas:
Install a pinned version of `pnpm` inside Librarian's local tools directory during `install nodejs` instead of relying on the host's global `pnpm` binary to ensure environment uniformity.
 
For #6638
Fixes #6846

### DIFF
--- a/internal/librarian/nodejs/install.go
+++ b/internal/librarian/nodejs/install.go
@@ -24,7 +24,6 @@ import (
 	"strings"
 
 	"github.com/googleapis/librarian/internal/cache"
-	"github.com/googleapis/librarian/internal/command"
 	"github.com/googleapis/librarian/internal/config"
 	"github.com/googleapis/librarian/internal/fetch"
 )
@@ -51,16 +50,17 @@ func Install(ctx context.Context, tools *config.Tools) error {
 		return errNoToolsSpecified
 	}
 
-	for _, cmd := range []string{"node", "npm"} {
+	for _, cmd := range []string{"node", "corepack"} {
 		if _, err := exec.LookPath(cmd); err != nil {
 			return fmt.Errorf("%s %w: %w", cmd, errMissingExecutable, err)
 		}
 	}
 
-	cacheDir, err := cache.Directory()
+	env, err := getPNPMEnv()
 	if err != nil {
 		return err
 	}
+
 	pnpmVersion := tools.PNPMVersion
 	if pnpmVersion == "" {
 		for _, tool := range tools.PNPM {
@@ -70,24 +70,23 @@ func Install(ctx context.Context, tools *config.Tools) error {
 			}
 		}
 	}
-	if pnpmVersion == "" {
-		if _, err := exec.LookPath("pnpm"); err != nil {
-			return fmt.Errorf("pnpm %w: %w", errMissingExecutable, err)
-		}
-	} else {
-		installDir, err := InstallDir()
+
+	if !isPNPMInstalled(ctx, env, pnpmVersion) {
+		binDir, err := getBinDir()
 		if err != nil {
 			return err
 		}
-		npmCacheDir := filepath.Join(cacheDir, "npm-cache")
-		if err := command.RunStreaming(ctx, "npm", "install", "--prefix", installDir, "--cache", npmCacheDir, "-g", "pnpm@"+pnpmVersion); err != nil {
-			return fmt.Errorf("failed to bootstrap pnpm: %w", err)
+		if err := os.MkdirAll(binDir, 0o755); err != nil {
+			return err
 		}
-	}
-
-	env, err := getPNPMEnv()
-	if err != nil {
-		return err
+		if err := runCorepack(ctx, env, "enable", "--install-directory", binDir, "pnpm"); err != nil {
+			return fmt.Errorf("failed to enable pnpm via corepack: %w", err)
+		}
+		if pnpmVersion != "" {
+			if err := runCorepack(ctx, env, "install", "-g", "pnpm@"+pnpmVersion); err != nil {
+				return fmt.Errorf("failed to install pnpm via corepack: %w", err)
+			}
+		}
 	}
 
 	for _, tool := range tools.PNPM {
@@ -161,6 +160,8 @@ func getPNPMEnv() ([]string, error) {
 	}
 
 	env := os.Environ()
+	env = append(env, "COREPACK_HOME="+filepath.Join(cacheDir, "corepack"))
+	env = append(env, "COREPACK_ENABLE_DOWNLOAD_PROMPT=0")
 	env = append(env, "PNPM_HOME="+installDir)
 	env = append(env, "PNPM_CONFIG_GLOBAL_BIN_DIR="+binDir)
 	env = append(env, "NPM_CONFIG_GLOBAL_BIN_DIR="+binDir)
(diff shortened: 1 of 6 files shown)


internal/librarian/nodejs/install.go
```diff
@@ -131,8 +162,12 @@ func getPNPMEnv() ([]string, error) {
 	env := os.Environ()
 	env = append(env, "PNPM_HOME="+installDir)
 	env = append(env, "PNPM_CONFIG_GLOBAL_BIN_DIR="+binDir)
+	env = append(env, "NPM_CONFIG_GLOBAL_BIN_DIR="+binDir)
 	env = append(env, "PNPM_CONFIG_GLOBAL_DIR="+filepath.Join(cacheDir, "pnpm-global"))
+	env = append(env, "NPM_CONFIG_GLOBAL_DIR="+filepath.Join(cacheDir, "pnpm-global"))
 	env = append(env, "PNPM_CONFIG_STORE_DIR="+filepath.Join(cacheDir, "pnpm-store"))
+	env = append(env, "NPM_CONFIG_STORE_DIR="+filepath.Join(cacheDir, "pnpm-store"))
+	env = append(env, "NPM_CONFIG_CACHE="+filepath.Join(cacheDir, "npm-cache"))
```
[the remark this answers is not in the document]
quirogas: We don't support windows.

internal/librarian/nodejs/install.go
```diff
@@ -50,20 +51,51 @@ func Install(ctx context.Context, tools *config.Tools) error {
...
 		}
 	}
 
+	cacheDir, err := cache.Directory()
+	if err != nil {
+		return err
+	}
+	pnpmVersion := tools.PNPMVersion
+	if pnpmVersion == "" {
+		for _, tool := range tools.PNPM {
```
JoeWang1127: It seems weird that the pnpm version is defined in one of the tools that needs to be installed by pnpm.

internal/librarian/nodejs/install.go
```diff
@@ -50,20 +51,51 @@ func Install(ctx context.Context, tools *config.Tools) error {
...
 	}
 
-	for _, cmd := range []string{"node", "pnpm"} {
+	for _, cmd := range []string{"node", "npm"} {
 		if _, err := exec.LookPath(cmd); err != nil {
 			return fmt.Errorf("%s %w: %w", cmd, errMissingExecutable, err)
 		}
 	}
 
+	cacheDir, err := cache.Directory()
```
JoeWang1127: Could you move this line to line 78 since cacheDir is only used there.

internal/librarian/nodejs/install.go
```diff
@@ -50,20 +51,51 @@ func Install(ctx context.Context, tools *config.Tools) error {
...
+	if pnpmVersion == "" {
+		if _, err := exec.LookPath("pnpm"); err != nil {
+			return fmt.Errorf("pnpm %w: %w", errMissingExecutable, err)
+		}
+	} else {
+		installDir, err := InstallDir()
+		if err != nil {
+			return err
+		}
+		npmCacheDir := filepath.Join(cacheDir, "npm-cache")
```
JoeWang1127: I think the installed binary should live within `installDir/nodejs_tools`, not `cacheDir`.

internal/librarian/nodejs/install.go
```diff
@@ -50,20 +51,51 @@ func Install(ctx context.Context, tools *config.Tools) error {
...
+		}
+	}
+
 	env, err := getPNPMEnv()
 	if err != nil {
 		return err
 	}
 
 	for _, tool := range tools.PNPM {
+		if tool.Name == "pnpm" {
```
JoeWang1127: It seems weird that pnpm is one of the tools that installed by pnpm.

internal/librarian/nodejs/install.go
```diff
@@ -131,8 +163,12 @@ func getPNPMEnv() ([]string, error) {
 	env := os.Environ()
 	env = append(env, "PNPM_HOME="+installDir)
 	env = append(env, "PNPM_CONFIG_GLOBAL_BIN_DIR="+binDir)
+	env = append(env, "NPM_CONFIG_GLOBAL_BIN_DIR="+binDir)
```
JoeWang1127, line 155: Why do we need NPM envs?

JoeWang1127 (commented):
I think this pull request need some refactor to merge.

Could you also elaborate why we need this change, rather than what we have in the playbook:
```
nvm install 22
nvm use 22
npm install -g pnpm@7
```

internal/librarian/nodejs/install.go
JoeWang1127: Also, line 66 - 69 could refactor to a helper function.

internal/librarian/nodejs/install.go
quirogas: pnpm  v7 relies on  NPM_CONFIG_*  environment variables. This can be removed once we move to pnpm 8+

internal/librarian/nodejs/install.go
quirogas: This is the case. All binaries are installed in `$LIBRARIAN_BIN/nodejs_tools/bin` directory.

internal/librarian/nodejs/install.go
quirogas: removed this loop

internal/librarian/nodejs/install.go
```diff
@@ -119,22 +138,68 @@ func getPNPMEnv() ([]string, error) {
...
+	env = append(env, "COREPACK_ENABLE_DOWNLOAD_PROMPT=0")
+	env = append(env, "PNPM_HOME="+installDir)
 	env = append(env, "PNPM_CONFIG_GLOBAL_BIN_DIR="+binDir)
+	env = append(env, "NPM_CONFIG_GLOBAL_BIN_DIR="+binDir)
 	env = append(env, "PNPM_CONFIG_GLOBAL_DIR="+filepath.Join(cacheDir, "pnpm-global"))
+	env = append(env, "NPM_CONFIG_GLOBAL_DIR="+filepath.Join(cacheDir, "pnpm-global"))
 	env = append(env, "PNPM_CONFIG_STORE_DIR="+filepath.Join(cacheDir, "pnpm-store"))
+	env = append(env, "NPM_CONFIG_STORE_DIR="+filepath.Join(cacheDir, "pnpm-store"))
+	env = append(env, "NPM_CONFIG_CACHE="+filepath.Join(cacheDir, "npm-cache"))
+	env = append(env, "PNPM_CONFIG_DANGEROUSLY_ALLOW_ALL_BUILDS=true")
```
JoeWang1127, line 161: This env is introduced in pnpm 10 (https://pnpm.io/settings#dangerouslyallowallbuilds) and it introduces unsafe features, why do we need to add this for pnpm 7?

JoeWang1127:
Could you take a look at this comment:
https://github.com/googleapis/librarian/pull/6847#pullrequestreview-4709216369

JoeWang1127:
Overall, this pull request introduces a different pattern from other languages.

Install command in other language packages doesn't install language runtime and package manager. Do we have a reason to install pnpm (package manager)?

internal/librarian/nodejs/install.go
suztomo: I didn't find NPM_CONFIG was needed when it runs with PNPM v7.

suztomo:
I doubt this is needed right now. Without this pull request, we can use a specific version of PNPM (declared in the playbook).

@quirogas Did you get a chance to find a way to produce the same code among people's environments? PNPM might not be the single cause of the code generation differences https://github.com/googleapis/librarian/issues/6638.

If we really need to declare PNPM version, then I think it should allow different versions in different languages as done in protoc (https://github.com/googleapis/librarian/issues/4815). PHP is going to use PNPM, I heard.

internal/librarian/nodejs/install.go
quirogas: That is because we we're not using version 7 previously. Newer versions use PNPM_* prefixed variables.

internal/librarian/nodejs/install.go
quirogas: Good catch! we should remove it.

quirogas:
Closing PR for now as feature is not needed.

Issue #6857, opened 2026-07-16
node(install): refactor pnpm install function to internal/tool/pnpm


zhumin8:
Currently to install pnpm tools needed for php, it is calling nodejs.Install
https://github.com/googleapis/librarian/blob/b26b68022f12d8a4859de0b5fbfa325c9813c2e7/internal/librarian/php/install.go#L83-L88

There is some refactoring opportunity in the nodejs side, we should be able to just call install from a `pnpm` package like we do for `pip`

https://github.com/googleapis/librarian/blob/b26b68022f12d8a4859de0b5fbfa325c9813c2e7/internal/librarian/php/install.go#L76-L82

zhumin8:
@JoeWang1127 I put in the minimal fix for now in #6858, without fully refactoring nodejs.Install. 
Do you have a separate issue tracking the refactoring?

JoeWang1127:
We can use this ticket to track the refactorization.

zhumin8:
Updated title to reflect. Is there a better milestone for this issue?

JoeWang1127:
Updated the milestone.

PR #6858, opened 2026-07-16
refactor(internal/librarian/nodejs): expose InstallPNPM for reuse in php

zhumin8:
Expose InstallPNPM in the nodejs package to allow installing PNPM tools directly by passing a slice of PNPMTool config.  This is a minimum fix to allow php.Install to reuse and install `tools.PNPM` properly.

For https://github.com/googleapis/librarian/issues/6857

### DIFF
--- a/internal/librarian/nodejs/install.go
+++ b/internal/librarian/nodejs/install.go
@@ -46,7 +46,15 @@ var (
 
 // Install installs Node.js tool dependencies.
 func Install(ctx context.Context, tools *config.Tools) error {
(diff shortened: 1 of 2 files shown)


JoeWang1127 (approved):
Thanks for the refactor.

I think we can further move the pnpm install function to a dedicated package, e.g., `internal/tool/pnpm`.

zhumin8:
> I think we can further move the pnpm install function to a dedicated package, e.g., internal/tool/pnpm.

Agree. Just wanted to limit scope of this.

PR #6865, opened 2026-07-16
feat(dart): implement publish

brianquinlan:
Publishes every Dart package that needs published using this algorithm:
1. Find the dependencies of all packages in the repository.
2. Order the packages so that no package appears before any of its dependencies.
3. For each package in order, check if it should be published:
    a. If it has not been published before, it should be published.
    b. If it has been published before, check if the published version is less than the
         repository version. If it is, it should be published.
4. If the package is being published, check for semver violations using dart-apitool
    (https://github.com/bmw-tech/dart_apitool).
5. For each package in order that should be published, run `dart pub publish`.

The complete publishing process should look something like this:

```bash
$ librarian update # update all SDK descriptions
$ librarian generate --all # regenerate all packages
$ git commit …
$ librarian bump --all # inc the version number of changed packages
$ git commit …
$ librarian publish # publish packages that were updated
$ librarian tag # create a git tag for each published package
```

The tests use shell scripts to fake `dart pub` and `dart-apitool`. I'm not a big fan of this approach but it is the approach used by the Rust publish tests and I wanted to be consistent.


natebosch:
Can you update the first comment on the PR with some details we can put into the commit message when we land this?

go/cl-descriptions is the internal citation I use all the time, I wonder if I should track down an external one...

natebosch (approved):

brianquinlan:
> Can you update the first comment on the PR with some details we can put into the commit message when we land this?
> 
> go/cl-descriptions is the internal citation I use all the time, I wonder if I should track down an external one...

Done.

Issue #6815, opened 2026-07-14
librarian: generate a single library as a presubmit test in each language (potentially increasing timeout)


sofisl:

sofisl:
these are examples of breakages on main in java and go that would have been prevented by adding this test: https://github.com/googleapis/librarian/actions/workflows/java.yaml?query=is%3Afailure+branch%3Amain, https://github.com/googleapis/librarian/actions/workflows/go.yaml?query=is%3Afailure+branch%3Amain

specifically, this PR would have been caught by integration tests: https://github.com/googleapis/librarian/actions/runs/28871931208

JoeWang1127:
IMO, if a change breaks the postsubmit, we should:
1. revert the change or fix forward for apparent errors, e.g., typos.
2. find out the root cause and write a unit test to prevent it happen again.

We can't depend on integration tests to catch everything.

sofisl:
I agree that if we break the postsubmit, we should revert. But reverting is costly and we should try to limit this action as much as possible.

I also agree that we should figure out a root cause and try to write unit tests. But sometimes unit tests are not appropriate in certain circumstances - that's why we have integration tests. Obviously we cannot depend on integration tests to catch everything. That is not what this is trying to do. We are trying to limit the times that we have to revert a PR because an integration test that ran on postsubmit could have been caught during presubmit with relatively little effort.

I'm confused why this seems to have so much pushback - it doesn't significantly increase the presubmit execution time?

JoeWang1127:
> But reverting is costly and we should try to limit this action as much as possible.

Could you elaborate which part of revert is costly? We use GitHub UI which is basically a few clicks.

sofisl:
1 PR is less than 3 PRs (original PR, revert, then fix).

sofisl:
Blocked by https://github.com/googleapis/librarian/issues/6508 (see https://chat.google.com/room/AAQA3zZCFy4/6xRhP1wwGk8), since we'll want to create integration tests for python in order to create a smoke test as well.

It seems Dart, Swift and Rust are not eligible for these kinds of smoke tests, and PHP and Ruby are still being set up so we will do those separately. Once I finish Python, I can call this done.

PR #6867, opened 2026-07-16
ci(java): generate a single library as a smoke test during presubmits

sofisl:
This PR generates a single library (secretmanager) as part of the presubmit workflows for Java and Go. The presubmit tests do not exceed 5 min execution time.

For #6815

### DIFF
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -16,39 +16,54 @@ on: [push, pull_request, merge_group]
 permissions:
   contents: read
 jobs:
-  test:
+  build-and-test:
+    name: ${{ matrix.job-name }}
     runs-on: ubuntu-24.04
-    # Presubmit jobs must complete within 5 minutes. See CONTRIBUTING.md.
-    timeout-minutes: 5
-    steps:
-      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
-        with:
-          persist-credentials: false
-      - uses: ./.github/actions/setup-librarian
-      - name: Run tests
-        run: go run ./tool/cmd/coverage ./internal/librarian/golang
-  integration:
-    runs-on: ubuntu-24.04
-    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+    # Presubmit jobs must complete within 5 minutes. Integration might take longer.
+    timeout-minutes: ${{ matrix.timeout }}
+    # Skip integration test unless it's a push to main
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - job-name: 'Unit Tests'
+            task: 'unit-test'
+            timeout: 5
+          - job-name: 'Smoke Test'
+            task: 'smoke-test'
+            timeout: 5
+          - job-name: 'Integration Test'
+            task: 'integration'
+            timeout: 60
     steps:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
         with:
           persist-credentials: false
       - uses: ./.github/actions/setup-librarian
       - name: Checkout google-cloud-go
+        if: matrix.task != 'unit-test'
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
         with:
           repository: googleapis/google-cloud-go
           path: google-cloud-go
           persist-credentials: false
       - name: Run librarian install
+        if: matrix.task != 'unit-test'
         working-directory: google-cloud-go
         run: librarian install
-      - name: Run librarian generate
+      - name: Run Unit Tests
+        if: matrix.task == 'unit-test'
+        run: go run ./tool/cmd/coverage ./internal/librarian/golang
+      - name: Generate a single library (smoke test)
+        if: matrix.task == 'smoke-test'
+        working-directory: google-cloud-go
+        run: librarian generate secretmanager
+      - name: Run librarian generate all (integration test)
+        if: matrix.task == 'integration' && github.event_name == 'push' && github.ref == 'refs/heads/main'
         working-directory: google-cloud-go
         run: librarian generate --all
   create-issue-on-failure:
-    needs: [integration]
+    needs: [build-and-test]
     permissions:
       contents: read
       issues: write


.github/workflows/java.yaml
```diff
@@ -91,6 +91,11 @@ jobs:
           test -x "$(which protoc-gen-java_gapic)"
       - name: Run internal/librarian/java tests and check coverage
         run: go run ./tool/cmd/coverage ./internal/librarian/java
+      - name: Generate a single library (integration test)
+        working-directory: google-cloud-java
+        env:
+          PYTHONPATH: ${{ github.workspace }}/google-cloud-java/sdk-platform-java/hermetic_build/library_generation/owlbot
+        run: librarian generate secretmanager
```
zhumin8: I feel this either be its own test or part of `intergration` below. 
Are we just verifying `librarian generate secretmanager` does not return error? 

Also, note that recent changes to speed things up updated "Checkout google-cloud-java" step to only checkout selected files needed for `install` command. Means that this test now is not exactly running re-generate as it seems to do, it is generate secretmanager as if it's a new library.

sofisl: If we create a separate test, we'd have to rerun the expensive setup steps. This could potentially delay our job, or it might cause issues if gh ever experiences other outages regarding their runners. However it could be a non-issue that I'm concerned about, I could ask about it in tomorrow's meeting.

I (personally) think it's ok this job does nothing else than mirror the integration test in a (small) way. I basically think that this job should basically just help us avoid the situations where we say "oh no, I've broken main postsubmits with my PR". So if it just mirrors a slice of what's happening in integration tests, I think that's fine.

JoeWang1127:
It's weird to include integration test to a workflow for unit tests.

Besides, what is the confident level this test gives us? Why we choose this library but not another one? Can we use this library to test all generation code path?

Also leave a comment in https://github.com/googleapis/librarian/issues/6815#issuecomment-5034918959

.github/workflows/java.yaml
zhumin8: > "Checkout google-cloud-java" step to only checkout selected files needed for install command. Means that this test now is not exactly running re-generate as it seems to do, it is generate secretmanager as if it's a new library.

This needs to be fixed.

.github/workflows/go.yaml
```diff
@@ -25,8 +25,20 @@ jobs:
...
+          path: google-cloud-go
+          persist-credentials: false
       - name: Run tests
         run: go run ./tool/cmd/coverage ./internal/librarian/golang
+      - name: Run librarian install
+        working-directory: google-cloud-go
+        run: librarian install
+      - name: Generate a single library (integration test)
+        working-directory: google-cloud-go
+        run: librarian generate secretmanager
```
noahdietz: As discussed in team meeting, let's move this to its own `job` entry named `smoketest` (with the previous checkout and setup-librarian steps as well), so that we can keep the `test` job for unit tests and super fast/easily retriable.

.github/workflows/java.yaml
```diff
@@ -91,6 +91,11 @@ jobs:
           test -x "$(which protoc-gen-java_gapic)"
       - name: Run internal/librarian/java tests and check coverage
         run: go run ./tool/cmd/coverage ./internal/librarian/java
+      - name: Generate a single library (integration test)
```
noahdietz: I'm supportive of either structure (keeping it in same `test` job or making its own `smoketest` job with all of the above identical setup) but let's reword this to `smoke test`

If we can "share" the setup stuff some how, that would be great, perhaps by making a custom action _a la_ `setup-librarian` ([here](https://github.com/googleapis/librarian/blob/main/.github/actions/setup-librarian/action.yaml)). This can be done in a separate PR though, just file a todo and add a comment.

.github/workflows/java.yaml
noahdietz: I mentioned this in https://github.com/googleapis/librarian/pull/6867/changes#r3625599422 but maybe we should pull some of this Java setup into a custom action `setup-java`  to make this complex setup more manageable / cleaner

noahdietz (commented):
@JoeWang1127 good questions! Here are my thoughts: 

> It's weird to include integration test to a workflow for unit tests.

Agreed, this is actually a [smoke test](https://en.wikipedia.org/wiki/Smoke_testing_(software)). As discussed in team sync, let's present these tests as such. I think this has been a miscommunication, hopefully that helps clarify.

> Besides, what is the confident level this test gives us? Can we use this library to test all generation code path?

See above classification of smoke test, I think that will help get us all on the same page.

> Why we choose this library but not another one?

At first, it probably doesn't matter which we pick, the class of errors we encountered that we want to catch with this are more related to config schema compatibility. We could go with a more robust case at first or just wait until we need to switch it/add a new one to cover another error case that this missed and we encountered.

.github/workflows/java.yaml
sofisl: @zhumin8 I don't know what you mean here:

> > "Checkout google-cloud-java" step to only checkout selected files needed for install command. Means that this test now is not exactly running re-generate as it seems to do, it is generate secretmanager as if it's a new library.
> 
> This needs to be fixed.

Do you mean that instead for this specific job we should check out the whole library and attempt to regenerate secretmanager?

As I said above, I think it's fine and even encouraged if the smoke test is testing in the same way the integration test is testing, since ultimately that's what we base our decision on whether or not to revert. And, if we checked out the whole repo during presubmits, we would potentially be delaying our test by a lot. Maybe I don't understand what you mean.

In the meantime I have created a new job and consolidated the setup steps by using github's `matrix` testing functionality! Let me know what you think.

sofisl: Oh nevermind I see what you're saying, you're saying only the presubmit job is sparsely checked and the integration is fully checked out. I'll fix it and confirm it takes a reasonable time.

noahdietz (approved):
LGTM, we will likely need to change the branch required checks after this goes in (same as Go), so let's make sure to time it so that we don't disrupt things. Please let @zhumin8 double check the Java setup stuff.

.github/workflows/java.yaml
```diff
@@ -90,49 +116,22 @@ jobs:
...
+        run: librarian generate secretmanager
+      - name: Run librarian generate all (integration test)
+        if: matrix.task == 'integration' && github.event_name == 'push' && github.ref == 'refs/heads/main'
         working-directory: google-cloud-java
         env:
           PYTHONPATH: ${{ github.workspace }}/google-cloud-java/sdk-platform-java/hermetic_build/library_generation/owlbot
         run: librarian generate --all
   create-issue-on-failure:
-    needs: [integration]
+    needs: [build-and-test]
```
zhumin8: Is this ever going to create issue? I think this reads as "Only run create-issue-on-failure if all matrix instances of build-and-test succeeded." When build-and-test fails (e.g., Integration Test fails on main) GitHub skips create-issue-on-failure altogether.

.github/workflows/java.yaml
```diff
@@ -16,16 +16,35 @@ on: [push, pull_request, merge_group]
...
+      matrix:
+        include:
+          - job-name: 'Unit Tests'
+            task: 'unit-test'
+            timeout: 5
+            sparse: true
+          - job-name: 'Smoke Test'
+            task: 'smoke-test'
+            timeout: 5
+            sparse: false
```
zhumin8: The original unit tests was passing 5mins limits occasionally, I'd expect this smoke test takes around the same time. I suggest we do sparse checkout for 'Smoke Tests" too, and in the "Checkout google-cloud-java (Sparse)" step, add `java-secretmanager/` to be checked out too.
Note, this does mean that if we add additional library for this smoke test in the future, we need to edit the sparse checkout rules too manually. But it improves checkout time from >1mins to <10s. I think it is acceptable. (Maybe make a note in comment)

noahdietz: I think that's a great idea, full clone is super long.

.github/workflows/java.yaml
sofisl: Since build-and-test is a matrix job, gh treats it as one test when reporting. If any fails it will become a failure.

So contains(needs.*.result, 'failure') will evaluate to true, and it will correctly create the issue if any single test in the matrix fails

zhumin8 (approved):

Issue #6773, opened 2026-07-13
php(generate): run synthtool after generated code


zhumin8:
Refactor generated code into `owl-bot-staging/` and run synthtool

zhumin8:
Most of work is done. A few smaller adjustment are needed. Will move to next week's milestone

zhumin8:
I cannot remember if there are still pending small changes. Will file separate issue if surfaces during testing.

PR #6869, opened 2026-07-16
feat(internal/librarian/php): run owlbot.py in generate phase

zhumin8:
Move generated code in the expected staging dir and invoke owlbot.py from generate as postprocessing step per library.
Introduced a `StagingSubdir` and make it explicit for now.  Follow up PR will populate `staging_subdir` in migrating librarian.yaml.

For #6773

### DIFF
--- a/internal/librarian/php/generate.go
+++ b/internal/librarian/php/generate.go
@@ -34,10 +34,12 @@ import (
 
 const (
 	commonResourcesProto = "google/cloud/common_resources.proto"
+	owlBotStagingDir     = "owl-bot-staging"
 )
 
 var (
 	errCommonResourcesUnconfigured = errors.New("common_resources must be set (either per-API or globally under default.php)")
+	errMissingStagingSubdir        = errors.New("staging_subdir is required for PHP configurations")
 )
 
 // Generate generates a PHP client library.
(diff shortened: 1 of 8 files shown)


internal/librarian/php/postprocess_test.go
```diff
@@ -0,0 +1,101 @@
...
+			},
+		},
+		{
+			name: "empty name",
+			library: &config.Library{
+				Output: "some/path",
+			},
+		},
+	}
+	for _, tc := range tests {
```
JoeWang1127:
```suggestion
	for _, test := range tests {
```

internal/librarian/php/generate.go
```diff
@@ -34,10 +34,12 @@ import (
 
 const (
 	commonResourcesProto = "google/cloud/common_resources.proto"
+	owlBotStagingDir     = "owl-bot-staging"
```
quirogas, line 37: Do we know if the owl-bot-staging directory name is required? If not, I think it would be better to rename it to something more appropriate, like librarian-staging.

internal/librarian/php/postprocess_test.go
```diff
@@ -0,0 +1,101 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			if err := postProcessLibrary(ctx, tc.library); err == nil {
```
JoeWang1127: Could you check the error? It's fine in follow ups.

JoeWang1127 (approved):

quirogas (approved):

internal/librarian/php/generate.go
zhumin8: Yes, it is the expected directory by owlbot/synthtool.

internal/librarian/php/postprocess_test.go
zhumin8: Started refactoring the test, but on second thought, removed checks for empty output and name. They are redundant at this stage, as these values are filled with defaults in upstream before reaching to language packages.

PR #6862, opened 2026-07-16
chore(ci): use sparse-checkout for Java presubmit test

zhumin8:
The Java presubmit test job was timing out often, checking out the full google-cloud-java monorepo took too long (over 1min in recent [runs](https://screenshot.googleplex.com/rvhLUBEzwDpPNdf)).

Configure actions/checkout to use sparse-checkout to only download librarian.yaml, pom.xml, and the sdk-platform-java directory. This provides the necessary files to build and install local tools while avoiding downloading the rest of the monorepo.

PR #6870, opened 2026-07-16
Revert "chore(ci): use sparse-checkout for Java presubmit test"

zhumin8:
Reverts googleapis/librarian#6862

It seems to be failing: https://github.com/googleapis/librarian/actions/runs/29531033304/job/87731114107?pr=6869.
The presubmit may be cached?


zhumin8:
I think I know the reason: the root pom.xml of google-cloud-java lists all 270+ child modules and the librarian install tool was executing mvn from the root directory of google-cloud-java.
Let me fix that in a followup

noahdietz (approved):

PR #6873, opened 2026-07-17
fix(nodejs): unify pnpm v7 and v8+ global bin environment configuration

quirogas:
set global binary directory environment variables for both PNPM v7 ( NPM_CONFIG_GLOBAL_BIN_DIR ,  npm_config_global_bin_dir ) and PNPM v8+ ( PNPM_HOME ,  PNPM_CONFIG_GLOBAL_BIN_DIR ). Adds a unit test in  install_test.go  to verify environment configuration. PNPM v7 and PNPM v8+ use different configuration variable names for global binary directories. Setting both ensures that installed tool symlinks are placed directly in Librarian's isolated cache  binDir  across all supported PNPM versions.

### DIFF
--- a/internal/librarian/nodejs/install_test.go
+++ b/internal/librarian/nodejs/install_test.go
@@ -283,55 +283,45 @@ func TestGetToolsEnv(t *testing.T) {
 }
 
 func TestGetPNPMEnv(t *testing.T) {
-	for _, test := range []struct {
-		name string
-	}{
-		{
-			name: "configures unified pnpm and npm global bin environment variables",
-		},
-	} {
-		t.Run(test.name, func(t *testing.T) {
-			cacheDir := t.TempDir()
-			binDir := t.TempDir()
-			t.Setenv("LIBRARIAN_CACHE", cacheDir)
-			t.Setenv("LIBRARIAN_BIN", binDir)
+	cacheDir := t.TempDir()
+	binDir := t.TempDir()
+	t.Setenv("LIBRARIAN_CACHE", cacheDir)
+	t.Setenv("LIBRARIAN_BIN", binDir)
 
-			envList, err := getPNPMEnv()
-			if err != nil {
-				t.Fatal(err)
-			}
+	envList, err := getPNPMEnv()
+	if err != nil {
+		t.Fatal(err)
+	}
 
-			wantBinDir := filepath.Join(binDir, "nodejs_tools", "bin")
-			wantGlobalDir := filepath.Join(cacheDir, "pnpm-global")
-			wantStoreDir := filepath.Join(cacheDir, "pnpm-store")
+	wantBinDir := filepath.Join(binDir, "nodejs_tools", "bin")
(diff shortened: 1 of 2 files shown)


JoeWang1127:
Could you elaborate why we need to set envs for two versions of pnpm? Don't we only use pnpm 7?

quirogas:
> Could you elaborate why we need to set envs for two versions of pnpm?

V7 and 8+ have different standards; since node uses V7 while languages like PHP do not, we must support both.

JoeWang1127:
> V7 and 8+ have different standards; since node uses V7 while languages like PHP do not, we must support both.

Why does php need another version of pnpm? IIUC, pnpm is a package manager and the packaged downloaded by pnpm should be the same regardless of the version of pnpm.

quirogas:
> > V7 and 8+ have different standards; since node uses V7 while languages like PHP do not, we must support both.
> 
> Why does php need another version of pnpm? IIUC, pnpm is a package manager and the packaged downloaded by pnpm should be the same regardless of the version of pnpm.

pnpm version 7 is older, and we are eventually transitioning to a newer version. PHP is likely already using a newer version and should continue to do so. This change adds support for version 7, which was previously causing silent bugs, while maintaining existing support for version 8+. Removing support for version 8 is undesirable as we eventually want to move to a newer version too, but it's not currently in scope for this issue.

quirogas:
We can  remove support for pnpm 7 once Node no longer supports it, but we need to support pnpm 7 for now.

internal/librarian/nodejs/install_test.go
```diff
@@ -280,3 +281,46 @@ func TestGetToolsEnv(t *testing.T) {
 		})
 	}
 }
+
+func TestGetPNPMEnv(t *testing.T) {
+	cacheDir := t.TempDir()
+	binDir := t.TempDir()
+	t.Setenv("LIBRARIAN_CACHE", cacheDir)
+	t.Setenv("LIBRARIAN_BIN", binDir)
+
```
JoeWang1127, line 290: nit: remove this empty line.

internal/librarian/nodejs/install_test.go
```diff
@@ -280,3 +281,46 @@ func TestGetToolsEnv(t *testing.T) {
...
+	envList, err := getPNPMEnv()
+	if err != nil {
+		t.Fatal(err)
+	}
+
```
JoeWang1127, line 295: nit: remove this empty line.

internal/librarian/nodejs/install_test.go
```diff
@@ -280,3 +281,46 @@ func TestGetToolsEnv(t *testing.T) {
...
+	wantBinDir := filepath.Join(binDir, "nodejs_tools", "bin")
+	wantGlobalDir := filepath.Join(cacheDir, "pnpm-global")
+	wantStoreDir := filepath.Join(cacheDir, "pnpm-store")
+
```
JoeWang1127, line 299: nit: remove this empty line.

JoeWang1127 (approved):

PR #6874, opened 2026-07-17
fix(nodejs): require cached tools and prevent host binary fallback

quirogas:
Adds cache verification for Node.js generator tools ( gapic-generator-typescript ,  compileProtos , and  gapic-node-processing ) before execution during client library generation. Currently, when a required tool was missing from Librarian's isolated cache ( ~/.cache/librarian/bin/nodejs_tools/bin ), execution silently fell back to searching the host machine's system  PATH . This caused non-deterministic behavior and silent drift across developer machines and CI runners executing unmanaged ambient binaries. With this change, Librarian verifies that the required tools exist in its cache and returns an explicit, actionable error instructing the user to run  librarian install nodejs  if they are missing.

### DIFF
--- a/internal/librarian/nodejs/generate.go
+++ b/internal/librarian/nodejs/generate.go
@@ -82,7 +82,36 @@ func Generate(ctx context.Context, cfg *config.Config, library *config.Library, 
 	return nil
 }
 
+var (
+	errToolNotInstalled = errors.New("tool not installed in librarian cache")
+)
+
+func requireCachedTool(toolName string) error {
+	binDir, err := getBinDir()
+	if err != nil {
+		return fmt.Errorf("failed to get bin directory: %w", err)
+	}
+	toolPath := filepath.Join(binDir, toolName)
+	info, err := os.Stat(toolPath)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return fmt.Errorf("tool %s %w", toolName, errToolNotInstalled)
+		}
+		return err
+	}
+	if info.IsDir() {
+		return fmt.Errorf("tool path %s is a directory: %w", toolPath, errToolNotInstalled)
+	}
+	return nil
+}
+
 func generateAPI(ctx context.Context, api *config.API, library *config.Library, googleapisDir, repoRoot string) error {
+	for _, requiredTool := range []string{"gapic-generator-typescript", "compileProtos", "gapic-node-processing"} {
+		if err := requireCachedTool(requiredTool); err != nil {
+			return err
+		}
+	}
+
 	version := filepath.Base(api.Path)
 	stagingDir := filepath.Join(repoRoot, "owl-bot-staging", library.Name, version)
 	if err := os.MkdirAll(stagingDir, 0755); err != nil {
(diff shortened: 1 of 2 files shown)


internal/librarian/nodejs/generate_test.go
```diff
@@ -1700,3 +1700,59 @@ func TestMoveKeep_Errors(t *testing.T) {
 		})
 	}
 }
+
+func TestRequireCachedTool(t *testing.T) {
```
JoeWang1127, line 1704: Could you separate the error cases: https://github.com/googleapis/librarian/blob/main/doc/howwewritego.md#separate-error-tests

JoeWang1127 (commented):
It is nice to check the install directory to make sure the tools are there before execution.

However, don't we need to specify the absolute path when calling the command?

For example, does this code still using `gapic-node-processing` from  $PATH?
```go
if err := command.RunWithEnv(ctx, toolsEnv, "gapic-node-processing", combineArgs...); err != nil {
   return fmt.Errorf("combine-library: %w", err)
}
```

quirogas:
> It is nice to check the install directory to make sure the tools are there before execution.
> 
> However, don't we need to specify the absolute path when calling the command?
> 
> For example, does this code still using `gapic-node-processing` from $PATH?
> 
> ```go
> if err := command.RunWithEnv(ctx, toolsEnv, "gapic-node-processing", combineArgs...); err != nil {
>    return fmt.Errorf("combine-library: %w", err)
> }
> ```

Yes! Good catch. I've updated the tool execution to use an absolute path to make sure it's only. using the tool at the expected location y and eliminate the possibility  of falling back on a tool installation in another directory.

internal/librarian/nodejs/generate.go
```diff
@@ -82,7 +82,41 @@ func Generate(ctx context.Context, cfg *config.Config, library *config.Library,
...
+
+func requireCachedTool(toolName string) (string, error) {
+	binDir, err := getBinDir()
+	if err != nil {
+		return "", fmt.Errorf("failed to get bin directory: %w", err)
+	}
+	toolPath := filepath.Join(binDir, toolName)
+	info, err := os.Stat(toolPath)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
```
JoeWang1127: How about returning the error without this check?

internal/librarian/nodejs/generate_test.go
```diff
@@ -1700,3 +1700,66 @@ func TestMoveKeep_Errors(t *testing.T) {
...
+		t.Errorf("requireCachedTool(%q) = %q, want %q", "compileProtos", gotPath, toolPath)
+	}
+}
+
+func TestRequireCachedTool_Error(t *testing.T) {
+	for _, test := range []struct {
+		name      string
+		setup     func(t *testing.T, binDir string)
+		toolName  string
+		wantErrIs error
```
JoeWang1127:
```suggestion
		wantErr error
```

internal/librarian/nodejs/generate.go
quirogas: This error is triggered when the tool is not find in the cache, which could indicate that they haven't run `librarian install nodejs` or the cache was deleted/changed.

JoeWang1127: Is it the same case for other kind of errors? We can just return the message to let them run `librarian install`.

quirogas: Yes, I've updated the code to reflect this change.

JoeWang1127 (approved):

PR #6877, opened 2026-07-17
feat(internal/librarian/nodejs): support src_dir for pnpm tool sourcebuilds

quirogas:
When installing Node.js generator tools from source archives or monorepos, `installPNPMToolFromSource` hardcodes the build working directory to `core/generator/gapic-generator-typescript`. This prevents other monorepo tools (such as gapic-tools in core/packages/tools and gapic-node-processing in core/packages/gapic-node-processing) from being built from source with frozen lockfiles.

This change modifies `installPNPMToolFromSource` to respect tool.SrcDir from configuration before falling back to the default generator subdirectory. When src_dir is specified in librarian.yaml, Librarian runs the tool's build steps inside that specific subdirectory of the fetched archive.

### DIFF
--- a/internal/librarian/nodejs/install.go
+++ b/internal/librarian/nodejs/install.go
@@ -177,7 +177,11 @@ func installPNPMToolFromSource(ctx context.Context, env []string, tool *config.P
 	}
 
 	// Run build steps.
-	genDir := filepath.Join(dir, gapicGeneratorSubdir)
+	subdir := tool.SrcDir
+	if subdir == "" {
+		subdir = gapicGeneratorSubdir
+	}
+	genDir := filepath.Join(dir, subdir)
 	for _, cmd := range tool.Build {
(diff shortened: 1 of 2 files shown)


JoeWang1127:
Hi @quirogas, could you make this change in `tool/pnpm` package?

I copied the code in nodejs package there for a starter.

We should create a dedicated package for pnpm so it can be used by other languages, e.g., php.

See https://github.com/googleapis/librarian/issues/6857.

quirogas:
> Hi @quirogas, could you make this change in `tool/pnpm` package?
> 
> I copied the code in nodejs package there for a starter.
> 
> We should create a dedicated package for pnpm so it can be used by other languages, e.g., php.
> 
> See #6857.

I'll do this in a follow up PR to keep concerns separate.

JoeWang1127 (approved):

PR #6881, opened 2026-07-17
fix(tool/cmd/migrate): resolve staging subdir for library-root destinations

zhumin8:
The migration tool now extracts the staging subdirectory configuration for PHP APIs from the repository's OwlBot configuration. This ensures that the generated librarian.yaml correctly preserves the staging directories previously used by OwlBot.
Note: current generate implementation expect staging subdir be explicit set. We may change that in the future, but for now we populate this for all libraries.

For #6773

### DIFF
--- a/tool/cmd/migrate/php_test.go
+++ b/tool/cmd/migrate/php_test.go
@@ -16,6 +16,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"os"
 	"path/filepath"
 	"testing"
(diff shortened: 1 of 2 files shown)


tool/cmd/migrate/php_test.go
```diff
@@ -237,6 +279,25 @@ func TestExtractAPIsFromOwlBot_Error(t *testing.T) {
 				}
 				return path
 			},
+			// wantErr is nil as we only assert that a YAML parsing error is returned.
+			wantErr: nil,
```
JoeWang1127, line 283: I think this is fine since this code will be removed after the migration.

JoeWang1127 (approved):

PR #6894, opened 2026-07-17
feat(swift): convert message structs

nolanleastin:
Implement support in sidekick for generating native Swift message structs and their conversion extensions (init(proto:) / toProto()) mapping basic scalar fields (numbers, booleans, strings, bytes).


internal/sidekick/swift/annotate_field.go
```diff
@@ -238,3 +245,31 @@ func (c *codec) fieldPackage(field *api.Field) (string, error) {
...
+	// Nested Message and Enum type fields are handled in a subsequent PR.
+	case api.TypezMessage, api.TypezEnum:
+		return
+	default:
+		if field.Optional {
+			ann.ToGapicStatement = fmt.Sprintf("self.%s = proto.has%s ? proto.%s : nil", fieldName, protoFieldNamePascal, pFieldName)
+			ann.ToProtoStatement = fmt.Sprintf("if let %s = self.%s { proto.%s = %s }", fieldName, fieldName, pFieldName, fieldName)
+		} else {
+			ann.ToGapicStatement = fmt.Sprintf("self.%s = proto.%s", fieldName, pFieldName)
+			ann.ToProtoStatement = fmt.Sprintf("proto.%s = self.%s", pFieldName, fieldName)
```
coryan: For all other things we implement the optional switch in the mustache template. Any particular reason you want to move away from that?

internal/sidekick/swift/annotate_field.go
```diff
@@ -238,3 +245,31 @@ func (c *codec) fieldPackage(field *api.Field) (string, error) {
...
+	protoFieldNamePascal := protoFieldNamePascal(field.Name)
+
+	// Oneof fields are handled in a subsequent PR.
+	if field.IsOneOf {
+		return
+	}
+	// Map and Repeated fields are handled in subsequent PRs.
+	if field.Map || field.Repeated {
+		return
+	}
```
coryan: Maybe a `TODO(#...)` for these comments.

internal/sidekick/swift/annotate_field.go
coryan: FWIW,  you can classify a field using `IsOneOf`, and for the fields that are not oneofs, only one of these is set at a time:  `Map`, `Repeated`, or:

https://github.com/googleapis/librarian/blob/fbc93ee987c1e7b43779d460b9c688423f63f933/internal/sidekick/api/model.go#L1024-L1027

internal/sidekick/swift/annotate_message.go
```diff
@@ -69,6 +70,36 @@ type messageAnnotations struct {
...
+	ProtoTypeName     string
+	ModulePath        string
+}
+
+// ConvertImports returns the sorted list of import statements for message conversions.
+func (ann *messageAnnotations) ConvertImports() []string {
+	importMap := map[string]bool{
+		"import Foundation":             true,
+		"import GoogleCloudGax":         true,
+		"internal import SwiftProtobuf": true,
```
coryan: This is very strange, if we are going to hard-code text like this, we should just add it to the mustache template.

internal/sidekick/swift/annotate_message.go
```diff
@@ -32,6 +32,7 @@ type messageAnnotations struct {
 	Model               *modelAnnotations
 	TypeURL             string
 	CustomSerialization bool
+	CopyrightYear       string
```
coryan: You should be able to get the copyright year from `Codec.Model.CopyrightYear`, no need to create a new field.

internal/sidekick/swift/annotate_enum.go
```diff
@@ -136,9 +137,23 @@ func (c *codec) annotateEnum(enum *api.Enum, model *modelAnnotations) error {
...
 	enum.Codec = annotations
 	return nil
 }
+
+// ConvertImports returns the sorted list of import statements for enum conversions.
+func (ann *enumAnnotations) ConvertImports() []string {
+	imports := []string{
+		"import Foundation",
+		"import GoogleCloudGax",
+		"internal import SwiftProtobuf",
```
coryan: Ditto, it is strange to hard-code things like this in the code, just move it to the mustache template.

internal/sidekick/swift/generate_conversions_test.go
```diff
@@ -38,3 +41,71 @@ func TestGenerateConversions_MissingModulePath(t *testing.T) {
...
+	}
+	contentStr := string(content)
+
+	// Check output imports
+	if !strings.Contains(contentStr, "internal import StorageControlProtos") {
+		t.Errorf("expected generated file to import StorageControlProtos")
+	}
+
+	// Check conversion logic
+	wantInit := "  internal init(proto: ProtoType) throws {\n    self.name = proto.name\n    self.metageneration = proto.metageneration\n    self.self_ = proto.hasSelf_p ? proto.self_p : nil\n  }"
```
coryan, line 104: For things like this, continue using `extractBlock()` it produces better error messages on partial matches, in this case something like:

```swift
  got := extractBlock(t, contentStr, "internal init(proto: ProtoType)", "\n  }")
```

will let you match with `wantInit` and print the lines that do not match.

internal/sidekick/swift/codec.go
```diff
@@ -215,3 +217,29 @@ func (c *codec) addDependency(dep *Dependency) (*Dependency, error) {
 	}
 	return dep, nil
 }
+
+func (c *codec) protoMessageTypeName(m *api.Message) string {
```
coryan: Please move these new functions to their own files.  Files grow too easily and then navigation is a headache.

coryan (changes requested):
Prefer using the mustache templates to generate code. It is annoying, but less annoying than finding the right printf in the code.

internal/sidekick/swift/annotate_enum.go
nolanleastin: agree.

internal/sidekick/swift/annotate_enum_test.go
```diff
@@ -203,3 +203,46 @@ func TestAnnotateEnum_ModulePath(t *testing.T) {
...
+	}
+
+	if ann.ModulePath != "TestProtos" {
+		t.Errorf("ann.ModulePath = %q, want %q", ann.ModulePath, "TestProtos")
+	}
+
+	wantProtoTypeName := "TestProtos.Test_OuterMessage.InnerEnum"
+	if ann.ProtoTypeName != wantProtoTypeName {
+		t.Errorf("ann.ProtoTypeName = %q, want %q", ann.ProtoTypeName, wantProtoTypeName)
+	}
```
coryan: nit: in general, this repo prefers using `cmp.Diff()`, but I admit that might be overkill here.

internal/sidekick/swift/annotate_field_test.go
```diff
@@ -81,7 +81,7 @@ func TestAnnotateField(t *testing.T) {
 				t.Fatal(err)
 			}
 
-			if diff := cmp.Diff(test.want, field.Codec, cmpopts.IgnoreFields(fieldAnnotations{}, "Name", "DocLines", "Model")); diff != "" {
+			if diff := cmp.Diff(test.want, field.Codec, cmpopts.IgnoreFields(fieldAnnotations{}, "Name", "DocLines", "Model", "ProtoFieldName", "ProtoFieldNamePascal")); diff != "" {
```
coryan: Consider adding the expected `ProtoFieldName` and `ProtoFieldNamePascal` instead of ignoring them.

internal/sidekick/swift/annotate_message_test.go
```diff
@@ -160,7 +160,7 @@ func TestAnnotateMessage(t *testing.T) {
 			if err := codec.annotateModel(); err != nil {
 				t.Fatal(err)
 			}
-			if diff := cmp.Diff(test.want, test.message.Codec, cmpopts.IgnoreFields(messageAnnotations{}, "Model", "DependsOn")); diff != "" {
+			if diff := cmp.Diff(test.want, test.message.Codec, cmpopts.IgnoreFields(messageAnnotations{}, "Model", "DependsOn", "ProtoTypeName", "ModulePath")); diff != "" {
```
coryan: Ditto, maybe add the expected names?

coryan (approved):
Consider using the existing tests to verify the field and message names work as you expect.

PR #6901, opened 2026-07-17
fix(librarian): enable `google/cloud/batch/v1` for all

coryan:
It seems odd that this API is only enabled for a subset of the languages. I thought the intent was to enable all google/cloud APIs for everybody.


coryan:
@noahdietz does this look right to you?

noahdietz (approved):
Yeah looks right.

noahdietz:
Another one of those APIs that doesn't have a `publishing` stanza so its a bit of a mystery what was necessary.

Issue #6835, opened 2026-07-15
rust: support auto-generating bidirectional gRPC streaming for Rust GAPIC clients


suzmue:
## Context & Goal
Currently, the Rust generator in `librarian` skips gRPC streaming methods when generating GAPIC client libraries. To support high-throughput bidirectional streaming APIs (such as Speech-to-Text v2 `StreamingRecognize` and Dialogflow v2 `StreamingDetectIntent`), `librarian` needs to support auto-generating bidirectional (`BidiStreaming`) gRPC client methods, stubs, and transport dispatchers.

This issue tracks the generator implementation required in `librarian` to support bidirectional gRPC streaming for Rust GAPIC clients.

## Proposed Design & Approach

  - Keep all experimental streaming code generation **off by default** using a dedicated configuration option (`include_bidi_streaming_methods: true` in `librarian.yaml`).
  - Automatically generate internal, private (`pub(crate)`) raw Prost structs (`src/prost.rs` and `src/prost/`) and local `ToProto` / `FromProto` conversion trait implementations (`src/convert.rs`) inside target crates to prevent raw `prost` types from leaking into public client library APIs. This builds on the existing prost and convert_prost templates, but needs to be dynamically generated when streaming types are detected.
  - Generate bidirectional streaming method signatures on service stubs, builders, and client transports using GAX stream wrapper types.

## Related Issues
- `google-cloud-rust` umbrella issue: https://github.com/googleapis/google-cloud-rust/issues/2318

suzmue:
This is complete

PR #6914, opened 2026-07-20
feat(rust): add hybrid transport initialization for bidi streaming

suzmue:
Add dual gRPC and REST transport client initialization to the Rust generator, gated under the google_cloud_unstable_gapic_streaming compiler cfg flag.

The HasBidiStreaming service annotation is evaluated against the API definition prior to method filtering, ensuring streaming transport stubs are generated for bidi-streaming services while individual method implementations remain ungenerated.

For #6835


internal/sidekick/rust/templates/crate/src/transport.rs.mustache
```diff
@@ -35,16 +35,29 @@ use crate::Error;
...
+    grpc_inner: gaxi::grpc::Client,
+    {{/Codec.HasBidiStreaming}}
 }
 
 {{#Codec.PerServiceFeatures}}
 #[cfg(feature = "{{Codec.FeatureName}}")]
 {{/Codec.PerServiceFeatures}}
 impl std::fmt::Debug for {{Codec.Name}} {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::result::Result<(), std::fmt::Error> {
+        {{^Codec.HasBidiStreaming}}
```
coryan, line 49: I was going to suggest some cleanup here, but then I realized you may want the code to stay unchanged for existing clients.

Maybe add a `{{! Preserve code structure for existing clients }}` or something along those lines?

internal/sidekick/rust/templates/crate/src/transport.rs.mustache
```diff
@@ -53,6 +66,7 @@ impl std::fmt::Debug for {{Codec.Name}} {
 {{/Codec.PerServiceFeatures}}
 impl {{Codec.Name}} {
     pub async fn new(config: gaxi::options::ClientConfig) -> crate::ClientBuilderResult<Self> {
+        {{^Codec.HasBidiStreaming}}
```
coryan, line 69: As Gemini says, I think you can simplify the mustache template here and still keep the generated code untouched for most services.

coryan (approved):
Some optional stuff to simplify  the mustache templates, feel free to merge as-is make the changes and then merge.

internal/sidekick/rust/codec_test.go
```diff
@@ -2369,10 +2370,30 @@ func TestGenerateMethod_Streaming(t *testing.T) {
 			},
 			want: true,
 		},
+		{
+			name:                        "includes bidirectional streaming when includeBidiStreamingMethods is enabled",
```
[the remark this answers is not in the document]
suzmue: fixed.

internal/sidekick/rust/templates/crate/src/transport.rs.mustache
suzmue: Yeah this was exactly my goal, I wanted to have the least churn possible (no changes at all to clients that don't enable this feature), until things stabilize more.

Ultimately, I think a small amount of changes to simplify the templates would be ok, but I don't want to do that multiple times so will save those changes until the end.

internal/sidekick/rust/templates/crate/src/transport.rs.mustache
suzmue: Will handle this in a follow up.

internal/sidekick/rust/templates/crate/src/transport.rs.mustache
```diff
@@ -55,18 +68,51 @@ impl {{Codec.Name}} {
...
+        let grpc_inner = gaxi::grpc::Client::new(config, crate::DEFAULT_HOST).await?;
+        {{/Codec.DetailedTracingAttributes}}
+        {{/Codec.HasBidiStreaming}}
+        Ok(Self {
+            inner,
+            {{#Codec.HasBidiStreaming}}
+            #[cfg(google_cloud_unstable_gapic_streaming)]
+            grpc_inner,
+            {{/Codec.HasBidiStreaming}}
+        })
```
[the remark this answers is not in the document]
suzmue: addressed partially. want to keep things unchanged for existing code.

PR #6929, opened 2026-07-20
fix(internal/librarian/php): split protoc compilation for gapic and proto

### LINKED ISSUE #6924: php(generate): split protoc compilation to prevent dependency over-generation
Currently, the PHP generator runs protoc in a single invocation with both `--gapic_out` and `--php_out` applied to the union of main protos and `additional_protos`.
  This causes protoc to generate PHP message classes for dependency APIs (such as `Location`), which:
    1. Deviates from the legacy Bazel behavior where dependency protos were only used for GAPIC generation and not compiled to PHP classes.
    2.  The duplicate Location classes did not show up in the git diff of the generated library because they were never actually copied to the final repository—they only existed in the temporary staging directory.
    3. Even though Location was never copied, its mere presence in the staging directory triggered a synthtool recursive search issue when there are multiple root-level namespaces (e.g. `Location` alongside `SecretManager`) in the staging directory.

The implementation in `internal/librarian/php/generate.go` should be updated to:
    1. Run protoc with `--gapic_out` on both main protos and `additional_protos` to generate the client and mixins.
    2. Run protoc separately with `--php_out` on ONLY the main protos to generate package-specific message classes.

### LINKED ISSUE #6927: php(generate): generator does not compile protobuf messages
Currently, the PHP generator in `internal/librarian/php/generate.go` only invokes `protoc` with `--gapic_out` (running `gapic-generator-php`). It does not invoke `--php_out`, meaning it does not generate the protobuf message classes (like requests, responses, and metadata) from the proto definitions.

During the implementation of the `Clean` step (which deletes autogenerated files), these message classes were removed and not restored, exposing the gap in the generator.

Fix this by running `protoc` with `--php_out` for the main protos (using a split-run compilation to avoid over-generating dependency protos like `Location`).

zhumin8:
This change splits the protoc execution into two separate runs: one for the GAPIC client generation (running on main, additional, and common resources protos) and one for the Proto message generation (running on main protos only).

By running them separately and only passing main protos to --php_out, we prevent the over-generation of dependency protobuf classes (like Location) inside the library package. 

Fixes #6924 
Fixes #6927

### DIFF
--- a/internal/librarian/php/generate.go
+++ b/internal/librarian/php/generate.go
@@ -120,8 +120,12 @@ func Generate(ctx context.Context, cfg *config.Config, library *config.Library, 
 			return err
 		}
 		// Cleanup output zips for subsequent APIs in the same library package
-		_ = os.Remove(gapicZipPath)
-		_ = os.Remove(protoZipPath)
+		if err := os.Remove(gapicZipPath); err != nil && !errors.Is(err, os.ErrNotExist) {
+			return fmt.Errorf("failed to remove gapic zip: %w", err)
+		}
+		if err := os.Remove(protoZipPath); err != nil && !errors.Is(err, os.ErrNotExist) {
+			return fmt.Errorf("failed to remove proto zip: %w", err)
+		}
 	}
 	if err := postProcessLibrary(ctx, library); err != nil {
 		return fmt.Errorf("failed to postprocess: %w", err)
--- a/internal/librarian/php/generate.go
+++ b/internal/librarian/php/generate.go
@@ -95,31 +95,39 @@ func Generate(ctx context.Context, cfg *config.Config, library *config.Library, 
 	if err := os.MkdirAll(stagingDir, 0755); err != nil {
 		return err
 	}
-	outputZipPath := filepath.Join(tempDir, "output.zip")
+	gapicZipPath := filepath.Join(tempDir, "output.zip")
+	protoZipPath := filepath.Join(tempDir, "proto.zip")
 	srcCfg := sources.NewSourceConfig(src, library.Roots)
 	for _, api := range library.APIs {
 		if api.PHP == nil || api.PHP.StagingSubdir == "" {
 			return fmt.Errorf("API %q: %w", api.Path, errMissingStagingSubdir)
 		}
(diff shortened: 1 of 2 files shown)


internal/librarian/php/generate.go
```diff
@@ -95,31 +95,37 @@ func Generate(ctx context.Context, cfg *config.Config, library *config.Library,
...
+			gapicDestDir: gapicDestDir,
+			protoDestDir: protoDestDir,
 		}
 		if err := generateAPI(ctx, params); err != nil {
 			return err
 		}
-		// Cleanup output zip for subsequent APIs in the same library package
-		_ = os.Remove(outputZipPath)
+		// Cleanup output zips for subsequent APIs in the same library package
+		if err := os.Remove(gapicZipPath); err != nil && !errors.Is(err, os.ErrNotExist) {
```
JoeWang1127: nit: use a random name for the zip file or include api path (since they are unique) to eliminate this step.

JoeWang1127: Or, can we use a defer function?

internal/librarian/php/generate_test.go
```diff
@@ -344,28 +350,28 @@ func TestGatherTargetProtos_Error(t *testing.T) {
 					t.Fatal(err)
 				}
 			}
-			_, err := gatherTargetProtos(tempDir, test.apiPath, nil, true)
+			_, err := gatherGAPICProtos(tempDir, test.apiPath, nil, true)
 			if err == nil {
-				t.Fatal("gatherTargetProtos() expected error, got nil")
+				t.Fatal("gatherGAPICProtos() expected error, got nil")
```
JoeWang1127, line 355: Can you check the error?

JoeWang1127 (approved):

internal/librarian/php/generate.go
zhumin8: done. Moved construction into generateAPI so we can use defer.

internal/librarian/php/generate_test.go
zhumin8: will do a followup

Issue #6629, opened 2026-07-06
php: add generate command


JoeWang1127:
Add generate support for Php.

zhumin8:
Big trunk of initial works and setup is done. Moving to next milestone to polish and finish up owlbot postprocessing.

PR #6933, opened 2026-07-20
fix(internal/librarian/php): check error and add test case for TestGatherGAPICProtos_Error

zhumin8:
Introduces a new sentinel error `errNoProtos` to handle missing target proto files or directories, and updates the corresponding tests to verify this error. 

Followup to https://github.com/googleapis/librarian/pull/6929#discussion_r3616006700

For #6629

### DIFF
--- a/internal/librarian/php/generate_test.go
+++ b/internal/librarian/php/generate_test.go
@@ -332,11 +332,19 @@ func TestGatherGAPICProtos_Error(t *testing.T) {
 		name       string
 		setupFiles []string
 		apiPath    string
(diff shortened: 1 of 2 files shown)


internal/librarian/php/generate_test.go
```diff
@@ -354,6 +362,9 @@ func TestGatherGAPICProtos_Error(t *testing.T) {
 			if err == nil {
```
JoeWang1127: Could you remove this check?

JoeWang1127 (approved):

Issue #6659, opened 2026-07-07
php: support clean functionality


JoeWang1127:
Add clean support for Php, this function should remove all generated files in a Php library.

zhumin8:
An initial survey from google-cloud-php code:


  ### 1. File Location & Namespace Structure

  • Generated Files: Always live inside versioned namespace directories corresponding to the proto definitions.
      • Examples:  src/V1/ ,  src/V2/ ,  src/V1beta2/ ,  src/Admin/Database/V1/ .
  • Handwritten Files: Typically live directly in the root of the library's  src/  directory.
      • Examples (under  Spanner ):  src/SpannerClient.php ,  src/Database.php ,  src/Backup.php .
      • Note: Some simpler libraries (like  SecretManager ) are purely generated and have no handwritten files under  src/  at all.

  ──────
  ### 2. Header Comment Markers

  You can reliably check the file headers:

  • Generated Protobuf Message Files (e.g., request/response payloads) have a shell-style comment on the second line:
```
    <?php
    # Generated by the protocol buffer compiler.  DO NOT EDIT!
    # source: google/cloud/secretmanager/v1/service.proto
```
  • Generated Service Clients & Resource Configurations (under  Client/  or  resources/  directories) have an explicit warning comment:
```
    /*
     * GENERATED CODE WARNING
     * Generated by gapic-generator-php from the file ...
     */
    or:
    /*
     * GENERATED CODE WARNING
     * This file was automatically generated - do not edit!
     */
```

  • Handwritten Files will only have the standard Google copyright license header without any generation warning block.

zhumin8:
One more PR needed for migrate to populate keep items.

zhumin8:
So the typical source to migrate keep items is from owlbot.yaml `deep-preserve-regex`. There is no `deep-preserve-regex` present in any of the owlbot.yaml files in google-cloud-php.

Then additionally, checked `copy_excludes` settings in owlbot.py files.

 `copy_excludes` is a configuration in owlbot.py (passed to synthtool) that specifies file patterns that should not be copied from the staging directory to the final destination directory in the repository. Hypothetically,  a file risk be permanently lost during the build process if it meets all of the following conditions:

  1. It is excluded from copy via copy_excludes in owlbot.py (meaning the generator's version in staging will not be copied to the repo).
  2. It exists in the repository (presumably with manual changes).
  3. It still contains the generation warning markers (which tells the clean step it is safe to delete).
[...]

PR #6939, opened 2026-07-20
feat(internal/librarian/php): implement selective clean before generate

zhumin8:
Selective cleaning is implemented for PHP client libraries to run before code generation. This step deletes only autogenerated files (identified by code generator warning markers) and gapic_metadata.json within the src, tests, metadata, and samples directories.
    
This approach prevents the accumulation of stale generated files while preserving manually maintained configuration files (like composer.json, README.md, and owlbot.py) and custom wrapper classes that live in the same target directories. 

For https://github.com/googleapis/librarian/issues/6659


zhumin8:
Will do follow-up to add more test coverage for error cases.

internal/librarian/php/clean.go
```diff
@@ -15,12 +15,88 @@
...
+
+// cleanFiles walks the given subdirectory under the library output and removes
+// generated files. It preserves files listed in the keep set. A file is considered
+// generated if it is named "gapic_metadata.json" or if it is a ".php" file containing
+// either the GAPIC or Protobuf code generation warning markers.
+func cleanFiles(lib *config.Library, keepSet map[string]bool, subDir string) error {
+	dir := filepath.Join(lib.Output, subDir)
+	_, err := os.Stat(dir)
+	if err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
```
JoeWang1127, line 68: Maybe this is find for now, but do we want to return an error if a generated directory doesn't exist?

internal/librarian/php/clean_test.go
```diff
@@ -15,15 +15,127 @@
...
+		setupFiles  []string
+		contentMap  map[string]string
+		wantDeleted []string
+	}{
+		{
+			name: "gapic metadata is deleted",
+			setupFiles: []string{
+				"src/V1/gapic_metadata.json",
+				"src/gapic_metadata.json",
+				"gapic_metadata.json", // in root, won't be walked, so kept
```
JoeWang1127:
```suggestion
                // does not delete because it is in the root directory
				"gapic_metadata.json", // in root, won't be walked, so kept
```

internal/librarian/php/clean_test.go
```diff
@@ -15,15 +15,127 @@
...
+			setupFiles: []string{
+				"other/V1/ServiceClient.php",
+			},
+			contentMap: map[string]string{
+				"other/V1/ServiceClient.php": "<?php\n// " + string(gapicMarker) + "\nclass ServiceClient {}",
+			},
+			wantDeleted: nil, // 'other' is not in directoriesToClean
+		},
+	} {
+		t.Run(test.name, func(t *testing.T) {
```
JoeWang1127, line 99: t.Parallel()

internal/librarian/php/clean_test.go
```diff
@@ -15,15 +15,127 @@
...
+
+func verifyFileDeletions(t *testing.T, dir string, setupFiles, wantDeleted []string) {
+	t.Helper()
+	for _, file := range setupFiles {
+		fullPath := filepath.Join(dir, file)
+		_, err := os.Stat(fullPath)
+		if err != nil && !errors.Is(err, fs.ErrNotExist) {
+			t.Fatal(err)
+		}
+		got := err != nil
```
JoeWang1127, line 126: Does `got` be `true` when reaching this line?

internal/librarian/php/clean.go
zhumin8: I think keeping it as a no-op when the directory doesn't exist is safer and ensures Clean() remains idempotent. It should not be `Clean`'s responsibility to do structural validation.

internal/librarian/php/clean_test.go
zhumin8: `got` can be both `true` or `false` here, and is relative to wantDeleted (L122-124 only filters out unexpected errors):
- `got` is `true` means:  The file WAS deleted 
- `got` is `false` means:  The file was NOT deleted ("obey keep list" test case)

zhumin8:
Added test coverage for error cases. 
This `Clean` does not remove empty dirs afterwards. Will do a followup PR to refactor and reuse a helper method to do that similar to Java.

internal/librarian/php/clean.go
JoeWang1127: OK, could you add a comment on why this error is fine.

internal/librarian/php/clean_test.go
JoeWang1127: Ah, got it.

Could you add a comment on this test?

JoeWang1127 (approved):

Issue #6824, opened 2026-07-14
php: refactor install


sofisl:
I think we need more refactor on `Install` function but this is fine for the first iteration.

Some potential refactor:
1. extract `composer install` to a separate package to `internal/tool/composer` or `internal/tool/php`.
2. verify a tool before installation, can also happen in composer/php package.

_Originally posted by @JoeWang1127 in https://github.com/googleapis/librarian/pull/6799#pullrequestreview-4696687677_

sofisl:
also can delete min's workaround for installing via generator since she  is using migrate

sofisl:
also, remove wantErr in TestInstall case:

https://github.com/googleapis/librarian/pull/6941/changes#r3625559633

PR #6941, opened 2026-07-21
fix(internal/librarian/php): install dev dependencies and create wrapper script for php generator

### LINKED ISSUE #6903: php(install): install for generator not working properly
My local test run came across these issues:
1. Missing Runtime Dependencies: The installer was running composer install for the generator with the `--no-dev `flag. This stripped out all development dependencies. However, the generator's composer.json lists critical dependencies like google/gax. Without these, the generator would crash at runtime with PHP class-not-found errors.
https://github.com/googleapis/librarian/blob/357320f0789cb4bb7e17fe10dab3c5a7f1ba09fb/internal/librarian/php/install.go#L67-L69

2. Wrapper Naming Mismatch: The installer created a wrapper script under bin/google/gapic-generator-php (mismatching the naming convention), whereas the generation logic in generate.go was looking for the wrapper at bin/gapic-generator-php.

I got some temp fix in my local to continue with my tests. But since @sofisl is still working on install command and refactoring, assigning this one to you.

sofisl:
The gapic-generator-php tool has unique installation requirements and requires its dev dependencies (specifically google/gax) to run properly (as opposed to the other tools).  Previously, generic Composer tool installations stripped these out via  the \`--no-dev\` flag, causing ClassNotFound fatal errors during generation.

This change:
* Always installs dev deps
* Dynamically generates the required wrapper script directly in the librarian bin

For https://github.com/googleapis/librarian/issues/6824
Fixes #6903


internal/librarian/php/install.go
```diff
@@ -59,18 +60,40 @@ func Install(ctx context.Context, tools *config.Tools) error {
...
 		dir, err := fetch.Repo(ctx, tool.Repo, tool.Version, tool.SHA256)
 		if err != nil {
 			return fmt.Errorf("fetching %s: %w", tool.Name, err)
 		}
 
-		if err := command.RunInDir(ctx, dir, "composer", "install", "--no-dev", "--no-interaction", "--prefer-dist"); err != nil {
-			return err
+		args := []string{"install", "--no-interaction", "--prefer-dist"}
+		if !isGenerator {
+			args = append(args, "--no-dev")
```
JoeWang1127: Why can we exclude `--no-dev` for other tools?

internal/librarian/php/install.go
```diff
@@ -59,18 +60,40 @@ func Install(ctx context.Context, tools *config.Tools) error {
...
-		if err := createBinWrapper(tool.Name, destPath, bin); err != nil {
-			return err
+		if _, err := exec.LookPath("composer"); err != nil {
+			return fmt.Errorf("composer is not installed or not in PATH, which is required for PHP tool installation: %w", err)
+		}
+		if err := command.RunInDir(ctx, dir, "composer", args...); err != nil {
+			return fmt.Errorf("failed to run composer install: %w", err)
+		}
+		if isGenerator {
+			phpPath, err := exec.LookPath("php")
```
JoeWang1127: How about creating a command check helper function at the beginning of `Install` function?

internal/librarian/php/install.go
```diff
@@ -59,18 +60,40 @@ func Install(ctx context.Context, tools *config.Tools) error {
...
+			if err != nil {
+				return fmt.Errorf("failed to find php: %w", err)
+			}
+			destPath := filepath.Join(dir, "src", "Main.php")
+			wrapperName := filepath.Base(tool.Repo)
+			wrapperPath := filepath.Join(bin, wrapperName)
+			wrapperContent := fmt.Sprintf("#!/bin/bash\nexec %q -d display_errors=stderr -d memory_limit=1024M %q --side_loaded_root_dir \"$GOOGLEAPIS_DIR\" \"$@\"\n", phpPath, destPath)
+			if err := os.WriteFile(wrapperPath, []byte(wrapperContent), 0755); err != nil {
```
JoeWang1127: Can we reuse `createBinWrapper` function here?

internal/librarian/php/install.go
```diff
@@ -124,24 +147,18 @@ func installGenerator(ctx context.Context) (string, error) {
 		return "", err
 	}
 	// Ensure Composer dependencies are installed
-	vendorDir := filepath.Join(generatorDir, "vendor")
-	if _, err := os.Stat(vendorDir); err != nil {
-		if !errors.Is(err, os.ErrNotExist) {
-			return "", err
+	if _, err := exec.LookPath("composer"); err == nil {
```
JoeWang1127: Why do we need to modify this part since it is testing code.

internal/librarian/php/install_test.go
JoeWang1127, line 61: Can you remove this field since no test uses it.

JoeWang1127, line 236: Do you need to check anything in this test, e,g., the content of the wrapper?

internal/librarian/php/install.go
sofisl: Technically the only tool listed right now in `tools` is the generator, so this is kind of a moot point. But in general, usually we only need to run a dependency (i.e., like running a linter). We want to be safe and ensure the generator has all the dependencies it historically relied on during execution (for example, packages like google/gax are listed in its require-dev), so we don't risk breaking existing generation pipelines.

internal/librarian/php/install.go
sofisl: Good point! I had originally tried to clean up the composer installation logic in this temporary testing function by making it remove the vendor directory entirely, but you're right that it's unnecessary churn for code that is slated for removal anyway. I've reverted installGenerator back to its original state (with the os.Stat check intact) to avoid touching it.

internal/librarian/php/install_test.go
sofisl: Since it doesn't have to do with this PR, I'll add it to the refactoring steps later.

sofisl: Sounds good!

internal/librarian/php/install.go
```diff
@@ -47,6 +48,15 @@ func Install(ctx context.Context, tools *config.Tools) error {
 		return err
 	}
 
+	var phpPath string
+	if len(tools.Composer) > 0 {
+		var err error
+		phpPath, err = checkRequiredCommands()
```
JoeWang1127, line 55:
```suggestion
		phpPath, err := checkRequiredCommands()
```

internal/librarian/php/install.go
```diff
@@ -47,6 +48,15 @@ func Install(ctx context.Context, tools *config.Tools) error {
 		return err
 	}
 
+	var phpPath string
+	if len(tools.Composer) > 0 {
+		var err error
```
JoeWang1127, line 54: Do you need to define a variable?

internal/librarian/php/install.go
```diff
@@ -59,17 +69,33 @@ func Install(ctx context.Context, tools *config.Tools) error {
...
+			return fmt.Errorf("failed to run composer install: %w", err)
+		}
+
+		var wrapperName, wrapperContent string
+		if isGenerator {
+			destPath := filepath.Join(dir, "src", "Main.php")
+			wrapperName = filepath.Base(tool.Repo)
+			wrapperContent = fmt.Sprintf("#!/bin/bash\nexec %q -d display_errors=stderr -d memory_limit=1024M %q --side_loaded_root_dir \"$GOOGLEAPIS_DIR\" \"$@\"\n", phpPath, destPath)
+		} else {
+			destPath := filepath.Join(dir, "vendor", "bin", tool.Name)
```
JoeWang1127: Is it a convention to put the wrapper in bin/vendor/bin/{toolName}?

Can we just use bin/{toolName} for simplicity?

internal/librarian/php/install.go
```diff
@@ -162,18 +199,26 @@ func generatorDir(ctx context.Context) (string, error) {
...
 	wrapperPath := filepath.Join(binDir, wrapperName)
-	content := fmt.Sprintf("#!/bin/sh\nexec %q \"$@\"\n", destPath)
 	if err := os.MkdirAll(filepath.Dir(wrapperPath), 0755); err != nil {
 		return fmt.Errorf("failed to create directory for wrapper: %w", err)
 	}
-	_ = os.Remove(wrapperPath)
 	f, err := os.OpenFile(wrapperPath, os.O_WRONLY|os.O_CREATE|os.O_EXCL, 0755)
 	if err != nil {
-		return err
+		if os.IsExist(err) {
```
JoeWang1127:
```suggestion
		if errors.Is(err, fs.ErrNotExist) {
```

JoeWang1127: Could you explain what does line 209 - 214 do?

internal/librarian/php/install.go
zhumin8: Note we have a case: #6978
From my limited knowledge, this also won't work with `--no-dev`?

Why not just go with the default? what is the benefit of keeping this check and adding `no-dev`?

internal/librarian/php/install.go
zhumin8: If generator is the only use case now, what about deleting this else block for simplicity/low maintance.
I don't think we tested anything other than generator either, so I won't say we are 100% confident about this code path.

internal/librarian/php/install.go
```diff
@@ -89,6 +115,17 @@ func Install(ctx context.Context, tools *config.Tools) error {
 	return nil
 }
 
+func checkRequiredCommands() (string, error) {
+	if _, err := exec.LookPath("composer"); err != nil {
+		return "", fmt.Errorf("composer is not installed or not in PATH, which is required for PHP tool installation: %w", err)
```
zhumin8: Error language used L120 and L124 seems unnecessary different. Let's use concise and direct language for both.

internal/librarian/php/install_test.go
```diff
@@ -74,6 +77,19 @@ func TestInstall(t *testing.T) {
 				writeExecutable(t, filepath.Join(bin, "composer"), "#!/bin/sh\nexit 0\n")
 				t.Setenv("PATH", bin+string(os.PathListSeparator)+os.Getenv("PATH"))
 			},
+			check: func(t *testing.T) {
+				repoDir := filepath.Join(os.Getenv("LIBRARIAN_CACHE"), "github.com/googleapis/gapic-generator-php@v1.21.2")
+				wrapperPath := filepath.Join(repoDir, "wrapper.sh")
+				b, err := os.ReadFile(wrapperPath)
+				if err != nil {
+					t.Fatalf("failed to read wrapper: %v", err)
```
zhumin8: use `t.Fatal(err)` per https://github.com/googleapis/librarian/blob/main/doc/howwewritego.md#use-tfatal-or-terror-for-simple-error-handling

internal/librarian/php/install_test.go
```diff
@@ -107,6 +123,20 @@ func TestInstall(t *testing.T) {
 				writeExecutable(t, filepath.Join(bin, "pip"), "#!/bin/sh\nexit 0\n")
 				t.Setenv("PATH", bin+string(os.PathListSeparator)+os.Getenv("PATH"))
 			},
+			check: func(t *testing.T) {
+				binDir := filepath.Join(os.Getenv("LIBRARIAN_BIN"), "php_tools", "bin")
+				wrapperPath := filepath.Join(binDir, "fake-composer-tool")
+				b, err := os.ReadFile(wrapperPath)
+				if err != nil {
+					t.Fatalf("failed to read wrapper: %v", err)
```
zhumin8: use `t.Fatal(err)` per https://github.com/googleapis/librarian/blob/main/doc/howwewritego.md#use-tfatal-or-terror-for-simple-error-handling

internal/librarian/php/install_test.go
```diff
@@ -148,6 +178,64 @@ func TestInstall(t *testing.T) {
 				writeExecutable(t, filepath.Join(bin, "pnpm"), "#!/bin/sh\nexit 0\n")
 				t.Setenv("PATH", bin+string(os.PathListSeparator)+os.Getenv("PATH"))
 			},
+			check: func(t *testing.T) {
+				binDir := filepath.Join(os.Getenv("LIBRARIAN_BIN"), "php_tools", "bin")
+				wrapperPath := filepath.Join(binDir, "fake-composer-tool")
+				b, err := os.ReadFile(wrapperPath)
+				if err != nil {
+					t.Fatalf("failed to read wrapper: %v", err)
```
zhumin8: use `t.Fatal(err)` per https://github.com/googleapis/librarian/blob/main/doc/howwewritego.md#use-tfatal-or-terror-for-simple-error-handling

same below

zhumin8 (commented):
nit: The two fixes, `--no-dev` flag and wrapper logic fixes seems different enough, perhaps separate into 2 PR is easiler to review.

super nit: in description, you have "For https://github.com/googleapis/librarian/issues/6903 Fixes https://github.com/googleapis/librarian/issues/6903" repeating.

internal/librarian/php/install.go
sofisl: Yeah in my experience in Node, we generally tried not to install dev dependencies unless absolutely necessary since it was costly and time consuming. I figured this would be the case for PHP but when I asked Jetski about your new situation, it seems that this tool will also need to install dev dependencies. So I'll just make it the default.

internal/librarian/php/install.go
sofisl: Composer installs the tool inside the `vendor` directory, the wrapper itself is just in bin/{toolName}

sofisl: Ok fair enough. I'll add a comment about this code path saying that this currently will *only* work for the generator.

internal/librarian/php/install.go
sofisl: we use it in line 91 so we can't do this!

internal/librarian/php/install.go
sofisl: yes, (learning about go): since we're reusing phpPath below, if we did :=, we'd be scoping those variables locally and breaking the build. So we have to use normal local assigment.

internal/librarian/php/install.go
```diff
@@ -59,17 +68,29 @@ func Install(ctx context.Context, tools *config.Tools) error {
...
+
 		dir, err := fetch.Repo(ctx, tool.Repo, tool.Version, tool.SHA256)
 		if err != nil {
 			return fmt.Errorf("fetching %s: %w", tool.Name, err)
 		}
 
-		if err := command.RunInDir(ctx, dir, "composer", "install", "--no-dev", "--no-interaction", "--prefer-dist"); err != nil {
-			return err
+		args := []string{"install", "--no-interaction", "--prefer-dist"}
+		if err := command.RunInDir(ctx, dir, "composer", args...); err != nil {
```
zhumin8: nit: `args` only used here. why not inline it?

internal/librarian/php/install.go
```diff
@@ -59,17 +68,29 @@ func Install(ctx context.Context, tools *config.Tools) error {
 		if tool.Repo == "" {
 			return fmt.Errorf("%w: composer tool %s", errMissingRepo, tool.Name)
 		}
+
```
zhumin8: nit: remove empty line

internal/librarian/php/install.go
```diff
@@ -59,17 +68,29 @@ func Install(ctx context.Context, tools *config.Tools) error {
...
-		if err := command.RunInDir(ctx, dir, "composer", "install", "--no-dev", "--no-interaction", "--prefer-dist"); err != nil {
-			return err
+		args := []string{"install", "--no-interaction", "--prefer-dist"}
+		if err := command.RunInDir(ctx, dir, "composer", args...); err != nil {
+			return fmt.Errorf("failed to run composer install: %w", err)
 		}
 
-		destPath := filepath.Join(dir, "vendor", "bin", tool.Name)
-		if err := createBinWrapper(tool.Name, destPath, bin); err != nil {
+		// TODO: We will need to add branching logic here when we add other tools
```
zhumin8: link the issue in "TODO(issue link):"
ref. https://github.com/googleapis/librarian/blob/main/CONTRIBUTING.md#leaving-a-todo

internal/librarian/php/install.go
```diff
@@ -59,17 +68,29 @@ func Install(ctx context.Context, tools *config.Tools) error {
...
+		// (like the upcoming google-cloud tool in #6978).
+		// Currently, this assumes the tool is the gapic-generator-php. This specific
+		// wrapper logic will not work for generic Composer tools because:
+		// 1. It hardcodes the executable entry point to "src/Main.php" (ignoring Composer's vendor/bin/ paths).
+		// 2. It injects specific PHP configurations (e.g. memory_limit=1024M) required to prevent the generator from crashing.
+		// 3. It automatically injects the "--side_loaded_root_dir" argument which other tools will not expect.
```
zhumin8: What is  "--side_loaded_root_dir" argument and why is it needed? 
Did a bit digging, looks like this is only needed because we are passing in relative paths for config files
  • $opts['grpc_service_config'] = "google/cloud/ces/v1/ces_grpc_service_config.json"
  • $opts['service_yaml'] = "google/cloud/ces/v1/ces_v1.yaml"

If so we should be able to remove this by:
  1. Modify generate.go to resolve and pass absolute paths in --gapic_out.
  2. Modify install.go to remove --side_loaded_root_dir "$GOOGLEAPIS_DIR" from the generated wrapper.sh.

I will file a followup issue for this. For now, can you add a comment for this argument?

internal/librarian/php/install.go
```diff
@@ -162,18 +194,19 @@ func generatorDir(ctx context.Context) (string, error) {
...
 	if err := os.MkdirAll(filepath.Dir(wrapperPath), 0o755); err != nil {
 		return fmt.Errorf("failed to create directory for wrapper: %w", err)
 	}
 	_ = os.Remove(wrapperPath)
 	f, err := os.OpenFile(wrapperPath, os.O_WRONLY|os.O_CREATE|os.O_EXCL, 0o755)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to create wrapper script: %w", err)
 	}
 	defer f.Close()
```
zhumin8: check the defer error.

internal/librarian/php/install.go
```diff
@@ -59,17 +68,29 @@ func Install(ctx context.Context, tools *config.Tools) error {
...
 		}
 
-		destPath := filepath.Join(dir, "vendor", "bin", tool.Name)
-		if err := createBinWrapper(tool.Name, destPath, bin); err != nil {
+		// TODO: We will need to add branching logic here when we add other tools
+		// (like the upcoming google-cloud tool in #6978).
+		// Currently, this assumes the tool is the gapic-generator-php. This specific
+		// wrapper logic will not work for generic Composer tools because:
+		// 1. It hardcodes the executable entry point to "src/Main.php" (ignoring Composer's vendor/bin/ paths).
+		// 2. It injects specific PHP configurations (e.g. memory_limit=1024M) required to prevent the generator from crashing.
```
zhumin8: It's useful to know why this limit. I suppose from https://github.com/googleapis/gapic-generator-php/commit/685b419f2220e2d19c74e7f1464067f995cf1a95
Can you add in comment?

internal/librarian/php/install.go
JoeWang1127: Could you create an issue?

internal/librarian/php/install.go
JoeWang1127: Does this configuration exist in Bazel target?

internal/librarian/php/install.go
zhumin8: filed https://github.com/googleapis/librarian/issues/7000

JoeWang1127: Could you explain what does this configuration do?

internal/librarian/php/install.go
```diff
@@ -47,6 +47,15 @@ func Install(ctx context.Context, tools *config.Tools) error {
 		return err
 	}
 
+	var phpPath string
+	if len(tools.Composer) > 0 {
```
JoeWang1127, line 53: How about we verify `tools.Composer` is not empty before install?

We can't run generate without installing tools anyways.

internal/librarian/php/install_test.go
```diff
@@ -74,6 +77,19 @@ func TestInstall(t *testing.T) {
...
 			},
+			check: func(t *testing.T) {
+				repoDir := filepath.Join(os.Getenv("LIBRARIAN_CACHE"), "github.com/googleapis/gapic-generator-php@v1.21.2")
+				wrapperPath := filepath.Join(repoDir, "wrapper.sh")
+				b, err := os.ReadFile(wrapperPath)
+				if err != nil {
+					t.Fatal(err)
+				}
+				phpPath, _ := exec.LookPath("php")
+				want := fmt.Sprintf("#!/bin/bash\nexec %q -d display_errors=stderr -d memory_limit=1024M %q --side_loaded_root_dir \"$GOOGLEAPIS_DIR\" \"$@\"\n", phpPath, filepath.Join(repoDir, "src/Main.php"))
```
zhumin8: This string is repeated several times, extract mini helper to avoid duplicating?

internal/librarian/php/install_test.go
```diff
@@ -74,6 +77,19 @@ func TestInstall(t *testing.T) {
...
+				if diff := cmp.Diff(want, string(b)); diff != "" {
+					t.Errorf("wrapper content mismatch (-want +got):\n%s", diff)
```
JoeWang1127:
```suggestion
					t.Errorf("mismatch (-want +got):\n%s", diff)
```

internal/librarian/php/install_test.go
```diff
@@ -105,8 +121,24 @@ func TestInstall(t *testing.T) {
...
+				b, err := os.ReadFile(wrapperPath)
+				if err != nil {
+					t.Fatal(err)
+				}
+				repoDir := filepath.Join(os.Getenv("LIBRARIAN_CACHE"), "github.com/fake/fake-tool@1.0.0")
+				destPath := filepath.Join(repoDir, "src", "Main.php")
+				phpPath, _ := exec.LookPath("php")
+				want := fmt.Sprintf("#!/bin/bash\nexec %q -d display_errors=stderr -d memory_limit=1024M %q --side_loaded_root_dir \"$GOOGLEAPIS_DIR\" \"$@\"\n", phpPath, destPath)
+				if diff := cmp.Diff(want, string(b)); diff != "" {
+					t.Errorf("wrapper content mismatch (-want +got):\n%s", diff)
```
JoeWang1127:
```suggestion
					t.Errorf("mismatch (-want +got):\n%s", diff)
```

internal/librarian/php/install_test.go
```diff
@@ -146,8 +178,68 @@ func TestInstall(t *testing.T) {
...
```
JoeWang1127:
```suggestion
					t.Errorf("mismatch (-want +got):\n%s", diff)
```

internal/librarian/php/install_test.go
```diff
@@ -146,8 +178,68 @@ func TestInstall(t *testing.T) {
...
+				b, err := os.ReadFile(wrapperPath)
+				if err != nil {
+					t.Fatal(err)
+				}
+				repoDir := filepath.Join(os.Getenv("LIBRARIAN_CACHE"), "github.com/googleapis/gapic-generator-php@1.0.0")
+				destPath := filepath.Join(repoDir, "src", "Main.php")
+				phpPath, _ := exec.LookPath("php")
+				want := fmt.Sprintf("#!/bin/bash\nexec %q -d display_errors=stderr -d memory_limit=1024M %q --side_loaded_root_dir \"$GOOGLEAPIS_DIR\" \"$@\"\n", phpPath, destPath)
+				if diff := cmp.Diff(want, string(b)); diff != "" {
+					t.Errorf("wrapper content mismatch (-want +got):\n%s", diff)
```
JoeWang1127:
```suggestion
					t.Errorf("mismatch (-want +got):\n%s", diff)
```

internal/librarian/php/install.go
sofisl: Got you. I think then is probably not better as a TODO. This is in response to your request to remove the branching logic earlier - I think we will need to write a branch eventually because this logic is specific to the generator. But I don't have the work planned since we aren't onboarding another tool yet. Anyways, I'll just remove the "TODO" part.

internal/librarian/php/install.go
sofisl: Will do.

sofisl: see above!

internal/librarian/php/install.go
sofisl: see above as well

internal/librarian/php/install.go
sofisl: So after implementing this, tests break because we're kind of relying on Min's fallback of using the hardcoded generator wrapper. However we want to get rid of it anyways, so I added a copy instruction in the test to copy over the migration librarian.yaml to test it during install (which is what we do for Go). @zhumin8 please lmk what you think.

sofisl: Actually this is just too much for this PR. I'll do this when I refactor and remove Min's fallback. Added a TODO.

sofisl:
I've decided to revert any changes to generate since the PR was becoming too large. Any further refactor/improvement suggestions I'll add to #6824 and do separately.

internal/librarian/php/install.go
```diff
@@ -63,14 +74,29 @@ func Install(ctx context.Context, tools *config.Tools) error {
...
+			// wrapper logic will not work for generic Composer tools because:
+			// 1. It hardcodes the executable entry point to "src/Main.php" (ignoring Composer's vendor/bin/ paths).
+			// 2. It injects specific PHP configurations (e.g. memory_limit=1024M) required to prevent the generator from crashing.
+			// See https://github.com/googleapis/gapic-generator-php/commit/685b419f2220e2d19c74e7f1464067f995cf1a95
+			// 3. It automatically injects the "--side_loaded_root_dir" argument which other tools will not expect.
+			// (this argument is to pass through relative paths for config files)
+			// TODO(https://github.com/googleapis/librarian/issues/7000): Remove the --side_loaded_root_dir once we pass full paths to generator
+			destPath := filepath.Join(dir, "src", "Main.php")
+			wrapperContent := phpWrapperContent(phpPath, destPath)
+
```
JoeWang1127: nit: remove this empty line.

internal/librarian/php/install_test.go
```diff
@@ -97,25 +100,41 @@ func TestInstall(t *testing.T) {
...
+				binDir := filepath.Join(os.Getenv("LIBRARIAN_BIN"), "php_tools", "bin")
+				wrapperPath := filepath.Join(binDir, "gapic-generator-php")
+				b, err := os.ReadFile(wrapperPath)
+				if err != nil {
+					t.Fatal(err)
+				}
+				repoDir := filepath.Join(os.Getenv("LIBRARIAN_CACHE"), "github.com/googleapis/gapic-generator-php@1.0.0")
+				destPath := filepath.Join(repoDir, "src", "Main.php")
+				phpPath, _ := exec.LookPath("php")
+				want := phpWrapperContent(phpPath, destPath)
```
JoeWang1127, line 124: It is possible to hardcode the wrapper content so we don't rely on `phpWrapperContent` for testing.

internal/librarian/php/install.go
```diff
@@ -161,19 +198,29 @@ func generatorDir(ctx context.Context) (string, error) {
 	return fetch.Repo(ctx, "github.com/googleapis/gapic-generator-php", generatorVersion, generatorSHA256)
 }
 
+// phpWrapperContent generates the bash script content for the PHP tool wrapper.
+func phpWrapperContent(phpExecutable, entrypoint string) string {
```
JoeWang1127, line 201:
```suggestion
func wrapperContent(phpExecutable, entrypoint string) string {
```

JoeWang1127 (commented):
One last comment on the tests, thanks.

internal/librarian/php/install_test.go
sofisl: I think this is going in direct contrast with min's suggestion, unless I'm misunderstanding?

https://github.com/googleapis/librarian/pull/6941#discussion_r3641215109

JoeWang1127 (approved):

internal/librarian/php/install_test.go
JoeWang1127: Ah, my concern is that the expected string should be static.

However, I think leave it for now is fine. We can come back to it later.

internal/librarian/php/install.go
```diff
@@ -63,14 +74,28 @@ func Install(ctx context.Context, tools *config.Tools) error {
...
+
+		if wrapperName == "gapic-generator-php" {
+			// Currently, this assumes the tool is the gapic-generator-php. This specific
+			// wrapper logic will not work for generic Composer tools because:
+			// 1. It hardcodes the executable entry point to "src/Main.php" (ignoring Composer's vendor/bin/ paths).
+			// 2. It injects specific PHP configurations (e.g. memory_limit=1024M) required to prevent the generator from crashing.
+			// See https://github.com/googleapis/gapic-generator-php/commit/685b419f2220e2d19c74e7f1464067f995cf1a95
+			// 3. It automatically injects the "--side_loaded_root_dir" argument which other tools will not expect.
+			// (this argument is to pass through relative paths for config files)
+			// TODO(https://github.com/googleapis/librarian/issues/7000): Remove the --side_loaded_root_dir once we pass full paths to generator
```
zhumin8, line 91: nit: mind moving relevant comments to `phpWrapperContent`?
This can be done in followup.

zhumin8 (approved):

PR #6872, opened 2026-07-16
ci(node): generate a single library as a smoke test during presubmits, and generate --all in postsubmits

sofisl:
This PR generates a single library (secretmanager) as part of the presubmit workflows for Node. It also generates all libraries as a postsubmit/integration test.

For https://github.com/googleapis/librarian/issues/6815

PR #6942, opened 2026-07-21
feat(internal/librarian/generate): cap goroutine concurrency during `generate --all`

### LINKED ISSUE #6952: node: generate --all breaks with non-allowlisted APIs and goroutine tasks overflow
when attempting to create pre- and post- submit tests for Node: https://github.com/googleapis/librarian/pull/6872

i ran into these issues:

- https://github.com/googleapis/librarian/actions/runs/29781733203/job/88484310001
- https://github.com/googleapis/librarian/actions/runs/29783811456/job/88490821427

I opened this attempt to fix it, but it is more general (touches more libraries): https://github.com/googleapis/librarian/pull/6942

so wanted to check in first if this is the right solution.

sofisl:
This change limits goroutine concurrency to the number of available CPUs across all language generators to prevent resource exhaustion during parallel builds.

I discovered this issue while attempting to run `generate --all` for Nodejs in #6872.

Unblocks https://github.com/googleapis/librarian/pull/6872
Fixes #6952

### DIFF
--- a/internal/librarian/generate.go
+++ b/internal/librarian/generate.go
@@ -44,6 +44,7 @@ var (
 	errSkipGenerate            = errors.New("library has skip_generate set")
 	errNoPreviewVariant        = errors.New("library does not have a preview variant")
 	errUnsupportedLanguage     = errors.New("language does not support generation")
+	errNoLibrariesToGenerate   = errors.New("no libraries to generate: all libraries have skip_generate set")
 )
 
 func generateCommand() *cli.Command {
@@ -156,7 +157,7 @@ func runGenerate(ctx context.Context, cfg *config.Config, all bool, libraryName 
 	}
 	if len(libraries) == 0 {
 		if all {
-			return errors.New("no libraries to generate: all libraries have skip_generate set")
+			return errNoLibrariesToGenerate
 		}
 		for _, lib := range cfg.Libraries {
 			if lib.Name == baseName {
--- a/internal/librarian/generate.go
+++ b/internal/librarian/generate.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"runtime"
 	"strings"
 
 	"github.com/googleapis/librarian/internal/config"
@@ -30,6 +31,7 @@ import (
 	"github.com/googleapis/librarian/internal/librarian/ruby"
 	"github.com/googleapis/librarian/internal/librarian/rust"
 	"github.com/googleapis/librarian/internal/librarian/swift"
+	"github.com/googleapis/librarian/internal/serviceconfig"
 	"github.com/googleapis/librarian/internal/sources"
 	"github.com/googleapis/librarian/internal/yaml"
 	"github.com/urfave/cli/v3"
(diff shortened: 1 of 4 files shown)


JoeWang1127:
This pull request tries to so many things:
- Skip generation for APIs that are not allowlisted for a specific language when the all flag is used.
- Limit goroutine concurrency to the number of available CPUs across all language generators to prevent resource exhaustion during parallel builds.
- Exports ErrNotAllowed from the internal/serviceconfig package to support these allowlist checks.

My comments on these changes:
1. Could you separate them into different pull requests?

1. Could you elaborate why we want to do this since surface this issue would be an indicator that we need to update sdk.yaml.
>Skip generation for APIs that are not allowlisted for a specific language when the all flag is used.

1. Do you observe any related issues? Could you paste the run log or link?
> Limit goroutine concurrency to the number of available CPUs across all language generators to prevent resource exhaustion during parallel builds.

1. Returns an exported error will change our API so we should have a good reason for this. Since this is related to my 2nd comment, we can discuss there.
> Exports ErrNotAllowed from the internal/serviceconfig package to support these allowlist checks.

sofisl:
> This pull request tries to so many things:
> 
> * Skip generation for APIs that are not allowlisted for a specific language when the all flag is used.
> * Limit goroutine concurrency to the number of available CPUs across all language generators to prevent resource exhaustion during parallel builds.
> * Exports ErrNotAllowed from the internal/serviceconfig package to support these allowlist checks.
> 
> My comments on these changes:
> 
> 1. Could you separate them into different pull requests?
> 2. Could you elaborate why we want to do this since surface this issue would be an indicator that we need to update sdk.yaml.
> 
> > Skip generation for APIs that are not allowlisted for a specific language when the all flag is used.
> 
> 1. Do you observe any related issues? Could you paste the run log or link?
> 
> > Limit goroutine concurrency to the number of available CPUs across all language generators to prevent resource exhaustion during parallel builds.
> 
> 1. Returns an exported error will change our API so we should have a good reason for this. Since this is related to my 2nd comment, we can discuss there.
> 
> > Exports ErrNotAllowed from the internal/serviceconfig package to support these allowlist checks.

Hi Joe, 

1. Yes, happy to separate them - although there would only be two PRs, since the exported error needs to go with skipping the generation for APIs that are not allowlisted. So the only separation would be for the goroutine concurrency changes. (FWIW, I *already* separated these changes from #6872, where the problem was initially found)

2. I listed the observed issue in the PR body - #6872. Here is the failure for skipping non-allowlisted APIs: https://github.com/googleapis/librarian/actions/runs/29781733203/job/88484310001. And, here is the failure for limiting the goroutine; it is not very descriptive: https://github.com/googleapis/librarian/actions/runs/29783811456/job/88490821427

sofisl:
I'm going to wait to separate until we have an answer on https://github.com/googleapis/librarian/issues/6952

sofisl:
Updated in https://github.com/googleapis/librarian/pull/6959, so we'll only be focusing on goroutine concurrency limits now.

quirogas (approved):
LGMT! 

One thing to keep in mind is that this will increase our generation time, but it make generation more stable. I believe our generations are I/O bound so we could provably benefit from higher number of `gorutines` without having to cap at the number of CPUs, but this would be a benchmarking type efford -- worth considering in the future if we want to speed up things.

JoeWang1127 (approved):

PR #6841, opened 2026-07-15
feat(sidekick/rust): generate LRO `Poller` for bigquery `InsertJob`

haphungw:
Add a core implementation of `lro::Poller` for BigQuery `InsertJob`. 

This maps the custom BQ REST job methods into the standard `lro::Poller` interface.

Staged @ https://github.com/googleapis/google-cloud-rust/pull/6067

PR #6948, opened 2026-07-21
feat(sidekick/rust): make request builder fields `pub(crate)` for gapic extensions

haphungw:
Expose the internal `RequestBuilder` on generated builders like `InsertJob`. 

By exposing these fields to the rest of the crate, handwritten extensions `src/operation.rs` can directly inspect requests.

Staged @ https://github.com/googleapis/google-cloud-rust/pull/6120

### DIFF
--- a/internal/sidekick/rust/annotate_method.go
+++ b/internal/sidekick/rust/annotate_method.go
@@ -49,6 +49,7 @@ type methodAnnotation struct {
 	IsBigQueryInsertJob       bool
 	ClientSideStreaming       bool
 	ServerSideStreaming       bool
+	IsBigQueryV2              bool
 }
 
 // HasGrpcResourceNameArgs returns true if the method has gRPC resource name arguments.
@@ -322,6 +323,7 @@ func (c *codec) annotateMethod(m *api.Method) (*methodAnnotation, error) {
 		IsBigQueryInsertJob:       m.ID == ".google.cloud.bigquery.v2.JobService.InsertJob",
 		ClientSideStreaming:       m.ClientSideStreaming,
 		ServerSideStreaming:       m.ServerSideStreaming,
+		IsBigQueryV2:              m.Service != nil && m.Service.Package == "google.cloud.bigquery.v2",
 	}
 
 	if err := c.annotateResourceNameGeneration(m, annotation); err != nil {
--- a/internal/sidekick/rust/annotate_service.go
+++ b/internal/sidekick/rust/annotate_service.go
@@ -58,6 +58,8 @@ type serviceAnnotations struct {
 	InternalBuilders bool
 	// If true, the service has at least one bidirectional streaming RPC in the API definition.
 	HasBidiStreaming bool
+	// If true, this is the BigQuery v2 service.
+	IsBigQueryV2 bool
 }
 
 // BuilderVisibility returns the visibility for client and request builders.
@@ -142,6 +144,7 @@ func (c *codec) annotateService(s *api.Service) (*serviceAnnotations, error) {
 		DetailedTracingAttributes: c.detailedTracingAttributes,
 		InternalBuilders:          c.internalBuilders,
 		HasBidiStreaming:          hasBidiStreaming,
+		IsBigQueryV2:              s.Package == "google.cloud.bigquery.v2",
 	}
 	s.Codec = ann
 	return ann, nil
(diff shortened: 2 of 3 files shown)


internal/sidekick/rust/templates/crate/src/builder.rs.mustache
```diff
@@ -62,9 +62,9 @@ pub mod {{Codec.ModuleName}} {
     /// Common implementation for [crate::client::{{Codec.Name}}] request builders.
     #[derive(Clone, Debug)]
     pub(crate) struct RequestBuilder<R: std::default::Default> {
-        stub: std::sync::Arc<dyn super::super::stub::dynamic::{{Codec.Name}}>,
```
dbolduc, line 65: Q: why is this necessary? I thought we could implement the job-level retry loop without looking at the builder's internals.

if necessary...

1. I think we should also have the handwritten LRO code in `google-cloud-rust`

   https://github.com/googleapis/librarian/pull/6841#discussion_r3597329621

2. This change is gonna hit every library, and lead to N version bumps. I am thinking we should gate the change in a mustache template, so that only BQ-v2 is affected.

   Then maybe we get rid of the gate when there is a change that affects all libraries.

haphungw: > without looking at the builder's internals

I did try to avoid touching the internals too.
However, we only have access to a .`with_request(req)` setter and no getter, so it is impossible to extract and mutate the request payload once the user passes it it.

For requests with a custom `job_id`, we would need to provide a new `job_id` (by getting the original and appending a suffix) to retry.


> we should also have the handwritten LRO code in `google-cloud-rust`

the handwritten poller lives `operation` module and the generated builder is in `builders` module


> we should gate the change in a mustache template

I agree, fixed accordingly.

dbolduc: > For requests with a custom `job_id`, we would need to provide a new `job_id` (by getting the original and appending a suffix) to retry.

Ack, although I would be fine with generating a new, random job ID.

What will we do if the application's Job ID is already at the character limit? We cannot append a suffix.

> > we should also have the handwritten LRO code in `google-cloud-rust`
> 
> the handwritten poller lives `operation` module and the generated builder is in `builders` module

I have proof that this doesn't matter:

https://github.com/dbolduc/google-cloud-rust/commit/15fed311811a77eb4edef93d68bfe2f6b3f928f6

internal/sidekick/rust/annotate_method.go
```diff
@@ -49,6 +49,7 @@ type methodAnnotation struct {
 	IsBigQueryInsertJob       bool
 	ClientSideStreaming       bool
 	ServerSideStreaming       bool
+	IsBigQueryV2              bool
```
dbolduc, line 52: consider calling it `ExtendRequestBuilder`. We might want to use it for a crate that isn't BQ

internal/sidekick/rust/annotate_method.go
```diff
@@ -322,6 +323,7 @@ func (c *codec) annotateMethod(m *api.Method) (*methodAnnotation, error) {
 		IsBigQueryInsertJob:       m.ID == ".google.cloud.bigquery.v2.JobService.InsertJob",
 		ClientSideStreaming:       m.ClientSideStreaming,
 		ServerSideStreaming:       m.ServerSideStreaming,
+		IsBigQueryV2:              m.Service != nil && m.Service.Package == "google.cloud.bigquery.v2",
```
dbolduc, line 326: unit test?

internal/sidekick/rust/annotate_service.go
```diff
@@ -142,6 +144,7 @@ func (c *codec) annotateService(s *api.Service) (*serviceAnnotations, error) {
 		DetailedTracingAttributes: c.detailedTracingAttributes,
 		InternalBuilders:          c.internalBuilders,
 		HasBidiStreaming:          hasBidiStreaming,
+		IsBigQueryV2:              s.Package == "google.cloud.bigquery.v2",
```
dbolduc, line 147: unit test?

internal/sidekick/rust/annotate_service.go
```diff
@@ -58,6 +58,8 @@ type serviceAnnotations struct {
 	InternalBuilders bool
 	// If true, the service has at least one bidirectional streaming RPC in the API definition.
 	HasBidiStreaming bool
+	// If true, this is the BigQuery v2 service.
+	IsBigQueryV2 bool
```
dbolduc, line 62: same about considering `ExtendRequestBuilder`, (similar to `ExtendGrpcTransport` above)

dbolduc (approved):
please add unit tests.

Renaming the codec field is for your consideration. You can choose not to.

internal/sidekick/rust/templates/crate/src/builder.rs.mustache
haphungw: That is a good point. Then I think we can generate random ID with `uuid` instead.

Closing this PR for redundancy.

PR #6958, opened 2026-07-21
feat(rust): generate method signatures for bidi streaming

suzmue:
Generate method signatures, stub traits, and transport implementations for bidirectional streaming RPCs when include_bidi_streaming_methods is enabled.

Bidi streaming methods are generated on Client, Stub, DynamicStub, Tracing, and Transport types, gated under the google_cloud_unstable_gapic_streaming compiler cfg flag. RequestBuilder generation is omitted for bidi streaming methods because stream handles are created directly.

Stub default implementations and transport placeholders use async { unimplemented!() } and unimplemented!() respectively in place of full streaming transport definitions.

For #6835


internal/sidekick/rust/templates/crate/src/stub.rs.mustache
```diff
@@ -52,14 +52,28 @@ pub(crate) mod dynamic;
...
+    #[cfg(google_cloud_unstable_gapic_streaming)]
     /// Implements [super::client::{{Codec.ServiceNameToPascal}}::{{Codec.Name}}].
+    fn {{Codec.Name}}(
+        &self,
+        _options: crate::RequestOptions,
+    ) -> impl std::future::Future<Output = (
+        google_cloud_gax::streaming::RequestSender<{{InputType.Codec.QualifiedName}}>,
+        google_cloud_gax::streaming::ResponseReceiver<{{OutputType.Codec.QualifiedName}}>,
+    )> + Send {
+        async { unimplemented!() }
```
coryan: why not use `gaxi::unimplemented::unimplemented_stub()` here?

internal/sidekick/rust/templates/crate/src/transport.rs.mustache
```diff
@@ -127,6 +127,20 @@ impl {{Codec.Name}} {
...
+    {{#Codec.IsBidiStreaming}}
+    #[cfg(google_cloud_unstable_gapic_streaming)]
+    async fn {{Codec.Name}}(
+        &self,
+        _options: crate::RequestOptions,
+    ) -> (
+        google_cloud_gax::streaming::RequestSender<{{InputType.Codec.QualifiedName}}>,
+        google_cloud_gax::streaming::ResponseReceiver<{{OutputType.Codec.QualifiedName}}>,
+    ) {
+        unimplemented!()
```
coryan, line 140: I assume this will happen in a future PR? So maybe:

```suggestion
        {{! TODO(....) - implement the body }}
        unimplemented!()
```

coryan (approved):
Address the comments and then merge.

internal/sidekick/rust/templates/crate/src/stub.rs.mustache
suzmue: It's not the right return types, so needed to do some wrangling. I did add it because I want the error message, may consider adding a helper for the correct type since it will look cleaner so added a TODO, but have now included the functionality.

internal/sidekick/rust/templates/crate/src/transport.rs.mustache
suzmue: Correct, added. Added a TODO for tracing as well.

PR #6960, opened 2026-07-21
ci(go): add presubmit smoke test for go, and refactor go.yaml

sofisl:
This change modifies Librarian to use a matrix strategy for the Go tests in order to share setup instructions across unit, smoke, and integration tests. It also adds a smoke test for presubmits

For #6815


JoeWang1127 (approved):

noahdietz (approved):
LGTM but we will need to change the required checks for the branch rules after this is submitted, I believe...

PR #6991, opened 2026-07-22
fix(librarian): add to individual release for python

### LINKED ISSUE #6971: python: librarian add isolates new libraries in individual release
# Proposal: Onboard new Python client libraries into individual `release-please` config by default

## Background & Problem Statement
Currently, when new Python client libraries are onboarded into the `google-cloud-python` monorepo via `librarian add`, they are automatically added to the bulk release configuration ([`.release-please-bulk-manifest.json`](https://github.com/googleapis/google-cloud-python/blob/main/.release-please-bulk-manifest.json)). 

Once an onboarding/generation PR is merged, the new client library appears in the standard automated bulk release PR. However, releasing a **new** library requires manual internal onboarding and follow-up steps (e.g., exit gate config CL & propagation, initial PyPI registration, etc). If those steps are incomplete or delayed, any release PR containing the new package will fail to publish that new package.

## The Bulk Release Bottleneck
Because new client libraries sit in the bulk release configuration from day one, a single unreleased or blocked library effectively **halts the entire bulk release** for all other packages in the repository. 

When there is a backlog of multiple new client libraries being onboarded, each library would have to be individually configured elsewhere and released before the repository-wide bulk release could be cut. While `release-please` supports targeting a single package via the `--path` flag, this does not prevent new, unconfigured libraries in the bulk manifest from being a part of standard automated bulk releases.

## Proposed Solution
Update `librarian add` (and standard Python onboarding procedures) so that new client libraries are placed directly into the **individual release configuration** ([`release-please-individual-config.json`](https://github.com/googleapis/google-cloud-python/blob/main/release-please-individual-config.json) and [`.release-please-individual-manifest.json`](https://github.com/googleapis/google-cloud-python/blob/main/.release-please-individual-manifest.json)) from the start.

### Key Benefits
1. **Unblocks Bulk Releases:** New libraries are isolated from the primary release pipeline. An incomplete onboarding, pending PyPI registration, or failed release for a new client will never block routine releases for existing libraries.
2. **Flexible Onboarding Timelines:** Teams can complete the required manual and internal onboarding steps for new libraries in a reasonable timeframe without the pressure of holding up repository-wide releases.
3. **Safer First Releases:** Individual release PRs can be reviewed, debugged, and merged independently.

### Trade-offs / Next Steps
* **Post-Onboarding Migration:** Once a new client library has completed its initial release and all internal onboarding prerequisites are satisfied, a follow-up PR will be needed to move its configuration from `release-please-individual-config.json` into the bulk manifest. While this adds a small amount of manual toil (opening a quick PR), it takes under a minute and significantly outweighs the cost of blocking bulk releases.

noahdietz:
New python libraries are appended to the _individual_ `release-please` config instead of the bulk. They can be manually moved to the bulk release config after they are published for the first time.

Tested locally to produce https://github.com/noahdietz/google-cloud-python/commit/da98a3749e9f63e1d8e97cfdce6d7dab37473833.

Fixes https://github.com/googleapis/librarian/issues/6971


internal/librarian/release_please_test.go
```diff
@@ -24,7 +24,7 @@ import (
 	"github.com/googleapis/librarian/internal/config"
 )
 
-func TestHasBulkReleasePleaseConfigs(t *testing.T) {
+func TestHasReleasePleaseConfigs(t *testing.T) {
```
JoeWang1127, line 27: Do you think we should separate test per language, otherwise this test will keep growing?

internal/librarian/release_please_test.go
```diff
@@ -469,18 +497,19 @@ func TestSyncToReleasePlease_Errors(t *testing.T) {
 	} {
 		t.Run(test.name, func(t *testing.T) {
 			tmp := t.TempDir()
-			manifestPath := filepath.Join(tmp, ".release-please-bulk-manifest.json")
-			configPath := filepath.Join(tmp, "release-please-bulk-config.json")
+			cfg := &config.Config{
```
JoeWang1127, line 500: Similar comment here, could we create a test for python?

JoeWang1127 (approved):

internal/librarian/release_please_test.go
noahdietz: I think this one is fine as is, the test cases and necessary inputs are almost identical - this is table-driven tests working well :)

internal/librarian/release_please_test.go
noahdietz: I agree, I think this one should be separated. The fact that the test logic is all changed up indicates the test has outgrown its generality. I'll refactor in this PR, please hold.

noahdietz: Actually, I think we can make this table driven by including the `config` in the test case. The `releasePleaseFiles` are already switched on language. This could be similar to the other tests.

noahdietz: Actually, we don't need each language being tested here, that's why the `config.LanguagePython` was hardcoded previously. We just need one. Regardless of what the file name is for a specific, we are just exercising the error modes. I think we should just use LanguageFake here to signify that.

noahdietz: Actually (again) since this was already hardcoded to Python, I think that was the intention of the test. We only update the logic here to use the `releasePleaseFiles` API rather than hard coded ones in order to reflect that we do not use the same config by default anymore.

I'm going to leave this as is.

Issue #6969, opened 2026-07-22
nodejs: `librarian generate` overwrites `index.ts` barrel exports on multi-API Node.js packages


quirogas:
When `librarian generate` executes on a Node.js client library configured with **multiple proto API paths targeting the same version namespace** in `librarian.yaml` (for example, a data service and its corresponding admin services), subsequent generation passes overwrite `src/v*/index.ts` instead of merging exports. This causes main data client exports (such as `FirestoreClient`, `BigtableClient`, `DatastoreClient`, and `SpannerClient`) to be dropped from `src/v*/index.ts` and `src/index.ts`, breaking CommonJS and TypeScript imports for consumers of the package.

## 🔍 Steps to Reproduce
    1. In `librarian.yaml`, define a library containing multiple API paths under the same version directory (e.g. `v2`):
       ```yaml
       name: google-cloud-bigtable-api
       apis:
         - path: google/bigtable/v2
         - path: google/bigtable/admin/v2

  2. Run librarian generate google-cloud-bigtable-api.
  3. Inspect packages/google-cloud-bigtable-api/src/v2/index.ts and packages/google-cloud-bigtable-api/src/index.ts.

## 💥 Expected vs. Actual Behavior

### Expected Behavior

  src/v2/index.ts should export all GAPIC client classes generated across all proto paths defined for that version:

    export { BigtableClient } from './bigtable_client';
    export { BigtableInstanceAdminClient } from './bigtable_instance_admin_client';
    export { BigtableTableAdminClient } from './bigtable_table_admin_client';
### Actual Behavior

The second generation pass (google/bigtable/admin/v2) overwrites src/v2/index.ts with only the admin client exports:

    // BigtableClient is missing!
    export { BigtableInstanceAdminClient } from './bigtable_instance_admin_client';
    export { BigtableTableAdminClient } from './bigtable_table_admin_client';

## Impacted Multi-API Packages

  • @google-cloud/bigtable-api (google/bigtable/v2 + google/bigtable/admin/v2)
  • @google-cloud/firestore-api (google/firestore/v1 + google/firestore/admin/v1)
  • @google-cloud/datastore-api (google/datastore/v1 + google/datastore/admin/v1)
  • @google-cloud/spanner-api (google/spanner/v1 + google/spanner/admin/database/v1 + google/spanner/admin/instance/v1 + google/spanner/executor/v1)

##  Root Cause

[...]

PR #6993, opened 2026-07-23
fix(nodejs): isolate staging subdirectories per api version

quirogas:
Construct unique staging subdirectory paths per API pass (owl-bot-staging/<lib>/<index>_<api_slug>_<version>) when generating Node.js libraries. This prevents sequential gapic-generator runs for multi-API packages from overwriting each other before post-processing.

For https://github.com/googleapis/librarian/issues/6969

### DIFF
--- a/internal/librarian/nodejs/generate.go
+++ b/internal/librarian/nodejs/generate.go
@@ -102,9 +102,9 @@ func requireCachedTool(toolName string) (string, error) {
 	return toolPath, nil
 }
 
-func buildStagingSubdirName(index int, apiPath, version string) string {
+func buildStagingSubdirName(index int, apiPath string) string {
 	slug := strings.ReplaceAll(apiPath, "/", "_")
-	return fmt.Sprintf("%d_%s_%s", index, slug, version)
+	return fmt.Sprintf("%d_%s", index, slug)
 }
 
 func generateAPI(ctx context.Context, apiIndex int, api *config.API, library *config.Library, googleapisDir, repoRoot string) error {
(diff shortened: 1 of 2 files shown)


quirogas:
Adding do not merge, because we first have to wait for the release of `gapic-node-processing` to go through.

internal/librarian/nodejs/generate.go
```diff
@@ -102,7 +102,12 @@ func requireCachedTool(toolName string) (string, error) {
 	return toolPath, nil
 }
 
-func generateAPI(ctx context.Context, api *config.API, library *config.Library, googleapisDir, repoRoot string) error {
+func buildStagingSubdirName(index int, apiPath string) string {
+	slug := strings.ReplaceAll(apiPath, "/", "_")
+	return fmt.Sprintf("%d_%s", index, slug)
+}
+
+func generateAPI(ctx context.Context, apiIndex int, api *config.API, library *config.Library, googleapisDir, repoRoot string) error {
```
JoeWang1127: Could you combine the inputs (excluding ctx) to a struct?

internal/librarian/nodejs/generate.go
```diff
@@ -58,14 +58,14 @@ func Generate(ctx context.Context, cfg *config.Config, library *config.Library,
 		return fmt.Errorf("failed to create output directory: %w", err)
 	}
 	repoRoot := filepath.Dir(filepath.Dir(outdir))
-	for _, api := range library.APIs {
+	for i, api := range library.APIs {
```
JoeWang1127, line 61: Why can we use TempDir to create a random staging directory?

quirogas: We could. This directory name is already somewhat misleading. I believe other languages follow this pattern as well, so we could  address it consistently across all languages rather than just for Node. Let me know you think, and we can use this comment to create a tracking issue.

JoeWang1127: OK, I'll find a consistent way to define staging dir.

This is fine for now.

JoeWang1127 (approved):

PR #6997, opened 2026-07-23
feat(dart): generate a testing skill

brianquinlan:
Adds a skill to help agents generate tests using fakes. The generated skill for is `package: google_cloud_ai_generativelanguage_v1beta` is:

````markdown
---
name: google_cloud_ai_generativelanguage_v1beta-tests
description: >-
  Use this skill when writing tests for code that uses
  package:google_cloud_ai_generativelanguage_v1beta
---

## Testing with Fakes

Most unit tests should use fakes instead of making network requests to Google
services.

- Code that uses `CacheService` can be tested by injecting the
  fake `FakeCacheService`.
- Code that uses `DiscussService` can be tested by injecting the
  fake `FakeDiscussService`.
- Code that uses `FileService` can be tested by injecting the
  fake `FakeFileService`.
- Code that uses `GenerativeService` can be tested by injecting the
  fake `FakeGenerativeService`.
- Code that uses `ModelService` can be tested by injecting the
  fake `FakeModelService`.
- Code that uses `PermissionService` can be tested by injecting the
  fake `FakePermissionService`.
- Code that uses `PredictionService` can be tested by injecting the
  fake `FakePredictionService`.
- Code that uses `RetrieverService` can be tested by injecting the
  fake `FakeRetrieverService`.
- Code that uses `TextService` can be tested by injecting the
  fake `FakeTextService`.

Import the fakes from the testing library:

```dart
import 'package:google_cloud_ai_generativelanguage_v1beta/testing.dart';
```

### Option A: Using Constructor Closures (Recommended)
You can inject behavior by passing optional function callbacks to the fake's
constructor. Methods that are not provided will throw an `UnsupportedError`.

```dart
import 'package:google_cloud_ai_generativelanguage_v1beta/generativelanguage.dart';
import 'package:google_cloud_ai_generativelanguage_v1beta/testing.dart';

final fake = FakeFileService(
  createFile: (request) async {
    // Assert request contents here if needed.
    return CreateFileResponse();
  },
);
```

### Option B: Subclassing the Fake
For more complex test setups or shared states, you can subclass the fake and
override its methods. Methods that are not overriden will throw an
`UnsupportedError`.

```dart
import 'package:google_cloud_ai_generativelanguage_v1beta/generativelanguage.dart';
import 'package:google_cloud_ai_generativelanguage_v1beta/testing.dart';

final class MyFakeFileService extends FakeFileService {
  @override
  Future<CreateFileResponse> createFile(
    CreateFileRequest request,
  ) async {
    // Assert request contents here if needed.
    return CreateFileResponse();
  }
}
```

## A Simple Test

```dart
import 'package:google_cloud_ai_generativelanguage_v1beta/generativelanguage.dart';
import 'package:google_cloud_ai_generativelanguage_v1beta/testing.dart';
import 'package:test/test.dart';

Future<void> functionUnderTest(FileService service) async {
  // Application logic here.
  await service.createFile(CreateFileRequest());
  // More application logic here.
}

void main() {
  test('test', () async {
    final fake = FakeFileService(
      createFile: (request) async {
          // Assert request contents here.
          return CreateFileResponse();
      },
    );
    // Instead of verifying that `functionUnderTest`, you should verify the
    // relevant properties of the result.
    await expectLater(functionUnderTest(fake), completes);
  });
}
````

### CODE REVIEW

**Comment by natebosch** [comment]
> ```
> name: google_cloud_ai_generativelanguage_v1beta-tests
> ```

@jakemac53 should we replace the underscores here?

**Comment by jakemac53** [comment]
> @jakemac53 should we replace the underscores here?

Technically yes only hyphens are allowed in skill names. But I don't think it actually breaks any clients.

File: internal/sidekick/dart/templates/skills/tests.md.mustache [hunk 1]
```diff
@@ -0,0 +1,94 @@
...
+{{#Codec.ExampleMethodName}}
+
+Import the fakes from the testing library:
+
+```dart
+import 'package:{{Codec.PackageName}}/testing.dart';
+```
+
+### Option A: Using Constructor Closures (Recommended)
+You can inject behavior by passing optional function callbacks to the fake's
```
natebosch, line 27: [nit]

```suggestion
### Option A: Using Constructor Closures (Recommended)

You can inject behavior by passing optional function callbacks to the fake's
```

internal/sidekick/dart/templates/skills/tests.md.mustache
```diff
@@ -0,0 +1,94 @@
...
+    // Assert request contents here if needed.
+    {{#Codec.ExampleMethodReturnsValue}}
+    return {{Codec.ExampleMethodResponseType}}();
+    {{/Codec.ExampleMethodReturnsValue}}
+  },
+);
+```
+
+### Option B: Subclassing the Fake
+For more complex test setups or shared states, you can subclass the fake and
```
natebosch, line 46: [nit]

```suggestion
### Option B: Subclassing the Fake

For more complex test setups or shared states, you can subclass the fake and
```

internal/sidekick/dart/templates/skills/tests.md.mustache
```diff
@@ -0,0 +1,94 @@
+---
+name: {{Codec.PackageName}}-tests
```
natebosch: I think we might want to replace `_` with `-` in the package name if possible.

natebosch (approved):

internal/sidekick/dart/templates/skills/tests.md.mustache
brianquinlan: This is the name of the skill, not the name of the package. This naming convention is described at https://pub.dev/packages/skills

jakemac53: The skills package has actually changed to allow hyphens to be used, just for the purpose of spec compliance. But, it will only work in version 1.0.0 of the package.

brianquinlan: The examples use hyphens. My understanding was that @natebosch thought that the package name contained a hyphen but it is just the skill name.

jakemac53: My understanding is that `Codec.PackageName` might contain underscores, which should technically be replaced by hyphens.

brianquinlan: OK, I totally misunderstood. The idea is that we should replace "_" with "-" in `{{Codec.PackageName}}`.

I can do that.

Issue #6657, opened 2026-07-07
ruby: support clean functionality


JoeWang1127:
Add clean support for Ruby, this function should remove all generated files in a Ruby library.

See https://github.com/googleapis/librarian/blob/939ae6f16e82cadcdd09aa111d7f71a0e81062da/internal/librarian/generate.go#L149

Reference Go's implementation. The clean function must not delete files in the keep list in librarian.yaml.

PR #6943, opened 2026-07-21
feat(internal/librarian/ruby): implement clean functionality

suztomo:
Implements Ruby library cleaning using a targeted approach inspired by Golang's implementation (internal/librarian/golang/clean.go).

Removes known generated root files (README.md, AUTHENTICATION.md, Rakefile, Gemfile, Gemfile.lock, *.gemspec, gapic_metadata.json, .rubocop.yml, .yardopts, .toys.rb, etc.) and generated subdirectories (lib, proto_docs, snippets, test) while leaving unlisted root files (such as CHANGELOG.md and CHANGES.md) and paths specified in the keep list untouched.

For https://github.com/googleapis/librarian/issues/6657

PR #6999, opened 2026-07-23
feat(internal/librarian/ruby): call clean function during generation

suztomo:
Calls `ruby.Clean(library)` at the start of `ruby.Generate()` to clean pre-existing generated files before regenerating a Ruby client library.

### Test Case
In `TestGenerate`, pre-existing files are created in the target output directory to verify cleaning behavior:
- `AUTHENTICATION.md`: A generated root file which is expected to be cleaned up when `Generate()` runs.
- `CHANGELOG.md`: A non-generated root file which should not be regenerated and must be preserved across generation.

For https://github.com/googleapis/librarian/issues/6657

### DIFF
--- a/internal/librarian/ruby/generate.go
+++ b/internal/librarian/ruby/generate.go
@@ -38,6 +38,9 @@ func Generate(ctx context.Context, cfg *config.Config, library *config.Library, 
 	if len(library.APIs) == 0 {
 		return errNoAPIs
 	}
+	if err := Clean(library); err != nil {
+		return fmt.Errorf("failed to clean output directory: %w", err)
+	}
 	outDir, err := filepath.Abs(library.Output)
 	if err != nil {
 		return fmt.Errorf("failed to get absolute path of output directory: %w", err)
--- a/internal/librarian/ruby/generate_test.go
+++ b/internal/librarian/ruby/generate_test.go
@@ -324,6 +324,16 @@ func TestGenerate(t *testing.T) {
 		t.Fatal(err)
 	}
 	outDir := t.TempDir()
+	// AUTHENTICATION.md is a generated file.
+	oldFile := filepath.Join(outDir, "AUTHENTICATION.md")
+	if err := os.WriteFile(oldFile, []byte("old"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+	// CHANGELOG.md should not be regenerated.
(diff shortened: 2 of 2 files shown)


internal/librarian/ruby/generate.go
```diff
@@ -38,6 +38,9 @@ func Generate(ctx context.Context, cfg *config.Config, library *config.Library,
 	if len(library.APIs) == 0 {
 		return errNoAPIs
 	}
+	if err := Clean(library); err != nil {
+		return fmt.Errorf("failed to clean output directory: %w", err)
+	}
```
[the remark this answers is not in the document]
suztomo: I like that idea.

internal/librarian/ruby/generate.go
```diff
@@ -67,6 +67,9 @@ func Generate(ctx context.Context, cfg *config.Config, library *config.Library,
 			return fmt.Errorf("api %q: %w", api.Path, err)
 		}
 	}
+	if err := Clean(library); err != nil {
```
JoeWang1127: I think `Clean` function should be called before `Generate`: https://github.com/JoeWang1127/librarian/blob/main/internal/librarian/generate.go#L150

suztomo: While that's more intuitive, deleting files after successful generation is less destructive upon a generation failure. Quoting below from gemini-code-assist and I agree.

> To prevent destructive data loss on generation failure, consider calling Clean(library) only after the generation has successfully completed, right before filesystem.MoveAndMerge(tempDir, outDir).

> Since the generator writes all files to a temporary directory first, keeping the existing output directory intact until we are sure the new files are successfully generated makes the process much more robust. If generateAPI fails (e.g., due to a syntax error in a proto file or a missing tool), the user's existing generated files will remain completely untouched.

JoeWang1127: Why we can't use `git restore .` to restore the removed file after a failed generation?

If you prefer clean after generation success, could you make the function unexported, i.e., `clean` to avoid calling it from other package.

suztomo:
Closing as PR #6943 already registered ruby.Clean(library) in cleanLibraries() in internal/librarian/generate.go.

internal/librarian/ruby/generate.go
suztomo: I forgot about clean already being called from internal/librarian/ruby/generate.go. Closed this pull request.

Issue #6508, opened 2026-06-18
python: move tools install list to google-cloud-python/librarian.yaml


noahdietz:
As described in #5158, we need to move the "hardcoded" list of tools to install from `internal/librarian/python/librarian.yaml` to the `google-cloud-python/librarian.yaml` https://github.com/googleapis/librarian/blob/2bdb3b5f262bad05ce2a569828f06b9445ab78bd/internal/librarian/python/librarian.yaml#L15-L21

This will impact our Docker image building workflow for Python: https://github.com/googleapis/librarian/blob/2bdb3b5f262bad05ce2a569828f06b9445ab78bd/cmd/librarian/Dockerfile#L192-L194

cc @hj690

parthea:
Since the goal here (and in #5158) is to make the language repo the source of truth for its toolchain, could we take this opportunity to stop pulling gapic-generator from PyPI?

Instead, could we configure the toolchain to use the local `packages/gapic-generator` directory directly?

The huge benefit of using the local path is that when we make changes to the generator, we can roll out those changes to all generated libraries in the exact same PR. I believe this will lead to much smoother integrations, as the onus of updating the generated code lands directly on the author of the generator change rather than requiring a staggered rollout with multiple PRs.

cc @zhumin8

JoeWang1127:
Yes, this is actually my preference.

cc: @hj690 as hongda will look into how to implement this.

zhumin8:
> Instead, could we configure the toolchain to use the local `packages/gapic-generator` directory directly?

Definitely doable,  in fact, Java does that today.
https://github.com/googleapis/google-cloud-java/blob/acff1f2d620484c6f62fab6007d8ef1818772336/librarian.yaml#L40
IMO, it should be Python team's choice to develop at a released generator version or at HEAD. 
Note that developing with local path means that each generator change PR needs to come together with the changes of the generated code (at current stage means Python developer manually run librarian generate on their changes). It might be a bit harder to review the generator change PR with all the changes too.

daniel-sanche:
This would be a great improvement. I opened a bug in our list for next semester planning

JoeWang1127:
In order to use protoc defined in librarian.yaml, we need to update python package to find the protoc installation location.

After implementing this feature, adding protoc configuration in librarian.yaml in google-cloud-python, then removing embed librarian.yaml in librarian.

hj690:
Un-assign myself as I'm not working on this atm.

[...]

PR #7019, opened 2026-07-24
feat(python): read python tools from librarian.yaml config

hj690:
First attempt to read Python tools from librarian.yaml if provided in the repository configuration. Falls back to embedded librarian.yaml if no tools are specified in librarian.yaml. 

For #6508

### DIFF
--- a/internal/librarian/python/install_test.go
+++ b/internal/librarian/python/install_test.go
@@ -28,13 +28,11 @@ func TestInstall(t *testing.T) {
 		t.Fatal(err)
 	}
 	t.Setenv("PATH", bin+string(os.PathListSeparator)+os.Getenv("PATH"))
-
 	t.Run("fallback to embedded librarian.yaml", func(t *testing.T) {
 		if err := Install(t.Context(), nil); err != nil {
 			t.Fatal(err)
 		}
 	})
-
 	t.Run("use tools from config", func(t *testing.T) {
 		tools := &config.Tools{
 			Pip: []*config.PipTool{
@@ -46,4 +44,3 @@ func TestInstall(t *testing.T) {
 		}
 	})
 }
-
--- a/internal/librarian/python/install_test.go
+++ b/internal/librarian/python/install_test.go
@@ -15,11 +15,13 @@
 package python
 
 import (
+	"errors"
 	"os"
 	"path/filepath"
 	"testing"
 
 	"github.com/googleapis/librarian/internal/config"
+	"github.com/googleapis/librarian/internal/tool/pip"
 )
 
 func TestInstall(t *testing.T) {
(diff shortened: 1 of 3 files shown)


internal/librarian/python/install_test.go
```diff
@@ -26,8 +28,19 @@ func TestInstall(t *testing.T) {
 		t.Fatal(err)
 	}
 	t.Setenv("PATH", bin+string(os.PathListSeparator)+os.Getenv("PATH"))
-
-	if err := Install(t.Context()); err != nil {
-		t.Fatal(err)
-	}
+	t.Run("fallback to embedded librarian.yaml", func(t *testing.T) {
```
JoeWang1127: Could you convert this test to table driven test: https://github.com/googleapis/librarian/blob/main/doc/howwewritego.md#table-driven-tests

hj690: ah yup, updated.

internal/librarian/python/install_test.go
```diff
@@ -27,7 +29,30 @@ func TestInstall(t *testing.T) {
...
+		{
+			name: "use tools from config",
+			tools: &config.Tools{
+				Pip: []*config.PipTool{
+					{Name: "ruff", Version: "0.14.14"},
+				},
+			},
+		},
+	}
+	for _, tc := range tests {
```
JoeWang1127:
```suggestion
	for _, test := range tests {
```

internal/librarian/python/install_test.go
```diff
@@ -27,7 +29,30 @@ func TestInstall(t *testing.T) {
 	}
 	t.Setenv("PATH", bin+string(os.PathListSeparator)+os.Getenv("PATH"))
 
-	if err := Install(t.Context()); err != nil {
-		t.Fatal(err)
+	tests := []struct {
```
JoeWang1127: Could you use anonymous struct: [anonymous](https://github.com/googleapis/librarian/blob/main/doc/howwewritego.md#table-driven-tests)

internal/librarian/python/install_test.go
```diff
@@ -27,7 +29,30 @@ func TestInstall(t *testing.T) {
...
+		name    string
+		tools   *config.Tools
+		wantErr bool
```
JoeWang1127: Do you expect error in this test? If not, could you remove this field, see https://github.com/googleapis/librarian/blob/main/doc/howwewritego.md#separate-error-tests

internal/librarian/python/install_test.go
hj690: updated 👍

internal/librarian/python/install_test.go
hj690: good call, no I don't expect error in this test, and removed this field now.

JoeWang1127:
Could you fix the unit tests?

internal/librarian/python/install_test.go
```diff
@@ -27,7 +29,27 @@ func TestInstall(t *testing.T) {
...
+			name: "use tools from config",
+			tools: &config.Tools{
+				Pip: []*config.PipTool{
+					{Name: "ruff", Version: "0.14.14"},
+				},
+			},
+		},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			if err := Install(t.Context(), test.tools); err != nil {
```
JoeWang1127, line 52: I think we should check the error but we can do it in follow ups.

JoeWang1127 (approved):

internal/librarian/python/install_test.go
hj690: make sense, added negative test cases in [1af3144](https://github.com/googleapis/librarian/pull/7019/commits/1af3144f09280d9e8dca259155eef1f25eb6bd06)

PR #7026, opened 2026-07-24
fix(internal/librarian/ruby): pass client library version to snippet metadata

### LINKED ISSUE #7020: ruby: pass client library version to snippet metadata generator
### Bug Report: Snippet Metadata Client Library Version Reset to Empty String (`""`)

During `librarian generate`, the `"version"` field in generated snippet metadata (`snippets/snippet_metadata_*.json`) is reset to an empty string `""` (e.g. from `"1.9.0"` to `""`).

#### Affected File
`google-cloud-asset-v1/snippets/snippet_metadata_google.cloud.asset.v1.json`

```json
{
  client_library: {
    name: google-cloud-asset-v1,
-   version: 1.9.0,
+   version: ,
    language: RUBY,
```

#### Root Cause
In `internal/librarian/ruby/generate.go`, client library version information (from `library.Version` or version file) is not passed to `gapic-generator-cloud` or updated in the snippet metadata post-processing step (similar to `snippetmetadata.UpdateAllLibraryVersions` in Go / Node.js).

#### How OwlBot avoided this problem
In `google-cloud-ruby`, OwlBot post-processing ran via `owlbot-postprocessor` from [`googleapis/ruby-common-tools`](https://github.com/googleapis/ruby-common-tools).
The post-processor automatically installed the default modifier [`preserve_snippet_metadata_release_versions`](https://github.com/googleapis/ruby-common-tools/blob/main/owlbot-postprocessor/lib/owlbot/common_modifiers.rb#L107-L117):

```ruby
def preserve_snippet_metadata_release_versions name: nil
  path = [%r{^snippets/snippet_metadata_[\w.]+\.json$}]
  name ||= "preserve_snippet_metadata_release_versions"
  modifier path: path, name: name do |src, dest|
    if src && dest
      match = /"version": "(\d+\.\d+\.\d+)"/.match dest
      src = src.sub(/"version": "(\d+\.\d+\.\d+)?"/, "\"version\": \"#{match[1]}\"") if match
    end
    src
  end
end
```
During post-processing, it extracted the existing release version from `dest` (which Release Please updated on releases) and substituted it back into `src`.

In `librarian`, generation runs via Go (`internal/librarian/ruby/generate.go`) and moves generated files directly into `outDir` without running `owlbot-postprocessor` modifiers, causing the version to be lost unless `snippetmetadata.UpdateAllLibraryVersions(outDir, library.Version)` is called.

#### Sub-issue
Sub-issue under https://github.com/googleapis/librarian/issues/6633.

suztomo:
During librarian generate, snippet metadata files in Ruby libraries were leaving version set to empty string because gapic-generator-ruby outputs version as empty string. Call snippetmetadata.UpdateAllLibraryVersions after generation to set the library version.

Additionally, export the `SnippetMetadata` and `ClientLibrary` struct types in package `snippetmetadata` for typed unmarshaling and test assertions across the codebase.

The exported `snippetmetadata.SnippetMetadata` struct can also benefit existing test fixtures and assertions in:
- `internal/librarian/golang/bump_test.go` (snippet metadata test fixtures and version assertions)
- `internal/librarian/python/bump_test.go` (Python snippet metadata version bump test assertions)
- `internal/librarian/nodejs/generate_test.go` (Node.js generated snippet metadata test assertions)

The changes will be in an upcoming pull request.

Fixes https://github.com/googleapis/librarian/issues/7020

### DIFF
--- a/internal/librarian/ruby/generate.go
+++ b/internal/librarian/ruby/generate.go
@@ -27,6 +27,7 @@ import (
 	"github.com/googleapis/librarian/internal/config"
 	"github.com/googleapis/librarian/internal/filesystem"
 	"github.com/googleapis/librarian/internal/serviceconfig"
+	"github.com/googleapis/librarian/internal/snippetmetadata"
 	"github.com/googleapis/librarian/internal/sources"
 	"github.com/googleapis/librarian/internal/tool/protoc"
 )
@@ -79,6 +80,11 @@ func Generate(ctx context.Context, cfg *config.Config, library *config.Library, 
 	}
 	if err := filesystem.MoveAndMergeWithKeep(tempDir, outDir, outDir, keepFunc); err != nil {
 		return fmt.Errorf("failed to move generated files: %w", err)
+	}
+	if library.Version != "" {
+		if err := snippetmetadata.UpdateAllLibraryVersions(outDir, library.Version); err != nil {
+			return fmt.Errorf("failed to update snippet metadata versions: %w", err)
+		}
 	}
 	return nil
 }
--- a/internal/librarian/ruby/generate.go
+++ b/internal/librarian/ruby/generate.go
@@ -81,10 +81,8 @@ func Generate(ctx context.Context, cfg *config.Config, library *config.Library, 
 	if err := filesystem.MoveAndMergeWithKeep(tempDir, outDir, outDir, keepFunc); err != nil {
 		return fmt.Errorf("failed to move generated files: %w", err)
 	}
-	if library.Version != "" {
-		if err := snippetmetadata.UpdateAllLibraryVersions(outDir, library.Version); err != nil {
-			return fmt.Errorf("failed to update snippet metadata versions: %w", err)
-		}
+	if err := snippetmetadata.UpdateAllLibraryVersions(outDir, library.Version); err != nil {
+		return fmt.Errorf("failed to update snippet metadata versions: %w", err)
 	}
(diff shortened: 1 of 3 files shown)


internal/librarian/ruby/generate.go
```diff
@@ -80,6 +81,11 @@ func Generate(ctx context.Context, cfg *config.Config, library *config.Library,
 	if err := filesystem.MoveAndMergeWithKeep(tempDir, outDir, outDir, keepFunc); err != nil {
 		return fmt.Errorf("failed to move generated files: %w", err)
 	}
+	if library.Version != "" {
```
JoeWang1127: Is there a use case that library.Version is empty?

JoeWang1127 (approved):

internal/librarian/ruby/generate.go
suztomo: I saw `librarian add` didn't create a version field somewhere but it's confusing. Let's not create such case in Ruby.

noahdietz: > I saw `librarian add` didn't create a version field somewhere but it's confusing. Let's not create such case in Ruby.

librarian add should set the initial version (we do for all other languages), even if its 0.0.0 to let release-please bump it to 0.1.0 when the onboarding `feat` commit lands.

suztomo: Done.

PR #7036, opened 2026-07-24
fix(internal/librarian/php): optimize generated file marker detection

### LINKED ISSUE #7017: php(generate): clean to only read starting lines to find marker
currently, in internal/librarian/php/clean.go, cleans marker by reading the whole file. 
In reality, the generated markers are either at the top or right below licence header. We should update to only ready top X lines and find the marker.

This is relevant if decision in [go/sdk-librarian-php](http://goto.google.com/sdk-librarian-php) is to keep marker based cleaning.

zhumin8:
Marker detection in PHP files is optimized by only reading the first 50 lines of each file instead of loading the entire file content into memory. This reduces I/O and memory usage when cleaning up generated files in large repositories.

In reality from google-cloud-php, markers always appear line 2 for Protobuf files and line 19 for GAPIC files. A limit of 50 lines provides a safe margin.

Fixes #7017

### DIFF
--- a/internal/librarian/php/clean.go
+++ b/internal/librarian/php/clean.go
@@ -30,6 +30,9 @@ import (
 const (
 	gapicMetadataFile = "gapic_metadata.json"
 	phpExtension      = ".php"
+	// Limit reading to 50 lines. In google-cloud-php, markers are found early:
+	// line 2 for Protobuf files and line 19 for GAPIC files.
+	maxLinesToCheck = 50
 )
 
 var (
@@ -99,7 +102,7 @@ func cleanFiles(lib *config.Library, keepSet map[string]bool, subDir string) err
 		if !strings.HasSuffix(d.Name(), phpExtension) {
 			return nil
 		}
-		hasMarker, err := fileHasMarker(path, 50)
+		hasMarker, err := fileHasMarker(path)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


internal/librarian/php/clean.go
```diff
@@ -98,13 +99,36 @@ func cleanFiles(lib *config.Library, keepSet map[string]bool, subDir string) err
 		if !strings.HasSuffix(d.Name(), phpExtension) {
 			return nil
 		}
-		content, err := os.ReadFile(path)
+		hasMarker, err := fileHasMarker(path, 50)
```
noahdietz: Nit: please put `50` in a `const` witha comment. I have beef with magic numbers :)

internal/librarian/php/clean.go
```diff
@@ -98,13 +99,36 @@ func cleanFiles(lib *config.Library, keepSet map[string]bool, subDir string) err
...
+		if hasMarker {
 			return os.Remove(path)
 		}
 		return nil
 	})
 }
+
+// fileHasMarker checks if the file at path contains either gapicMarker or
+// protobufMarker within its first maxLines lines.
+func fileHasMarker(path string, maxLines int) (bool, error) {
```
noahdietz: Any reason for `maxLines` to be configurable? Do we have files where we know the marker is beyond 50? If not, i'd just remove the param and use the `const` in the code directly.

noahdietz (approved):
couple of nits but lgtm

PR #7038, opened 2026-07-24
fix(internal/librarian/php): remove php fallback generator installation

sofisl:
This PR removes the fallback hardocded generator installation for PHP in lieu of the new generator installation command.

For #6630

### DIFF
--- a/.github/actions/setup-librarian/action.yaml
+++ b/.github/actions/setup-librarian/action.yaml
@@ -33,11 +33,4 @@ runs:
         go install tool
         go install ./cmd/librarian
       shell: bash
-    - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
-      with:
-        node-version: "22"
-    - name: "Set up pnpm via corepack"
-      run: |
-        corepack enable pnpm
-        corepack prepare pnpm@7.32.2 --activate
-      shell: bash
+
--- a/.github/actions/setup-librarian/action.yaml
+++ b/.github/actions/setup-librarian/action.yaml
@@ -33,4 +33,3 @@ runs:
         go install tool
         go install ./cmd/librarian
       shell: bash
-
--- a/.github/actions/setup-librarian/action.yaml
+++ b/.github/actions/setup-librarian/action.yaml
@@ -33,3 +33,11 @@ runs:
         go install tool
         go install ./cmd/librarian
       shell: bash
+    - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
+      with:
+        node-version: "22"
+    - name: "Set up pnpm via corepack"
+      run: |
+        corepack enable pnpm
+        corepack prepare pnpm@7.32.2 --activate
+      shell: bash
--- a/internal/librarian/php/generate_test.go
+++ b/internal/librarian/php/generate_test.go
@@ -88,10 +88,7 @@ func requirePHPGenerator(t *testing.T) {
 	testhelper.RequireCommand(t, "protoc")
 	testhelper.RequireCommand(t, "python3")
 	testhelper.RequireCommand(t, "php")
-	bin, err := binDir()
-	if err != nil {
-		t.Fatalf("failed to locate bin dir: %v", err)
-	}
+	bin, _ := binDir()
 	wrapperPath := filepath.Join(bin, "gapic-generator-php")
 	if _, err := os.Stat(wrapperPath); err != nil {
 		t.Skip("skipping test: PHP generator is not installed (run 'librarian install php' first)")
--- a/internal/librarian/php/generate_test.go
+++ b/internal/librarian/php/generate_test.go
@@ -88,7 +88,10 @@ func requirePHPGenerator(t *testing.T) {
 	testhelper.RequireCommand(t, "protoc")
 	testhelper.RequireCommand(t, "python3")
 	testhelper.RequireCommand(t, "php")
-	bin, _ := binDir()
+	bin, err := binDir()
+	if err != nil {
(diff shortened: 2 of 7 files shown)


.github/actions/setup-librarian/action.yaml
```diff
@@ -33,3 +33,11 @@ runs:
         go install tool
         go install ./cmd/librarian
       shell: bash
+    - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
+      with:
+        node-version: "22"
+    - name: "Set up pnpm via corepack"
```
zhumin8: Why do we need to setup pnpm in this action? IIUC, this should be strictly setting up librarian and Go.

noahdietz: For clarity: This is moved into setup-librarian bc now both node and php need it, rather than just node?

noahdietz (dismissed):

internal/librarian/php/generate_test.go
```diff
@@ -88,11 +88,11 @@ func requirePHPGenerator(t *testing.T) {
 	testhelper.RequireCommand(t, "protoc")
 	testhelper.RequireCommand(t, "python3")
 	testhelper.RequireCommand(t, "php")
-	genDir, err := generatorDir(t.Context())
+	bin, err := binDir()
 	if err != nil {
-		t.Skipf("skipping test: failed to locate PHP generator: %v", err)
+		t.Fatalf("failed to locate bin dir: %v", err)
```
JoeWang1127:
```suggestion
		t.Fatal(err)
```

internal/librarian/php/install.go
```diff
@@ -37,25 +37,23 @@ const (
...
 
 // Install installs the PHP generator tool dependencies.
 func Install(ctx context.Context, tools *config.Tools) error {
-	if tools == nil || (len(tools.Composer) == 0 && len(tools.Pip) == 0) {
-		_, err := installGenerator(ctx)
-		return err
+	if tools == nil {
+		return errMissingTools
+	}
+	if len(tools.Composer) == 0 {
```
zhumin8, line 52: only Composer is checked and return early if not found. Let's also check for pip as it is also required right now.

.github/actions/setup-librarian/action.yaml
noahdietz: let's use Min's comment

noahdietz: My guess is bc now multiple languages need it and we want them to be using the same version. 

We also install protoc in this action a few lines above.

zhumin8: My point is we want to keep stuff that needed by every language here: Go, protoc, librarian. We don't setup python here (used by at least Python, Java and PHP) 
`pnpm` is needed by more than one language, but including here would also install it for languages that don't use it?

noahdietz: Yeah that's reasonable. @sofisl let's keep this in node and php setup steps please :)

sofisl: I see that reasoning about not installing dependencies unnecessarily, but I do see some value in sharing this code so that Node and PHP do not diverge, since currently we give guidance to follow the same process as Node for installation in the PHP Generator Developer Guide.

What if we were to include some common installation script (either in action.yaml or somewhere else) that has the steps to install common deps (like pip, node, etc.) and set that boolean as true in each testing script, like `include_pnpm: true`?

internal/librarian/php/install.go
sofisl: Also did it for pnpm since technically we require it as well.

.github/actions/setup-librarian/action.yaml
sofisl: Actually, let me know what you think about this, but I guess regardless it would need to be in a separate PR. If you're OK with this I'll create an issue to do a refactor.

JoeWang1127:
Could you fix the linter check?

internal/librarian/php/install.go
```diff
@@ -37,25 +37,31 @@ const (
...
+	errMissingPip      = errors.New("tools.pip configuration is missing")
+	errMissingPNPM     = errors.New("tools.pnpm configuration is missing")
 )
 
 // Install installs the PHP generator tool dependencies.
 func Install(ctx context.Context, tools *config.Tools) error {
-	if tools == nil || (len(tools.Composer) == 0 && len(tools.Pip) == 0) {
-		_, err := installGenerator(ctx)
-		return err
+	if tools == nil {
```
JoeWang1127, line 49: nit: group all checks in a separate function.

JoeWang1127 (approved):

.github/actions/setup-librarian/action.yaml
```diff
@@ -33,3 +33,4 @@ runs:
         go install tool
         go install ./cmd/librarian
       shell: bash
+
```
noahdietz: nit remove this diff plzz

internal/librarian/php/install.go
sofisl: Hm, I might leave it separate just for it to be more readable!

noahdietz: Might be personal preference, but I actually think its more readable to have a "chain" of `if` preconditions as they are. Putting it in a function makes the code less readable -> you have to jump to another function just to see what must be true for the code that follows to execute. An example of how this disjointedness would manifest is that we'd likely want a separate function to cover the lines for this `func checkTools`. I'd rather see an `TestInstall_Errors` :)

Maybe this was too long of a comment for a Nit :P

noahdietz (approved):

Issue #6229, opened 2026-05-31
sidekick: consider a different default for the swift package name


coryan:
The default package name is derived from the protobuf package name, that produces unsatisfactory results when the proto package name includes compound words, for example:

- `google.cloud.secretmanager.v1` becomes `GoogleCloudSecretmanagerV1`. `GoogleCloudSecretManagerV1` (uppercase `M`) is better
-  `google.cloud.resourcemanager.v3` becomes `GoogleCloudResourcemanagerV3`. `GoogleCloudResourceManagerV3` (uppercase `M`) is better

It seems that most services have a component in their package name that is formed by mixing multiple words.  Combined with the `PascalCase` naming in Swift, this produces sad results.

We could manually override these names, but that is very tedious and error-prone.  Instead, we could change the default to use the `csharp_namespace`:

https://github.com/googleapis/googleapis/blob/0054cdcbd8ea44298f329d916d8173dd736fbaaa/google/cloud/secretmanager/v1/service.proto#L29

or `ruby_package` options to create better names:

https://github.com/googleapis/googleapis/blob/0054cdcbd8ea44298f329d916d8173dd736fbaaa/google/cloud/secretmanager/v1/service.proto#L36

We have to make a decision and document it. If we change the default name we should make that change before any public previews.

coryan:
After some internal discussion, we decided to use the PHP, Ruby, and C# annotations to create better module names.

coryan:
Some special cases:

```
too many names: ["Google.Cloud.AlloyDB.V1" "Google.Cloud.AlloyDb.V1"]
too many names: ["Google.Cloud.Aiplatform.V1" "Google.Cloud.AIPlatform.V1"]
too many names: ["Google.Cloud.Bigquery.AnalyticsHub.V1" "Google.Cloud.BigQuery.AnalyticsHub.V1"]
too many names: ["Google.Cloud.Bigquery.DataPolicies.V1" "Google.Cloud.BigQuery.DataPolicies.V1"]
too many names: ["Google.Cloud.BigQuery.DataPolicies.V2" "Google.Cloud.Bigquery.DataPolicies.V2"]
too many names: ["Google.Cloud.BigQuery.DataTransfer.V1" "Google.Cloud.Bigquery.DataTransfer.V1"]
too many names: ["Google.Cloud.Config.V1" "Google.Cloud.ConfigService.V1"]
[...]

PR #7047, opened 2026-07-25
fix(sidekick/swift): modules and library name

coryan:
I made a mistake when introducing library names: modules cannot have library names, so they should not try to set the value.

Towards #6229


internal/sidekick/api/test.go
```diff
@@ -76,6 +76,7 @@ func NewTestAPI(messages []*Message, enums []*Enum, services []*Service) *API {
 	return model
 }
 
+// WithPackageName changes the package name of an API instance.
```
dbolduc, line 79: thanks

dbolduc (approved):

PR #7049, opened 2026-07-25
fix(sidekick/rust): support oneof fields in binding substitutions

### LINKED ISSUE #7045: sidekick: rust support for `oneof` fields as path parameters
from https://github.com/googleapis/google-cloud-rust/pull/6167#top

----
> > Docs error is interesting:
> 
> In exciting ways ... That is not a doc error, it is an error in the code! Several other builds caught the same thing, so at least we have that going for us...
> 
> > ```
> > error[E0615]: attempted to take value of method `address_query` on type `&model::GeocodeAddressRequest`
> > --
> >   | --> src/generated/maps/geocode/v4/src/transport.rs:212:47
> >   | \|
> >   | 212 \|                         Some(&req).map(\|m\| &m.address_query).map(\|s\| s.as_str()),
> >   | \|                                               ^^^^^^^^^^^^^ method, not a field
> >   | \|
> >   | help: use parentheses to call the method
> > ```
> > 
> >   
> > in `src/generated/maps/geocode/v4/src/transport.rs:286:43`
> 
> `address_query` appears in the binding:
> 
> https://github.com/googleapis/googleapis/blob/09bc253ea6b336c1a214487684f4eb303bd7fc62/google/maps/geocode/v4/geocode_service.proto#L48-L53
> 
> Yikes, it is a oneof field:
> 
> https://github.com/googleapis/googleapis/blob/09bc253ea6b336c1a214487684f4eb303bd7fc62/google/maps/geocode/v4/geocode_service.proto#L91-L97
> 
> Sounds to me like sidekick does not handle these for Rust. Please open a bug...

Basically sidekick generates bad Rust code for `oneof` path parameters.

dbolduc:
Support oneof fields in a binding substitution.

The accessor for them returns on `Option<&T>`, so we need to chain that.

Here is the diff in `google-maps-geocode-v4`, which now builds clean:

https://github.com/dbolduc/google-cloud-rust/commit/da7ab7f99434cd8938bb530159ac8bc232ec99ba

cc: @noahdietz 

## Testing

The unit testing is fine. We at least verify that different code is produced for a oneof. But the real verification is that the code compiles.

I think ideally, we would want a gapic-showcase request with a oneof field to verify the code not only compiles, but is correct, but eh, I am too lazy for that.

Fixes #7045

### DIFF
--- a/internal/sidekick/api/test.go
+++ b/internal/sidekick/api/test.go
@@ -76,6 +76,7 @@ func NewTestAPI(messages []*Message, enums []*Enum, services []*Service) *API {
 	return model
 }
 
+// WithPackageName changes the package name of an API instance.
 func (a *API) WithPackageName(name string) *API {
 	a.PackageName = name
 	return a
--- a/internal/sidekick/rust/annotate_method.go
+++ b/internal/sidekick/rust/annotate_method.go
@@ -406,7 +406,9 @@ func makeAccessors(fields []string, m *api.Method) ([]string, error) {
 		if field == nil {
 			return nil, fmt.Errorf("invalid routing/path field (%q) for request message %s", rustFieldName, message.ID)
 		}
-		if field.Optional {
+		if field.IsOneOf {
+			accessors = append(accessors, fmt.Sprintf(".and_then(|m| m.%s())", rustFieldName))
+		} else if field.Optional {
 			accessors = append(accessors, fmt.Sprintf(".and_then(|m| m.%s.as_ref())", rustFieldName))
 		} else {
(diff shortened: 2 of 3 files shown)


coryan (approved):

coryan:
> The unit testing is fine. We at least verify that different code is produced for a oneof. But the real verification is that the code compiles.

In Swift we write a tests like this:

https://github.com/googleapis/librarian/blob/c8a1e3e31a79c8e8121886110fe0b9f0c845002e/internal/sidekick/swift/generate_stub_test.go#L206-L212

Not quite a golden file test, but close to it.

PR #7050, opened 2026-07-25
refactor(sidekick/swift): configure codec using structs

coryan:
For historical reasons the `Swift` codec was receiving some configuration parameters via the `Codec` map, and some configuration parameters via structures in the `config.*` package.

With this change we stop using the `Codec` map. Less parsing and conversions this way.

A lot of the tests were setting the `copyright-year` parameter, but it was not needed, so I removed it too.

Towards #6229

### DIFF
--- a/internal/config/swift.go
+++ b/internal/config/swift.go
@@ -38,6 +38,9 @@ type SwiftPackage struct {
 	// Each module defines a source proto path, and output location.
 	Modules []*SwiftModule `yaml:"modules,omitempty"`
 
+	// PackageNameOverride overrides the package name.
+	PackageNameOverride string `yaml:"package_name_override,omitempty"`
+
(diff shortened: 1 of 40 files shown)


internal/config/swift.go
```diff
@@ -38,6 +38,9 @@ type SwiftPackage struct {
 	// Each module defines a source proto path, and output location.
 	Modules []*SwiftModule `yaml:"modules,omitempty"`
 
+	// PackageNameOverride overrides the package name.
+	PackageNameOverride string `yaml:"package_name_override,omitempty"`
```
nolanleastin, line 46: nit: why would you want to do this? maybe explain in comment

nolanleastin (approved):

internal/config/swift.go
coryan: Done.

Issue #6896, opened 2026-07-17
php(generate): derive/fill `staging_subdir` for regular cases


zhumin8:
Now with https://github.com/googleapis/librarian/pull/6869, I made `staging_subdir` as required and it is fully populated via migrate script. 
But in reality, most cases these are derivable from API path (version part of it). We may only need to explicitly specify the outliers in librarian.yaml

e.g., 
```
      - path: google/geo/type
        php:
          staging_subdir: .
```
```
      - path: google/identity/accesscontextmanager/type
        php:
          staging_subdir: type-protos
```

JoeWang1127:
Adding a (potentially) special case:
```
- path: google/bigtable/admin/v2
  php:
    staging_subdir: v2/Admin
```

PR #7069, opened 2026-07-27
feat(internal/librarian/php): fill default staging subdir

JoeWang1127:
Default configuration filling is implemented for PHP libraries, automatically deriving the staging subdirectory for each API when it is not explicitly specified in librarian.yaml.

This ensures PHP library configurations have consistent staging paths without requiring manual specification for every API.

For https://github.com/googleapis/librarian/issues/6896

### DIFF
--- a/internal/librarian/php/default.go
+++ b/internal/librarian/php/default.go
@@ -20,6 +20,7 @@ import (
 	"github.com/googleapis/librarian/internal/config"
 )
 
+// Fill populates default PHP configuration values for the library.
 func Fill(lib *config.Library) *config.Library {
 	fillStagingSubdir(lib)
 	return lib
@@ -27,11 +28,11 @@ func Fill(lib *config.Library) *config.Library {
 
 func fillStagingSubdir(lib *config.Library) {
 	if lib.PHP == nil {
-		return
+		lib.PHP = &config.PHPPackage{}
 	}
 	for _, api := range lib.APIs {
 		// Do not override existing library configuration.
-		if api.PHP == nil || api.PHP.StagingSubdir != "" {
+		if api.PHP.StagingSubdir != "" {
 			continue
 		}
 		name := filepath.Base(api.Path)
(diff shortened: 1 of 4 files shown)


internal/librarian/php/default.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/googleapis/librarian/internal/config"
+)
+
+// Fill populates default PHP configuration values for the library.
+func Fill(lib *config.Library) *config.Library {
+	fillStagingSubdir(lib)
+	return lib
+}
+
+func fillStagingSubdir(lib *config.Library) {
```
noahdietz: This should return the `*config.Library` to indicate that it has been modified: https://github.com/googleapis/librarian/blob/main/doc/howwewritego.md#make-mutations-explicit

internal/librarian/php/default.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"path/filepath"
+
+	"github.com/googleapis/librarian/internal/config"
+)
+
+// Fill populates default PHP configuration values for the library.
+func Fill(lib *config.Library) *config.Library {
+	fillStagingSubdir(lib)
+	return lib
```
noahdietz: With the other comment addressed, this can be `return fillStagingSubdir(lib)`

internal/librarian/php/default_test.go
```diff
@@ -0,0 +1,128 @@
...
+							StagingSubdir: "v3beta1",
+						},
+					},
+				},
+			},
+		},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			fillStagingSubdir(test.lib)
+			if diff := cmp.Diff(test.want, test.lib); diff != "" {
```
noahdietz: With the above return value change, let's make this `got := fillStagingSubdir(test.lib)` and `cmp.Diff(test.want, go)` for readability

noahdietz (approved):
couple of comments, but LGTM

internal/librarian/php/default.go
JoeWang1127: Done.

internal/librarian/php/default.go
JoeWang1127: Done.

internal/librarian/php/default_test.go
JoeWang1127: Done.

PR #7070, opened 2026-07-27
chore(tool/cmd/migrate): normalize staging subdir in php migration

### LINKED ISSUE #6896: php(generate): derive/fill `staging_subdir` for regular cases

JoeWang1127:
During PHP library migration, staging subdirectories are now normalized when creating API configurations. If the resolved staging subdirectory matches the base name of the API path, it is omitted from the resulting configuration.

This prevents redundant `staging_subdir` fields from being written to librarian.yaml when they can be automatically derived during code generation.

Fixes #6896


tool/cmd/migrate/php.go
```diff
@@ -151,10 +151,11 @@ func createAPIConfig(path string, dest string) (*config.API, error) {
 	if stagingSubdir == "" {
 		return nil, fmt.Errorf("%w: path %s from destination %q", errUnableToResolveStagingSubdir, path, dest)
```
JoeWang1127, line 152: I didn't remove this error because we still need an non-empty directory for the normalization step.

noahdietz (approved):

PR #7071, opened 2026-07-27
feat(rust): generate conversion traits for hybrid crates

suzmue:
Filter model messages and enums in filterModelToStreaming to include only types reachable from bidirectional streaming RPCs when generating convert.rs conversion traits.

Detect unsupported google.protobuf.Any fields in streaming reachability paths and return errors with actionable resolution instructions referencing skipped_ids in librarian.yaml.

For #6835


internal/librarian/rust/generate.go
```diff
@@ -112,19 +113,153 @@ func generateProstHybrid(ctx context.Context, model *api.API, library *config.Li
...
+	}
+
+	hybridModel := *model
+	hybridModel.Messages = language.FilterSlice(model.Messages, func(m *api.Message) bool {
+		return streamingMsgs[m.ID]
+	})
+	hybridModel.Enums = language.FilterSlice(model.Enums, func(e *api.Enum) bool {
+		return streamingEnums[e.ID]
+	})
+	return &hybridModel, nil
```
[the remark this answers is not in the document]
suzmue: Done.

internal/librarian/rust/generate.go
```diff
@@ -112,19 +113,189 @@ func generateProstHybrid(ctx context.Context, model *api.API, library *config.Li
...
 	hybridConfig.Codec["include-file"] = "includes.rs"
+	postProcess := fmt.Sprintf(`let name = format!("{destination}/includes.rs");
+let content = std::fs::read_to_string(&name).expect("error reading includes.rs");
+let content = content.replace("include!(\"%s.rs\");", "include!(\"%s.rs\");\n            include!(\"../convert.rs\");");
+std::fs::write(&name, content).expect("error writing includes.rs");`, model.PackageName, model.PackageName)
+	if existing, ok := hybridConfig.Codec["post-process-protos"]; ok && existing != "" {
+		hybridConfig.Codec["post-process-protos"] = existing + "\n" + postProcess
+	} else {
+		hybridConfig.Codec["post-process-protos"] = postProcess
+	}
```
coryan: Can we make this a mustache template? Or at least move the format string to a constant to make the code more readable?

internal/sidekick/rust/templates/convert-prost/oneof.mustache
coryan, line 1: Seems like this fix could happen by itself? And it will affect existing code (though probably none of the existing code uses oneofs and protobuf).

internal/librarian/rust/generate_test.go
```diff
@@ -848,6 +848,177 @@ func TestGenerateProstHybrid(t *testing.T) {
...
+				InputTypeID:         streamingMsg.ID,
+				OutputTypeID:        streamingMsg.ID,
+				InputType:           streamingMsg,
+				OutputType:          streamingMsg,
+				ClientSideStreaming: true,
+				ServerSideStreaming: true,
+			},
+		},
+	}
+	model := api.NewTestAPI([]*api.Message{streamingMsg, unusedMsg}, []*api.Enum{}, []*api.Service{bidiService})
```
coryan: There are some new-ish helpers to create `api.*` types.  They do a better job of keeping things like `InputTypeID` and `InputType` consistent:

https://github.com/googleapis/librarian/blob/cab06aa50fffdbd48790c499922517c23ec8f2ec/internal/sidekick/api/test.go#L210-L216

Consider using them instead of manually initializing the fields.

internal/librarian/rust/generate.go
```diff
@@ -112,19 +113,189 @@ func generateProstHybrid(ctx context.Context, model *api.API, library *config.Li
...
+
+	// Construct hybridModel for prost conversion code generation.
+	hybridModel := api.API{
+		Name:        model.Name,
+		PackageName: model.PackageName,
+		Title:       model.Title,
+		Description: model.Description,
+		Revision:    model.Revision,
+		Services:    model.Services,
```
coryan: Do you need all the services? What if an API has many services some of which have no bidi methods?

internal/librarian/rust/generate.go
```diff
@@ -112,19 +113,189 @@ func generateProstHybrid(ctx context.Context, model *api.API, library *config.Li
 		return nil
 	}
 
+	hybridModel, err := filterModelToStreaming(model)
+	if err != nil {
+		return fmt.Errorf("filtering model for streaming: %w", err)
```
coryan: I don't have strong opinions, but most of this project would just return `err` here.

internal/librarian/rust/generate.go
coryan: This file is getting long. Consider a `librarian/rust/generate_post_hybrid.go` file.

coryan (changes requested):
The only thing that gives me pause is the filtering of services without bidi methods. Maybe there is a good reason why we don't need to filter them, if so, just let me know.

I think the other comments are cosmetic / nits.

internal/sidekick/rust/templates/convert-prost/oneof.mustache
suzmue: It does affect existing code. I will split this out into a separate PR, wanted to get feedback if this is the route we should go or if we'd prefer to add the helpers. If you have thoughts let me know.

internal/librarian/rust/generate_test.go
suzmue: Done! Thanks for the pointer.

internal/librarian/rust/generate.go
suzmue: Done.

internal/librarian/rust/generate.go
suzmue: Done.

internal/librarian/rust/generate.go
suzmue: Done and added to test.

internal/librarian/rust/generate.go
suzmue: I moved it into the mustache template and combined with include-file since I am using  both just for this feature.

internal/sidekick/rust/templates/convert-prost/oneof.mustache
coryan: I think this is the right approach, the `from_*()` helpers will create a lot of extra code, which would mostly go unused.

internal/librarian/rust/generate_prost_hybrid_test.go
```diff
@@ -0,0 +1,220 @@
...
+)
+
+func TestGenerateProstHybrid(t *testing.T) {
+	testhelper.RequireCommand(t, "protoc")
+	testhelper.RequireCommand(t, "cargo")
+	msg := api.NewTestMessage("Request").WithPackage("google.cloud.test.v1")
+
+	chatMethod := api.NewTestMethod("Chat").WithInput(msg).WithOutput(msg)
+	chatMethod.ClientSideStreaming = true
+	chatMethod.ServerSideStreaming = true
```
coryan: nit: you could add a `WithBidiStreaming()` helper if you wanted to.

coryan (approved):
I wonder if you could change the point where `includes.rs` is used instead of changing the code itself.

internal/librarian/rust/generate_prost_hybrid_test.go
suzmue: Done.

suzmue:
> I wonder if you could change the point where `includes.rs` is used instead of changing the code itself.

Yeah this was the simplest / cleanest way that I found. Will look for a better way in the future.

internal/librarian/rust/generate_prost_hybrid_test.go
suzmue: done.

PR #7075, opened 2026-07-28
feat(sidekick/rust): add used_if: hybrid_services dependency resolution

suzmue:
Support used_if: hybrid_services in package_dependencies configuration to
allow conditionally enabling gaxi features for hybrid REST+gRPC crates.

When include_bidi_streaming_methods is enabled and a service contains
bidirectional streaming RPCs, packages configured with used_if: hybrid_services
are activated additively alongside used_if: services rules.

This will be used in librarian.yaml to enable the _internal-grpc-client
feature on google-cloud-gax-internal for hybrid services alongside the
standard _internal-http-client feature.

To support multiple package_dependencies rules for the same dependency name
without option map collisions, librarian joins definitions using a semicolon
delimiter in codec["package:<name>"]. Sidekick parses semicolon-delimited
package options into separate rules and merges all active features for the
package into a single Cargo.toml entry.

For #6835

### DIFF
--- a/internal/librarian/rust/codec.go
+++ b/internal/librarian/rust/codec.go
@@ -178,8 +178,16 @@ func newLibraryCodec(library *config.Library) map[string]string {
 		codec["package-name-override"] = library.Name
 	}
 	if library.Rust != nil {
+		depsByName := make(map[string][]string)
+		var names []string
 		for _, dep := range library.Rust.PackageDependencies {
-			codec["package:"+dep.Name] = formatPackageDependency(dep)
+			if _, seen := depsByName[dep.Name]; !seen {
+				names = append(names, dep.Name)
+			}
+			depsByName[dep.Name] = append(depsByName[dep.Name], formatPackageDependency(dep))
+		}
+		for _, name := range names {
+			codec["package:"+name] = strings.Join(depsByName[name], ";")
 		}
 		if len(library.Rust.DisabledRustdocWarnings) > 0 {
 			codec["disabled-rustdoc-warnings"] = strings.Join(library.Rust.DisabledRustdocWarnings, ",")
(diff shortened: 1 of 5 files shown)


dbolduc (approved):

suzmue:
There is a bug in this PR, converting to draft.

suzmue:
Hi @dbolduc, I've fixed a big issue with this PR not working, since multiple features are not supported in the librarian config. Instead of adding them, I decided to make the configuration additive, instead of adding a features plural definition because this makes the configuration simpler.

1. **Additive Triggers:** Rather than making `services` and `hybrid_services` mutually exclusive, they are now additive. A hybrid service activates both `used_if: services` and `used_if: hybrid_services` rules. This simplifies needing to repeat the async_trait, google-cloud-gax, tracing, etc dependency configs.
2. **Semicolon-Delimited Codec Options:** In `librarian`, multiple `package_dependencies` rules for the same dependency (e.g., `gaxi`) are joined with a `;` in `codec["package:<name>"]`. This prevents Go map key overwrites when passing options to `sidekick`. 

### Example Configuration (`librarian.yaml`)
```yaml
default:
  rust:
    package_dependencies:
      - name: gaxi
        package: google-cloud-gax-internal
        feature: _internal-http-client
        used_if: services
      - name: gaxi
        package: google-cloud-gax-internal
        feature: _internal-grpc-client
        used_if: hybrid_services
```

For hybrid crates, this generates:
```toml
gaxi                 = { workspace = true, features = ["_internal-http-client", "_internal-grpc-client"] }
```

suzmue:
Currently I am keeping the other dependency configurations changes that are made internally not in librarian.yaml. I think for the sake of consistency, I am going to add this to the features from within sidekick instead of making the change in librarian.yaml for now. 

Abandoning this PR.

PR #7078, opened 2026-07-28
feat(sidekick/rust): implement bidi streaming transport stubs

suzmue:
Implement bidirectional streaming transport stubs in generated Rust crates and add required workspace dependencies (`futures`, `http`, `tokio`, `tokio-stream`, and the `_internal-grpc-client` feature on `gaxi`) behind `#[cfg(google_cloud_unstable_gapic_streaming)]`.

Also add TODOs for follow-up work to propagate request serialization and stream setup errors to `RequestSender::send()`. This will require changes to the constructor and likely will move the proto conversion
to the RequestSender, if we can manage this while erasing the prost types.

For #6835

### DIFF
--- a/internal/sidekick/rust/annotate_model.go
+++ b/internal/sidekick/rust/annotate_model.go
@@ -236,6 +236,14 @@ func annotateModel(model *api.API, codec *codec) (*modelAnnotations, error) {
 			bidiStreamingServices = append(bidiStreamingServices, s)
 		}
 	}
+	hasBidiStreaming := codec.templateOverride == "" && codec.includeBidiStreamingMethods && slices.ContainsFunc(model.Services, (*api.Service).HasBidiStreaming)
+	if hasBidiStreaming {
+		for _, pkg := range codec.extraPackages {
+			if pkg.name == "gaxi" && !slices.Contains(pkg.features, "_internal-grpc-client") {
+				pkg.features = append(pkg.features, "_internal-grpc-client")
+			}
+		}
+	}
 
 	ann := &modelAnnotations{
 		PackageName:           codec.packageName(model),
@@ -245,7 +253,7 @@ func annotateModel(model *api.API, codec *codec) (*modelAnnotations, error) {
 		RequiredPackages:      requiredPackages(codec.extraPackages),
 		ExternPackages:        externPackages(codec.extraPackages),
 		HasLROs:               hasLROs,
-		HasBidiStreaming:      codec.templateOverride == "" && codec.includeBidiStreamingMethods && len(bidiStreamingServices) > 0,
+		HasBidiStreaming:      hasBidiStreaming,
 		BidiStreamingServices: bidiStreamingServices,
 		CopyrightYear:         codec.generationYear,
 		BoilerPlate: append(license.HeaderBulk(),
--- a/internal/sidekick/rust/annotate_model.go
+++ b/internal/sidekick/rust/annotate_model.go
@@ -30,6 +30,8 @@ import (
 // errQuickstartServiceNotFound is returned when the requested quickstart service override is not found.
 var errQuickstartServiceNotFound = errors.New("quickstart_service_override not found")
 
+const gaxiPackageName = "gaxi"
+
 type modelAnnotations struct {
 	PackageName           string
 	PackageVersion        string
@@ -239,7 +241,7 @@ func annotateModel(model *api.API, codec *codec) (*modelAnnotations, error) {
 	hasBidiStreaming := codec.templateOverride == "" && codec.includeBidiStreamingMethods && slices.ContainsFunc(model.Services, (*api.Service).HasBidiStreaming)
 	if hasBidiStreaming {
 		for _, pkg := range codec.extraPackages {
-			if pkg.name == "gaxi" && !slices.Contains(pkg.features, "_internal-grpc-client") {
+			if pkg.name == gaxiPackageName && !slices.Contains(pkg.features, "_internal-grpc-client") {
 				pkg.features = append(pkg.features, "_internal-grpc-client")
 			}
 		}
--- a/internal/sidekick/rust/annotate_model_test.go
+++ b/internal/sidekick/rust/annotate_model_test.go
@@ -16,6 +16,7 @@ package rust
 
 import (
 	"errors"
+	"slices"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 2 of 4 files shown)


suzmue:
https://github.com/googleapis/google-cloud-rust/pull/6191 has an integration test with showcase with this working.

internal/sidekick/rust/templates/crate/src/transport.rs.mustache
```diff
@@ -131,13 +131,80 @@ impl super::stub::{{Codec.Name}} for {{Codec.Name}} {
...
     ) -> (
         google_cloud_gax::streaming::RequestSender<{{InputType.Codec.QualifiedName}}>,
         google_cloud_gax::streaming::ResponseReceiver<{{OutputType.Codec.QualifiedName}}>,
     ) {
-        {{! TODO(#6835) - implement the body }}
-        unimplemented!()
+        use gaxi::prost::{FromProto, ToProto};
+        use futures::stream::StreamExt as _;
+
+        let (req_tx, mut req_rx) = tokio::sync::mpsc::channel::<{{InputType.Codec.QualifiedName}}>(100);
```
coryan: The magic 100 numbers is my fault, but makes me sad. Why do we know that 100 is enough, and not too much? Should this be configurable?

internal/sidekick/rust/annotate_model_test.go
```diff
@@ -604,6 +611,21 @@ func TestModelAnnotationsHasBidiStreaming(t *testing.T) {
...
+						gaxiPkg = pkg
+						break
+					}
+				}
+				if gaxiPkg == nil {
+					t.Fatalf("gaxi package not found in extraPackages")
+				}
+				if !slices.Equal(gaxiPkg.features, test.wantGaxiFeatures) {
+					t.Errorf("gaxi features = %v, want %v", gaxiPkg.features, test.wantGaxiFeatures)
+				}
```
coryan: I think the guidelines in the `howwewritego.md` file recommend:
 
```suggestion
				if diff := cmp.Diff( test.wantGaxiFeatures, gaxiPkg.features); diff != "" {
					t.Errorf("mismatch (-want +got):\n%s", diff)
				}
```

internal/sidekick/rust/annotate_model.go
```diff
@@ -229,6 +229,15 @@ func annotateModel(model *api.API, codec *codec) (*modelAnnotations, error) {
 		}
 	}
 
+	hasBidiStreaming := codec.templateOverride == "" && codec.includeBidiStreamingMethods && slices.ContainsFunc(model.Services, (*api.Service).HasBidiStreaming)
+	if hasBidiStreaming {
+		for _, pkg := range codec.extraPackages {
+			if pkg.name == "gaxi" && !slices.Contains(pkg.features, "_internal-grpc-client") {
```
coryan: Seems like we need a symbolic name or constant for `gaxi`

internal/sidekick/rust/annotate_model_test.go
```diff
@@ -604,6 +611,21 @@ func TestModelAnnotationsHasBidiStreaming(t *testing.T) {
...
+				var gaxiPkg *packagez
+				for _, pkg := range codec.extraPackages {
+					if pkg.name == "gaxi" {
+						gaxiPkg = pkg
+						break
+					}
+				}
+				if gaxiPkg == nil {
+					t.Fatalf("gaxi package not found in extraPackages")
+				}
```
coryan:
```suggestion
                 idx := slices.IndexFunc(codec.extraPackages, func(a *packagez) bool { return pkg.name == "gaxi" }))
                 if idx == -1 {
                     t.Fatalf("gaxi package not found in extraPackages")
                 }
                 gaxiPkg := codec.extraPackages[idx]
```

I mean, not as nice as Rust, but not as terrible as C...

coryan (changes requested):

internal/sidekick/rust/templates/crate/src/transport.rs.mustache
suzmue: Sigh I agree. I have added a TODO, I think probably it should be configurable.

I would like to proceed with this as it is right now because we have something that works (bar some todos) and I can add my integration test for showcase.

internal/sidekick/rust/templates/crate/src/transport.rs.mustache
```diff
@@ -131,13 +131,78 @@ impl super::stub::{{Codec.Name}} for {{Codec.Name}} {
...
+                .await;
+
+            match result {
+                Ok(response) => {
+                    let mut response_stream = response.into_inner();
+                    while let Some(res) = response_stream.next().await {
+                        let item = res
+                            .map_err(gaxi::grpc::from_status::to_gax_error)
+                            .and_then(|m| m.cnv().map_err(google_cloud_gax::error::Error::deser));
+                        if resp_tx.send(item).await.is_err() {
```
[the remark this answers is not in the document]
suzmue: Done.

internal/sidekick/rust/annotate_model.go
suzmue: Made it a constant. I had an alternative implementation where I added this in librarian.yaml so its less magic: https://github.com/googleapis/librarian/pull/7075

internal/sidekick/rust/annotate_model_test.go
suzmue: done.

internal/sidekick/rust/annotate_model_test.go
suzmue: done.

internal/sidekick/rust/annotate_model_test.go
```diff
@@ -604,6 +611,17 @@ func TestModelAnnotationsHasBidiStreaming(t *testing.T) {
...
 			}
+			if len(test.wantGaxiFeatures) > 0 {
+				idx := slices.IndexFunc(codec.extraPackages, func(pkg *packagez) bool {
+					return pkg.name == gaxiPackageName
+				})
+				if idx == -1 {
+					t.Fatalf("gaxi package not found in extraPackages")
+				}
+				if diff := cmp.Diff(test.wantGaxiFeatures, codec.extraPackages[idx].features); diff != "" {
+					t.Errorf("gaxi features mismatch (-want +got):\n%s", diff)
```
coryan: The guidelines as super strict:


```suggestion
					t.Errorf("mismatch (-want +got):\n%s", diff)
```

https://github.com/googleapis/librarian/blob/main/doc/howwewritego.md#use-cmpdiff-for-comparisons

coryan (approved):
A small nit, please fix before merging.

PR #7093, opened 2026-07-28
fix(sidekick/rust): fix prost oneof field conversion in templates

suzmue:
Update the prost conversion template for oneof fields to construct the model enum variant directly rather than invoking a setter method, which we don't generate universally.

For #6835


suzmue:
Here is the result of using this updated template on the existing crates: https://github.com/googleapis/google-cloud-rust/compare/main...suzmue:oneof-regen?expand=1

dbolduc (approved):

PR #7097, opened 2026-07-28
feat(rust): exclude non-streaming types from prost generation

suzmue:
Exclude non-streaming protobuf message and enum types from generated prost output files when hybrid streaming generation is enabled.

The filterModelToStreaming helper now collects all non-well-known-type IDs that are unreachable from bidirectional streaming RPCs. These unused type IDs are passed to the sidekick rust_prost generator and configured with prost_build extern_path mappings to crate::prost::unused::NonexistentType, preventing prost from emitting unused struct definitions and unused output files.

For #6835

### DIFF
--- a/internal/librarian/rust/generate_prost_hybrid.go
+++ b/internal/librarian/rust/generate_prost_hybrid.go
@@ -39,7 +39,7 @@ func generateProstHybrid(ctx context.Context, model *api.API, library *config.Li
 		return nil
 	}
 
-	hybridModel, err := filterModelToStreaming(model)
+	hybridModel, unusedTypes, err := filterModelToStreaming(model)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


internal/librarian/rust/generate_prost_hybrid.go
```diff
@@ -50,6 +50,9 @@ func generateProstHybrid(ctx context.Context, model *api.API, library *config.Li
 		hybridConfig.Codec = make(map[string]string)
 	}
 	hybridConfig.Codec["convert-include-package"] = model.PackageName
+	if len(unusedTypes) > 0 {
+		hybridConfig.Codec["unused-types"] = strings.Join(unusedTypes, "\n")
```
coryan, line 54: FWIW, in Swift we moved away from using the `Codec` map to pass parameters and use the `config.SwiftPackage` directly. It is a bit of a yak shaving, but you could ask Gemini to refactor the code in Rust to do the same....

coryan (approved):

internal/librarian/rust/generate_prost_hybrid.go
suzmue: Will save for a follow up.

PR #7107, opened 2026-07-28
feat(sidekick/swift): error on suspect library names

coryan:
If the library name (e.g. `GoogleCloudSecretmanagerV1`) is suspect, then stop the generation and ask the human to manually override the name. Many library names are bad because their capitalization is wrong. In the example, `manager` should be spelled with an uppercase `M`, but the generator only has `secretmanager` as a source and does not speak English.

Fortunately, we have a good source of alternative names: the C#, PHP, and Ruby annotations. When these annotations do not match the default name, we generate an error. The human running `generate` can override the name.

We prefer a manual override over automatically using the annotations for other languages because it would be sad to break Swift just because C# renamed something.

Towards #6229

### DIFF
--- a/internal/sidekick/swift/library_name.go
+++ b/internal/sidekick/swift/library_name.go
@@ -23,6 +23,8 @@ import (
 	"github.com/iancoleman/strcase"
 )
 
+const requiresLibraryNameOverrideFormat = "Other languages with PascalCase style use different names for the %s library.\nConsider these alternatives and use library_name_override to silence this error:\n%s"
+
 // LibraryName returns the Swift library (and module) name for the API.
 func LibraryName(api *api.API, swiftCfg *config.SwiftPackage) (string, error) {
 	if swiftCfg != nil && swiftCfg.LibraryNameOverride != "" {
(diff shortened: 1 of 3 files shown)


suzmue (approved):

internal/sidekick/swift/library_name.go
```diff
@@ -23,6 +23,11 @@ import (
 	"github.com/iancoleman/strcase"
 )
 
+const requiresLibraryNameOverrideFormat = `default library name for %s needs override.
+Other languages with PascalCase style deviate from the default name for this library,
+most likely, that indicates the default library name is not a good choice. Consider
+these alternatives and use library_name_override to silence this error:\n%s`
```
nolanleastin: go raw strings enclosed in "`" wont process the \n escaped character. use a literal newline instead.

internal/sidekick/swift/library_name.go
```diff
@@ -31,11 +36,28 @@ func LibraryName(api *api.API, swiftCfg *config.SwiftPackage) (string, error) {
...
 	}
-	return strings.Join(parts, ""), nil
+	libraryName := strings.Join(parts, "")
+	alternatives := map[string]string{
+		"C#":   strings.ReplaceAll(api.CsharpNamespace, ".", ""),
+		"PHP":  strings.ReplaceAll(api.PhpNamespace, "\\", ""),
+		"Ruby": strings.ReplaceAll(api.RubyPackage, "::", ""),
+	}
+	var messages []string
+	for lang, alt := range alternatives {
```
nolanleastin: nit: golang iterates over map keys randomly. since we are appending errors, they we output won't be deterministic.

nolanleastin (approved):

internal/sidekick/swift/library_name.go
coryan: Fixed, thanks.

internal/sidekick/swift/library_name.go
coryan: Fixed, though I don't think it matters much.

coryan (commented):
This also needs to wait until all the libraries are renamed.

PR #7109, opened 2026-07-28
feat(internal/tool/composer): add verification for composer tools

sofisl:
This change modifies Librarian to verify the configuration of composer tools before installation. It checks that essential fields such as the tool name and version are provided, bubbling up an error if the configuration is invalid.

For https://github.com/googleapis/librarian/issues/6824

### DIFF
--- a/internal/tool/composer/composer.go
+++ b/internal/tool/composer/composer.go
@@ -30,8 +30,14 @@ import (
 // ErrMissingRepo indicates that a repository URL is missing for a Composer tool.
 var ErrMissingRepo = errors.New("repo URL missing")
 
+// errInvalidTool indicates that a Composer tool configuration is invalid.
+var errInvalidTool = errors.New("invalid tool configuration")
+
 // Install installs a list of Composer tools into the environment.
 func Install(ctx context.Context, tools []*config.ComposerTool, phpPath, bin string) error {
+	if err := verify(tools); err != nil {
+		return err
+	}
 	for _, tool := range tools {
 		if tool.Repo == "" {
 			return fmt.Errorf("%w: composer tool %s", ErrMissingRepo, tool.Name)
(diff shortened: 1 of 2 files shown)


internal/tool/composer/composer.go
```diff
@@ -30,12 +30,15 @@ import (
 // ErrMissingRepo indicates that a repository URL is missing for a Composer tool.
 var ErrMissingRepo = errors.New("repo URL missing")
 
+// ErrInvalidTool indicates that a Composer tool configuration is invalid.
+var ErrInvalidTool = errors.New("invalid tool configuration")
```
noahdietz: Nit: Mind grouping this `var` with the other in a `var( )` declaration rather than individual ones?

internal/tool/composer/composer_test.go
```diff
@@ -45,6 +45,16 @@ func TestInstall(t *testing.T) {
 			},
 			wantErr: ErrMissingRepo,
 		},
+		{
```
noahdietz, line 38: Nit: do we need this test case? I think the other was added to cover the condition you replaced by calling `verify`, but now we just need one case here to cover that new condition you moved. All of the diff error modes are covered by TestVerify_Error

General comment that you do not need to act on: We should probably revisit some of our code to extract the error mode tests from the "functional" tests.

internal/tool/composer/composer_test.go
```diff
@@ -142,3 +152,49 @@ func TestCreateBinWrapper(t *testing.T) {
...
+				{Name: "gapic-generator-php", Version: "1.0.0", Repo: ""},
+			},
+			wantErr: ErrMissingRepo,
+		},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			err := verify(test.tools)
+			if !errors.Is(err, test.wantErr) {
+				t.Errorf("verify() error = %v, wantErr = %v", err, test.wantErr)
+			}
```
noahdietz: Nit: `gotErr`


```suggestion
			gotErr := verify(test.tools)
			if !errors.Is(gotErr, test.wantErr) {
				t.Errorf("verify() error = %v, wantErr = %v", gotErr, test.wantErr)
			}
```

noahdietz (approved):
Just some nits, lgtm

internal/tool/composer/composer_test.go
sofisl: yea good point! will move/clean!

PR #7115, opened 2026-07-29
feat(internal/librarian/ruby): add tidy function for ruby configuration

### LINKED ISSUE #6919: ruby: remove empty ruby struct in tidy
medium

If `vb` is non-nil but `vb.EnvPrefix` is empty, setting `lib.APIs[0].Ruby` will result in an empty `ruby:` block in the generated configuration. It is cleaner to only initialize the `Ruby` API configuration when a non-empty environment prefix is found.

```suggestion
			if vb != nil && vb.EnvPrefix != "" {
				lib.APIs[0].Ruby = &config.RubyAPI{
					EnvPrefix: vb.EnvPrefix,
				}
			}
```

_Originally posted by @gemini-code-assist[bot] in https://github.com/googleapis/librarian/pull/6917#discussion_r3614970946_

JoeWang1127:
A Tidy function is added to the `internal/librarian/ruby` package to normalize Ruby-specific library configurations.

It sorts and deduplicates additional proto file lists and nilifies empty Ruby configuration structs at both the API and library levels.

Fixes #6919


internal/librarian/ruby/tidy.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"slices"
+
+	"github.com/googleapis/librarian/internal/config"
+	"github.com/googleapis/librarian/internal/yaml"
+)
+
+// Tidy tidies Ruby-specific configuration for a library.
+func Tidy(lib *config.Library) (*config.Library, error) {
```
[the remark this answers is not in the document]
JoeWang1127: Done.

internal/librarian/ruby/tidy.go
```diff
@@ -0,0 +1,72 @@
...
+
+func tidyAdditionalProtos(protos []string) []string {
+	if len(protos) == 0 {
+		return nil
+	}
+	slices.Sort(protos)
+	return slices.Compact(protos)
+}
+
+func clearIfEmpty[T any](v *T) error {
```
noahdietz, line 62: If I'm understanding: We use a pointer generic so that we can actually clear a variable containing a pointer to empty struct/slice. Is that right? For example, if `foo := &RubyAPI{}`, we want to set `foo` to `nil`, and that's why we need a "reference to a reference"?

If so, can we add some commentary to this code? It's quite useful, but a little nuanced :)

internal/librarian/ruby/tidy.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/googleapis/librarian/internal/config"
+	"github.com/googleapis/librarian/internal/yaml"
+)
+
+// Tidy tidies Ruby-specific configuration for a library.
+func Tidy(lib *config.Library) (*config.Library, error) {
+	for _, api := range lib.APIs {
+		if api.Ruby == nil {
+			continue
+		}
```
noahdietz: Nit: it might be good to move this check into `tidyAPI` to make that code a bit safer. Right now, it assumes `api.Ruby` is non-nil and does unchecked dereferences. If we use that API elsewhere someday, we want it to be safe as is. WDYT?

noahdietz (approved):
One nit and request for clarifying commentary, but lgtm

internal/librarian/ruby/tidy.go
JoeWang1127: Yes, your understanding is correct. I'll add a comment here.

internal/librarian/ruby/tidy.go
```diff
@@ -0,0 +1,72 @@
...
+	if v == nil {
+		return nil
+	}
+	empty, err := yaml.Empty(*v)
+	if err != nil {
+		return err
+	}
+	if empty {
+		var zero T
+		*v = zero
```
codyoss: nit: Since we declare 1.26 I think you can make use of the new `new` function: https://go.dev/doc/go1.26#language

*v = *new(T)

JoeWang1127: Thanks for the suggestion, I've made the change.

internal/librarian/ruby/tidy.go
JoeWang1127: Done.

internal/librarian/ruby/tidy.go
JoeWang1127: Done.

PR #7119, opened 2026-07-29
feat(internal/librarian/php): implement code formatting using prettier

### LINKED ISSUE #6774: php(generate): format code with Prettier
Synthtool runs Prettier during post-processing via owlbot.py. If a second pass is needed, librarian calls Prettier directly:
npm exec --yes --package=@prettier/plugin-php@^0.19 \
  -- prettier '**/Client/*' --write \
  --parser=php --single-quote --print-width=120

zhumin8:
Formatting support for generated PHP libraries is implemented using Prettier and the @prettier/plugin-php plugin. The formatter is executed in the output directory on files matching **/Client/*. Options for running the formatter intended to keep consistent with the current status of on-the-fly invocation in [owlbot.py](https://github.com/googleapis/google-cloud-php/blob/bb0e3de6a995221c212af207c705ae2bcd9efab4/Ces/owlbot.py#L36-L47)

Fixes https://github.com/googleapis/librarian/issues/6774

### DIFF
--- a/internal/librarian/php/format.go
+++ b/internal/librarian/php/format.go
@@ -16,12 +16,48 @@ package php
 
 import (
 	"context"
+	"fmt"
+	"os"
+	"path/filepath"
 
+	"github.com/googleapis/librarian/internal/command"
 	"github.com/googleapis/librarian/internal/config"
+	"github.com/googleapis/librarian/internal/librarian/nodejs"
 )
 
 // Format formats a generated PHP library.
 func Format(ctx context.Context, library *config.Library) error {
-	// TODO(https://github.com/googleapis/librarian/issues/6629): implement PHP formatting
-	return nil
+	nodeInstallDir, err := nodejs.InstallDir()
+	if err != nil {
+		return err
+	}
+	prettierPath := filepath.Join(nodeInstallDir, "bin", "prettier")
+	if _, err := os.Stat(prettierPath); err != nil {
+		return fmt.Errorf("prettier not found at %s: %w", prettierPath, err)
+	}
+
+	outdir, err := filepath.Abs(library.Output)
+	if err != nil {
+		return fmt.Errorf("failed to resolve output directory path: %w", err)
+	}
+
+	binDir := filepath.Dir(prettierPath)
+	env := map[string]string{
+		"PATH": binDir + string(os.PathListSeparator) + os.Getenv("PATH"),
+	}
+
+	// In pnpm 8, global packages are installed in PNPM_HOME/global/5.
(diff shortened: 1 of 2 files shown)


internal/librarian/php/format.go
```diff
@@ -16,12 +16,62 @@ package php
...
+func prettierToolPaths() (string, string, error) {
+	nodeInstallDir, err := nodejs.InstallDir()
+	if err != nil {
+		return "", "", err
+	}
+	prettierPath := filepath.Join(nodeInstallDir, "bin", "prettier")
+	if _, err := os.Stat(prettierPath); err != nil {
+		return "", "", fmt.Errorf("prettier not found at %s: %w", prettierPath, err)
+	}
+	// In pnpm 8, global packages are installed in PNPM_HOME/global/5.
```
JoeWang1127: I think we use pnpm 7.

internal/librarian/php/format_test.go
```diff
@@ -15,16 +15,122 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	got := string(gotBytes)
+	if diff := cmp.Diff(want, got); diff != "" {
+		t.Errorf("mismatch (-want +got):\n%s", diff)
+	}
+}
+
+func TestFormat_Error(t *testing.T) {
```
JoeWang1127: We don't have to use a table driven test since we only have one test case.

internal/librarian/php/format_test.go
```diff
@@ -15,16 +15,122 @@
...
+	prettierPath, _, err := prettierToolPaths()
+	if err != nil {
+		t.Skipf("prettier tools not available: %v", err)
+	}
+	if !phpSupported(t.Context(), prettierPath) {
+		t.Skip("prettier does not support PHP (missing plugin?)")
+	}
+}
+
+func TestPrettierEnv(t *testing.T) {
```
JoeWang1127: Could you move this test above the helper functions?

internal/librarian/php/format.go
zhumin8: Thanks for pointing that out, indeed we use 7.32.2
https://github.com/googleapis/librarian/blob/f52568e7df469fa81067a70f9d1ec2058246246f/.github/workflows/php.yaml#L39

Updated the comment to be more exact. This path structure is true for the 7.32.2 version we used too.

In the meantime, I found an issue with the test setup that was silently skipping it. Fixed in [a01b669](https://github.com/googleapis/librarian/pull/7119/commits/a01b6694650da68a20d8d26e472fac9ddbab3f38)

JoeWang1127 (approved):

PR #7122, opened 2026-07-29
feat(internal/librarian/php): add google-cloud-php/dev as a tool to install

### LINKED ISSUE #6978: php(install): install step for dev/google-cloud tool
We need to run dev/google-cloud tool when onboarding a new library in php (when generate for the first time).
Read more at https://github.com/googleapis/google-cloud-php/blob/main/dev/README.md

This tool is only needed for running `librarian generate` on a newly added library config.

Install should make sure this tool is properly configured to use. 

from ai/local test: 
 Before Librarian can run php dev/google-cloud component:new, the environment (CI runner or developer machine) must run:

    composer install

  in the root of the google-cloud-php repository to populate the root vendor/ directory.

sofisl:
This change adds google-cloud-php/dev as a `local_path` to install its dependencies that are necessary for library generation. This ensures that PHP-specific tools have their dependencies installed before being executed.

Fixes https://github.com/googleapis/librarian/issues/6978


internal/librarian/php/install.go
```diff
@@ -56,20 +56,29 @@ func Install(ctx context.Context, tools *config.Tools) error {
...
 	phpPath, err := checkRequiredCommands()
 	if err != nil {
 		return err
 	}
-
+	_, err = os.Stat("composer.json")
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
+		return fmt.Errorf("failed to stat composer.json: %w", err)
+	}
+	if err == nil {
```
JoeWang1127: Do you need this check?

internal/librarian/php/install.go
```diff
@@ -56,20 +56,29 @@ func Install(ctx context.Context, tools *config.Tools) error {
...
+		cmd := exec.CommandContext(ctx, "composer", "install")
```
JoeWang1127: Why we have a separate `composer install` apart from line 82?

internal/librarian/php/install.go
sofisl: you're right, this could be simplified!

internal/librarian/php/install.go
sofisl: we're installing in the PHP monorepo vs. installing the dpeendencies for the global PHP tools.

I added some comments and also wrote a wrapper function so that it more closely mirrors the installation below.

JoeWang1127: I still feel it's not right to have two `composer install`.

Can we add an entry in the tool section? We could extend the composer.Install to accept local composer.json, something similar to pip local install: https://github.com/googleapis/librarian/blob/main/internal/tool/pip/pip.go#L41.

sofisl: ok, so in investigating Python's implementation, i see they have a separate config called local path. So to do that, I'll create a separate PR to make sure this one is an appropriate size. PTAL at https://github.com/googleapis/librarian/pull/7131

internal/tool/composer/composer.go
```diff
@@ -76,6 +82,17 @@ func Install(ctx context.Context, tools []*config.ComposerTool, phpPath, bin str
 			return fmt.Errorf("tool installation for non-generator composer tools is not yet supported")
 		}
 	}
+	for _, path := range paths {
```
JoeWang1127: How about creating a helper function for this for loop?

internal/tool/composer/composer.go
```diff
@@ -54,8 +59,9 @@ func Install(ctx context.Context, tools []*config.ComposerTool, phpPath, bin str
 		if err != nil {
 			return err
 		}
-		if err := command.RunInDir(ctx, dir, "composer", "install", "--no-interaction", "--prefer-dist"); err != nil {
-			return fmt.Errorf("failed to run composer install: %w", err)
+		paths = append(paths, dir)
+		if tool.Name == "google-cloud-php" {
```
JoeWang1127: This is another tool apart from `gapic-generator-php`, how about removing the else logic in line 81. Also, update the comment in line 68.

internal/tool/composer/composer.go
```diff
@@ -76,6 +82,17 @@ func Install(ctx context.Context, tools []*config.ComposerTool, phpPath, bin str
 			return fmt.Errorf("tool installation for non-generator composer tools is not yet supported")
 		}
 	}
+	for _, path := range paths {
+		if _, err := os.Stat(filepath.Join(path, "composer.json")); err != nil {
+			if errors.Is(err, os.ErrNotExist) {
```
JoeWang1127: How about combing these two errors?

internal/tool/composer/composer.go
sofisl: Yeah technically there's going to be a slight merge conflict with https://github.com/googleapis/librarian/pull/7141, since it actually removes the special processing between the generator and other tools. So we should merge in https://github.com/googleapis/librarian/pull/7141 first.

internal/tool/composer/composer.go
sofisl: Do you think that's necessary even when flattening the error messages?

If you really think it's necessary, then I'd probably refactor the whole Install function in a separate PR. However I think that #7141 which removes the special treating of the generator and with flattening the error messages, the Install function will be fairly easy to understand on its own.

tool/cmd/migrate/librarian_php.yaml
```diff
@@ -27,6 +27,8 @@ default:
     common_resources: true
 tools:
   composer:
+    - name: google-cloud-php
+      local_path: "."
```
zhumin8: Can we specify `/dev`? I suppose the tool scripts should be standalone and not depending on google-cloud-php code outside of /dev dir?

tool/cmd/migrate/librarian_php.yaml
```diff
@@ -27,6 +27,8 @@ default:
     common_resources: true
 tools:
   composer:
+    - name: google-cloud-php
```
zhumin8: "google-cloud-php" sounds slightly confusing to me. Imagine reading this section from google-cloud-php/librarian.yaml, I'd be confused on what it is referring to with repo name.
Since I believe this is just arbitrary name, can we call it along the lines of "google-cloud-php/dev" or "google-cloud-php-dev-scripts"?

zhumin8:
I am slightly confused by the PR title and description. IIUC, this change is to support install composer tool from local path, specifically we need to install google-cloud-php/dev. The "automatically run composer install" confuses me.

tool/cmd/migrate/librarian_php.yaml
sofisl: yeah I hear what you're saying. I changed the names!

tool/cmd/migrate/librarian_php.yaml
sofisl: yeah good idea!

.github/workflows/php.yaml
```diff
@@ -38,7 +38,10 @@ jobs:
           corepack enable pnpm
           corepack prepare pnpm@7.32.2 --activate
       - name: Setup test configuration
-        run: cp ./tool/cmd/migrate/librarian_php.yaml librarian.yaml
+        run: |
+          cp ./tool/cmd/migrate/librarian_php.yaml librarian.yaml
+          mkdir dev
+          echo '{"config": {"vendor-dir": "php-vendor"}}' > dev/composer.json
```
zhumin8, line 44: hmm, this works, but it concerns me a bit. 
I see `librarian_php.yaml` as a temp file during migration (I know this is not new addition now, sorry missed it earlier). Can you use separate files in test data for testing? Or maybe we move it to /testdata after migration? 

I don't think our test should expect dir structure/files to be present outside of the /testdata. Basically, if a test needs this file be present, add that as part of the test setup.

sofisl: yes I definitely think we should move it to /testdata after migration! I can create a f/u for it. But for now, it makes sense to me to have it in the test since we are testing the migratory .yaml which will test our real workflow without depending on an external library. Otherwise we'd need to update the testData and the migrate librarian.yaml when making a change.

zhumin8: I agree with your point for librarian.yaml. 

Can we remove these extra setup and set it up in the test that needs it?
```
mkdir dev
          echo '{"config": {"vendor-dir": "php-vendor"}}' > dev/composer.json
```

zhumin8 (approved):

sofisl:
In thinking more about this, i'm not sure how this will work with #7141  if we change the branching logic. To me it seems quite ugly to say if google-cloud-php/dev, then continue, else if generator, look in main.php, else fail. In talking to Jetski it recommended an Entrypoint option which I kind of like. I'll discuss with @JoeWang1127 on Monday.

.github/workflows/php.yaml
sofisl: The mkdir dev and dev/composer.json setup is actually required by the librarian install php command that runs directly in the workflow step, rather than by a specific Go test. Because it uses the librarian_php.yaml template which has a local composer dependency on dev, the install step will fail if we don't mock it.

Instead of moving it to a Go test, I can just create a dummy dev/composer.json inside a testdata folder and copy it over in the workflow. I'll include this change in the follow-up PR when we move the migratory .yaml to testdata!

JoeWang1127 (approved):

PR #7129, opened 2026-07-29
feat(swift): Improve module generation

nolanleastin:
Remove default message conversions from the conversion logic.

Separate gapic generation from conversion generation.

Some changes to ensure imports are correct for dependencies.

Generates https://github.com/googleapis/google-cloud-swift/pull/76

### DIFF
--- a/internal/sidekick/swift/field_type_name_test.go
+++ b/internal/sidekick/swift/field_type_name_test.go
@@ -484,3 +484,43 @@ func TestFieldTypeName_ExternalNestedMessage(t *testing.T) {
 		t.Errorf("mismatch (-want +got):\n%s", diff)
 	}
 }
+
+func TestFullyQualifiedMessageTypeName(t *testing.T) {
+	msg := &api.Message{
+		Name:    "TestMessage",
+		Package: "google.cloud.test.v1",
+		ID:      ".google.cloud.test.v1.TestMessage",
+	}
+	model := api.NewTestAPI([]*api.Message{msg}, nil, nil)
+	model.PackageName = "google.cloud.test.v1"
+
+	t.Run("standalone library with LibraryName", func(t *testing.T) {
+		c := newTestCodec(t, model, nil)
+		c.LibraryName = "TestLibrary"
+		c.Module = false
+
+		got, err := c.fullyQualifiedMessageTypeName(msg)
+		if err != nil {
+			t.Fatal(err)
+		}
+		want := "TestLibrary.TestMessage"
+		if got != want {
+			t.Errorf("got %q, want %q", got, want)
+		}
+	})
+
+	t.Run("module with empty LibraryName", func(t *testing.T) {
+		c := newTestCodec(t, model, nil)
+		c.LibraryName = ""
+		c.Module = true
+
+		got, err := c.fullyQualifiedMessageTypeName(msg)
+		if err != nil {
+			t.Fatal(err)
+		}
+		want := "TestMessage"
+		if got != want {
+			t.Errorf("got %q, want %q", got, want)
+		}
+	})
+}
--- a/internal/sidekick/swift/generate_conversions.go
+++ b/internal/sidekick/swift/generate_conversions.go
@@ -17,7 +17,6 @@ package swift
 import (
 	"context"
 	"fmt"
-	"path/filepath"
 
 	"github.com/googleapis/librarian/internal/config"
 	"github.com/googleapis/librarian/internal/sidekick/api"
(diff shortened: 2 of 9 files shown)


internal/sidekick/swift/templates/convert/convert_message.mustache
```diff
@@ -17,6 +17,7 @@ extension {{Codec.ParameterTypeName}} {
   internal typealias ProtoType = {{Codec.ProtoTypeName}}
 
   internal init(proto: ProtoType) throws {
+    self.init()
```
coryan, line 20: This was a bit surprising, I thought all the fields have a default? Maybe just so I learn something, why is this needed?

internal/sidekick/swift/field_type_name_test.go
```diff
@@ -484,3 +484,43 @@ func TestFieldTypeName_ExternalNestedMessage(t *testing.T) {
...
+		c.Module = false
+
+		got, err := c.fullyQualifiedMessageTypeName(msg)
+		if err != nil {
+			t.Fatal(err)
+		}
+		want := "TestLibrary.TestMessage"
+		if got != want {
+			t.Errorf("got %q, want %q", got, want)
+		}
```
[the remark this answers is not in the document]
coryan: slop

internal/sidekick/swift/field_type_name_test.go
```diff
@@ -484,3 +484,43 @@ func TestFieldTypeName_ExternalNestedMessage(t *testing.T) {
...
+		c.Module = true
+
+		got, err := c.fullyQualifiedMessageTypeName(msg)
+		if err != nil {
+			t.Fatal(err)
+		}
+		want := "TestMessage"
+		if got != want {
+			t.Errorf("got %q, want %q", got, want)
+		}
```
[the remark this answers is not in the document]
coryan: slop

internal/sidekick/swift/generate_conversions.go
```diff
@@ -43,12 +42,6 @@ func GenerateConversions(ctx context.Context, model *api.API, outdir string, lib
 		}
 		return string(contents), nil
 	}
-	if err := codec.generateMessages(outdir, model, provider); err != nil {
```
coryan, line 46: Ah, I see. Note to self, the librarian.yaml file has two modules, one that generates the GAPIC and one that generates the conversions, so we don't need this code.

coryan (approved):
> Generates https://github.com/googleapis/google-cloud-swift/pull/76

^^ That was super helpful, thanks. Next time consider marking that PR as a draft.

internal/sidekick/swift/generate_conversions.go
nolanleastin: Right. Initially we had both, but then when I was generating the dependency conversion (e.g. Interval+Conversion.swift), Interval.swift was also being re-generated inside of storage. 

I had some hacky code to prevent that, ... but decided to go with this direction instead. At least to me, it makes sense to not have the `generateMessages` inside of the `generateConversion` method. 

But I'm flexible.

coryan: No, this is great. I just was wonder about the deleted code. :shipit:

internal/sidekick/swift/templates/convert/convert_message.mustache
nolanleastin: It was surprising to me, too.

This was the error I was getting:

```
google-cloud-swift/packages/storage/Sources/GoogleCloudStorage/generated/StorageControl/Convert/GoogleType/Interval+Convert.swift:29:3: error: 'self.init' isn't called on all paths before returning from initializer
27 |
28 |   internal init(proto: ProtoType) throws {
29 |   }
   |   `- error: 'self.init' isn't called on all paths before returning from initializer
```

My friend gemini told me that


> In Swift, an initializer defined in a struct extension must guarantee that the struct is fully initialized before  returning. It can do this in one of two ways:
> 
> Manually initialize every single stored property of the struct.
> Delegate to a designated initializer by calling self.init().
> If an initializer body does neither, the compiler rejects it with: error: 'self.init' isn't called on all paths before 
> returning from initializer"

The problem is that Interval+Convert is not yet setting the field values (https://github.com/googleapis/google-cloud-swift/pull/76/changes#diff-2e98b597304397e032e8c264c037679fbb194016c45ff4208b018643f67f979c); hence, the failure. (I'll address that in a follow-up PR, just ... one thing at a time I guess)

nolanleastin: Some options:
1. Keep the self.init() as it's harmless
2. I could remove the interval module for now, if you'd prefer we not have this self.init() scattered (....I did want to have at least a POC of generating dependency converter)
3. I could leave as is, and remove the self.init() later

coryan: Ack, I vote to keep the `self.init()`

PR #7131, opened 2026-07-29
feat(internal/tool/composer): support local_path in ComposerTool configuration

### LINKED ISSUE #7132: php: one composer install
I still feel it's not right to have two `composer install`.

Can we add an entry in the tool section? We could extend the composer.Install to accept local composer.json, something similar to pip local install: https://github.com/googleapis/librarian/blob/main/internal/tool/pip/pip.go#L41.

_Originally posted by @JoeWang1127 in https://github.com/googleapis/librarian/pull/7122#discussion_r3678409656_

sofisl:
This change adds `local_path` support to the Composer tool configuration (analogous to the pip tool) so that we can support more installation cases (specifically, installing dependencies in the monorepo). This is split out from https://github.com/googleapis/librarian/pull/7122#discussion_r3678409656.

Fixes https://github.com/googleapis/librarian/issues/7132

### DIFF
--- a/internal/tool/composer/composer.go
+++ b/internal/tool/composer/composer.go
@@ -41,6 +41,19 @@ func Install(ctx context.Context, tools []*config.ComposerTool, phpPath, bin str
 		return err
 	}
 	for _, tool := range tools {
+		if tool.LocalPath != "" {
+			absPath, err := filepath.Abs(tool.LocalPath)
+			if err != nil {
+				return fmt.Errorf("failed to resolve absolute path for %s: %w", tool.LocalPath, err)
+			}
+			if _, err := os.Stat(absPath); err != nil {
+				return fmt.Errorf("local composer path not found: %w", err)
+			}
+			if err := command.RunInDir(ctx, absPath, "composer", "install", "--no-interaction", "--prefer-dist"); err != nil {
+				return fmt.Errorf("failed to run composer install: %w", err)
+			}
+			continue
+		}
 		dir, err := fetch.Repo(ctx, tool.Repo, tool.Version, tool.SHA256)
 		if err != nil {
 			return fmt.Errorf("fetching %s: %w", tool.Name, err)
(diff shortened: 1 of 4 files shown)


internal/tool/composer/composer.go
```diff
@@ -90,11 +106,24 @@ func createBinWrapper(wrapperName, content, binDir string) error {
...
-			return fmt.Errorf("%w: composer tool %s", ErrMissingRepo, tool.Name)
+		hasLocal := tool.LocalPath != ""
+		hasRemote := tool.Version != "" || tool.Repo != "" || tool.SHA256 != ""
+		if hasLocal && hasRemote {
+			return fmt.Errorf("%w: cannot specify both local_path and version/repo/sha256: %+v", ErrInvalidTool, tool)
+		}
+		if !hasLocal && !hasRemote {
+			return fmt.Errorf("%w: must specify either local_path or version/repo/sha256: %+v", ErrInvalidTool, tool)
+		}
+		if hasRemote {
```
JoeWang1127, line 128: Do we need to check sha256 for remote tools?

internal/tool/composer/composer.go
```diff
@@ -41,14 +41,30 @@ func Install(ctx context.Context, tools []*config.ComposerTool, phpPath, bin str
 		return err
 	}
 	for _, tool := range tools {
-		dir, err := fetch.Repo(ctx, tool.Repo, tool.Version, tool.SHA256)
-		if err != nil {
-			return fmt.Errorf("fetching %s: %w", tool.Name, err)
+		var dir string
```
JoeWang1127, line 44: How about putting line 44 - 63 to a helper function, e,g, `toolDir`?

internal/tool/composer/composer.go
sofisl: good catch!

internal/tool/composer/composer.go
sofisl: good idea, I think that's still maybe too much for a single function so I just created a localPath helper function to break up the branching logic readability

JoeWang1127 (approved):

PR #7140, opened 2026-07-30
fix(swift): fix sub-module imports

nolanleastin:
Fix incorrect dependency import generation for module. 

Introducing `TargetPackageName` to `codec` and updating `skipDependency()` to evaluate self-imports against the Swift SPM target name. 

yields https://github.com/googleapis/google-cloud-swift/pull/76

### DIFF
--- a/internal/sidekick/swift/codec.go
+++ b/internal/sidekick/swift/codec.go
@@ -267,7 +267,6 @@ func (c *codec) skipDependency(dep *Dependency) bool {
 	// During conversion generation, the raw Protobuf stubs module (c.ModulePath, e.g., "StorageControlProtos")
 	// is already statically imported via internal import statements in the conversion file template.
 	// We skip it dynamically to avoid emitting duplicate "import StorageControlProtos" statements.
-	// if c.Module && dep.Name == c.ModulePath {
 	if c.Module && dep.Name == c.ModulePath {
 		return true
 	}
--- a/internal/sidekick/swift/codec.go
+++ b/internal/sidekick/swift/codec.go
@@ -58,16 +58,16 @@ type codec struct {
 	// contains a single target and module with the same names as the library.
 	LibraryName string
 
-	// TargetPackageName is the PascalCase name of the Swift SPM target/package being built
+	// TargetLibraryName is the PascalCase name of the Swift SPM target/library being built
 	// (e.g. "GoogleCloudSecretManagerV1", "GoogleCloudStorage", or "GoogleCloudWkt").
 	//
-	// We need TargetPackageName to correctly identify self-imports in skipDependency.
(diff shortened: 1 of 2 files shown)


internal/sidekick/swift/codec.go
```diff
@@ -58,6 +58,17 @@ type codec struct {
 	// contains a single target and module with the same names as the library.
 	LibraryName string
 
+	// TargetPackageName is the PascalCase name of the Swift SPM target/package being built
```
coryan: I have been thinking about `GoogleCloudSecretManagerV1` as the product (which happens to be a library) and `google-cloud-secretmanarger-v1` as the package name. Because that is how they appear in the package file:

https://github.com/googleapis/google-cloud-swift/blob/03d3ccd34315804ada422618d6696c73c09d0890/generated/google-cloud-secretmanager-v1/Package.swift#L43

Can you call this `TargetLibraryName`?

coryan: I realize that it is somewhat confusing because `librarian` calls `google-cloud-secretmanager-v1` "a library", but Swift calls it (I think) "a package".

internal/sidekick/swift/codec.go
```diff
@@ -196,12 +207,17 @@ func newCodec(model *api.API, library *config.Library, module *config.SwiftModul
 		result.ModulePath = module.ModulePath
 	}
 
-	if !result.Module {
+	if result.Module {
+		if library != nil && library.Name != "" {
+			result.TargetPackageName = pascalCaseNoMangling(library.Name)
```
coryan: So if we build a module inside `google-cloud-foobar` then this will be `GoogleCloudFoobar` ? Is that what you want or `GoogleCloudFooBar` ?  I think it is fine for storage, but not for something like `google-cloud-bigquery` ?

coryan: This may be a problem to figure out in the future...

coryan (changes requested):
I may be wrong about these things, so please check my math.

internal/sidekick/swift/codec.go
nolanleastin: Yes thank you! I like it your suggestion. Naming is hard.

internal/sidekick/swift/codec.go
nolanleastin: Alright fixed. 

After running `librarian generate --all`, had to make updates to librarian.yaml:

https://github.com/googleapis/google-cloud-swift/pull/76/changes#diff-464bebf2dae71c02c30ec0ece355447da9494ccd906ffd3127e3efb9a7c50fcf

coryan (approved):

PR #7141, opened 2026-07-30
fix(internal/librarian/php): remove --side_loaded_root_dir option in install

### LINKED ISSUE #7000: php(install/generate): remove --side_loaded_root_dir option in install
> What is  "--side_loaded_root_dir" argument and why is it needed? 
> Did a bit digging, looks like this is only needed because we are passing in relative paths for config files
>   • $opts['grpc_service_config'] = "google/cloud/ces/v1/ces_grpc_service_config.json"
>   • $opts['service_yaml'] = "google/cloud/ces/v1/ces_v1.yaml"
> 
> If so we should be able to remove this by:
>   1. Modify generate.go to resolve and pass absolute paths in --gapic_out.
>   2. Modify install.go to remove --side_loaded_root_dir "$GOOGLEAPIS_DIR" from the generated wrapper.sh.
> 
> I will file a followup issue for this. For now, can you add a comment for this argument? 

 _Originally posted by @zhumin8 in [#6941](https://github.com/googleapis/librarian/pull/6941/changes#r3641036785)_

The --side_loaded_root_dir option is used to resolve relative paths of config files (e.g., service.yaml, grpc_service_config.json) when the generator runs in a different working directory than the build execution root.

Since the generator already supports resolving absolute paths when --side_loaded_root_dir is omitted, we can 
     - Update librarian/internal/librarian/php/install.go to remove --side_loaded_root_dir from the generated wrapper.sh.
     - Update librarian/internal/librarian/php/generate.go to pass absolute paths for config files in --gapic_out.

sofisl:
This PR completes the refactoring effort to make the PHP Composer tool installation logic fully generic. 
Now that wrapper script generation is decoupled from the `gapic-generator-php` package name (completed in PR #7156), we can safely remove the hardcoded `--side_loaded_root_dir "$GOOGLEAPIS_DIR"` flag from the bash wrapper script template.  Because the wrapper simply forwards all arguments (`"$@"`), this change makes the PHP wrapper logic completely generic and cleanly reusable for any future Composer tool that specifies an `entrypoint`.

In order to remove the option, we've plumbed the absolute path of the necessary arguments for the generator.

Fixes https://github.com/googleapis/librarian/issues/7000


internal/librarian/php/generate.go
```diff
@@ -276,7 +276,7 @@ func gatherProtos(root string) ([]string, error) {
 	return protos, nil
 }
 
-func gapicOpts(apiMetadata *serviceconfig.API, grpcConfigPath string) []string {
+func gapicOpts(apiMetadata *serviceconfig.API, grpcConfigPath, googleapisDir string) []string {
```
JoeWang1127: Why do we need to modify this function? This seems a unrelated change.

internal/tool/composer/composer.go
```diff
@@ -58,22 +58,28 @@ func Install(ctx context.Context, tools []*config.ComposerTool, phpPath, bin str
...
+		destPath := filepath.Join(dir, "vendor", "bin", wrapperName)
+		if _, err := os.Stat(destPath); err != nil {
+			if !errors.Is(err, os.ErrNotExist) {
+				return fmt.Errorf("failed to stat tool binary %s: %w", destPath, err)
 			}
-		} else {
-			return fmt.Errorf("tool installation for non-generator composer tools is not yet supported")
+			fallbackPath := filepath.Join(dir, "src", "Main.php")
+			if _, err := os.Stat(fallbackPath); err != nil {
+				if errors.Is(err, os.ErrNotExist) {
```
JoeWang1127: How about combining these errors?

internal/librarian/php/generate.go
sofisl: Yes, we need this change because the PHP protoc generation runs from the current working directory (the root of the librarian repository), rather than from inside googleapisDir.

The grpcConfigPath and apiMetadata.ServiceConfig returned by the serviceconfig package are paths relative to googleapisDir. Because of this, the gapic-generator-php plugin won't be able to find the grpc_service_config or service_yaml files unless we use googleapisDir to construct the full, absolute paths before passing them to the generator options.

zhumin8: nit: perhaps pre-calculate abs path and just pass in `filepath.Join(googleapisDir, grpcConfigPath)` and `filepath.Join(googleapisDir, apiMetadata.ServiceConfig)` to gapicOpts?

internal/tool/composer/composer.go
```diff
@@ -58,22 +58,25 @@ func Install(ctx context.Context, tools []*config.ComposerTool, phpPath, bin str
...
-			// TODO(https://github.com/googleapis/librarian/issues/7000): Remove the --side_loaded_root_dir once we pass full paths to generator
-			destPath := filepath.Join(dir, "src", "Main.php")
-			wrapperContent := phpWrapperContent(phpPath, destPath)
-			if err := createBinWrapper(wrapperName, wrapperContent, bin); err != nil {
-				return err
+
+		// Determine the executable entry point. By default, composer tools expose
+		// their binaries in vendor/bin/. Some tools (like gapic-generator-php)
+		// might not publish a bin, so we fallback to src/Main.php if present.
+		destPath := filepath.Join(dir, "vendor", "bin", wrapperName)
```
zhumin8: I am a bit confused on these changes, and if they are related to "remove --side_loaded_root_dir option"? 
It looks like a behavior change to support the previous unsupported case "tool installation for non-generator composer tools"? I might be missing context, can you please explain why this is necessary in the description, or if un-related, can you separate out to another PR?

sofisl: Since we are removing the side_loaded_root_dir, it practically eliminates all "special-casing" of the generator, which means we can remove the `if wrapperName` branch. The only piece of logic that is different is whether we're using Main.php or the vendor directory as the wrapper, so that's where we have a fallback path.

internal/librarian/php/generate.go
sofisl: Done, created a separate function to precalculate.

internal/tool/composer/composer.go
zhumin8: I don't oppose the change. 
But I think it's a behavior change that worth mention in PR and be discussed (or separated to a followup to be focused). 
We are supporting `vendor/bin` dir which I suppose is the default? But we don't have a use case for it? I may have missed it, but seems we also don't have a test for this case?

sofisl: Actually, you're right that we still have no need to support vendor/bin. I just removed it, but also removed the special case-ing of the generator. Instead I added an error message that basically says hey, not found, I only looked in main.php, and a comment that basically this is still only supporting the generator. If we ever want to support another composer tool, we'll need to change this.

zhumin8 (approved):

sofisl:
In thinking more about this, i'm not sure how this will work with #7122 if we change the branching logic. To me it seems quite ugly to say if google-cloud-php/dev, then continue, else if generator, look in main.php, else fail. In talking to Jetski it recommended an Entrypoint option which I kind of like. I'll discuss with @JoeWang1127 on Monday.

internal/librarian/php/generate.go
```diff
@@ -284,7 +289,14 @@ func gatherProtos(root string) ([]string, error) {
 	return protos, nil
 }
 
-func gapicOpts(apiMetadata *serviceconfig.API, grpcConfigPath string) []string {
+func absConfigPath(baseDir, configPath string) string {
+	if configPath == "" {
+		return ""
+	}
+	return filepath.Join(baseDir, configPath)
```
JoeWang1127: Do you want to use filepath.Abs to make sure the returned path is indeed an absolute one?

sofisl: TIL, thank you!

JoeWang1127 (approved):

internal/tool/composer/composer.go
sofisl: Ok did a bunch of refactoring (mostly listed in https://github.com/googleapis/librarian/pull/7156#issue-5033986650) and also removed the vendor/bin support (since we don't have an active use case). Thanks for thinking through this!

PR #7159, opened 2026-08-01
refactor(install): extract pnpm installation logic to internal/tool/pnpm

quirogas:
Extract generic `pnpm` package installation logic from `internal/librarian/nodejs/install.go` into a shared standalone package at `internal/tool/pnpm`.

Previously, PHP client library generation (`internal/librarian/php/install.go`) invoked `nodejs.InstallPNPM`, creating a cross-package dependency between PHP and Node.js toolchains. Extracting this functionality into `internal/tool/pnpm`—patterned after `internal/tool/pip`—centralizes binary resolution, caching, and error handling into a dedicated tool package that both language generators consume independently. This modular foundation cleanly paves the way for upcoming declarative upgrades to pnpm v8+.

For https://github.com/googleapis/librarian/issues/6857


internal/librarian/php/install.go
```diff
@@ -79,8 +79,7 @@ func Install(ctx context.Context, tools *config.Tools) error {
 		return err
 	}
 	// Install PNPM tools.
```
JoeWang1127: I think we can remove this comment as well.

internal/tool/pnpm/pnpm.go
```diff
@@ -67,16 +93,23 @@ func Install(ctx context.Context, pnpmTools []*config.PNPMTool, binDir string) e
 // This enables complete environment caching and restore on CI runners,
 // while permanently avoiding persistent side-effects on the host machine
 // (it does not modify the user's personal ~/.config/pnpm/rc files).
-func envs(binDir string) ([]string, error) {
+func getPNPMEnv(binDir string) ([]string, error) {
```
JoeWang1127: nit: `env` maybe better, considering this is `pnpm` package, we don't need to mention `pnpm` in function name.

internal/tool/pnpm/pnpm.go
```diff
@@ -117,10 +150,14 @@ func installFromSource(ctx context.Context, env []string, tool *config.PNPMTool,
 	}
 
 	// Run build steps.
-	buildDir := filepath.Join(dir, srcDir)
+	subdir := tool.SrcDir
+	if subdir == "" {
```
JoeWang1127: Do you think it's better to return an error if subdir is empty for a package that requires local build?

internal/librarian/nodejs/generate_test.go
JoeWang1127, line 1: Does change in this file related to pnpm install? If not, can we make a separate pull request?

internal/librarian/php/install.go
quirogas: Removed the redundant comment in the latest commit to keep the installation routine concise.

internal/tool/pnpm/pnpm.go
quirogas: Great suggestion. Renamed the function to `env` across the package and tests to avoid package naming stutter.

internal/tool/pnpm/pnpm.go
quirogas: Agreed. Removed the hardcoded Node-specific fallback and added `ErrMissingSrcDir`, which now returns an explicit error if `SrcDir` is empty for packages requiring local builds.

internal/librarian/nodejs/generate_test.go
quirogas: Thank you for catching this. These coverage additions for `movePackageFromStaging` have been reverted from this branch to ensure this PR remains strictly atomic to pnpm installation refactoring. We can propose them in a separate cleanup PR later.

JoeWang1127:
Can you fix the unit tests?

JoeWang1127 (approved):

PR #7160, opened 2026-08-02
refactor(internal/librarian/php): move `DefaultOutput` to internal/librarian/php/defaut.go

JoeWang1127:
Move DefaultOutput to internal/librarian/php/defaut.go.

This will leave `generate.go` only contains generate-related functions.

### DIFF
--- a/internal/librarian/php/default.go
+++ b/internal/librarian/php/default.go
@@ -20,6 +20,12 @@ import (
 	"github.com/googleapis/librarian/internal/config"
 )
 
+// DefaultOutput derives an output path from a library name and a default
+// output directory.
+func DefaultOutput(name, defaultOutput string) string {
+	return filepath.Join(defaultOutput, name)
+}
+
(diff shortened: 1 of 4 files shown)


internal/librarian/php/default.go
```diff
@@ -20,6 +20,12 @@ import (
 	"github.com/googleapis/librarian/internal/config"
 )
 
+// DefaultOutput derives an output path from a library name and a default
+// output directory.
+func DefaultOutput(name, defaultOutput string) string {
```
codyoss, line 25: nit: Approved since this is a refactor, but I see no value in this being a function as we are just passing the name inputs to another function.

codyoss (approved):

internal/librarian/php/default.go
JoeWang1127: We will have more logic to determine the output.

Issue #7152, opened 2026-07-31
php(generate): derive parameters from API path for dev tool


zhumin8:
--proto-package (Versioned proto package, e.g., google.cloud.speech.v2, from proto/api path)
--api-version (API version, e.g., v2, from API path)

PR #7168, opened 2026-08-03
chore(internal/librarian/php): derive onboarding parameters from proto

JoeWang1127:
An initialization parameter builder is added to the PHP onboarding logic to aggregate library component names, namespace, and protobuf package directly from source proto files.

For #7152


internal/librarian/php/init.go
```diff
@@ -24,50 +25,75 @@ import (
...
 	}
-	return backupNamespace(apiPath), nil
-}
-
-// componentName returns the component name from a namespace.
-func componentName(namespace string) string {
-	if comp, ok := strings.CutPrefix(namespace, `Google\Cloud\`); ok {
-		return comp
+	if pkg == "" {
+		return "", "", errNoProtoPackage
```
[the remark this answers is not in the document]
JoeWang1127: Done.

internal/librarian/php/init.go
```diff
@@ -24,50 +25,74 @@ import (
...
+	return &initParams{
+		componentName: deriveComponentName(ns),
+		namespace:     ns,
+		protoPackage:  pkg,
+	}, nil
+}
+
+// parseProto reads the proto package and php_namespace option from the first .proto file
+// in the API directory. If the option is not found, it generates a fallback namespace
+// from the API path.
```
zhumin8, line 54: Can we update the doc comment to cover both return values and error behavior?

internal/librarian/php/init.go
```diff
@@ -24,50 +25,74 @@ import (
 )
 
 var (
-	namespaceRe     = regexp.MustCompile(`php_namespace\)?\s*=\s*"([^"]+)"`)
-	versionSuffixRe = regexp.MustCompile(`\\V\d+.*$`)
+	namespaceRe       = regexp.MustCompile(`php_namespace\)?\s*=\s*"([^"]+)"`)
+	versionSuffixRe   = regexp.MustCompile(`[\\.][vV]\d+.*$`)
+	packageRe         = regexp.MustCompile(`^package\s+([^\s;]+)\s*;`)
```
zhumin8, line 30: I think we want to derive package name from the API path directly instead of parsing proto. It's 99% the same, when they are not, we can use override in config to specify.
cc. @noahdietz

PR #7175, opened 2026-08-04
fix(swift): use let for messages without converted fields

nolanleastin:
Conditionally emit `let proto = ProtoType()` instead of `var proto = ProtoType()`
in `toProto()` when a message contains no fields that emit conversion statements
(such as messages with only repeated or oneof fields, or zero fields).

generates: https://github.com/googleapis/google-cloud-swift/pull/112

### DIFF
--- a/internal/sidekick/swift/annotate_message.go
+++ b/internal/sidekick/swift/annotate_message.go
@@ -36,7 +36,7 @@ type messageAnnotations struct {
 	// HasConvertedFields is true if any field in the message emits code in
 	// toProto(), which currently includes only singular, non-oneof fields.
 	// This prevents the Swift compiler error "variable 'proto' was never mutated"
-	// for messages containing only repeated or oneof fields by emitting "let"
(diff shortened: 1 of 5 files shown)


internal/sidekick/swift/annotate_message.go
```diff
@@ -137,6 +147,13 @@ func (c *codec) annotateMessage(message *api.Message, model *modelAnnotations) e
 	if len(message.Fields) != 0 {
 		sampleField = camelCase(message.Fields[0].Name)
 	}
+	hasConvertedFields := false
+	for _, f := range message.Fields {
+		if !f.IsOneOf && f.Singular() {
+			hasConvertedFields = true
+			break
+		}
+	}
```
coryan: Go is not as nice as Rust or Swift, but it can be nicer than C89:

```suggestion
	hasConvertedFields := slices.IndexFunc(message.Fields, func(f *api.Field) bool { return !f.IsOneOf && f.Singular() }) != -1
```

coryan (approved):
A few nits ...

Issue #6218, opened 2026-05-29
java: improve library generation efficiency


JoeWang1127:
Today the `librarian generate -all` requires more than an hour, e.g., https://github.com/googleapis/google-cloud-java/actions/runs/26604650503/job/78396681746.

We should try to improve the efficency, e.g., run generate in parallel.

We need investigates on which part slows down the progress and find ways to improve.

zhumin8:
I did some testing on this, the code change is straight forward, but there is real risk of crashing the system when RAM running out.

  Java generation is more resource-heavy than Go/Python because it must spawn a new JVM instance for each generated library module, causing substantial CPU and memory overhead. Local testing of Java generation with a concurrency limit of  4  successfully completed in 30 minutes (a 2x–3x speedup compared to sequential runs) but peaked at 22.09 GB of RAM, which would crash the system without concurrency limits. Since we run librarian is run on both local (can use 120GB RAM machines) and resource-constrained GitHub Actions runners (limited to 16GB RAM), we should consider introduce a  --concurrency flag to allow dynamically adjusting the parallel worker count based on the host environment's hardware limits.

I recall there were discussion around a concurrency flag when Go/Python implemented parallel generation. Is it true that we decided to not do it for simplicity and Python/Go were lightweight execution anyway?


In the meantime, to make Java developer's life easier,  I can add to this workflow, to allow manual trigger from a branch and write a commit with changes.
https://github.com/googleapis/google-cloud-java/blob/main/.github/workflows/librarian_generation_check.yaml
(filed https://github.com/googleapis/librarian/issues/6568 for this)

zhumin8:
Unassign as I am not currently working on this.
For posterity, discussed and prioritizing other work at the moment.
Also, introducing a --concurrency flag  is undesired, especially it is only used by Java. 
We can try to auto detect cores and RAM size and setup concurrency threads

**Effort**: depending on solution and how deep we explore. Maybe timebox 1-2 days.

blakeli0:
This is a pretty open-ended problem and there could be very different ways to solve it. A few quick thoughts:
- IIRC, the Java generator is pretty efficient, so are the protoc/grpc plugins. 
- It is most likely due to the IO operations after the Java files are generated. e.g. Move files from one folder to another, post-processing etc. There could be unnecessary IO operations.
- Parallelization could improve it to some extent, but for IO heavy operations, it usually does not do much.

zhumin8:
[...]

PR #7186, opened 2026-08-04
perf(java): batch format all libraries using argument file

yangyzs:
Batch format all generated Java client libraries in a single `google-java-format` invocation using an argument file (@<filename>). 

This avoids per-library JVM startup overhead when formatting multiple libraries and prevents exceeding OS command-line length limits (ARG_MAX). 

For https://github.com/googleapis/librarian/issues/6218


internal/librarian/java/format.go
```diff
@@ -25,21 +25,33 @@ import (
...
-	args := append([]string{"--replace"}, files...)
+	tmpFile, err := os.CreateTemp("", "gjf-args-*.txt")
+	if err != nil {
+		return fmt.Errorf("failed to create temp file for format args: %w", err)
+	}
+	defer os.Remove(tmpFile.Name())
+	content := "--replace\n" + strings.Join(allFiles, "\n")
+	if err := os.WriteFile(tmpFile.Name(), []byte(content), 0600); err != nil {
+		return fmt.Errorf("failed to write format args: %w", err)
+	}
```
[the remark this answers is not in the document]
yangyzs: Fixed.

internal/librarian/java/format.go
```diff
@@ -25,21 +25,37 @@ import (
...
+	if err != nil {
+		return fmt.Errorf("failed to create temp file for format args: %w", err)
+	}
+	defer os.Remove(tmpFile.Name())
+	content := "--replace\n" + strings.Join(allFiles, "\n")
+	if _, err := tmpFile.WriteString(content); err != nil {
+		tmpFile.Close()
+		return fmt.Errorf("failed to write format args: %w", err)
+	}
+	if err := tmpFile.Close(); err != nil {
```
JoeWang1127: Do you need this since we have a defer function to remove this file.

internal/librarian/java/format.go
```diff
@@ -25,21 +25,37 @@ import (
...
+			return fmt.Errorf("failed to find java files for formatting in %q: %w", lib.Name, err)
+		}
+		allFiles = append(allFiles, files...)
 	}
-	if len(files) == 0 {
+	if len(allFiles) == 0 {
 		return nil
 	}
-	args := append([]string{"--replace"}, files...)
+	tmpFile, err := os.CreateTemp("", "gjf-args-*.txt")
```
JoeWang1127: Could you create a helper function for creating a file and write all java files.

Could you also include a code comment here why we need to create a file, rather than adding strings to the formatter.

internal/librarian/java/format.go
yangyzs: Yes. We need tmpFile.Close() to save and unlock the file so google-java-format can read it safely. defer os.Remove(argFile) in formatBatch handles deleting the file after formatting is done.

internal/librarian/java/format.go
yangyzs: Done. Created `createArgFile` function with unit tests and added comments.

yangyzs:
In the latest change, we batched Java formatting into chunks of 2,000 files using temporary argument files (@filename).

- We use @filename argument files to avoid hitting OS command-line size limits (ARG_MAX). 
- We chunk files in batches of 2,000 because passing all 20,000+ files in a single batch causes JVM heap exhaustion on CI runners.

.github/workflows/java.yaml
```diff
@@ -116,7 +116,7 @@ jobs:
         working-directory: google-cloud-java
         run: librarian generate secretmanager
       - name: Run librarian generate all (integration test)
-        if: matrix.task == 'integration' && github.event_name == 'push' && github.ref == 'refs/heads/main'
+        if: matrix.task == 'integration'
```
JoeWang1127: Could you restore the change?

internal/librarian/java/format_test.go
```diff
@@ -128,3 +153,42 @@ func TestCollectJavaFiles(t *testing.T) {
...
+			files: []string{},
+			want:  "--replace\n",
+		},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			argFile, err := createArgFile(test.files)
+			if err != nil {
+				t.Fatal(err)
+			}
+			defer os.Remove(argFile)
```
JoeWang1127: Do you need this defer func if the file is created by `os.CreateTemp`?

internal/librarian/java/format.go
```diff
@@ -25,26 +25,72 @@ import (
...
+		}
+	}
+	return nil
+}
+
+// formatBatch formats a single chunk of Java files using an argument file.
+func formatBatch(ctx context.Context, files []string) error {
+	// Write file paths to an argument file (@filename) to avoid
+	// exceeding OS command-line length limits (ARG_MAX).
+	argFile, err := createArgFile(files)
```
JoeWang1127: Can we batch the arguments to avoid command line length limits? This is more simple than creating batch files.

yangyzs: Great Suggestion. Made Changes.

internal/librarian/java/format_test.go
yangyzs: Resolved in the lastest commit. We switched to direct CLI argument batching, so this part are no longer needed.

internal/librarian/java/format.go
```diff
@@ -25,22 +25,35 @@ import (
...
+	var allFiles []string
+	for _, lib := range libraries {
+		files, err := collectJavaFiles(lib.Output)
+		if err != nil {
+			return fmt.Errorf("failed to find java files for formatting in %q: %w", lib.Name, err)
+		}
+		allFiles = append(allFiles, files...)
 	}
-	if len(files) == 0 {
+	if len(allFiles) == 0 {
```
JoeWang1127: I think this check is unnecessary.

JoeWang1127 (approved):

internal/librarian/java/format.go
yangyzs: Good catch. Removed.

PR #7188, opened 2026-08-04
feat(swift): support repeated fields in Protobuf conversions

nolanleastin:
Extend the Swift sidekick generator to support converting `Repeated` (collection) fields (`Repeated == true && Map == false`) between generated Swift struct models and SwiftProtobuf types.

generates https://github.com/googleapis/google-cloud-swift/pull/133

### DIFF
--- a/internal/sidekick/swift/templates/convert/convert_message.mustache
+++ b/internal/sidekick/swift/templates/convert/convert_message.mustache
@@ -58,6 +58,21 @@ extension {{Codec.ParameterTypeName}} {
     {{/IsObject}}
     {{/Optional}}
     {{/Singular}}
+    {{#Repeated}}
+    {{^Map}}
+    {{#IsObject}}
+    self.{{Codec.Name}} = try proto.{{Codec.ProtoFieldName}}.map { try .init(proto: $0) }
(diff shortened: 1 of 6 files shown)


internal/sidekick/swift/templates/convert/convert_message.mustache
```diff
@@ -58,6 +58,21 @@ extension {{Codec.ParameterTypeName}} {
     {{/IsObject}}
     {{/Optional}}
     {{/Singular}}
+    {{#Repeated}}
+    {{^Map}}
```
coryan: You don't need to check for `{{^Map}}` if you already checked for `{{#REpeated}}`. `Map`, `Repeated` and `Singular` are mutually exclusive.

internal/sidekick/swift/templates/convert/convert_message.mustache
```diff
@@ -109,6 +124,21 @@ extension {{Codec.ParameterTypeName}} {
...
```
coryan: Ditto.

coryan (approved):
A couple of nits to simplify the mustache templates, otherwise LGTM.

Issue #7025, opened 2026-07-24
php(generate): derive component name on the fly as output dir


zhumin8:
We are moving away from using library.name as output dir for php. 
Add logic to derive component name on the fly to use as output dir.

Ref [go/sdk-librarian-php:component-name](http://goto.google.com/sdk-librarian-php:component-name)

zhumin8:
Doing this in 3 steps: 
- Add the schema override field and populate baseline migration output (include component_name for each library). (#7196)
- Enable librarian generate to derive component names on the fly, with fallback to component_name config override
- Reuse PR 2 logic in the migration tool so librarian.yaml stays clean and omits component_name when it matches the derived name.

Then proceed with #7024

PR #7197, opened 2026-08-05
feat(internal/librarian/php): derive component name with config override in generate

zhumin8:
Add ComponentNameForLibrary to resolve the PHP component name for staging and output paths. Use component_name as override if explicitly set. Otherwise, the component name is derived on the fly from the php_namespace option of the primary API proto file.
    
Update Generate and postProcessLibrary to use the resolved component name when creating and cleaning up owl-bot-staging directories.

For #7025

### DIFF
--- a/internal/librarian/php/generate.go
+++ b/internal/librarian/php/generate.go
@@ -82,7 +82,7 @@ func Generate(ctx context.Context, cfg *config.Config, library *config.Library, 
 	}
 	srcCfg := sources.NewSourceConfig(src, library.Roots)
 	googleapisDir := srcCfg.Root("googleapis")
-	componentName, err := ComponentNameForLibrary(googleapisDir, library)
+	componentName, err := componentNameForLibrary(googleapisDir, library)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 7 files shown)


internal/librarian/php/init.go
```diff
@@ -55,6 +56,23 @@ func newInitParams(googleapisDir string, api *config.API) (*initParams, error) {
 	}, nil
 }
 
+// ComponentNameForLibrary resolves the component name for a PHP library.
+// If library.PHP.ComponentName is set, it is returned as an explicit override.
+// Otherwise, the component name is derived on the fly from the php_namespace of the library's primary API.
+func ComponentNameForLibrary(googleapisDir string, library *config.Library) (string, error) {
```
JoeWang1127: Can we use unexported function name?

JoeWang1127 (approved):

internal/librarian/php/generate.go
```diff
@@ -75,18 +75,27 @@ func Generate(ctx context.Context, cfg *config.Config, library *config.Library,
...
+		}
+	}
+
+	srcCfg := sources.NewSourceConfig(src, library.Roots)
+	googleapisDir := srcCfg.Root("googleapis")
+	componentName, err := ComponentNameForLibrary(googleapisDir, library)
+	if err != nil {
+		return err
+	}
+
```
JoeWang1127: nit: remove this empty line.

PR #7198, opened 2026-08-05
feat(internal/librarian/php): derive proto package and version in initParams

### LINKED ISSUE #7152: php(generate): derive parameters from API path for dev tool

zhumin8:
Adds protoPackage and apiVersion fields  to initParams and populates them in from API path. 

The protoPackage is derived as an unversioned proto package prefix by stripping the version from the API path, and  allow an explicit override using the `proto_package` field  in PHP API configuration.
  
The PHP migration tool is also updated to include explicit protoPackage overrides for existing APIs whose declared proto package names differ from their directory paths, specifically MapsFleetEngine, MapsFleetEngineDelivery, and Translate.
    

Fixes https://github.com/googleapis/librarian/issues/7152

### DIFF
--- a/doc/config-schema.md
+++ b/doc/config-schema.md
@@ -450,6 +450,7 @@ This document describes the schema for the librarian.yaml.
 | :--- | :--- | :--- |
 | `additional_protos` | list of string | Is a list of additional proto files to include in generation. |
 | `common_resources` | bool (optional) | Indicates whether to include common resources in generation. Must be configured either globally or per-API. |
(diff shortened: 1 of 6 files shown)


zhumin8:
I could split into smaller PRs, but thought combining may be faster considering time lag in reviews as I am async.
Hopefully the commits helps break things down for review.

JoeWang1127 (approved):

PR #7205, opened 2026-08-05
feat(swift): protobuf conversions support oneofs

nolanleastin:
Extend the Swift sidekick generator to support converting Protobuf `oneof` fields between clean Swift struct models and `SwiftProtobuf` message types.

generates https://github.com/googleapis/google-cloud-swift/pull/167

### DIFF
--- a/internal/sidekick/swift/templates/convert/convert_message.mustache
+++ b/internal/sidekick/swift/templates/convert/convert_message.mustache
@@ -73,6 +73,26 @@ extension {{Codec.ParameterTypeName}} {
     {{/Repeated}}
     {{/IsOneOf}}
     {{/Fields}}
+    {{#OneOfs}}
+    if let oneof = proto.{{Codec.PropertyName}} {
+      switch oneof {
+      {{#Fields}}
+      case .{{Codec.ProtoFieldName}}(let value):
+        {{#IsObject}}
+        self.{{Codec.OneOfPropertyName}} = .{{Codec.Name}}(try .init(proto: value))
+        {{/IsObject}}
+        {{#IsEnum}}
+        self.{{Codec.OneOfPropertyName}} = .{{Codec.Name}}(.init(proto: value))
+        {{/IsEnum}}
+        {{^IsObject}}
+        {{^IsEnum}}
+        self.{{Codec.OneOfPropertyName}} = .{{Codec.Name}}(value)
+        {{/IsEnum}}
+        {{/IsObject}}
+      {{/Fields}}
+      }
+    }
+    {{/OneOfs}}
   }
 
   internal func toProto() throws -> ProtoType {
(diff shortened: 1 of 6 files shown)


internal/sidekick/swift/templates/convert/convert_message.mustache
```diff
@@ -137,6 +157,28 @@ extension {{Codec.ParameterTypeName}} {
...
+    {{#OneOfs}}
+    if let oneof = self.{{Codec.PropertyName}} {
+      switch oneof {
+      {{#Fields}}
+      case .{{Codec.Name}}(let value):
+        {{#IsObject}}
+        if let value = value {
+          proto.{{Group.Codec.PropertyName}} = .{{Codec.ProtoFieldName}}(try value.toProto())
+        }
+        {{/IsObject}}
```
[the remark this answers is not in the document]
nolanleastin: Omitting if let value = value causes a Swift compiler error: value of optional type 'Status?' must be unwrapped to refer to member 'toProto'."

coryan, line 169: Gemini says this is possible:


```suggestion
      {{#IsObject}}
      case .{{Codec.Name}}(let value?):
          proto.{{Group.Codec.PropertyName}} = .{{Codec.ProtoFieldName}}(try value.toProto())
      {{/IsObject}}
```

Note the `?` in `(let value?)`.  Which generates nicer code. You would need more changes in the template, of course.

coryan (commented):
Can you check if my search for "swift combine enum and optional in pattern matching" yielded good results?

internal/sidekick/swift/templates/convert/convert_message.mustache
nolanleastin: Hmm. I tried this. I had some compiler errors, though. 

```
error: switch must be exhaustive 321 | proto.percentageIncrease = self.percentageIncrease 322 | if let oneof = self.details { 323 | switch oneof { | |- error: switch must be exhaustive | |- note: add missing case: '.contribution(.none)' | |- note: add missing case: '.error(.none)' | `- note: add missing cases 324 | case .contribution(let value?): 325 | proto.details = .contribution(try value.toProto())
```

So I can do a couple of options ... : 

1. Keep as is. Pros: handles everything. Cons: verbose/less nice code.
2. Add a .none case e.g.: 
```
case .error(let value?):
  proto.details = .error(try value.toProto())
case .error(.none):
  break
```

This is also not very nice code imo. 

3. Add a default: break case e.g.

```
case .error(let value?):
  proto.details = .error(try value.toProto())
default:
  break
```

I don't like this approach as it doesn't explicitly handle all cases if/when new fields are added.

I think we keep it as is. WDYT?

coryan: Blegh... I speculate that you could use `.error(nil)` but that looks worse. Keep as-is SGTM.

coryan:
:shipit:

coryan (approved):

PR #7206, opened 2026-08-05
feat(swift): support skipped_ids in configuration

nolanleastin:
This unblocks storage v2 generation by providing the ability to skip duplicated object types.

Example librarian.yaml usage:
```yaml
libraries:
  - name: google-cloud-storage
    swift:
      modules:
        - output: packages/storage/Sources/GoogleCloudStorage/generated/Storage
          api_path: google/storage/v2
          skipped_ids:
            - .google.storage.v2.Storage
```

helps generate https://github.com/googleapis/google-cloud-swift/pull/150


internal/sidekick/swift/generate.go
```diff
@@ -52,11 +52,13 @@ func Generate(ctx context.Context, model *api.API, outdir string, library *confi
 	if err := codec.generateEnums(outdir, model, provider); err != nil {
 		return err
 	}
-	if err := codec.generateServices(outdir, model, provider); err != nil {
-		return err
-	}
-	if err := codec.generateClients(outdir, model, provider); err != nil {
-		return err
+	if !codec.Module {
```
coryan: I was assuming we could do these changes via configuration, as in:

https://github.com/googleapis/google-cloud-rust/blob/d38f414729194405a5c68e9cde2c53a8cefb1898/librarian.yaml#L1514

Is that possible?

coryan: If the feature is missing, can we add the feature to do this via config vs. a hard-coded thing?

nolanleastin: Ahhh. Done!

coryan (approved):

PR #7212, opened 2026-08-06
feat(internal/librarian/rust): support external type conversions 

### LINKED ISSUE #7144: rust: LatLng is not conversions are not being generated in hybrid crates
Reproducible when generating streaming RPCs for crates: google-cloud-dialogflow-v2 and google-cloud-dialogflow-cx-v3

This causes a compile time error. In firestore, we configure librarian.yaml to generate the necessary conversions. We will maybe want to reuse the underlying mechanism.

suzmue:
External message and enum types referenced by bidirectional streaming RPCs in hybrid crates were not being generated in convert.rs because filterModelToStreaming only included messages in the target package.

Populate ExternalMessages and ExternalEnums in hybridModel for external top-level messages and enums. Override RelativeName for ExternalMessages and ExternalEnums so ToProto and FromProto implementations generate cleanly.

Fixes #7144

### DIFF
--- a/internal/librarian/rust/generate_prost_hybrid.go
+++ b/internal/librarian/rust/generate_prost_hybrid.go
@@ -15,6 +15,7 @@
 package rust
 
 import (
+	"cmp"
 	"context"
 	"fmt"
 	"maps"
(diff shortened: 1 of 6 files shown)


coryan (commented):
> External message and enum types referenced by bidirectional streaming RPCs in hybrid crates were not being generated in convert.rs because filterModelToStreaming only included messages in the target package.

How does the orphan rule apply or not apply here?  I can see how defining traits for *protos* in external packages works: while the package is external, the proto is defined in the package.

But you cannot define an external trait for a type defined in a separate crate. How does that work?

suzmue:
> How does the orphan rule apply or not apply here? I can see how defining traits for _protos_ in external packages works: while the package is external, the proto is defined in the package.
> 
> But you cannot define an external trait for a type defined in a separate crate. How does that work?

This works because the protos are generated in our crate and are a part of the definition (From https://doc.rust-lang.org/reference/items/implementations.html#r-items.impl.trait.orphan-rule.def, T1 is local)

We already have this in firestore:
-  The protos are generated: https://github.com/googleapis/google-cloud-rust/blob/main/src/firestore/src/generated/protos/firestore/google.r%23type.rs
- And conversions are written: https://github.com/googleapis/google-cloud-rust/blob/main/src/firestore/src/generated/convert/type/convert.rs
- Librarian.yaml for the conversions: https://github.com/googleapis/google-cloud-rust/blob/ecd5aebbf26589df2793b868a24dc62b9a765c5e/librarian.yaml#L877-L883 
- 
For simplicity in this hybrid crate I am putting the conversions in the same file.

coryan (approved):

PR #7221, opened 2026-08-06
feat(internal/sidekick/rust): initialize bidi streaming RPCs eagerly

suzmue:
Initialize bidirectional streaming RPCs eagerly on send instead of deferring transport dialing until the first message is written to the channel in a background task.

Stream builders now support field setters and with_request to configure the initial request message, and send initiates the stream immediately, surfacing connection errors to the caller.

This change is intended to be the minimal change needed to move the initialization logic and does not address other TODOs remaining in the transport implementation. This includes configuration for the channel values, which may mean we will not use RequestBuilder directly in the final implementation.

For #6835

### DIFF
--- a/internal/sidekick/rust/generate_bidi_streaming_test.go
+++ b/internal/sidekick/rust/generate_bidi_streaming_test.go
@@ -1,0 +1,267 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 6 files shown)


coryan (approved):
In Swift we started writing unit tests to verify the mustache template output matches expectations.... For example:

https://github.com/googleapis/librarian/blob/5a7775a1535a48df6a6fc0197cd8a1d3ac045df7/internal/sidekick/swift/generate_enum_swift_test.go#L90-L103

I think these are less annoying than golden file tests, and will avoid somebody accidentally changing the template to return the wrong thing.

The downside is that they trigger errors if you change spaces or indentation in the mustache templates.

Consider starting a few such tests with this PR.

PR #7222, opened 2026-08-06
feat(swift): conversions support map fields

nolanleastin:
Generates https://github.com/googleapis/google-cloud-swift/pull/244


coryan (approved):
Can we see the effects in the generated code?

coryan:
> Can we see the effects in the generated code?

Thanks, they look good. :shipit:

PR #7241, opened 2026-08-10
feat(internal/librarian/php): call dev tool from librarian generate on first generation

### LINKED ISSUE #7153: php(generate): use library.output for a library directory vs. `componentName`
if output dir not existing, call dev tool to generate files that are only generated once.

sofisl:
This change allows us to generate initial files if the library is brand new by using the newly installed dev tool (#7122). The upstream `dev/google-cloud component:new` tool hardcodes its scaffolding output to a directory named after the component and does not natively support an output directory override. To fix this, `initComponentIfMissing` now checks for the existence of `library.Output` instead of `componentName`. If missing, it allows the dev tool to scaffold the default directory, then manually moves the result to the configured `library.Output` destination. 

Fixes https://github.com/googleapis/librarian/issues/7153

### DIFF
--- a/internal/librarian/php/init.go
+++ b/internal/librarian/php/init.go
@@ -70,9 +70,8 @@ func initComponentIfMissing(ctx context.Context, library *config.Library, google
 	if err := initComponent(ctx, params); err != nil {
 		return "", err
 	}
-	// The dev tool hardcodes the scaffold output strictly to params.componentName.
-	// If a custom library.Output is requested, we must manually move the scaffolded
-	// directory to the correct target location.
+	// The dev tool always scaffolds to params.componentName.
+	// Move it manually if a custom output directory is requested.
 	if targetDir != params.componentName {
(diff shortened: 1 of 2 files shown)


internal/librarian/php/init.go
```diff
@@ -58,17 +58,30 @@ func initComponentIfMissing(ctx context.Context, library *config.Library, google
...
-	_, err = os.Stat(params.componentName)
-	if err == nil {
-		// Component exists, return the component name.
-		return params.componentName, nil
+	targetDir := library.Output
+	if targetDir == "" {
+		targetDir = params.componentName
 	}
-	if !errors.Is(err, fs.ErrNotExist) {
+	if _, err := os.Stat(targetDir); err == nil {
```
JoeWang1127, line 65: nit: do you think a flatten `if` is better than a `if-else if`?

JoeWang1127 (commented):
I feel we should ignore `Output` to in favor of `componentNaem` for php libraries to maintain backward compatibility.

Could we put on hold this change for now?

PR #7244, opened 2026-08-11
feat(internal/librarian/php): install and execute the post-processor

### LINKED ISSUE #7103: php(generate):  install and run php post-processor
Start noticing when testing.
(I did not see this in test early last week (?), but tools are pinned. Did not find out cause yet.)

For examples,
```
--- a/Ces/src/V1/WidgetTool.php
+++ b/Ces/src/V1/WidgetTool.php
@@ -1,5 +1,6 @@
 <?php
 # Generated by the protocol buffer compiler.  DO NOT EDIT!
+# NO CHECKED-IN PROTOBUF GENCODE
 # source: google/cloud/ces/v1/widget_tool.proto
```

sofisl:
This PR installs the PHP post-processor. Althugh it appears we are configuring a duplicate gapic-generator-php tool, these tools are actually distinct but colocated within the exact same repository. Thanks to our existing caching mechanism, the shared repository is only downloaded once. Therefore, this PR mainly adds tests to confirm the post-processor is run, and adds the post-processor as a tool in the migrate/librarian.yaml.

Fixes #7103


internal/librarian/php/install_test.go
```diff
@@ -113,7 +126,16 @@ func TestInstall(t *testing.T) {
...
 				if _, err := os.Stat(wrapperPath); err != nil {
-					t.Errorf("wrapper file %s not found: %v", wrapperPath, err)
+					t.Error(err)
+				}
+				postProcessorPath := filepath.Join(binDir, "php-post-processor")
+				info, err := os.Stat(postProcessorPath)
+				if err != nil {
+					t.Error(err)
+				} else {
+					if info.Mode().Perm() != 0o755 {
```
JoeWang1127: nit: can we flatten this if statement?

internal/librarian/php/postprocess_test.go
```diff
@@ -176,3 +190,62 @@ func TestPostProcess_CleanupError(t *testing.T) {
...
+	}
+	lib := &config.Library{
+		Name:   "SecretManager",
+		Output: destDir,
+		PHP: &config.PHPPackage{
+			ComponentName: "SecretManager",
+		},
+	}
+	err := postProcessLibrary(ctx, lib, lib.PHP.ComponentName)
+	if err == nil {
```
JoeWang1127: Do we still need this check?

internal/librarian/php/postprocess.go
```diff
@@ -49,5 +49,13 @@ func postProcessLibrary(ctx context.Context, library *config.Library, componentN
 	if err := command.RunInDir(ctx, library.Output, "python3", "owlbot.py"); err != nil {
 		return fmt.Errorf("failed to run owlbot.py: %w", err)
 	}
+	bin, err := binDir()
+	if err != nil {
+		return fmt.Errorf("failed to get bin dir: %w", err)
+	}
+	postProcessor := filepath.Join(bin, "php-post-processor")
+	if err := command.RunInDir(ctx, library.Output, postProcessor, "--input", "."); err != nil {
```
JoeWang1127: We need to use component name as library output to maintain backward compatibility, do we need to change `library.Output`?

It's fine we fix it in a follow up.

sofisl: No that's a good call! changed it and added a TODO to address once we change to use library.Output

JoeWang1127 (approved):

PR #7247, opened 2026-08-12
feat(internal/tool/protoc): add BinaryPath and BinaryPathOrSystem helpers

quirogas:
When repositories pin a `protoc` version in `librarian.yaml`, `librarian install` downloads and caches that exact binary in the Librarian cache directory. However, several generation pipelines (such as Node.js GAPIC generation) need to supply the `protoc` binary path as a CLI flag (`--protoc=<path>`) rather than executing it directly through `protoc.Run`. Previously, `internal/tool/protoc` only offered execution helpers, forcing generators to look up `"protoc"` in `$PATH` and causing non-deterministic output based on the host environment's ambient binary. This change introduces `BinaryPath` and `BinaryPathOrSystem` helper functions to resolve the absolute path to the Librarian-installed `protoc` binary matching the version configured in `librarian.yaml`, while falling back to system `$PATH` when unconfigured.

### Changes

- Added `BinaryPath(version string) (string, error)` to resolve the absolute path to the cached `protoc` binary installed by Librarian for a given version.
- Added `BinaryPathOrSystem(pc *config.Protoc) (string, error)` to resolve the path for a configured `tools.protoc` definition from `librarian.yaml` or look up `"protoc"` in `$PATH` if unconfigured.
- Added a fast-path file existence check in `Install` to skip redundant downloads when the binary is already cached.
- Updated `Run` and `RunOrSystem` to use `BinaryPath` and `BinaryPathOrSystem`.
- Added unit tests in `protoc_test.go`.


internal/tool/protoc/protoc_test.go
```diff
@@ -71,6 +74,178 @@ func TestInstallDir(t *testing.T) {
...
 
+func TestBinaryPath(t *testing.T) {
+	binaryName := protocBinaryName()
+	for _, test := range []struct {
+		name         string
+		version      string
+		librarianBin string
+		cacheDir     string
+		want         string
+		wantErr      bool
```
JoeWang1127: Could you separate the error case: https://github.com/googleapis/librarian/blob/main/doc/howwewritego.md#separate-error-tests

internal/tool/protoc/protoc_test.go
```diff
@@ -71,6 +74,178 @@ func TestInstallDir(t *testing.T) {
...
+					t.Errorf("mismatch (-want +got):\n%s", diff)
+				}
+			}
+		})
+	}
+}
+
+func TestBinaryPathOrSystem(t *testing.T) {
+	binaryName := protocBinaryName()
+
```
JoeWang1127, line 129: nit: remove this empty line.

internal/tool/protoc/protoc_test.go
```diff
@@ -71,6 +74,178 @@ func TestInstallDir(t *testing.T) {
...
+			},
+		},
+		{
+			name: "system protoc missing returns error",
+			pc:   nil,
+			setupPATH: func(t *testing.T) string {
+				t.Setenv("PATH", t.TempDir())
+				return ""
+			},
+			wantErr: true,
```
JoeWang1127: Could you separate this error case?

internal/tool/protoc/protoc_test.go
```diff
@@ -71,6 +74,178 @@ func TestInstallDir(t *testing.T) {
...
+	binaryName := protocBinaryName()
+	binDir := t.TempDir()
+	t.Setenv("LIBRARIAN_BIN", binDir)
+	version := "33.2"
+	protocDir := filepath.Join(binDir, "protoc", "v"+version, "bin")
+	if err := os.MkdirAll(protocDir, 0o755); err != nil {
+		t.Fatal(err)
+	}
+	testhelper.WriteExecutable(t, filepath.Join(protocDir, binaryName), "#!/bin/sh\nexit 0\n")
+	// An invalid SHA256 and unresolvable network call would fail if download was attempted
```
JoeWang1127: I'm wondering whether should we check the binary sha before execution, I thought about this but the sha in `librarian.yaml` is different than the binary sha.

This is out of scope of this pull request.

internal/tool/protoc/protoc_test.go
```diff
@@ -71,6 +74,178 @@ func TestInstallDir(t *testing.T) {
...
+	}
+}
+
+func TestRunOrSystem(t *testing.T) {
+	if runtime.GOOS == osWindows {
+		t.Skip("skipping execution test on Windows")
+	}
+	binaryName := protocBinaryName()
+	createFakeSystemExecutable(t, binaryName)
+
```
JoeWang1127, line 219: nit: remove this empty line.

internal/tool/protoc/protoc_test.go
```diff
@@ -135,6 +310,22 @@ func TestDownloadAndExtract(t *testing.T) {
 	}
 }
 
+func protocBinaryName() string {
```
JoeWang1127, line 289: t.Helper()

JoeWang1127 (commented):
Looks good, just some comments about tests.

internal/tool/protoc/protoc_test.go
quirogas: Can you explain why we should use  `t.Helper()` here. From my understanding we do not need to add one here because the function does  cause a test to fail or emit logs, at least that is my interpretation from reading Go style Best Practices.


> Test helpers are functions that do setup or cleanup tasks. All failures that occur in test helpers are expected to be failures of the environment (not from the code under test) — for example when a test database cannot be started because there are no more free ports on this machine. For functions like these, calling t.Helper is often appropriate to [mark them as a test helper](https://g3doc.corp.google.com/go/g3doc/style/decisions#mark-test-helpers). See [error handling in test helpers](https://g3doc.corp.google.com/go/g3doc/style/best-practices.md?cl=head#test-helper-error-handling) for more details.

I've added it regardless, but I'd appreciate if you could help me understand it better.

JoeWang1127: Ah yes, thanks for the explanation. This is indeed redundant.

JoeWang1127 (approved):

PR #7248, opened 2026-08-12
feat(internal/librarian/python): use protoc version from librarian.yaml

quirogas:
Python generation currently executes `"protoc"` directly from `$PATH` during API generation, ignoring any `tools.protoc.version` pinned in `librarian.yaml` and leading to non-deterministic generation and potential incompatibilities across different developer workstations and CI runners. This change passes the configured `tools.protoc` from `librarian.yaml` through `Generate` and `generateAPI`, using `protoc.BinaryPathOrSystem` to invoke the exact pinned `protoc` executable.

### Changes

- Extracted `cfg.Tools.Protoc` in `Generate` and passed it to `generateAPI`.
- Updated `generateAPI` to resolve the executable path via `protoc.BinaryPathOrSystem(pc)` before invoking `command.RunInDir`.
- Added unit tests in `generate_test.go` verifying execution of configured protoc binaries and error propagation when unconfigured binaries are missing from `$PATH`.

For https://github.com/googleapis/librarian/issues/6508


internal/librarian/python/generate_test.go
```diff
@@ -788,6 +799,40 @@ func TestGenerateAPI_Error(t *testing.T) {
...
+	pc := &config.Protoc{Version: version}
+	err := generateAPI(
+		t.Context(),
+		&config.API{Path: "google/cloud/secretmanager/v1"},
+		&config.Library{Name: "secretmanager", Output: repoRoot},
+		pc,
+		googleapisDir,
+		repoRoot,
+	)
+	if err == nil {
```
JoeWang1127: Do you need this check?

internal/librarian/python/generate_test.go
```diff
@@ -788,6 +799,40 @@ func TestGenerateAPI_Error(t *testing.T) {
...
+		t.Fatal("expected error from stub protoc, got nil")
+	}
+	var exitErr *exec.ExitError
+	if !errors.As(err, &exitErr) {
+		t.Fatalf("expected *exec.ExitError from stub protoc, got: %v", err)
+	}
+	if got, want := exitErr.ExitCode(), 42; got != want {
```
JoeWang1127, line 820: Why do you need this check, `exitErr` is predefined.

JoeWang1127 (approved):

internal/librarian/python/generate_test.go
quirogas: Yes, we need to make sure the error code is not another generic error code.

internal/librarian/python/generate_test.go
quirogas: we don't , than you for pointing that out.

PR #7266, opened 2026-08-13
fix(sidekick/rust): map empty `POST` and `PUT` bodies to `wkt::Empty`

haphungw:
Generate `None::<wkt::Empty>` whenever an HTTP `POST`, `PUT`, or `PATCH` request does not specify a body field in its annotations.

This ensures that we can send empty mutation requests, while `GET` and `DELETE` remain `None::<gaxi::http::NoBody>`, 

See https://github.com/googleapis/google-cloud-rust/pull/6364#pullrequestreview-4923253057


internal/sidekick/rust/codec.go
```diff
@@ -768,6 +768,12 @@ func (c *codec) fullyQualifiedEnumValueName(v *api.EnumValue, sourceSpecificatio
 
 func bodyAccessor(m *api.Method) string {
 	if m.PathInfo.BodyFieldPath == "" {
+		for _, binding := range m.PathInfo.Bindings {
```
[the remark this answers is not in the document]
haphungw: done

internal/sidekick/rust/codec.go
```diff
@@ -767,7 +767,16 @@ func (c *codec) fullyQualifiedEnumValueName(v *api.EnumValue, sourceSpecificatio
...
+		return "None::<gaxi::http::NoBody>"
+	}
 	if m.PathInfo.BodyFieldPath == "" {
+		for _, binding := range m.PathInfo.Bindings {
+			verb := strings.ToUpper(binding.Verb)
+			if verb == "POST" || verb == "PUT" || verb == "PATCH" {
+				return "None::<wkt::Empty>"
+			}
+		}
 		return "None::<gaxi::http::NoBody>"
```
dbolduc: You can directly swap out `gaxi::http::NoBody` for `wkt::Empty`.

Calling into `gaxi::http::handle_empty()` will deal with the verb[^1].

[^1]: You have the right instinct that this sort of logic should be done at code generation time vs. runtime, but that requires some re-architecting, which is probably not worth doing.

dbolduc (changes requested):

internal/sidekick/rust/codec.go
haphungw: ack

internal/sidekick/rust/codec.go
```diff
@@ -767,8 +767,8 @@ func (c *codec) fullyQualifiedEnumValueName(v *api.EnumValue, sourceSpecificatio
 }
 
 func bodyAccessor(m *api.Method) string {
-	if m.PathInfo.BodyFieldPath == "" {
-		return "None::<gaxi::http::NoBody>"
+	if m.PathInfo == nil || m.PathInfo.BodyFieldPath == "" {
```
dbolduc: why did we add `if m.PathInfo == nil` ? We did not need it before

dbolduc (approved):

internal/sidekick/rust/codec.go
haphungw: https://github.com/googleapis/librarian/pull/7266#discussion_r3773237689
this seems reasonable to me so i added it

dbolduc: Ah thanks for sharing.

> If `include-grpc-only-methods` is enabled, `generateMethod` can return `true` even if `m.PathInfo` is `nil`.

yes, but if it is a gRPC only method, we will not be dealing with an HTTP body.

> When generating code for such gRPC-only methods, `bodyAccessor` will be called.

No, it won't be called.

haphungw:
https://github.com/googleapis/google-cloud-rust/pull/6364 is a better approach to fix this.

PR #7272, opened 2026-08-13
fix(internal/librarian/php): run post-processor in API staging directories before owlbot

### LINKED ISSUE #7269: php: post-processor running on wrong output directory
The PHP diffs (missing build methods and leftover # NO CHECKED-IN PROTOBUF GENCODE headers) occur because librarian runs php-post-processor on the root staging directory (owl-bot-staging/<Library>) instead of the API-specific staging subdirectory (owl-bot-staging/<Library>/<StagingSubdir>).

The post-processor expects to find proto/src at the root of its working directory, but protoc outputs it inside the StagingSubdir (e.g. v1/proto/src).

Fix in librarian: In librarian/internal/librarian/php/postprocess.go, iterate over library.APIs and run php-post-processor inside filepath.Join(stagingDir, api.PHP.StagingSubdir) instead of the root stagingDir.

sofisl:
The PHP post-processor is now executed inside the API staging directories before owlbot.py is run. This ensures that the post-processor modifies the newly generated API files prior to any file movement or cleanup performed by owlbot.py.

Fixes https://github.com/googleapis/librarian/issues/7269

### DIFF
--- a/internal/librarian/php/postprocess.go
+++ b/internal/librarian/php/postprocess.go
@@ -37,7 +37,7 @@ func postProcessLibrary(ctx context.Context, library *config.Library, componentN
 			err = errors.Join(err, cleanupErr)
 		}
 	}()
-
+	
 	// TODO(https://github.com/googleapis/librarian/issues/7153): We need to use component name as library output to maintain backward compatibility. Change this to library.Output when ready.
 	owlbotPy := filepath.Join(componentName, "owlbot.py")
 	if _, err := os.Stat(owlbotPy); err != nil {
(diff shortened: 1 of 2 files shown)


internal/librarian/php/postprocess.go
```diff
@@ -47,16 +47,34 @@ func postProcessLibrary(ctx context.Context, library *config.Library, componentN
...
+	if err := command.RunInDir(ctx, componentName, "python3", "owlbot.py"); err != nil {
+		return fmt.Errorf("failed to run owlbot.py: %w", err)
+	}
+
+	return nil
+}
+
+func runPostProcessors(ctx context.Context, library *config.Library, stagingDir, postProcessor string) error {
+	for _, api := range library.APIs {
+		if api.PHP == nil {
```
JoeWang1127: I think the staging dir is filled before running generate: https://github.com/JoeWang1127/librarian/blob/main/internal/librarian/php/default.go#L31, so we don't need these checks

internal/librarian/php/postprocess.go
```diff
@@ -47,16 +47,34 @@ func postProcessLibrary(ctx context.Context, library *config.Library, componentN
...
+			continue
+		}
+		apiStagingDir := filepath.Join(stagingDir, api.PHP.StagingSubdir)
+		if _, err := os.Stat(apiStagingDir); err == nil {
```
JoeWang1127: Same as above, we can remove this check.

JoeWang1127 (approved):

PR #7307, opened 2026-08-17
fix(internal/filesystem): skip unzip execution for empty archives

### LINKED ISSUE #7308: php: generate failed with proto-only client
When generating `ShoppingCommonProtos`, the command failed with:
```
joewa@joewa-glinux-medium:~/workspace/google-cloud-php$ ../librarian/librarian generate shopping-type
librarian: generate library "shopping-type" (php): failed to extract generated output to owl-bot-staging/ShoppingCommonProtos: /usr/bin/unzip -q -o /tmp/librarian-php-1651765465/google_shopping_type-gapic.zip -d owl-bot-staging/ShoppingCommonProtos: warning [/tmp/librarian-php-1651765465/google_shopping_type-gapic.zip]:  zipfile is empty
```

JoeWang1127:
Unzip is updated to inspect the zip archive contents using archive/zip before invoking the system unzip binary, immediately returning nil if the archive contains zero files. The archive reader is closed before running the external unzip process to prevent holding open a file handle during command execution.

This prevents errors when unzipping empty archives and avoids unnecessary subprocess calls.

Fixes #7308

### DIFF
--- a/internal/filesystem/filesystem.go
+++ b/internal/filesystem/filesystem.go
@@ -148,7 +148,9 @@ func Unzip(ctx context.Context, src, dest string) error {
 	empty := len(r.File) == 0
 	// Do not use defer because we want to close the file before running unzip to avoid
 	// file contention.
-	r.Close()
+	if err := r.Close(); err != nil {
+		return err
+	}
 	if empty {
 		return nil
 	}
--- a/internal/filesystem/filesystem.go
+++ b/internal/filesystem/filesystem.go
@@ -145,8 +145,11 @@ func Unzip(ctx context.Context, src, dest string) error {
 	if err != nil {
 		return err
 	}
-	defer r.Close()
-	if len(r.File) == 0 {
+	empty := len(r.File) == 0
+	// Do not use defer because we want to close the file before running unzip to avoid
+	// file contention.
+	r.Close()
+	if empty {
 		return nil
 	}
 	return command.Run(ctx, "unzip", "-q", "-o", src, "-d", dest)
--- a/internal/filesystem/filesystem.go
+++ b/internal/filesystem/filesystem.go
@@ -146,8 +146,7 @@ func Unzip(ctx context.Context, src, dest string) error {
 		return err
 	}
 	empty := len(r.File) == 0
-	// Do not use defer because we want to close the file before running unzip to avoid
-	// file contention.
+	// Avoid file contention with potential unzip call by force closing the reader.
 	if err := r.Close(); err != nil {
 		return err
 	}
--- a/internal/filesystem/filesystem.go
+++ b/internal/filesystem/filesystem.go
@@ -16,6 +16,7 @@
 package filesystem
 
 import (
+	"archive/zip"
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


internal/filesystem/filesystem_test.go
```diff
@@ -356,6 +356,17 @@ func TestUnzip_Permissions(t *testing.T) {
...
 
+func TestUnzip_Empty(t *testing.T) {
+	t.Parallel()
+	tmp := t.TempDir()
+	zipPath := filepath.Join(tmp, "empty.zip")
+	destDir := filepath.Join(tmp, "dest")
+	createZip(t, zipPath, nil, nil)
+	if err := Unzip(t.Context(), zipPath, destDir); err != nil {
+		t.Fatalf("Unzip() error = %v", err)
+	}
```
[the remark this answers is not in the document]
JoeWang1127: Done.

internal/filesystem/filesystem.go
```diff
@@ -138,7 +139,19 @@ func CopyFile(src, dest string) error {
...
+	if err != nil {
+		return err
+	}
+	empty := len(r.File) == 0
+	// Do not use defer because we want to close the file before running unzip to avoid
+	// file contention.
+	r.Close()
+	if empty {
+		return nil
+	}
```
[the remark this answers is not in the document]
JoeWang1127: Go for a simple approach.

internal/filesystem/filesystem.go
```diff
@@ -138,7 +139,21 @@ func CopyFile(src, dest string) error {
...
+	empty := len(r.File) == 0
+	// Do not use defer because we want to close the file before running unzip to avoid
+	// file contention.
+	if err := r.Close(); err != nil {
+		return err
+	}
+	if empty {
+		return nil
+	}
```
noahdietz, line 155: Can this be moved to make things read more linearly?
```suggestion
	// Do not use defer because we want to close the file before running unzip to avoid
	// file contention.
	if err := r.Close(); err != nil {
		return err
	}
	if len(r.File) == 0 {
		return nil
	}
```

Not sure if `r.File` is still set after `r.Close()`. Mind checking?

internal/filesystem/filesystem.go
```diff
@@ -138,7 +139,21 @@ func CopyFile(src, dest string) error {
...
 // Unzip unzips the src archive into dest directory using the system unzip command.
+// If the archive contains 0 files, it returns nil without invoking unzip.
 func Unzip(ctx context.Context, src, dest string) error {
+	r, err := zip.OpenReader(src)
+	if err != nil {
+		return err
+	}
+	empty := len(r.File) == 0
+	// Do not use defer because we want to close the file before running unzip to avoid
+	// file contention.
```
noahdietz:
```suggestion
	// Avoid file contention with potential unzip call by force closing the reader.
```
Nit: simplify comment, focus on why, not "the code we aren't using"

codyoss (approved):

internal/filesystem/filesystem.go
JoeWang1127: Done.

internal/filesystem/filesystem.go
JoeWang1127: Yes, I think we can read the metadata (r.File) after closing it.

However, idiomatically and for readability, inspecting fields of a resource after closing it is considered a use-after-close code smell.

Capturing `empty := len(r.File) == 0` before calling `r.Close()` is clearer and avoids any reader/linter confusion.

noahdietz: Fair enough

noahdietz (approved):

PR #7312, opened 2026-08-18
feat(internal/sidekick/rust): refactor bidi streaming templates

suzmue:
Refactor bidirectional streaming templates in Rust sidekick to lazily initialize the transport and stub synchronously. Bidirectional streaming builders now wrap BidiStreamBuilder without storing an initial request or generating request field setters. All requests are sent asynchronously through RequestSender.

The existing implementation was broken because services may not send RepsonseHeaders until multiple requests were sent, causing a deadlock that could not be worked around.

For #6835

### DIFF
--- a/internal/sidekick/rust/generate_bidi_streaming_test.go
+++ b/internal/sidekick/rust/generate_bidi_streaming_test.go
@@ -17,7 +17,6 @@ package rust
 import (
 	"os"
 	"path/filepath"
-	"strconv"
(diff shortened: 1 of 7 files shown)


suzmue:
https://github.com/googleapis/google-cloud-rust/pull/6436 is the changes in google-cloud-rust.

dbolduc (approved):

PR #7319, opened 2026-08-18
feat(swift): support service name overrides for doc link resolution

nolanleastin:
for https://github.com/googleapis/google-cloud-swift/issues/308 

generates https://github.com/googleapis/google-cloud-swift/pull/471

### DIFF
--- a/internal/config/swift.go
+++ b/internal/config/swift.go
@@ -71,7 +71,8 @@ type SwiftPackage struct {
 	// other packages.
 	PackageNameOverride string `yaml:"package_name_override,omitempty"`
 
-	// NameOverrides contains codec-level overrides for type and service names (e.g., {".google.storage.v2.Storage": "StorageControl"}).
+	// NameOverrides contains codec-level overrides for service names (e.g., {".google.storage.v2.Storage": "StorageControl"}).
+	// TODO(https://github.com/googleapis/google-cloud-swift/issues/308): Support overriding other symbol types (e.g., messages, enums, oneofs) if needed.
 	NameOverrides map[string]string `yaml:"name_overrides,omitempty"`
 
 	// PerServiceTraits enables per-service compile-time flags.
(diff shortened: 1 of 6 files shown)


internal/config/swift.go
```diff
@@ -71,6 +71,9 @@ type SwiftPackage struct {
 	// other packages.
 	PackageNameOverride string `yaml:"package_name_override,omitempty"`
 
+	// NameOverrides contains codec-level overrides for type and service names (e.g., ".google.storage.v2.Storage=StorageAdmin").
+	NameOverrides string `yaml:"name_overrides,omitempty"`
```
coryan: Did you copy this from Rust?  For historical reasons Rust uses strings for all its configs, but we can use something with a better structure here, maybe `map[string]string` ?

internal/config/swift.go
```diff
@@ -171,6 +174,9 @@ type SwiftModule struct {
 
 	// ModulePath is the module import path or target containing stubs (used by convert-swift and gRPC transports).
 	ModulePath string `yaml:"module_path,omitempty"`
+
+	// NameOverrides contains codec-level overrides for type and service names for this module.
+	NameOverrides string `yaml:"name_overrides,omitempty"`
```
coryan: Ditto.

internal/config/swift.go
coryan: Also it seems that only service names are supported at the moment, that is fine, but maybe document that with a TODO here or in the codec?

coryan (changes requested):

internal/sidekick/swift/codec.go
```diff
@@ -215,6 +227,22 @@ func newCodec(model *api.API, library *config.Library, module *config.SwiftModul
...
+	}
+	if nameOverridesStr != "" {
+		result.NameOverrides = make(map[string]string)
+		for override := range strings.SplitSeq(nameOverridesStr, ",") {
+			tokens := strings.Split(override, "=")
+			if len(tokens) == 2 {
+				result.NameOverrides[tokens[0]] = tokens[1]
+			}
+		}
+	}
```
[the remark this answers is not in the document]
nolanleastin: old

coryan (approved):

PR #7326, opened 2026-08-18
feat(migrate/php): extract oldest copyright year for PHP libraries

### LINKED ISSUE #6795: php(generate): survey copyright year status and add to config
`google-cloud-php` does not use current year as Java does. We need to specify a copyright year in librarian.yaml per library.

However, from initial survey, looks like within a single library, files may have different copyright year.

Looking at a few examples:
 For the  `SecretManager`   library, 
 1. All generated client code under  src/  has a uniform year:
  Every single generated client file and descriptor config under  src/V1/  and  src/V1beta2/  had  Copyright 2024 . This indicates that the last time these APIs were
  generated and committed to the repository was in 2024.
  2. Mixed years exist in samples/snippets:
  The  2022  year is present in some generated sample snippet files (e.g.  samples/V1/SecretManagerServiceClient/enable_secret_version.php ), which were generated in 2022
  when those specific snippets were introduced and have not been updated since.

The  Vision  library follows the exact same pattern:

  1. Client & Config Files ( src/V1/ ):
  All files under  and  are uniform and have  Copyright 2024 , indicating the main client generation was last executed in 2024.
  2. Snippets & Samples ( samples/ ):
  The generated sample snippet files under  contain  Copyright 2022 , showing they have not been regenerated since their original snippet creation run in 2022.

  For the  `GeminiDataAnalytics`  library, we see the following copyright years under  src/ has a Versioned Split:
  •  V1beta  Client & Config Files:
  All files under  and  have  Copyright 2025 .
  •  V1  Client & Config Files:
  All files under  and  have  Copyright 2026 .

Needs a bit more analysis, but likely we need to consider updating copyright year in google-cloud-php to per library. Alternatively, let librarian to support CopyrightYear per API path.

sofisl:
Librarian's clean step deletes the target directory before generation to remove stale files, which breaks the legacy synthtool/owlbot.py copyright year preservation logic and leads to yearly copyright churn. To fix this, we've introduced the `copyright_year` field in `librarian.yaml` to explicitly set copyright years during post-processing. This decision is made in accordance with the doc "[Migrating PHP to Librarian](https://docs.google.com/document/d/1kxoL0p8JBqwTXoz4mQJH3bj3jHIOPJrSrsvoUZBgyUk/edit?tab=t.0#heading=h.6aw374b79k8s)".

Fixes #6795

### DIFF
--- a/internal/license/license.go
+++ b/internal/license/license.go
@@ -18,6 +18,7 @@ package license
 import (
 	"bytes"
 	"fmt"
+	"regexp"
 )
 
 // HasHeader returns true if the content contains the Apache 2.0 license header.
@@ -49,4 +50,21 @@ func HeaderBulk() []string {
 		" See the License for the specific language governing permissions and",
 		" limitations under the License.",
 	}
+}
+
+var copyrightYearRegexp = regexp.MustCompile(`Copyright (\d{4})`)
+
+// OldestYear returns the oldest copyright year found in the content.
+// It returns an empty string if no copyright year is found.
+func OldestYear(content []byte) string {
+	var oldest string
+	for _, match := range copyrightYearRegexp.FindAllSubmatch(content, -1) {
+		if len(match) > 1 {
+			year := string(match[1])
+			if oldest == "" || year < oldest {
+				oldest = year
+			}
+		}
+	}
+	return oldest
 }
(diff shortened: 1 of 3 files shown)


sofisl:
Let's wait for the response in  https://github.com/googleapis/librarian/issues/6795

PR #7331, opened 2026-08-19
impl(internal/sidekick/rust): use bidi_streaming helper in transport

suzmue:
Update the Rust transport template for bidirectional streaming RPCs to delegate to the new bidi_streaming helper method on gaxi::grpc::Client instead of emitting inline channel and conversion boilerplate.

For #6835


suzmue:
Using https://github.com/googleapis/google-cloud-rust/pull/6457

coryan (approved):

PR #7343, opened 2026-08-19
feat(migrate/php): strip PA namespace during migration for non-generic APIs

### LINKED ISSUE #7330: php: strip all directories from library names, not just google-cloud
In librarian.yaml, we generate library names by stripping google/cloud. however that leaves some library names that aren't in google/cloud, like google/identity/accesscontextmanager, as they are: identity-accesscontextmanager. we should strip any PA function basically.

sofisl:
This PR modifies the PHP migration tool to strip the Product Area (PA) namespace from the generated library name. 
Previously, the PA was only stripped for `cloud` APIs. This updates the tool to strip the PA for any `google/` namespace (e.g., `google/identity/accesscontextmanager/v1` -> `accesscontextmanager`), ensuring consistent library naming across all Product Areas. I manually ran migration and this script will cover all current libraries in google-cloud-php

Fixes  https://github.com/googleapis/librarian/issues/7330

### DIFF
--- a/tool/cmd/migrate/php.go
+++ b/tool/cmd/migrate/php.go
@@ -22,7 +22,6 @@ import (
 	"io/fs"
 	"log"
 	"os"
-	"path"
 	"path/filepath"
 	"regexp"
 	"slices"
@@ -32,7 +31,6 @@ import (
 	"github.com/googleapis/librarian/internal/config"
 	"github.com/googleapis/librarian/internal/librarian"
 	"github.com/googleapis/librarian/internal/librarian/php"
-	"github.com/googleapis/librarian/internal/serviceconfig"
 	"github.com/googleapis/librarian/internal/yaml"
 )
 
@@ -262,7 +260,7 @@ func findPHPLibraries(repoPath string, googleapisDir string, globalDefaultCommon
 		if len(apis) == 0 {
 			continue
 		}
-		libraryName := deriveMigratedLibraryName(apis[0].Path)
+		libraryName := php.DefaultLibraryName(apis[0].Path)
 		lib := &config.Library{
 			Name:    libraryName,
 			Version: version,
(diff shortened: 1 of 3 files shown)


tool/cmd/migrate/php.go
```diff
@@ -260,7 +262,7 @@ func findPHPLibraries(repoPath string, googleapisDir string, globalDefaultCommon
 		if len(apis) == 0 {
 			continue
 		}
-		libraryName := php.DefaultLibraryName(apis[0].Path)
+		libraryName := deriveMigratedLibraryName(apis[0].Path)
```
JoeWang1127: Do we need to update `php.DefaultLibraryName` so that further library has the same naming pattern?

tool/cmd/migrate/php.go
```diff
@@ -370,3 +372,24 @@ func normalizeStagingSubdir(apiPath, stagingDir string) string {
...
 }
+
+// deriveMigratedLibraryName derives the library name from the API path for migration.
+// It strips the PA segment (e.g. cloud, identity, geo) to conform to the new naming scheme.
+func deriveMigratedLibraryName(apiPath string) string {
+	if serviceconfig.ExtractVersion(apiPath) != "" {
+		apiPath = path.Dir(apiPath)
+	}
+	parts := strings.Split(apiPath, "/")
+	if len(parts) >= 3 && parts[0] == "google" {
```
JoeWang1127: I think this implementation is a bit hard to understand, how about stripping known prefix, e.g., `google/cloud/`, `google/identity`?

sofisl: I guess what I don't love about that implementation is that we'd have to update it every time we encounter a new PA use case. In this way we don't have to update this again, and since we will only use this code once, I'm willing to sacrifice some readability for a one-shot solution :) I tested it manually and it does cover all our use cases currently!

tool/cmd/migrate/php.go
sofisl: good call!

sofisl:
Ah, I didn't realize this was done inconsistently in PHP. closing!

PR #7347, opened 2026-08-19
fix(internal/sidekick/rust): use anyhow::Result in bidi streaming samples

suzmue:
Update bidi streaming doc comment examples to use anyhow::Result instead of crate::Error to handle the SendError type returned by send()

### DIFF
--- a/internal/sidekick/rust/templates/crate/src/builder.rs.mustache
+++ b/internal/sidekick/rust/templates/crate/src/builder.rs.mustache
@@ -107,13 +107,15 @@ pub mod {{Codec.ModuleName}} {
     /// # use {{Model.Codec.PackageNamespace}}::builder::{{Service.Codec.ModuleName}}::{{Codec.BuilderName}};
     {{#Codec.IsBidiStreaming}}
     /// # use {{InputType.Codec.NameInExamples}};
+    /// # use google_cloud_gax::streaming::SendError;
     {{/Codec.IsBidiStreaming}}
     /// # async fn sample() -> {{Model.Codec.PackageNamespace}}::Result<()> {
     {{#Codec.IsBidiStreaming}}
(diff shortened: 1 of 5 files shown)


suzmue:
Depends on the changes in https://github.com/googleapis/google-cloud-rust/pull/6475 being submitted.

internal/sidekick/rust/templates/crate/src/builder.rs.mustache
```diff
@@ -107,13 +107,15 @@ pub mod {{Codec.ModuleName}} {
     /// # use {{Model.Codec.PackageNamespace}}::builder::{{Service.Codec.ModuleName}}::{{Codec.BuilderName}};
     {{#Codec.IsBidiStreaming}}
     /// # use {{InputType.Codec.NameInExamples}};
+    /// # use google_cloud_gax::streaming::SendError;
     {{/Codec.IsBidiStreaming}}
     /// # async fn sample() -> {{Model.Codec.PackageNamespace}}::Result<()> {
```
coryan, line 113: Alternatively:


```suggestion
    /// # async fn sample() -> anyhow::Result<()> {
```

That should be available already: 

https://github.com/googleapis/google-cloud-rust/blob/9ed933971e029df574bb76eac2a5360b727e3f26/src/generated/cloud/secretmanager/v1/Cargo.toml#L57-L58

coryan (changes requested):
I think I would prefer to change the return type and keep the example simpler. Thoughts?

internal/sidekick/rust/templates/crate/src/builder.rs.mustache
suzmue: SGTM

coryan (approved):

PR #7395, opened 2026-08-25
fix(rust): identfy new crates better

### LINKED ISSUE #7124: rust: Unblock release of `google-cloud-bigquery-v2`
We shouldn't run semver checks for any new crates.

We currently say a new crate is one that has a new manifest file:

https://github.com/googleapis/librarian/blob/0eb263a544ce5c99ca2ffc4f348570bd980de62a/internal/librarian/rust/publish.go#L166-L169

This works for like 99% of the libraries we onboard. We just happened to sit on `google-cloud-bigquery-v2` for a while. The manifest existed in the last release, it is just that the publish bit was `false`.

A robust fix probably looks like trying a GET request against crates.io to see if the crate exists.

I am also fine with a workaround because this problem is rare. Like parsing the error looking for "crate not found" or even a local change that says "if bigquery ..." 😱

dbolduc:
Find out a crate exists by checking `cargo info <crate> --registry crates-io` instead of checking if there is a new manifest file since the last release. This lets us add crates with `publish = false`, sit on them for a release, and publish them later.

Note that this change is also more accurate for crates that we are taking ownership of. e.g. `google-cloud-bigquery` is new to our repo, but there is a published `0.15.0` version we need to be semver compatible with. (We are, we bump the major)

Tested a dry run locally and it says:

```diff
--- a/internal/librarian/rust/publish.go
+++ b/internal/librarian/rust/publish.go
@@ -135,6 +135,7 @@ func publishCrates(ctx context.Context, params PublishParams, lastTag string, fi
                        return err
                }
        }
+       return fmt.Errorf("DEBUG")
        args := []string{"workspaces", "publish", "--skip-published", "--publish-interval=60", "--no-git-commit", "--from-git", "skip"}
        if params.DryRunKeepGoing {
                args = append(args, "--dry-run", "--keep-going")
@@ -169,11 +170,14 @@ func semverCheck(ctx context.Context, semverData semverData, name string) error
                if exitErr, ok := errors.AsType[*exec.ExitError](err); ok {
                        if exitErr.ExitCode() == 101 && strings.Contains(err.Error(), "could not find") {
                                // The crate was not found on crates.io, so we can skip the semver check.
+                               fmt.Println("SKIP: ", name)
                                return nil
                        }
                }
                return err
        }
+       fmt.Println("would check: ", name)
+       return nil
        if semverData.verbose {
                err = command.RunStreaming(ctx, command.Cargo, "semver-checks", "--all-features", "-p", name)
        } else {
```

and it correctly skips `google-cloud-bigquery-v2`:
```
SKIP:  google-cloud-bigquery-v2
```

Fixes #7124

### DIFF
--- a/internal/librarian/rust/publish.go
+++ b/internal/librarian/rust/publish.go
@@ -164,7 +164,7 @@ func runSemverChecks(ctx context.Context, semverData semverData) error {
 
 // semverCheck runs semver checks for a specific crate.
 func semverCheck(ctx context.Context, semverData semverData, name string) error {
-	_, err := command.Output(ctx, command.Cargo, "info", name, "--registry", "crates-io")
+	err := command.Run(ctx, command.Cargo, "info", name, "--registry", "crates-io")
 	if err != nil {
 		if exitErr, ok := errors.AsType[*exec.ExitError](err); ok {
 			if exitErr.ExitCode() == 101 && strings.Contains(err.Error(), "could not find") {
--- a/internal/librarian/rust/publish.go
+++ b/internal/librarian/rust/publish.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"os/exec"
 	"runtime"
 	"slices"
 	"strings"
@@ -150,9 +151,9 @@ func publishCrates(ctx context.Context, params PublishParams, lastTag string, fi
 func runSemverChecks(ctx context.Context, semverData semverData) error {
 	group, ctx := errgroup.WithContext(ctx)
 	group.SetLimit(max(runtime.NumCPU()/semverCheckCPUDivisor, 1))
-	for name, manifest := range semverData.manifests {
+	for name := range semverData.manifests {
 		group.Go(func() error {
-			if err := semverCheck(ctx, semverData, name, manifest); err != nil {
+			if err := semverCheck(ctx, semverData, name); err != nil {
 				return fmt.Errorf("%s: %w: %v", name, errSemverCheck, err)
 			}
 			return nil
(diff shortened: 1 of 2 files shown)


internal/librarian/rust/publish.go
```diff
@@ -162,12 +163,17 @@ func runSemverChecks(ctx context.Context, semverData semverData) error {
 }
 
 // semverCheck runs semver checks for a specific crate.
-func semverCheck(ctx context.Context, semverData semverData, name string, manifest string) error {
-	if git.IsNewFile(ctx, command.Git, semverData.lastTag, manifest) {
-		// If the manifest is new, we can skip semver checks, since there is no previous version to compare against.
-		return nil
+func semverCheck(ctx context.Context, semverData semverData, name string) error {
+	_, err := command.Output(ctx, command.Cargo, "info", name, "--registry", "crates-io")
```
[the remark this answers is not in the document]
dbolduc: Done. I am YOLOing this.

internal/librarian/rust/publish.go
```diff
@@ -162,12 +163,17 @@ func runSemverChecks(ctx context.Context, semverData semverData) error {
...
+func semverCheck(ctx context.Context, semverData semverData, name string) error {
+	err := command.Run(ctx, command.Cargo, "info", name, "--registry", "crates-io")
+	if err != nil {
+		if exitErr, ok := errors.AsType[*exec.ExitError](err); ok {
+			if exitErr.ExitCode() == 101 && strings.Contains(err.Error(), "could not find") {
+				// The crate was not found on crates.io, so we can skip the semver check.
+				return nil
+			}
+		}
+		return err
```
suzmue, line 175: Alternative: A lighter weight solution than crates.info to query index.crates.io directly, though more brittle.

Something like:

```suggestion
    // All our crates are google-cloud-* (len >= 4), so sparse index layout is stable:
	// https://index.crates.io/{name[0:2]}/{name[2:4]}/{name}
	url := fmt.Sprintf("https://index.crates.io/%s/%s/%s", name[:2], name[2:4], name)

	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
	if err != nil {
		return err
	}

	resp, err := http.DefaultClient.Do(req)
	if err != nil {
		return err
	}
	defer resp.Body.Close()

	switch resp.StatusCode {
	case http.StatusNotFound:
		// New crate, nothing published yet.
		return nil
	case http.StatusOK:
		// Published crate; continue to semver checks.
	default:
		return fmt.Errorf("crates index lookup failed for %q: status %d", name, resp.StatusCode)
	}
```

suzmue (approved):
PR title: typo (s/identfy/identify)

quirogas (approved):

internal/librarian/rust/publish.go
dbolduc: Why is this better?

There is no getting around `cargo`. We `cargo semver-check` on the next line.

dbolduc: To answer my own question: it is probably the running commands bit.

PR #7397, opened 2026-08-25
feat(tool/cmd/migrate/php): support longrunning operations in PHP

### LINKED ISSUE #7369: php: owlbot did not remove obsolete LRO methods in Dataform
# Title: 
[Dataform] Clean up orphaned LRO forwarding methods (Breaking Change)

# Description:
When migrating `google-cloud-php` generation from OwlBot to Librarian, Librarian's strict clean-generation process surfaced orphaned methods in `DataformClient.php` that should have been deleted previously. 

Specifically, the following LRO convenience methods (and their corresponding samples) are being deleted:
- `cancelOperation()`
- `deleteOperation()`
- `getOperation()`
- `listOperations()`

### Background
In the past, `gapic-generator-php` had a behavior where if an API specified `google.longrunning.Operations` as a mixin (which `Dataform` does, as `MoveFolder` returns an Operation), it would expose standard LRO methods directly on the main client class (`DataformClient`). 

In `gapic-generator-php` PR #834, the legacy V1 surface generation logic was removed. The new V2 surface correctly mandates that users interact with LROs exclusively via the `OperationsClient` (accessed via `$client->getOperationsClient()`), and stops generating these forwarding methods on the main client. 

However, because OwlBot only overwrites or adds files—and does not perform structural deletions—the old `cancelOperation` methods were left orphaned in the `google-cloud-php` repository even after they stopped being generated.

### Impact / Breaking Change
Librarian correctly removes these orphaned methods. This is a **client-side breaking change**. Any user who currently calls `$client->cancelOperation(...)` will encounter a PHP Fatal Error (`Call to undefined method`) upon upgrading. 

**Migration Path:**
Users must migrate their code to use the operations client. For example:
```diff
- $client->cancelOperation($request);
+ $client->getOperationsClient()->cancelOperation($request);
```

### Affected Files
* `Dataform/src/V1/Client/DataformClient.php`
* `Dataform/src/V1beta1/Client/DataformClient.php`
* `Dataform/samples/V1/DataformClient/cancel_operation.php`
* `Dataform/samples/V1/DataformClient/delete_operation.php`
* `Dataform/samples/V1/DataformClient/get_operation.php`
* `Dataform/samples/V1/DataformClient/list_operations.php`
* `Dataform/samples/V1beta1/DataformClient/cancel_operation.php`
* `Dataform/samples/V1beta1/DataformClient/delete_operation.php`
* `Dataform/samples/V1beta1/DataformClient/get_operation.php`
* `Dataform/samples/V1beta1/DataformClient/list_operations.php`

### Alternative Option: Preservation
If we determine this breaking change is too disruptive, we could theoretically preserve these methods. However, because Librarian's `keep` rules in `librarian.yaml` operate at the *file* level, we cannot simply `keep` the `DataformClient.php` file without losing all future generated updates for the main client. 

To properly preserve backwards compatibility, we would need to:
1. Revert the generator change or add a configuration option in `gapic-generator-php` to explicitly re-enable legacy LRO forwarding for this specific client.
2. OR implement a post-processing script to manually inject these forwarding methods back into the generated `DataformClient.php` file during generation.

sofisl:
This PR updates the migrate tool to allow generation of longrunning operation methods in PHP by mapping the LRO proto endpoints. This reverses previous hardcoded ignorance of LROs.

Ran `librarian generate --all` and the only LRO changes we found was for LongRunning,  but it fixes the diff in Dataform. I think this is OK since @JoeWang1127 shows a diff for LongRunning in the PHP Migration tracker.

Fixes #7369

### DIFF
--- a/tool/cmd/migrate/php.go
+++ b/tool/cmd/migrate/php.go
@@ -38,7 +38,7 @@ import (
 var librarianPHPYAML []byte
 
 var protoMappings = map[string]string{
-	"//google/cloud/location:location_proto": "google/cloud/location/locations.proto",
+	"//google/cloud/location:location_proto":     "google/cloud/location/locations.proto",
 	"//google/iam/v1:iam_policy_proto":           "google/iam/v1/iam_policy.proto",
 	"//google/longrunning:longrunning_php_proto": "google/longrunning/operations.proto",
 	"//google/longrunning:operations_proto":      "google/longrunning/operations.proto",
--- a/tool/cmd/migrate/php.go
+++ b/tool/cmd/migrate/php.go
@@ -41,7 +41,6 @@ var protoMappings = map[string]string{
 	"//google/cloud/location:location_proto":     "google/cloud/location/locations.proto",
 	"//google/iam/v1:iam_policy_proto":           "google/iam/v1/iam_policy.proto",
 	"//google/longrunning:longrunning_php_proto": "google/longrunning/operations.proto",
-	"//google/longrunning:operations_proto":      "google/longrunning/operations.proto",
 }
 
 // protoPackageOverrides maps API paths to explicit proto_package overrides.
--- a/tool/cmd/migrate/php.go
+++ b/tool/cmd/migrate/php.go
@@ -39,7 +39,9 @@ var librarianPHPYAML []byte
 
 var protoMappings = map[string]string{
 	"//google/cloud/location:location_proto": "google/cloud/location/locations.proto",
-	"//google/iam/v1:iam_policy_proto":       "google/iam/v1/iam_policy.proto",
+	"//google/iam/v1:iam_policy_proto":           "google/iam/v1/iam_policy.proto",
+	"//google/longrunning:longrunning_php_proto": "google/longrunning/operations.proto",
+	"//google/longrunning:operations_proto":      "google/longrunning/operations.proto",
 }
 
 // protoPackageOverrides maps API paths to explicit proto_package overrides.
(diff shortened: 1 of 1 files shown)


sofisl:
@JoeWang1127  Just a quick update: we ran `librarian generate --all` with this change, and the only LRO-specific diff we found was for `LongRunning` (which you already noted in the sheet). 

Because the migrate script currently overwrites the `keep` list for `LongRunning`, the generated output ends up deleting its handwritten files. Should we open a separate PR to hardcode preserving the `keep` list for `LongRunning` ?

tool/cmd/migrate/php.go
```diff
@@ -38,8 +38,10 @@ import (
 var librarianPHPYAML []byte
 
 var protoMappings = map[string]string{
-	"//google/cloud/location:location_proto": "google/cloud/location/locations.proto",
-	"//google/iam/v1:iam_policy_proto":       "google/iam/v1/iam_policy.proto",
+	"//google/cloud/location:location_proto":     "google/cloud/location/locations.proto",
+	"//google/iam/v1:iam_policy_proto":           "google/iam/v1/iam_policy.proto",
+	"//google/longrunning:longrunning_php_proto": "google/longrunning/operations.proto",
+	"//google/longrunning:operations_proto":      "google/longrunning/operations.proto",
```
JoeWang1127: Can we only add `//google/longrunning:longrunning_php_proto` to this mapping? I'm worried this line can affect other libraries as well.

JoeWang1127 (approved):

PR #7407, opened 2026-08-26
fix(tool/cmd/migrate/php): preserve Logging smoketest

### LINKED ISSUE #7409: php: preserve Logging's smoke test
since Logging is a hybrid library, it has some handwritten files, namely, a smoke test. We should add it to the migration script to include it.

sofisl:
Logging is a hybrid library that contains a handwritten system test (`tests/System/V2/LoggingServiceV2SmokeTest.php`). The migration script was dropping the `keep` configuration for it, causing `librarian generate logging` to accidentally delete the test file.

This PR adds `Logging` to the migration script `keep` exemptions so the file is preserved during generation.

Fixes #7409

### DIFF
--- a/tool/cmd/migrate/php.go
+++ b/tool/cmd/migrate/php.go
@@ -494,6 +494,8 @@ func specialCases(api *config.API) {
 
 func appendKeep(lib *config.Library) {
 	switch lib.Name {
+	case "logging":
+		lib.Keep = append(lib.Keep, "tests/System/V2/LoggingServiceV2SmokeTest.php")
 	case "bigtable":
 		lib.Keep = append(lib.Keep, "tests/Conformance/proxy/src/GPBMetadata/Google/Bigtable/Testproxy/TestProxy.php")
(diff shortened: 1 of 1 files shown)


tool/cmd/migrate/php.go
```diff
@@ -265,17 +265,13 @@ func findPHPLibraries(repoPath string, googleapisDir string, globalDefaultCommon
...
+			Version: version,
+			APIs:    apis,
+			Output:  name,
 		}
-		if name == "Datastore" {
-			lib.Keep = []string{"src/V1/TransactionOptions/ReadOnly.php"}
-		}
-		if name == "ErrorReporting" {
-			lib.Keep = []string{"tests/System/V1beta1/ReportErrorsServiceSmokeTest.php"}
+		if name == "Logging" {
```
JoeWang1127: The name should be `logging`.

JoeWang1127: Also, could you add this case to `appendKeep` function?

JoeWang1127 (approved):

PR #7413, opened 2026-08-26
fix(librarian): add validates path existence in sources

### LINKED ISSUE #7412: librarian: add should validate api path existence at target googleapis commit
Currently, `librairan add google/gshoe/v1` will work, and dutifully add a librarian entry and any supporting config entries for `google/gshoe/v1`.

The problem is, that path does not exist at all, nor does it exist at the `sources.googleapis.commit` of an SDK repo.

The solution is to verify that a given path exists in the `sources.googleapis` downloaded at the target `commit` as configured in `librarian.yaml`.

This also means that the `add`  command will need to ensure the `sources` are downloaded, using the cache of course.

noahdietz:
Introduces input API path existence validation to the `librarian add` command. 

When run, librarian will fetch the `sources.googleapis` configured in `librarian.yaml` (or use what already exists in the cache), and check if the given API path exists in the sources. If it does not, it will fail.

This means that calls to `librarian add` will now download `sources.googleapis` if necessary, when it wouldn't have before.

This PR also refactors some of the `add` tests to split between normal and error cases.

Fixes https://github.com/googleapis/librarian/issues/7412


internal/librarian/add.go
```diff
@@ -130,6 +136,32 @@ func runAdd(ctx context.Context, cfg *config.Config, api, explicitLibraryName st
 	return RunTidyOnConfig(ctx, ".", cfg)
 }
 
+// validateAPIPathExistence verifies that the given API path exists as a directory
+// in the configured googleapis source repository.
+func validateAPIPathExistence(ctx context.Context, cfg *config.Config, api string) error {
+	if api == "" {
+		return fmt.Errorf("%w: %s", errAPINotFound, api)
+	}
+	if cfg.Sources == nil || cfg.Sources.Googleapis == nil {
```
sofisl: This *feels* like it should be a validation check for the actual librarian.yaml, it shouldn't be happening in this lower level function.

```
Right now, the codebase checks sources.googleapis redundantly across multiple places with inconsistent errors:

RunTidyOnConfig
 returns errNoGoogleapiSourceInfo
LoadSources
 returns ErrMissingGoogleapisSource
PR 7413 in 
validateAPIPathExistence
 returns ErrMissingGoogleapisSource
Unifying this into a single validateSources(cfg.Sources) or a top-level validateConfig(cfg) would eliminate this duplication.
```

sofisl (commented):
When you attempt to add a nonexistent library now, it usually fails at generation with "protos not found". Should we change that code since we should be catching this at a higher level (i.e., maybe just returning the raw error, since it should no longer fail like this anymore?)

internal/librarian/add.go
noahdietz: Good point. I was between making this "standalone" in that it validates all expected inputs and can be called from anywhere, but that is overkill. Let me clean this up

noahdietz: Done.

sofisl (approved):

noahdietz:
> When you attempt to add a nonexistent library now, it usually fails at generation with "protos not found". Should we change that code since we should be catching this at a higher level (i.e., maybe just returning the raw error, since it should no longer fail like this anymore?)

I think this error mode is still valid, at least for now. This PR introduces a path existence check, but it does not validate its contents. That could be changed in the future to front load it, but I'm not inclined to move it just yet. This is a good thought though.

PR #7420, opened 2026-08-27
fix(internal/librarian/java): validate required libraries google-cloud-java and google-cloud-pom-parent

### LINKED ISSUE #6615: java: tidy should validate google-cloud-java and google-cloud-pom-parent present in config
The postgenerate step generates repo level pom.xml files, that sources version of google-cloud-java and google-cloud-pom-parent  from the library list.
These two libraries and their version are required. Check with tidy validate step.

sofisl:
The postgenerate step generates repo level pom.xml files, that sources versions of google-cloud-java and google-cloud-pom-parent from the library list. These two libraries and their version are required. Added a check within the tidy validate step to enforce this.

Fixes #6615

### DIFF
--- a/internal/librarian/java/default.go
+++ b/internal/librarian/java/default.go
@@ -210,6 +210,10 @@ var (
 		"google/iam/v3":     true,
 		"google/iam/v3beta": true,
 	}
+	// errLibraryNotFound is returned when a required library is not found in librarian.yaml.
+	errLibraryNotFound = errors.New("library not found in librarian.yaml")
+	// errLibraryMissingVersion is returned when a required library is missing a version.
+	errLibraryMissingVersion = errors.New("library must have a version")
 )
 
 // Validate checks that the Java-specific configuration for a library and global config is
@@ -220,8 +224,10 @@ func Validate(cfg *config.Config) error {
 	if cfg.Default == nil || cfg.Default.Java == nil || cfg.Default.Java.LibrariesBOMVersion == "" {
 		errs = append(errs, errBOMVersionMissing)
 	}
-
 	pathCount := make(map[string]int)
+	if verifyErrs := verifyRequiredLibraries(cfg); len(verifyErrs) > 0 {
+		errs = append(errs, verifyErrs...)
+	}
 	for _, library := range cfg.Libraries {
 		if library.Version != "" {
 			if _, err := semver.Parse(library.Version); err != nil {
@@ -245,7 +251,6 @@ func Validate(cfg *config.Config) error {
 					errs = append(errs, fmt.Errorf("%s: %w", api.Path, ErrOmitCommonResourcesConflict))
 				}
 			}
-
 		}
 	}
 	for path, count := range pathCount {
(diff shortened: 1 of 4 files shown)


internal/librarian/java/default.go
```diff
@@ -195,7 +195,12 @@ var (
 	// ErrCannotDeriveReleasedVersion is returned when released_version cannot be derived.
 	ErrCannotDeriveReleasedVersion = errors.New("cannot derive released version")
 	// errBOMVersionMissing is returned when libraries_bom_version is not set.
+	// errBOMVersionMissing is returned when libraries_bom_version is not set.
 	errBOMVersionMissing = errors.New("libraries bom version not found in config")
+	// ErrLibraryNotFound is returned when a required library is not found in librarian.yaml.
+	ErrLibraryNotFound = errors.New("library not found in librarian.yaml")
```
sofisl: It seems that howwewritego wants this to be exportable, but I think it also breaks tradition with errBOMVErsionMissing, so I'm not sure what's best to do.

internal/librarian/java/default.go
```diff
@@ -195,7 +195,12 @@ var (
 	// ErrCannotDeriveReleasedVersion is returned when released_version cannot be derived.
 	ErrCannotDeriveReleasedVersion = errors.New("cannot derive released version")
 	// errBOMVersionMissing is returned when libraries_bom_version is not set.
+	// errBOMVersionMissing is returned when libraries_bom_version is not set.
```
JoeWang1127: Is this redundant?

internal/librarian/java/default_test.go
```diff
@@ -511,13 +512,29 @@ func TestValidate(t *testing.T) {
 		},
 	} {
 		t.Run(test.name, func(t *testing.T) {
+			libs := []*config.Library{test.lib}
```
JoeWang1127: Can we add these two libs to the test cases to simplify the test setup?

internal/librarian/java/default_test.go
```diff
@@ -713,7 +762,13 @@ func TestValidate_GlobalConfig(t *testing.T) {
 		},
 	} {
 		t.Run(test.name, func(t *testing.T) {
-			cfg := &config.Config{Default: test.def}
+			cfg := &config.Config{
```
JoeWang1127: How about changing this test to a regular test since it only has one test case.

internal/librarian/tidy_test.go
```diff
@@ -928,7 +938,7 @@ func TestTidy_SkipGenerate(t *testing.T) {
 		t.Fatal(err)
 	}
 	if len(cfg.Libraries) != 1 {
```
JoeWang1127, line 915: Do we need to modify this line?

internal/librarian/java/default.go
```diff
@@ -227,7 +239,11 @@ func Validate(cfg *config.Config) error {
 					errs = append(errs, fmt.Errorf("%s: %w", api.Path, ErrOmitCommonResourcesConflict))
 				}
 			}
-
+		}
+	}
+	for _, req := range []string{rootLibrary, parentPOM} {
```
JoeWang1127: I think this implementation is fine.

However, we could create a helper function to encapsulate the verification logic of these libraries in one place. This could be a follow up change.

internal/librarian/java/default.go
JoeWang1127: I think we can use unexported var if it has no usage outside of the package.

internal/librarian/java/default.go
sofisl: added verification!

sofisl (commented):
Updated test that checks how many libraries were changed to conform to new count.

internal/librarian/java/default_test.go
```diff
@@ -527,80 +556,123 @@ func TestValidate(t *testing.T) {
...
+				{Name: "google-cloud-java"},
+				{Name: "google-cloud-pom-parent", Version: "1.2.3"},
+			},
+			wantErr: errLibraryMissingVersion,
+		},
+		{
+			name:          "missing both required libraries",
+			libraries:     []*config.Library{},
+			wantErr:       errLibraryNotFound,
+			wantErrString: "library not found in librarian.yaml: google-cloud-java\nlibrary not found in librarian.yaml: google-cloud-pom-parent",
```
JoeWang1127: Maybe an error is sufficient, what do you think?

JoeWang1127 (approved):

PR #7425, opened 2026-08-27
feat(rust): support default streaming configuration and explicit overrides

suzmue:
Add include_bidi_streaming_methods and include_server_streaming_methods to RustDefault and update them to pointer booleans across the Rust configuration structs.

This allows bi-directional and server-side streaming to be configured globally under default.rust in librarian.yaml, while allowing individual libraries or modules to explicitly disable streaming with false overrides. False was ambiguous because it could be unset or explicitly turned off. This will assist with enabling these globally.

### DIFF
--- a/internal/librarian/rust/codec.go
+++ b/internal/librarian/rust/codec.go
@@ -135,10 +135,10 @@ func buildCodec(library *config.Library, releaseLevel string) map[string]string 
 	if rust.IncludeStreamingMethods {
 		codec["include-streaming-methods"] = "true"
 	}
-	if rust.IncludeBidiStreamingMethods {
+	if rust.IncludeBidiStreamingMethods != nil && *rust.IncludeBidiStreamingMethods {
 		codec["include-bidi-streaming-methods"] = "true"
 	}
-	if rust.IncludeServerStreamingMethods {
+	if rust.IncludeServerStreamingMethods != nil && *rust.IncludeServerStreamingMethods {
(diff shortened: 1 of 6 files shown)


internal/librarian/rust/codec.go
```diff
@@ -321,10 +321,10 @@ func buildModuleCodec(library *config.Library, module *config.RustModule) map[st
 	if module.IncludeStreamingMethods {
 		codec["include-streaming-methods"] = "true"
 	}
-	if module.IncludeBidiStreamingMethods {
+	if module.IncludeBidiStreamingMethods != nil && *module.IncludeBidiStreamingMethods {
 		codec["include-bidi-streaming-methods"] = "true"
 	}
```
dbolduc, line 326: ah, so the idea is that we can easily flip the default to `true` with:

```go
	if module.IncludeBidiStreamingMethods == nil || *module.IncludeBidiStreamingMethods {
		codec["include-bidi-streaming-methods"] = "true"
	}
```

dbolduc (approved):

internal/librarian/rust/codec.go
suzmue: Basically, I wanted to set include-bidi-streaming-methods: true and then override with include-bidi-streaming-methods: false everywhere that it was broken (many fewer places to add and very clear where work is still needed). But with a boolean, overriding is hard because you don't know is it false? or is it unset? hence pointer to boolean change.

PR #7426, opened 2026-08-27
feat(rust): support bidi streaming in grpc-client

suzmue:
This change was verified by regenerating the pubsub client with streaming_pull.

For #6835

### DIFF
--- a/internal/sidekick/rust/annotate_model.go
+++ b/internal/sidekick/rust/annotate_model.go
@@ -293,7 +293,7 @@ func annotateModel(model *api.API, codec *codec) (*modelAnnotations, error) {
 			streamingServices = append(streamingServices, s)
 		}
 	}
-	hasBidiStreaming := codec.templateOverride == "" && len(bidiStreamingServices) > 0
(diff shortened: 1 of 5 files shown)


internal/sidekick/rust/annotate_model.go
```diff
@@ -293,7 +293,7 @@ func annotateModel(model *api.API, codec *codec) (*modelAnnotations, error) {
 			streamingServices = append(streamingServices, s)
 		}
 	}
-	hasBidiStreaming := codec.templateOverride == "" && len(bidiStreamingServices) > 0
+	hasBidiStreaming := (codec.templateOverride == "" || codec.templateOverride == "templates/grpc-client") && len(bidiStreamingServices) > 0
```
dbolduc, line 296: Same comment as before that this seems unnecessary.

dbolduc (approved):

internal/sidekick/rust/annotate_model.go
suzmue: Ack. I have a WIP that is regenerating all with these set as default, and am worried about adding more breakages by removing the check. And will verify I can remove this after.

PR #7429, opened 2026-08-28
feat(sidekick/rust): support grpc_rust client generation

joshuatants:
For https://github.com/googleapis/google-cloud-rust/issues/5991

Add code generation support for `google-cloud-storage` to use `GrpcRustClient` - a gRPC transport written using Google's `grpc_rust` library) - when `google_cloud_unstable_grpc_rust` is enabled.

- Add GrpcRust boolean option to RustModule configuration and codec.
- Update transport.rs.mustache to generate conditional GrpcClient type aliases.
- Add unit test coverage in codec_test.go and annotate_model_test.go.

### DIFF
--- a/internal/config/language.go
+++ b/internal/config/language.go
@@ -127,6 +127,8 @@ type RustDefault struct {
 // Each module specifies what proto source to use, which template to apply,
 // and where to output the generated code.
 type RustModule struct {
+	// GrpcRust indicates whether to use GrpcRustClient for gRPC transport when google_cloud_unstable_grpc_rust is enabled.
+	GrpcRust bool `yaml:"grpc_rust,omitempty"`
 	// DisabledRustdocWarnings specifies rustdoc lints to disable. An empty slice explicitly enables all warnings.
 	DisabledRustdocWarnings yaml.StringSlice `yaml:"disabled_rustdoc_warnings,omitempty"`
 
--- a/internal/sidekick/rust/templates/grpc-client/transport.rs.mustache
+++ b/internal/sidekick/rust/templates/grpc-client/transport.rs.mustache
@@ -51,19 +51,42 @@ mod info {
         });
 }
 
+{{#Codec.UseGrpcRust}}
+{{#Codec.PerServiceFeatures}}
+#[cfg(any({{#Codec.Services}}feature = "{{Codec.FeatureName}}",{{/Codec.Services}}))]
+{{/Codec.PerServiceFeatures}}
+#[cfg(google_cloud_unstable_grpc_rust)]
+type GrpcClient = gaxi::grpc::GrpcRustClient;
+{{#Codec.PerServiceFeatures}}
+#[cfg(any({{#Codec.Services}}feature = "{{Codec.FeatureName}}",{{/Codec.Services}}))]
+{{/Codec.PerServiceFeatures}}
+#[cfg(not(google_cloud_unstable_grpc_rust))]
+type GrpcClient = gaxi::grpc::Client;
+{{/Codec.UseGrpcRust}}
+
 {{/Codec.HasServices}}
 {{#Codec.Services}}
-/// Implements [{{Codec.Name}}](super::stub::{{Codec.Name}}) using a Tonic-generated client.
+/// Implements [{{Codec.Name}}](super::stub::{{Codec.Name}}) using a {{^Codec.UseGrpcRust}}Tonic-generated{{/Codec.UseGrpcRust}}{{#Codec.UseGrpcRust}}gRPC{{/Codec.UseGrpcRust}} client.
 {{#Codec.PerServiceFeatures}}
 #[cfg(feature = "{{Codec.FeatureName}}")]
 {{/Codec.PerServiceFeatures}}
 #[derive(Clone)]
 pub struct {{Codec.Name}} {
     {{^Codec.ExtendGrpcTransport}}
+    {{#Codec.UseGrpcRust}}
+    inner: GrpcClient,
+    {{/Codec.UseGrpcRust}}
+    {{^Codec.UseGrpcRust}}
     inner: gaxi::grpc::Client,
+    {{/Codec.UseGrpcRust}}
     {{/Codec.ExtendGrpcTransport}}
     {{#Codec.ExtendGrpcTransport}}
+    {{#Codec.UseGrpcRust}}
+    pub(crate) inner: GrpcClient,
+    {{/Codec.UseGrpcRust}}
+    {{^Codec.UseGrpcRust}}
     pub(crate) inner: gaxi::grpc::Client,
+    {{/Codec.UseGrpcRust}}
     {{/Codec.ExtendGrpcTransport}}
 }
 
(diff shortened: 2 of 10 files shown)


internal/sidekick/rust/templates/grpc-client/transport.rs.mustache
```diff
@@ -51,19 +51,42 @@ mod info {
...
+#[cfg(any({{#Codec.Services}}feature = "{{Codec.FeatureName}}",{{/Codec.Services}}))]
+{{/Codec.PerServiceFeatures}}
+#[cfg(google_cloud_unstable_grpc_rust)]
+type GrpcClient = gaxi::grpc::GrpcRustClient;
+{{#Codec.PerServiceFeatures}}
+#[cfg(any({{#Codec.Services}}feature = "{{Codec.FeatureName}}",{{/Codec.Services}}))]
+{{/Codec.PerServiceFeatures}}
+#[cfg(not(google_cloud_unstable_grpc_rust))]
+type GrpcClient = gaxi::grpc::Client;
+{{/Codec.UseGrpcRust}}
```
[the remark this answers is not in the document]
joshuatants: `google-cloud-rust:main` CI tests with `--cfg google_cloud_unstable_grpc_rust` enabled. Because `GrpcRustClient` only exists on the `grpc_rust` feature branch, referencing it on `main` would break workspace compilation.

internal/sidekick/rust/templates/grpc-client/transport.rs.mustache
```diff
@@ -85,13 +108,28 @@ impl {{Codec.Name}} {
...
+            {{/Codec.UseGrpcRust}}
         };
         {{/Codec.DetailedTracingAttributes}}
         {{^Codec.DetailedTracingAttributes}}
+        {{#Codec.UseGrpcRust}}
+        let inner = GrpcClient::new(config, DEFAULT_HOST).await?;
+        {{/Codec.UseGrpcRust}}
+        {{^Codec.UseGrpcRust}}
         let inner = gaxi::grpc::Client::new(config, DEFAULT_HOST).await?;
+        {{/Codec.UseGrpcRust}}
```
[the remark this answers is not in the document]
joshuatants: As above.

internal/sidekick/rust/templates/grpc-client/transport.rs.mustache
```diff
@@ -51,19 +51,42 @@ mod info {
...
+    {{/Codec.UseGrpcRust}}
     {{/Codec.ExtendGrpcTransport}}
     {{#Codec.ExtendGrpcTransport}}
+    {{#Codec.UseGrpcRust}}
+    pub(crate) inner: GrpcClient,
+    {{/Codec.UseGrpcRust}}
+    {{^Codec.UseGrpcRust}}
     pub(crate) inner: gaxi::grpc::Client,
+    {{/Codec.UseGrpcRust}}
     {{/Codec.ExtendGrpcTransport}}
```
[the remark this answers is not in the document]
joshuatants: As above.

internal/config/language.go
```diff
@@ -127,6 +127,8 @@ type RustDefault struct {
 // Each module specifies what proto source to use, which template to apply,
 // and where to output the generated code.
 type RustModule struct {
+	// GrpcRust indicates whether to use GrpcRustClient for gRPC transport when google_cloud_unstable_grpc_rust is enabled.
+	GrpcRust bool `yaml:"grpc_rust,omitempty"`
```
dbolduc, line 131: Q: why is `rust` in any of these names? Isn't it a given that everything we do in `google-cloud-rust` is rust?

dbolduc: I guess we already called `tonic` "grpc". So now that `grpc` comes along, we don't have a name for it.

internal/sidekick/rust/templates/grpc-client/transport.rs.mustache
```diff
@@ -85,13 +108,28 @@ impl {{Codec.Name}} {
     pub async fn new(config: gaxi::options::ClientConfig) -> crate::ClientBuilderResult<Self> {
         {{#Codec.DetailedTracingAttributes}}
         let inner = if gaxi::options::tracing_enabled(&config) {
+            {{#Codec.UseGrpcRust}}
+            GrpcClient::new_with_instrumentation(config, DEFAULT_HOST, &super::tracing::info::INSTRUMENTATION_CLIENT_INFO).await?
+            {{/Codec.UseGrpcRust}}
+            {{^Codec.UseGrpcRust}}
             gaxi::grpc::Client::new_with_instrumentation(config, DEFAULT_HOST, &super::tracing::info::INSTRUMENTATION_CLIENT_INFO).await?
+            {{/Codec.UseGrpcRust}}
```
dbolduc: Can't we say:

```suggestion
            GrpcClient::new_with_instrumentation(config, DEFAULT_HOST, &super::tracing::info::INSTRUMENTATION_CLIENT_INFO).await?
```

because `GrpcClient` means either `gaxi::grpc::Client` or `gaxi::grpc::GrpcRustClient`

internal/sidekick/rust/templates/grpc-client/transport.rs.mustache
dbolduc, line 65: The API surface for these things is identical, right?

Can we just rename the [existing `gaxi::grpc::Client`](https://github.com/googleapis/google-cloud-rust/blob/178dc275cc1fa66362e1a0eee31a4b7661b30f10/src/gax-internal/src/grpc.rs#L70) as like `gaxi::tonic::Client` (which is `pub(crate)`)

And then gaxi switches its definition based on the config flags:

```rs
// File: src/gax-internal/src/grpc.rs

mod grpc_rust;
mod tonic; // move existing client into here

pub mod grpc {
  #[cfg(not(google_cloud_unstable_grpc_rust))]
  pub use grpc_rust::GrpcRustClient as Client;
  #[cfg(google_cloud_unstable_grpc_rust)]
  pub use tonic::Client as Client;
}
```

dbolduc (changes requested):

internal/sidekick/rust/templates/grpc-client/transport.rs.mustache
joshuatants: Currently `GrpcRustClient` lives on the `grpc_rust` branch and is only used by Storage (https://github.com/googleapis/google-cloud-rust/blob/998bc75f7d6f055e1893efcbed896b8082a7793e/src/storage/src/storage/bidi.rs#L33-L41). 

1.  `server_streaming` isn't implemented because it's not used by Storage. So compilation for crates that use `server_streaming` will fail, breaking CI at the very least (CI is set up to run with `google_cloud_unstable_grpc_rust` enabled).

2.  Also, since `GrpcRustClient` and the underlying 'grpc-rust'  (https://crates.io/crates/grpc) library are still unstable I'd prefer to err on the side of caution and keep its use isolated to Storage.

internal/config/language.go
joshuatants: Partly yes, and also partly because the name of Google's gRPC library is `gRPC-Rust` (https://crates.io/crates/grpc), so I just thought it easier to follow that.

internal/sidekick/rust/templates/grpc-client/transport.rs.mustache
dbolduc: > Currently `GrpcRustClient` lives on the `grpc_rust` branch and is only used by Storage (https://github.com/googleapis/google-cloud-rust/blob/998bc75f7d6f055e1893efcbed896b8082a7793e/src/storage/src/storage/bidi.rs#L33-L41). 

Yeah, this is why I am not keen on changing librarian templates

> 1.  `server_streaming` isn't implemented because it's not used by Storage. So compilation for crates that use `server_streaming` will fail, breaking CI at the very least (CI is set up to run with `google_cloud_unstable_grpc_rust` enabled).

Should we change the CI in the `grpc_rust` branch to only test GCS?

We can stub out the function and leave it `unimplemented!()`

You could also introduce a feature flag for `gaxi` on the `grpc_rust` branch and only use the gRPC client when that and the `--cfg` flag are set.

> 2.  Also, since `GrpcRustClient` and the underlying 'grpc-rust'  (https://crates.io/crates/grpc) library are still unstable I'd prefer to err on the side of caution and keep its use isolated to Storage.

This isn't much different IMO

suzmue: > ```rust
> // File: src/gax-internal/src/grpc.rs
> 
> mod grpc_rust;
> mod tonic; // move existing client into here
> 
> pub mod grpc {
>   #[cfg(not(google_cloud_unstable_grpc_rust))]
>   pub use grpc_rust::GrpcRustClient as Client;
>   #[cfg(google_cloud_unstable_grpc_rust)]
>   pub use tonic::Client as Client;
> }
> ```

FYI: I like this as a long term solution, but this would break the CI on main for now (even with adding unimplemented) since we [test main against google_cloud_unstable_grpc_rust](https://github.com/googleapis/google-cloud-rust/blob/2cdd20f5076549f920a0c1b6d3272fd8e873c937/.gcb/builds/triggers/main.tf#L42) as well. All server streaming integration tests would need to be disabled before merging, which imo would be more disruptive than the proposed changes.

joshuatants: > You could also introduce a feature flag for `gaxi` on the `grpc_rust` branch and only use the gRPC client when that and the `--cfg` flag are set.

This might work. Give me some time to sketch out a solution and I'll get back here.

joshuatants: After discussion offline we'll go with the feature flag solution. Closing this PR as we no longer need changes to `librarian`.

PR #7432, opened 2026-08-28
feat(internal/librarian/python): embed local synthtool templates

### LINKED ISSUE #7424: python: inject synthtool env var into generation
Apparently, there is a race condition in the `librarian generate` implementation for `python` that is solved by specifying the path of a `librarian install` provided directory containing synthtool templates that would otherwise be downloaded and used by multiple routines (creating the race).

```
SYNTHTOOL_TEMPLATES="${HOME}/.cache/synthtool/synthtool/synthtool/gcp/templates"
```

Rather than requiring `librarian` be invoked with this env var set, we should instead update the `generate` impl for python to pass along that env var to exec calls if not already set in the environment.

This is a temporary solution to simplify the operator experience.

JoeWang1127:
Embed Synthtool templates into Librarian to enable hermetic, offline Python code generation without needing to download templates from GitHub at runtime.

The embedded templates are extracted to the tool cache directory during installation, and the `SYNTHTOOL_TEMPLATES` environment variable is passed to the post-processor so Synthtool resolves templates from the local cache.

Fixes https://github.com/googleapis/librarian/issues/7424


internal/librarian/python/install.go
```diff
@@ -52,3 +64,31 @@ func InstallDir() (string, error) {
...
+		return fmt.Errorf("failed to clean templates directory: %w", err)
+	}
+	if err := os.CopyFS(dest, sub); err != nil {
+		return fmt.Errorf("failed to extract templates: %w", err)
+	}
+	return nil
+}
+
+// templateDirectory gets the directory where templates are stored.
+func templateDirectory() (string, error) {
```
sofisl, line 88: nit, getTemplateDirectory()?

sofisl (approved):

internal/librarian/python/install.go
JoeWang1127: I feel `get` is redundant.

PR #7437, opened 2026-08-28
build(ci): add matrix for python tests

### LINKED ISSUE #6815: librarian: generate a single library as a presubmit test in each language (potentially increasing timeout)

sofisl:
This aligns python.yaml with the format used in PRs #6960, 6872, and 6979. It uses a strategy matrix to run unit tests, smoke tests (generating a single library), and integration tests (generating all libraries on push to main).

Fixes #6815

### DIFF
--- a/.github/workflows/python.yaml
+++ b/.github/workflows/python.yaml
@@ -66,7 +66,7 @@ jobs:
           sparse-checkout-cone-mode: false
           persist-credentials: false
       - name: "Checkout google-cloud-python (full)"
-        if: matrix.sparse == false
+        if: matrix.sparse == false && github.event_name == 'push' && github.ref == 'refs/heads/main'
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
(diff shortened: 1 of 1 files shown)


.github/workflows/python.yaml
```diff
@@ -36,17 +53,47 @@ jobs:
...
           repository: googleapis/google-cloud-python
           path: google-cloud-python
           depth: 1
           sparse-checkout: |
             librarian.yaml
+            .librarian
+            google-cloud-secret-manager
           sparse-checkout-cone-mode: false
           persist-credentials: false
+      - name: "Checkout google-cloud-python (full)"
```
hj690, line 68: nit: curious if this is only needed for Integration test? Should we apply the same `if` condition (push to main) here to avoid running the full clone for unit/smoke test?

hj690 (approved):

.github/workflows/python.yaml
sofisl: that's a good idea! fixed!

PR #7457, opened 2026-08-31
build(.github): avoid full checkout for integration tests on PRs

sofisl:
This PR applies the `github.event_name == 'push' && github.ref == 'refs/heads/main'` condition to the full repository checkout steps in Java, Node.js, and Ruby workflows, preventing an expensive and unnecessary clone during PR builds since the actual integration test step is skipped. 

This is identical to the fix applied for Python in #7437 (specifically https://github.com/googleapis/librarian/pull/7437#discussion_r3896955505 - thanks @hj690!).

This also includes a small change to a comment in the java workflow. Zizmor was not happy that the sha didn't match the comment exactly, so I changed it in this PR given that the change is not code-related (only updates the comment).

### DIFF
--- a/.github/workflows/java.yaml
+++ b/.github/workflows/java.yaml
@@ -97,7 +97,7 @@ jobs:
       - name: Add Java tools to PATH
         run: echo "$HOME/.cache/librarian/bin/java_tools/bin" >> $GITHUB_PATH
       - name: Run librarian install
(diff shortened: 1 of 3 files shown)


JoeWang1127:
Could you fix the tests?

hj690 (approved):

PR #7469, opened 2026-08-31
build(.github): add test matrix for PHP

### LINKED ISSUE #7468: php: create smoke and integration tests
once PHP generate is working, let's add a workflow to librarian

sofisl:
This pull request introduces a test matrix for PHP to align its CI testing strategy with other languages such as Python, Node.js, Ruby, and Java.
The test matrix includes:
- A Unit Tests task
- A Smoke Test task (generating the SecretManager library)
- An Integration Test task
To optimize the workflow on pull requests, this update utilizes sparse checkout for the unit and smoke tests. It also skips the expensive full clone and installation steps for the integration test on pull requests, executing them only on pushes to main.

Fixes https://github.com/googleapis/librarian/issues/7468


.github/workflows/php.yaml
```diff
@@ -16,29 +16,96 @@ on: [push, pull_request, merge_group]
...
+            task: 'unit-test'
+            timeout: 5
+            sparse: true
+          - job-name: 'Smoke Test'
+            task: 'smoke-test'
+            timeout: 5
+            sparse: true
+          - job-name: 'Integration Test'
+            task: 'integration'
+            timeout: 120
```
zhumin8: Do we need 2hrs? How long does it typically take?

.github/workflows/php.yaml
```diff
@@ -16,29 +16,96 @@ on: [push, pull_request, merge_group]
...
+        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
         with:
-          include_pnpm: 'true'
-          include_python: 'true'
-      - name: Setup test configuration
+          path: |
+            ~/.cache/librarian
+          key: librarian-php-tools-v2-${{ runner.os }}-${{ hashFiles('internal/config/**/*.go') }}-${{ hashFiles('google-cloud-php/librarian.yaml') }}
+      - name: Run librarian install
+        if: matrix.task != 'integration' || (github.event_name == 'push' && github.ref == 'refs/heads/main')
```
zhumin8: why is librarian install skipped for integration test?

.github/workflows/php.yaml
```diff
@@ -16,29 +16,96 @@ on: [push, pull_request, merge_group]
...
+            sparse: false
     steps:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
         with:
           persist-credentials: false
+      - uses: ./.github/actions/setup-librarian
+        with:
+          include_pnpm: 'true'
+          include_python: 'true'
+      - name: Checkout google-cloud-php (Sparse)
```
zhumin8, line 74: Do we need to checkout google-cloud-php for unit test?
https://github.com/googleapis/librarian/actions/runs/33650565460/job/100316212001

.github/workflows/php.yaml
JoeWang1127: It usually takes 20 mins.

I can decrease it to 30 mins.

.github/workflows/php.yaml
```diff
@@ -16,29 +16,96 @@ on: [push, pull_request, merge_group]
...
+          persist-credentials: false
+          sparse-checkout-cone-mode: false
+          sparse-checkout: |
+            librarian.yaml
+            composer.json
+            dev/
+            # Note: If changing the target of the smoke test below, update this sparse checkout rule.
+            SecretManager/
+      - name: Checkout google-cloud-php (Full)
+        if: matrix.sparse == false && (matrix.task != 'integration' || (github.event_name == 'push' && github.ref == 'refs/heads/main'))
```
zhumin8: nit: Because `matrix.sparse == false` is only defined for the Integration Test task, the left side of the OR (matrix.task != 'integration') is essentially redundant. 

Also, when an integration test runs on a PR (pull_request), this checkout step will be skipped entirely, but the job still spins up an environment and runs all steps without the code checked out. Is this intentially?

.github/workflows/php.yaml
JoeWang1127: This check means run librarian install if one of the following conditions apply:
1. unit test or smoke test
2. merge to main

We only want to run integration test at post submit so integration test meets the 2nd condition.

.github/workflows/php.yaml
```diff
@@ -16,29 +16,96 @@ on: [push, pull_request, merge_group]
...
+          php --version
+      - name: Run internal/librarian/php tests and check coverage
+        if: matrix.task == 'unit-test'
         run: go run ./tool/cmd/coverage ./internal/librarian/php
+      - name: Generate a single library (smoke test)
+        if: matrix.task == 'smoke-test' && !(github.event_name == 'push' && github.ref == 'refs/heads/main')
+        working-directory: google-cloud-php
+        run: librarian generate secretmanager
+      - name: Run librarian generate all (integration test)
+        if: matrix.task == 'integration' && github.event_name == 'push' && github.ref == 'refs/heads/main'
```
zhumin8: This creates an unnecessary GitHub Actions job on every PR that checks out code, installs PHP, and passes as a no-op without running any tests.
Can we exclude the integration task during PR runs entirely? Maybe separate integration-test into its own job alongside build-and-test.

.github/workflows/php.yaml
JoeWang1127: We need a sparse checkout for `librarian.yaml` since we need to run `librarian install` later.

.github/workflows/php.yaml
JoeWang1127: Makes sense, I'll split the checks.

.github/workflows/php.yaml
JoeWang1127: Removed this condition.

.github/workflows/php.yaml
```diff
@@ -16,29 +16,138 @@ on: [push, pull_request, merge_group]
...
+        with:
+          php-version: '8.2'
+          extensions: mbstring, xml, bcmath
+      - name: Cache Librarian Tools
+        id: cache-tools
+        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+        with:
+          path: |
+            ~/.cache/librarian
+          key: librarian-php-tools-v2-${{ runner.os }}-${{ hashFiles('internal/config/**/*.go') }}-${{ hashFiles('google-cloud-php/librarian.yaml') }}
```
zhumin8, line 54: Should the cache key be different for L98 and L136?

zhumin8: On second thought, I believe this is intended. Keeping the keys identical prevents redundant downloads and compiles across the three jobs.

.github/workflows/php.yaml
zhumin8: Off topic for this PR: I see redundant integration no-op runs for Ruby Python Node and Java. We might want to change and avoid unnecessary GitHub Actions jobs.

zhumin8 (approved):

.github/workflows/php.yaml
JoeWang1127: Created https://github.com/googleapis/librarian/issues/7484

JoeWang1127 (approved):

Issue #7373, opened 2026-08-21
rust: support auto-generating server gRPC streaming for Rust GAPIC clients


suzmue:
Follow up from #6835 to support server-side streaming.

suzmue:
This is complete

PR #7471, opened 2026-08-31
fix(internal/sidekick/rust): emit grpc client header for streaming apis

suzmue:
Streaming methods in Rust client crates execute over gRPC but previously sent the REST client header constant, which formats the telemetry header as rest/<version>-reqwest.

Define X_GOOG_API_CLIENT_GRPC_HEADER using ac.grpc_header_value() when streaming methods are generated, and pass this header to execute_bidi_streaming and execute_server_streaming so gRPC requests accurately identify as grpc/<version>-tonic.

For #6835 and For #7373

### DIFF
--- a/internal/sidekick/rust/templates/crate/src/lib.rs.mustache
+++ b/internal/sidekick/rust/templates/crate/src/lib.rs.mustache
@@ -164,6 +164,18 @@ pub(crate) mod info {
             };
             ac.rest_header_value()
         });
+{{#Codec.HasStreaming}}
+    #[allow(dead_code)]
+    pub(crate) static X_GOOG_API_CLIENT_GRPC_HEADER: std::sync::LazyLock<String> =
+        std::sync::LazyLock::new(|| {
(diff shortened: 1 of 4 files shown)


internal/sidekick/rust/templates/crate/src/lib.rs.mustache
```diff
@@ -164,6 +164,18 @@ pub(crate) mod info {
             };
             ac.rest_header_value()
         });
+{{#Codec.HasStreaming}}
+    #[allow(dead_code)]
+    pub(crate) static X_GOOG_API_CLIENT_GRPC_HEADER: std::sync::LazyLock<String> =
```
dbolduc, line 169: Non-blocking for this PR or this project... just an FYI in case you are inspired: I remember thinking these should be scoped to the service in `transport.rs`.

https://github.com/googleapis/google-cloud-rust/issues/2138

dbolduc (approved):

PR #7476, opened 2026-09-01
cleanup(rust): support grpc and http on a single template

suzmue:
This change allows choosing the default transport for a crate, to switch unary methods from grpc to http and back, configuring the relevant internal client types.

This also more accurately encapsulates when we need to include things because we are using gRPC vs because they are streaming. 

For googleapis/google-cloud-rust#6470


suzmue:
I'm not ready to send this for review. I think I want to make this change more incremental.

PR #7480, opened 2026-09-02
docs(internal/sidekick/rust): link to gaxi::grpc::Client in docs

suzmue:
This is internal only documentation.

### DIFF
--- a/internal/sidekick/rust/templates/crate/src/transport.rs.mustache
+++ b/internal/sidekick/rust/templates/crate/src/transport.rs.mustache
@@ -31,7 +31,7 @@ use crate::Error;
 
 {{/Codec.HasServices}}
 {{#Codec.Services}}
-/// Implements [{{Codec.Name}}](super::stub::{{Codec.Name}}) using a [gaxi::http::ReqwestClient].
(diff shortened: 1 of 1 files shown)


internal/sidekick/rust/templates/crate/src/transport.rs.mustache
```diff
@@ -31,7 +31,7 @@ use crate::Error;
 
 {{/Codec.HasServices}}
 {{#Codec.Services}}
-/// Implements [{{Codec.Name}}](super::stub::{{Codec.Name}}) using a [gaxi::http::ReqwestClient].
+/// Implements [{{Codec.Name}}](super::stub::{{Codec.Name}}) using a [gaxi::http::ReqwestClient]{{#Codec.HasStreaming}} and a [gaxi::grpc::Client]{{/Codec.HasStreaming}}.
```
coryan, line 34: Note that this is documenting an internal-only type, and the docs.rs documentation for gaxi is empty. So it is more consistent / accurate, but will not have much effect.

coryan (approved):

internal/sidekick/rust/templates/crate/src/transport.rs.mustache
suzmue: Yeah I knew this was internal only, I am just hoping to help ourselves a bit with this one.

PR #7481, opened 2026-09-02
cleanup(internal/sidekick/rust): support pure grpc services in default crates

suzmue:
Differentiate between gRPC-only, HTTP-only, and hybrid services in the default Rust crate template, and clarify where template generation depends on gRPC vs HTTP transport capabilities rather than streaming mechanics.

Previously, template logic often used streaming checks as a proxy for gRPC presence. This change clarifies those distinctions by checking for gRPC support where transport features are concerned (such as `grpc_inner`, `X_GOOG_API_CLIENT_GRPC_HEADER`, `mod prost`, and transport doc comments), while preserving streaming checks only where they genuinely relate to stream builder mechanics.

This allows services that only define gRPC methods (such as `apigeeconnect::v1::Tether`) to generate clean transport structs without instantiating or storing an unused `inner: gaxi::http::ReqwestClient`, while preserving the exact existing structure and zero diffs for pure HTTP services.

This also removes some service annotations to minimize chances for skew between the methods that exist in the current filtered service and the corresponding annotation.

For googleapis/google-cloud-rust#6470

### DIFF
--- a/internal/sidekick/rust/annotate_method.go
+++ b/internal/sidekick/rust/annotate_method.go
@@ -49,11 +49,17 @@ type methodAnnotation struct {
 	IsBigQueryInsertJob       bool
 	ClientSideStreaming       bool
 	ServerSideStreaming       bool
+	IsGrpc                    bool
+}
+
+// IsHttp returns true if the method is routed over HTTP transport.
+func (m *methodAnnotation) IsHttp() bool {
+	return !m.IsGrpc
 }
 
 // IsBidiStreaming returns true if the method is a bidirectional streaming RPC.
 func (m *methodAnnotation) IsBidiStreaming() bool {
-	return m.ClientSideStreaming && m.ServerSideStreaming
+	return m.IsGrpc && m.ClientSideStreaming && m.ServerSideStreaming
 }
 
 // FullBody returns true if the HTTP body is configured as "*".
@@ -64,7 +70,7 @@ func (m *methodAnnotation) FullBody() bool {
 // IsServerStreaming returns true if the method is a server-side streaming RPC
 // and is not bidirectional.
 func (m *methodAnnotation) IsServerStreaming() bool {
-	return !m.ClientSideStreaming && m.ServerSideStreaming
+	return m.IsGrpc && !m.ClientSideStreaming && m.ServerSideStreaming
 }
 
 // HasGrpcResourceNameArgs returns true if the method has gRPC resource name arguments.
@@ -347,6 +353,7 @@ func (c *codec) annotateMethod(m *api.Method) (*methodAnnotation, error) {
 		IsBigQueryInsertJob:       m.ID == ".google.cloud.bigquery.v2.JobService.InsertJob",
 		ClientSideStreaming:       m.ClientSideStreaming,
 		ServerSideStreaming:       m.ServerSideStreaming,
+		IsGrpc:                    c.methodUsesGrpc(m),
 	}
 
 	if err := c.annotateResourceNameGeneration(m, annotation); err != nil {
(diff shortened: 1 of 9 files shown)


internal/sidekick/rust/templates/crate/src/transport.rs.mustache
```diff
@@ -31,16 +31,18 @@ use crate::Error;
 
 {{/Codec.HasServices}}
 {{#Codec.Services}}
-/// Implements [{{Codec.Name}}](super::stub::{{Codec.Name}}) using a [gaxi::http::ReqwestClient]{{#Codec.HasStreaming}} and a [gaxi::grpc::Client]{{/Codec.HasStreaming}}.
+/// Implements [{{Codec.Name}}](super::stub::{{Codec.Name}}) using {{#Codec.IsPureGrpc}}a [gaxi::grpc::Client]{{/Codec.IsPureGrpc}}{{#Codec.IsPureHttp}}a [gaxi::http::ReqwestClient]{{/Codec.IsPureHttp}}{{#Codec.IsHybrid}}a [gaxi::http::ReqwestClient] and a [gaxi::grpc::Client]{{/Codec.IsHybrid}}.
```
dbolduc, line 34: or just:

```suggestion
/// Implements [{{Codec.Name}}](super::stub::{{Codec.Name}}).
```

and the code tells us which client is used.

dbolduc: Argh, your motivation was probably to keep the diff in check. If so, then ignore me.

internal/sidekick/rust/annotate_method.go
```diff
@@ -734,3 +741,15 @@ func isIdempotent(p *api.PathInfo) string {
...
 }
+
+func (c *codec) methodUsesGrpc(m *api.Method) bool {
+	if !c.templateSupportsGrpc() {
+		return false
+	}
+	if m.ClientSideStreaming || m.ServerSideStreaming {
+		return (m.ClientSideStreaming && m.ServerSideStreaming && c.includeBidiStreamingMethods) ||
+			(!m.ClientSideStreaming && m.ServerSideStreaming && c.includeServerStreamingMethods) ||
+			c.includeStreamingMethods
```
dbolduc, line 752: aside: we should probably consolidate these flags.

internal/sidekick/rust/annotate_method_test.go
```diff
@@ -596,18 +602,97 @@ func TestIsServerStreaming(t *testing.T) {
...
+			}
+		})
+	}
+}
+
+func TestMethodUsesGrpc(t *testing.T) {
+	msg := api.NewTestMessage("Message").WithPackage("test.v1")
+	unaryMethod := api.NewTestMethod("Unary").WithInput(msg).WithOutput(msg).WithPathTemplate(&api.PathTemplate{})
+	bidiMethod := api.NewTestMethod("Bidi").WithInput(msg).WithOutput(msg).WithBidiStreaming()
+	serverMethod := api.NewTestMethod("Server").WithInput(msg).WithOutput(msg).WithServerSideStreaming()
```
dbolduc, line 649: TIL: these model builders

dbolduc (approved):
Good clean up :)

internal/sidekick/rust/annotate_method_test.go
suzmue: They are very helpful (and Carlos reminds me to use them every time which is also very helpful).

internal/sidekick/rust/annotate_method.go
suzmue: Agreed. I wanted these to be temporary. Will try to eliminate them from our librarian.yaml ideally.

I do think I might need your help because some of these were added for the grpc-mock to be used by the grpc-mock template, so you may be hearing from me.

internal/sidekick/rust/templates/crate/src/transport.rs.mustache
suzmue: That was part of the motivation, but I could be convinced this is a silly effort and I should be fine with re-releasing all of our crates this month...

internal/sidekick/rust/annotate_method.go
dbolduc: > some of these were added for the grpc-mock

Yeah, I added `includeStreamingMethods`.

Basically, I needed the mock server to include all of the RPCs before the GAPICs were ready to support streaming.

When the default is to include all the RPCs in the model, it won't be necessary.

PR #7485, opened 2026-09-02
fix(internal/sidekick/rust): skip doc links for omitted methods and services

suzmue:
Do not generate intra-doc link definitions for methods that have been pruned from a service (via `included_ids` or `skipped_ids`), or services that have been pruned from the model.

Previously, `methodRustdocLink` looked up methods in the global ID map and only checked streaming/gRPC flags via `generateMethod(m)` without verifying whether the method survived model element pruning in `s.Methods`. This caused broken intra-doc links to non-existent methods (such as `crate::client::StorageControl::read_object` when `Storage.ReadObject` was omitted).

### DIFF
--- a/internal/sidekick/rust/codec_test.go
+++ b/internal/sidekick/rust/codec_test.go
@@ -1698,6 +1698,98 @@ func TestFormatDocCommentsRelativeCrossLinks(t *testing.T) {
 	}
 }
 
+func TestFormatDocCommentsSkippedMethodsAndServices(t *testing.T) {
+	input := `
+[CreateFoo][test.v1.SomeService.CreateFoo]
+[SkippedMethod][test.v1.SomeService.SkippedMethod]
+[SkippedService][test.v1.SkippedService]
+[SkippedServiceMethod][test.v1.SkippedService.SomeMethod]
+`
+	want := []string{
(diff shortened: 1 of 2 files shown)


internal/sidekick/rust/codec_test.go
```diff
@@ -1698,6 +1698,98 @@ func TestFormatDocCommentsRelativeCrossLinks(t *testing.T) {
...
+[CreateFoo][test.v1.SomeService.CreateFoo]
+[SkippedMethod][test.v1.SomeService.SkippedMethod]
+[SkippedService][test.v1.SkippedService]
+[SkippedServiceMethod][test.v1.SkippedService.SomeMethod]
+`
+	want := []string{
+		"/// [CreateFoo][test.v1.SomeService.CreateFoo]",
+		"/// [SkippedMethod][test.v1.SomeService.SkippedMethod]",
+		"/// [SkippedService][test.v1.SkippedService]",
+		"/// [SkippedServiceMethod][test.v1.SkippedService.SomeMethod]",
```
dbolduc, line 1712: comment: it's funny to say we want this. But I agree that it is preferable to leave the docs as-is than link to a type that doesn't exist.

dbolduc (approved):

Issue #7484, opened 2026-09-02
librarian: separate unit test, smoke test and integration test


JoeWang1127:
Off topic for this PR: I see redundant integration no-op runs for Ruby Python Node and Java. We might want to change and avoid unnecessary GitHub Actions jobs.

_Originally posted by @zhumin8 in https://github.com/googleapis/librarian/pull/7469#discussion_r3917801043_

zhumin8:
Here is a case where the no-op integration runs causes unnecessary hiccup when flaky happens. https://github.com/googleapis/librarian/actions/runs/33676889485/attempts/1?pr=7487

PR #7489, opened 2026-09-02
build(.github/workflows): separate java integration test into standalone job

zhumin8:
Move the Java integration test out of the build-and-test presubmit matrix into a dedicated integration-test job that only runs on pushes to main.
    
Previously, including the integration test in the presubmit matrix resulted in unnecessary presubmit runs that do not run tests, and skipping tool installation but unconditionally running tool verification, which caused failures on cache misses.

for https://github.com/googleapis/librarian/issues/7484


.github/workflows/java.yaml
```diff
@@ -112,15 +95,66 @@ jobs:
...
+      - name: Add Java tools to PATH
+        run: echo "$HOME/.cache/librarian/bin/java_tools/bin" >> $GITHUB_PATH
+      - name: Run librarian install
+        if: steps.cache-tools.outputs.cache-hit != 'true'
+        working-directory: google-cloud-java
+        run: |
+          librarian install
+      - name: Verify tools
+        run: |
+          protoc --version
```
JoeWang1127, line 160: `protoc` is installed to `$HOME/.cache/librarian/bin/protoc`, do we need to add this path to `GITHUB_PATH` as well? Or you can remove it from this section.

zhumin8: Good point. This is an pre-existing bug, see [recent log](https://screenshot-v2.corp.google.com/37jeu4taf77ko). This step not failing because protoc is also installed to PATH.
Updated workflow as suggested. 

Separately, might worth eval if install protoc is still needed in setup-librarian action [here](https://github.com/googleapis/librarian/blob/7f108509d90656c995f2a7c52dcb9e5d4e678fc9/.github/actions/setup-librarian/action.yaml#L37)?

zhumin8: created https://github.com/googleapis/librarian/issues/7493

JoeWang1127: > Separately, might worth eval if install protoc is still needed in setup-librarian action [here](https://github.com/googleapis/librarian/blob/7f108509d90656c995f2a7c52dcb9e5d4e678fc9/.github/actions/setup-librarian/action.yaml#L37)?

We can delete this if protoc is defined in all onboarded lanagues.

JoeWang1127 (approved):

PR #7492, opened 2026-09-03
feat(sidekick/rust): support grpc_client option 

joshuatants:


.github/workflows/java.yaml
```diff
@@ -102,6 +102,7 @@ jobs:
         run: |
           librarian install
       - name: Verify tools
+        if: matrix.task != 'integration' || (github.event_name == 'push' && github.ref == 'refs/heads/main')
```
dbolduc: cc: @blakeli0 

I am approving this PR because I want the other changes.

dbolduc (approved):

.github/workflows/java.yaml
coryan: The test is (or was) passing at HEAD.  We should not disable other tests just to push ours.  Can you rebase and see if that fixes the problem?

coryan (changes requested):

.github/workflows/java.yaml
suzmue: In my experience it has been flaky and re-running has done the trick.

joshuatants: Reverted the disabling. Passing now without rebasing. Looks like @suzmue is right about flakiness.

coryan (approved):

.github/workflows/java.yaml
coryan: Please file a bug (if one does not exist) explaining the flake.

Issue #7389, opened 2026-08-25
php: pass googleapis directory in `add` interface


JoeWang1127:
Currently, librarian will compute the component name on the fly and move the generated library to the directory that is the same as component name to maintain backwards compatibility. For example, for library `speech`, librarian will generate and move the code to `Speech` directory, which is also the same as component name.

To achieve this, we write the library output as component name using the migration tool since librarian generate is not allowed to update librarian.yaml and the component name is not derivable. The component name is computed by parsing proto in the API path.

However, for new libraries, we can't maintain this pattern because `librarian add` does not have access to googleapis directory to calculate the component name (also the library output).

The current workaround is adding the library output after running `librarian add`. We need to fix this issue before handing PHP add & generate to our oncall.

zhumin8:
After #7494, should revise generate code and remove unnecessary on-the-fly deriving logic. 
Also, generate should rely on output and library.PHP.ComponentName becomes completely redundant and can be removed.

PR #7494, opened 2026-09-03
feat(internal/librarian/php): populate library output on add

zhumin8:
The PHP add interface is updated to accept the googleapis directory and populate the library output directory with the derived component name when onboarding a library.

Previously, librarian add did not have access to the googleapis directory, preventing it from deriving the component name from proto definitions. Newly onboarded PHP libraries were left without an output path in librarian.yaml, requiring manual adjustments before generation (for formatting specifically). With this change, add resolves the component name from the proto php_namespace option or an explicit component name override and records it in the library output.

For https://github.com/googleapis/librarian/issues/7389

### DIFF
--- a/internal/librarian/php/add.go
+++ b/internal/librarian/php/add.go
@@ -15,6 +15,7 @@
 package php
 
 import (
+	"fmt"
 	"path"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


internal/librarian/php/add.go
```diff
@@ -50,5 +51,12 @@ func Add(lib *config.Library) *config.Library {
 			api.PHP.StagingSubdir = serviceconfig.ExtractVersion(api.Path)
 		}
 	}
-	return lib
+	if lib.Output == "" {
+		compName, err := ComponentNameForLibrary(googleapisDir, lib)
```
JoeWang1127, line 55: We can unexport this function, feel free to do it in a follow up.

JoeWang1127 (approved):

PR #7501, opened 2026-09-03
fix(internal/librarian/python): flatten nested gapic_version paths

### LINKED ISSUE #7497: python: add uses the wrong gapic_version file path in release-please config for nested subdirs
In python, we reduce API directory nesting for files like `gapic_version.py` by taking the would-be path segments following the PA and concatenating them with underscores e.g. `google/shopping/merchant/loyaltycustomer/v1` -> `google/shopping/merchant_loyaltycustomer_v1`.

`librarian add` release-please configs do not populate that properly which results in the `gapic_version.py` not being updated by release-please, requiring fixes like: https://github.com/googleapis/google-cloud-python/pull/18279

This needs to be correct bc its a subtle and very error prone issue, especially for deeply nested API paths like the example above.

JoeWang1127:
In Python libraries, APIs with nested subdirectories following the namespace (such as google/shopping/merchant/inventories/v1 or google/cloud/datacatalog/lineage/v1) are collapsed into underscore-separated directory names in generated GAPIC packages (e.g. merchant_inventories and datacatalog_lineage).

Previously, ReleasePleaseExtraFiles did not account for this directory flattening, leaving nested slash-separated directory paths in the release-please extra files configuration.

This change introduces `flattenNestedPath`, using `deriveGAPICNamespace` and `deriveGAPICName` to collapse nested subdirectories into their corresponding underscore-separated GAPIC directory names. This ensures gapic_version.py files for nested APIs are properly tracked and bumped by release-please.

Fixes https://github.com/googleapis/librarian/issues/7497


internal/librarian/python/add.go
```diff
@@ -17,7 +17,7 @@ package python
 import (
 	"errors"
 	"fmt"
-	"path"
+	"path/filepath"
```
codyoss: I believe for release-please, we actually do want this to be path.

internal/librarian/python/add.go
```diff
@@ -145,25 +145,17 @@ func versionless(apiPath string) string {
...
-
-		versionlessPath := api.Path
-		if version != "" {
-			versionlessPath = path.Dir(api.Path)
-		}
-
-		if !addedVersionless[versionlessPath] {
-			addedVersionless[versionlessPath] = true
-			extraFiles = append(extraFiles, versionlessPath+"/gapic_version.py")
+		flattenedPath := flattenNestedPath(api.Path)
```
codyoss: Should we make use of `deriveGAPICGenerationInfo` which honors some custom gapic options? Or make this aware of those options?

internal/librarian/python/add.go
JoeWang1127: Done.

internal/librarian/python/add.go
JoeWang1127: Good suggestion.

I raised https://github.com/googleapis/librarian/pull/7507 for a refactor first.

codyoss (approved):

codyoss (approved):

PR #7507, opened 2026-09-04
refactor(internal/librarian/python): consolidate GAPIC option helpers

JoeWang1127:
The GAPIC option parsing and derivation helper functions (`gapicNamespace`, `gapicName`, `deriveGAPICNamespace`, `deriveGAPICName`, and `findOption`) are moved from `generate.go` to a dedicated `gapic.go` file. A shared `apiOptions` helper is introduced to deduplicate fetching API options from the library configuration.

Previously, `deriveGAPICGenerationInfo` in `clean.go` duplicated the option lookup and fallback logic instead of sharing the implementation with `generate.go`.

Consolidating these helpers into `gapic.go` allows `clean.go` to reuse the shared functions.

For #7501


codyoss (approved):
Thanks!

PR #7512, opened 2026-09-04
fix(internal/librarian/rust): batch unsupported any field errors (#7479)

haphungw:
Collect all unsupported google.protobuf.Any field references encountered during streaming reachability traversal instead of returning immediately on the first error. Sort and deduplicate the offending fields, and format a descriptive error message with a librarian.yaml allow_streaming_any_types configuration snippet.

Forward work by @suzmue at #7479, adding back `visited[item.id] = true` and a unit test to confirm no cyclic recursion.


suzmue (approved):
Thank you!

PR #7514, opened 2026-09-04
feat(sidekick/rust): move x-goog-api-client header handling

alvarowolfx:
Towards https://github.com/googleapis/google-cloud-rust/issues/2138

### DIFF
--- a/internal/sidekick/rust/templates/crate/src/transport.rs.mustache
+++ b/internal/sidekick/rust/templates/crate/src/transport.rs.mustache
@@ -237,7 +237,7 @@ impl super::stub::{{Codec.Name}} for {{Codec.Name}} {
         {{#Codec.HasBindings}}
         use google_cloud_gax::error::binding::BindingError;
         use gaxi::path_parameter::PathMismatchBuilder;
-        use gaxi::http::reqwest::{Method, HeaderValue};
+        use gaxi::http::reqwest::Method;
         {{#Codec.HasBindingSubstitutions}}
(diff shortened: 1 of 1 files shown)


internal/sidekick/rust/templates/crate/src/transport.rs.mustache
```diff
@@ -372,16 +372,15 @@ impl super::stub::{{Codec.Name}} for {{Codec.Name}} {
             {{! TODO(#4156) - return idempotency from the above closure }}
             gaxi::http::default_idempotency(&method),
         );
-        let builder = builder
```
coryan, line 375: That cannot be right.  You would be setting `x-goog-api-client` N times, depending on how many system parameters are set....

coryan (changes requested):

internal/sidekick/rust/templates/crate/src/transport.rs.mustache
coryan: Oh, it *used* to be wrong...

coryan (dismissed):

internal/sidekick/rust/templates/crate/src/transport.rs.mustache
alvarowolfx: the `.header` thing was outside of `{{#Codec.SystemParameters}}`.

coryan: Yeah, ignore me.

PR #7516, opened 2026-09-04
fix(internal/librarian/php): use library output in generate and postprocessing

zhumin8:
The PHP generator and postprocessor now use library.Output directly rather than calculating the component name on the fly from protobuf definitions.

A non-empty library.Output is required; initComponentIfMissing validates this field and checks whether the component directory already exists on disk. If the directory exists, component initialization is skipped entirely. Staging and owlbot postprocessing use library.Output, allowing custom output paths and maintaining compatibility with existing configurations.

Note: this change is compatible with google-cloud-php because during migration, library.Output is already populated. see https://github.com/googleapis/librarian/issues/7153#issuecomment-5430491979, and [librarian.yaml](https://github.com/googleapis/google-cloud-php/blob/main/librarian.yaml).

For https://github.com/googleapis/librarian/issues/7389


zhumin8:
Next step to this is to remove the now obsolete library.PHP.ComponentName, will do in a separate PR.

JoeWang1127 (approved):

Issue #7496, opened 2026-09-03
nodejs: onboarding known non-cloud APIs should work out of the box


noahdietz:
Onboarding the non-cloud API google/shopping/merchant/loyaltycustomer/v1 in https://github.com/googleapis/google-cloud-node/pull/9245 required manual application of the `nodejs` block to include the `package_name` property setting it to `@google-shopping/loyaltycustomer`.

This is en example of a a well known non-Cloud API that has an "organization" to publish to (`@google-shopping`) and that new API libraries for it should be included in automatically by `librarian add`.

Other languages like Python and Java have a similar top-level mapping for tracking API path to the appropriate "namespace". We should do the same for `nodejs` add with existing non-Cloud APIs.

PR #7534, opened 2026-09-09
refactor(internal/serviceconfig): add MatchPrefix helper and migrate Java

zhumin8:
Add a shared MatchPrefix function to internal/serviceconfig that matches an API path against a map of prefixes by checking exact matches and walking up parent path segments.

Migrate Java group ID resolution in internal/librarian/java to use the shared prefix matcher, prepare for node to use the same prefix lookup behavior. By replacing map iteration with hierarchical path lookups, this eliminates non-deterministic prefix matching in Java and guarantees that more specific sub-paths always take precedence.

For https://github.com/googleapis/librarian/issues/7496

### DIFF
--- a/internal/serviceconfig/prefix_test.go
+++ b/internal/serviceconfig/prefix_test.go
@@ -26,7 +26,6 @@ func TestMatchPrefix(t *testing.T) {
 		Remainder string
 		OK        bool
 	}
-
 	prefixes := map[string]string{
 		"google/shopping/merchant": "@google-shopping",
 		"google/shopping":          "@google-shopping",
@@ -35,7 +34,6 @@ func TestMatchPrefix(t *testing.T) {
 		"google/chat":              "@google-apps",
 		"google/area120":           "@google/area120",
 	}
-
 	for _, test := range []struct {
 		name     string
 		apiPath  string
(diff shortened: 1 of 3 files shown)


internal/serviceconfig/prefix_test.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestMatchPrefix(t *testing.T) {
+	type result struct {
+		Value     string
+		Remainder string
+		OK        bool
+	}
+
```
JoeWang1127: nit: remove this empty line.

internal/serviceconfig/prefix_test.go
```diff
@@ -0,0 +1,120 @@
...
+	prefixes := map[string]string{
+		"google/shopping/merchant": "@google-shopping",
+		"google/shopping":          "@google-shopping",
+		"google/maps/isochrones":   "@google-maps",
+		"google/maps":              "@googlemaps",
+		"google/chat":              "@google-apps",
+		"google/area120":           "@google/area120",
+	}
+
```
JoeWang1127: nit: remove this empty line.

JoeWang1127 (approved):

PR #7538, opened 2026-09-09
feat(internal/tool/pip): force reinstall git packages

### LINKED ISSUE #7528: librarian(install): pip should force reinstall packages when installing from vcs urls
In `internal/tool/pip/pip.go`, `pip.Install` invokes `pip install <targets...>`. When a tool dependency is pinned to a Git URL with a specific commit hash (such as `gcp-synthtool@git+https://github.com/googleapis/synthtool@<sha>`) where the package's `setup.py` or
  `pyproject.toml` static version string remains unchanged across commits (e.g., `2020.2.4`), pip's default resolution behavior checks only package version numbers, not commit hashes.

In persistent local development environments (such as a developer's virtualenv), running `librarian install` reports `Requirement already satisfied` and **silently skips** fetching and installing the new commit hash. This causes local code generation to silently run outdated tooling while remote CI (in clean containers) runs the new commit, creating hard-to-diagnose generation differences.

I run into this when trying to re-generate google-cloud-php locally. Due to this setup, it picked up older synthtool prior to fix https://github.com/googleapis/synthtool/pull/2181. This should be rare, considering we don't actively develop synthtool and it should be the only pip installed tool that don't use a release version.

### Implementation Details

The implementation in `internal/tool/pip/pip.go` needs to ensure that VCS/Git URLs or commit updates are unconditionally reinstalled:

Update `Install` in `internal/tool/pip/pip.go` to pass `--force-reinstall` (or `--upgrade --force-reinstall`) when installing tool dependencies, or detect VCS targets (e.g., entries containing `@git+`, `git://`, or an explicit commit hash/package URL) and ensure `--force-reinstall --no-deps` is applied to those targets.

JoeWang1127:
Packages specified from git repositories are now installed separately using the `force-reinstall` flag. This ensures that pip fetches and reinstalls the exact commit referenced by the package URL even if a package with the same version number is already installed in the environment.

Previously, pip install processed all tools in a single batch invocation. If a git repository package shared a version number with an already-installed package, pip would report the requirement as satisfied and skip updating to the new commit. Standard non-git packages and local path targets continue to be installed together in a single batch.

Fixes https://github.com/googleapis/librarian/issues/7528


internal/tool/pip/pip.go
```diff
@@ -36,31 +37,58 @@ var (
...
-		if tool.Package != "" {
-			installTargets = append(installTargets, tool.Package)
-			continue
-		}
-		if tool.Version != "" {
-			installTargets = append(installTargets, fmt.Sprintf("%s==%s", tool.Name, tool.Version))
-			continue
+			stdPackages = append(stdPackages, absPath)
+		case tool.Package != "":
+			if strings.Contains(tool.Package, "git+https://github.com") {
```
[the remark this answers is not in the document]
JoeWang1127: Done.

internal/tool/pip/pip.go
```diff
@@ -36,31 +37,58 @@ var (
...
+		return nil
 	}
 	args := []string{"install"}
-	args = append(args, installTargets...)
+	args = append(args, targets...)
+	return runPip(ctx, args...)
+}
+
+// reinstallPackages reinstalls packages from git repositories.
+func reinstallPackages(ctx context.Context, targets []string) error {
```
zhumin8: This name sounds a bit confusing to me, as if it is called after `installPackages` or as retry. 
 Perhaps `forceInstallPackages()` or `installGitPackages()`?

internal/tool/pip/pip.go
```diff
@@ -36,31 +37,58 @@ var (
...
-		if tool.Package != "" {
-			installTargets = append(installTargets, tool.Package)
-			continue
-		}
-		if tool.Version != "" {
-			installTargets = append(installTargets, fmt.Sprintf("%s==%s", tool.Name, tool.Version))
-			continue
+			stdPackages = append(stdPackages, absPath)
+		case tool.Package != "":
+			if strings.Contains(tool.Package, "git+") {
```
zhumin8, line 54: nit: strings.HasPrefix(tool.Package, "git+")

zhumin8 (approved):

internal/tool/pip/pip.go
JoeWang1127: Done.

internal/tool/pip/pip.go
JoeWang1127: Package doesn't start with `git+`, e.g., `gcp-synthtool@git+https://github.com/googleapis/synthtool@e643ce8e20f8fe237a31a1754524ba987de72875`

PR #7553, opened 2026-09-10
feat(internal/librarian/nodejs): derive package name from custom prefixes

zhumin8:
Derive default Node.js package names for non-cloud APIs during library addition using custom_package_prefixes configured in librarian.yaml. Node.js constructs package names for bare scopes, scoped prefixes with hyphenated suffixes, and exact leaf package names without manual configuration.

Unrecognized non-cloud APIs log a structured warning prompting the user to manually configure package_name. 

For https://github.com/googleapis/librarian/issues/7496

### DIFF
--- a/internal/librarian/nodejs/add.go
+++ b/internal/librarian/nodejs/add.go
@@ -29,7 +29,7 @@ const defaultVersion = "0.0.0"
 func Add(cfg *config.Config, lib *config.Library) *config.Library {
 	lib.Version = defaultVersion
 	if cfg != nil {
-		lib = fillDefault(lib, cfg.Default)
+		lib = populatePackageName(lib, cfg.Default)
 	}
 	apiPath := lib.APIs[0].Path
 	if !strings.HasPrefix(apiPath, "google/cloud/") && (lib.Nodejs == nil || lib.Nodejs.PackageName == "") {
@@ -38,9 +38,9 @@ func Add(cfg *config.Config, lib *config.Library) *config.Library {
 	return lib
 }
 
-// fillDefault populates empty Node.js-specific fields in lib from [config.Default],
-// specifically from [config.NodejsDefault].
-func fillDefault(lib *config.Library, d *config.Default) *config.Library {
+// populatePackageName sets the Node.js package name on lib if empty, deriving it
+// from custom_package_prefixes in [config.Default].
+func populatePackageName(lib *config.Library, d *config.Default) *config.Library {
 	if lib.Nodejs != nil && lib.Nodejs.PackageName != "" {
 		return lib
 	}
--- a/internal/librarian/nodejs/add.go
+++ b/internal/librarian/nodejs/add.go
@@ -15,6 +15,7 @@
 package nodejs
 
 import (
+	"log/slog"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


internal/librarian/nodejs/add.go
```diff
@@ -25,11 +26,61 @@ import (
...
 	lib.Version = defaultVersion
+	if len(lib.APIs) == 0 {
+		return lib
+	}
+	if cfg != nil {
+		fillDefault(lib, cfg.Default)
+	}
+	apiPath := lib.APIs[0].Path
+	if !strings.HasPrefix(apiPath, "google/cloud/") && (lib.Nodejs == nil || lib.Nodejs.PackageName == "") {
+		slog.Warn("unrecognized non-cloud API path; please manually configure nodejs.package_name in librarian.yaml", "api", apiPath)
```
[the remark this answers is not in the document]
zhumin8: Don't think this is true.
This warning is necessary to give developer a head up before going into generate and error out.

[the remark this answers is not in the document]
zhumin8: There is something related this though. With current code, say developer ignored this warning and ran generate anyway, it does not fail immediately. Instead it will try to derive the package name as if it's cloud API I think. 
This should be fixed separately, move the derive logic upstream to apply defaults, and in generate to fail fast when nodejs.package_name is empty.

Will file an issue to track.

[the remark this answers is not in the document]
zhumin8: created https://github.com/googleapis/librarian/issues/7558

internal/librarian/nodejs/add.go
```diff
@@ -25,11 +26,58 @@ import (
 const defaultVersion = "0.0.0"
 
 // Add initializes Node.js-specific configuration for a library.
-func Add(lib *config.Library) *config.Library {
+func Add(cfg *config.Config, lib *config.Library) *config.Library {
 	lib.Version = defaultVersion
+	if len(lib.APIs) == 0 {
```
JoeWang1127: Is this check redundant?

internal/librarian/nodejs/add.go
```diff
@@ -25,11 +26,58 @@ import (
...
+		return lib
+	}
+	if cfg != nil {
+		lib = fillDefault(lib, cfg.Default)
```
JoeWang1127: Why we can't fill defaults during generation, like other languages?

zhumin8: There are 2 fill defaults: 1. fill in and write value to librarian.yaml.
2. fill defaults and generate.
For (2), I agree and filed https://github.com/googleapis/librarian/issues/7558, will work on next.
For (1), I think we want to write down the package name for non-cloud APIs in librarian.yaml, so human have a chance to review it before running generate. (this is the same as Java, let me check on what each languages do)

zhumin8: The closest example is Java for group_id and Python for Namespace. Java has the same logic and writes to librarian.yaml at `add`. Python's Namespace is not a field in librarian.yaml, so it is derived on the fly.

JoeWang1127: Oh, now I understand. 

nit: can we use another name to replace `fillDefault`?

JoeWang1127 (approved):

Issue #7217, opened 2026-08-06
librarian: refactor helper functions in add/generate to a dedicated package


JoeWang1127:
The add/generation logic is similar among languages, e.g., gather protos from API path.

We should identify common logic and refactor to a dedicated package.

PR #7572, opened 2026-09-11
refactor(internal/librarian/nodejs): adopt proto.Gather helper

JoeWang1127:
Replace `filepath.Glob` in `collectProtos` with the centralized `proto.Gather` helper function to consolidate proto gathering logic across language generators.

We need to update API configuration of `google-cloud-aiplatform` in google-cloud-node to
```
...
apis:
  - path: google/cloud/aiplatform/v1
  - path: google/cloud/aiplatform/v1beta1
      nodejs:
        exclude_protos:
          - google/cloud/aiplatform/v1beta1/schema/io_format.proto
...
```

For #7217


zhumin8 (approved):
The change itself LGTM. 
nit: In `TestCollectProtos`, could you add a test case targeting nested subdirectories to explicitly verify the recursive discovery behavior introduced by proto.Gather.

JoeWang1127:
> nit: In TestCollectProtos, could you add a test case targeting nested subdirectories to explicitly verify the recursive discovery behavior introduced by proto.Gather.

Added a test case. Btw, we already tested this in the unit tests of `proto.Gather`.

PR #7574, opened 2026-09-11
feat(internal/librarian/nodejs): support excluding protos from generation

JoeWang1127:
An `exclude_protos` option is added to NodejsAPI configuration, allowing APIs to exclude specific proto files from generation.

We plan to use `proto.Gather` to collect protos for code generation. However, `aiplatform` excludes `google/cloud/aiplatform/v1beta1/schema/io_format.proto`, thus we need this configuration before further refactor.

This change let the nodejs configurations align with other languages.

For #7217

### DIFF
--- a/internal/librarian/nodejs/generate.go
+++ b/internal/librarian/nodejs/generate.go
@@ -195,6 +195,9 @@ func resolveNodejsAPI(library *config.Library, api *config.API) *config.NodejsAP
 	if library.Nodejs != nil {
 		protos = append(protos, library.Nodejs.AdditionalProtos...)
 	}
+	if api.Nodejs != nil {
+		protos = append(protos, api.Nodejs.AdditionalProtos...)
+	}
 	res.AdditionalProtos = unique(protos)
 	return res
 }
--- a/internal/librarian/nodejs/generate.go
+++ b/internal/librarian/nodejs/generate.go
@@ -205,22 +205,22 @@ func resolveNodejsAPI(library *config.Library, api *config.API) *config.NodejsAP
 		}
 		res.OmitCommonResources = api.Nodejs.OmitCommonResources
 	}
-
 	var protos []string
 	if !omitCommon {
 		protos = append(protos, cloudCommonResourcesProto)
 	}
-
 	// Add package-level additional protos.
 	if library.Nodejs != nil {
 		protos = append(protos, library.Nodejs.AdditionalProtos...)
 	}
-
-	// Add API-level additional protos.
 	if api.Nodejs != nil {
 		protos = append(protos, api.Nodejs.AdditionalProtos...)
+		for _, excluded := range api.Nodejs.ExcludeProtos {
+			protos = slices.DeleteFunc(protos, func(p string) bool {
+				return p == excluded
+			})
+		}
 	}
-
 	res.AdditionalProtos = unique(protos)
 	return res
 }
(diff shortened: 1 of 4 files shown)


internal/librarian/nodejs/generate.go
```diff
@@ -205,22 +205,22 @@ func resolveNodejsAPI(library *config.Library, api *config.API) *config.NodejsAP
...
 	}
-
-	// Add API-level additional protos.
 	if api.Nodejs != nil {
 		protos = append(protos, api.Nodejs.AdditionalProtos...)
+		for _, excluded := range api.Nodejs.ExcludeProtos {
+			protos = slices.DeleteFunc(protos, func(p string) bool {
+				return p == excluded
+			})
+		}
```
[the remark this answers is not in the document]
JoeWang1127: Done.

internal/librarian/nodejs/generate.go
```diff
@@ -237,6 +214,29 @@ func unique(ss []string) []string {
 	return res
 }
 
+func collectProtos(absGoogleapisDir string, nodejsAPI *config.NodejsAPI) ([]string, error) {
+	apiDir := filepath.Join(absGoogleapisDir, nodejsAPI.Path)
+	protos, err := filepath.Glob(apiDir + "/*.proto")
+	if err != nil {
+		return nil, fmt.Errorf("failed to find protos: %w", err)
```
zhumin8: nit:  "failed to match proto pattern: %w"

JoeWang1127: Done.

internal/librarian/nodejs/generate_test.go
```diff
@@ -1624,6 +1624,62 @@ func TestResolveNodejsAPI(t *testing.T) {
 				AdditionalProtos:    []string{"pkg.proto", "dup.proto", "api.proto"},
 			},
 		},
+		{
+			name:    "exclude protos preserved",
```
zhumin8, line 1628: nit: Can you add a test case where a proto is listed both as AdditionalProtos and ExcludeProtos?

zhumin8 (commented):
Is the only motivation for adding support for excluding protos to unify/extract shared logic across lang implementations, or do we got a use case in nodejs expecting this?
If the former, can we simply implement a shared helper that can support an optional exclude-protos, and not exclude protos for nodejs? This way there is no behavior/functional changes to node, and is easily expandable in the future if needed.

internal/librarian/nodejs/generate_test.go
JoeWang1127: I can't think of a valid use case of adding and excluding a proto at the same time.

JoeWang1127:
> do we got a use case in nodejs expecting this

We have at least one case, `aiplatform` doesn't include `google/cloud/aiplatform/v1beta1/schema/io_format.proto`

I plan to use `proto.Gather` in nodejs but we need to exclude this proto.

internal/librarian/nodejs/generate_test.go
zhumin8: demon's advocate: we cannot stop wrongly configured librarian.yaml in the future.
I don't feel strongly about this.

zhumin8 (approved):
LGTM

> We have at least one case, aiplatform

Can you link the issue in description or add an explanation there?

PR #7575, opened 2026-09-11
chore(internal/librarian/nodejs): derive package names in defaults

### LINKED ISSUE #7558: nodejs: derive package names in defaults and require package_name at generate
There is something related this though. With current code, say developer ignored this warning and ran generate anyway, it does not fail immediately. Instead it will try to derive the package name as if it's cloud API I think. 
This should be fixed separately, move the derive logic upstream to apply defaults, and in generate to fail fast when nodejs.package_name is empty.

Will file an issue to track.

_Originally posted by @zhumin8 in https://github.com/googleapis/librarian/pull/7553#discussion_r3981736420_

zhumin8:
Derives standard Node.js package names (@google-cloud/<suffix>) during defaults resolution instead of during generator argument construction. The generator now requires nodejs.package_name to be set and fails fast with an explanatory error when it is missing.

Non-cloud APIs and custom scopes are populated during library onboarding via custom package prefixes or configured explicitly in librarian.yaml. Moving derivation upstream establishes a single source of truth for the entire generation pipeline and removes redundant runtime guesswork.

Fixes https://github.com/googleapis/librarian/issues/7558

### DIFF
--- a/internal/librarian/library_test.go
+++ b/internal/librarian/library_test.go
@@ -864,8 +864,12 @@ func TestApplyDefaults_Nodejs(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if got.Nodejs == nil || got.Nodejs.PackageName != "@google-cloud/secretmanager" {
-		t.Errorf("got package name %v, want %q", got.Nodejs, "@google-cloud/secretmanager")
+	if got.Nodejs == nil {
+		t.Error("got.Nodejs is nil")
+		return
+	}
+	if got.Nodejs.PackageName != "@google-cloud/secretmanager" {
+		t.Errorf("got package name %q, want %q", got.Nodejs.PackageName, "@google-cloud/secretmanager")
 	}
 }
 
--- a/internal/librarian/library_test.go
+++ b/internal/librarian/library_test.go
@@ -865,7 +865,7 @@ func TestApplyDefaults_Nodejs(t *testing.T) {
 		t.Fatal(err)
 	}
 	if got.Nodejs == nil {
-		t.Error("got.Nodejs is nil")
+		t.Fatal("got.Nodejs is nil")
 		return
 	}
 	if got.Nodejs.PackageName != "@google-cloud/secretmanager" {
--- a/internal/librarian/library_test.go
+++ b/internal/librarian/library_test.go
@@ -855,6 +855,20 @@ func TestApplyDefaults_Error(t *testing.T) {
 	}
 }
 
+func TestApplyDefaults_Nodejs(t *testing.T) {
+	lib := &config.Library{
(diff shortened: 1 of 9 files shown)


internal/librarian/nodejs/default.go
```diff
@@ -0,0 +1,39 @@
...
+// Fill populates Node.js-specific default values for the library.
+// For standard Google Cloud libraries (prefixed with "google-cloud-"), it derives
+// the package name as "@google-cloud/<suffix>". Non-cloud packages are not filled
+// and must be populated during "librarian add" (via custom_package_prefixes) or
+// configured explicitly in librarian.yaml.
+func Fill(lib *config.Library) *config.Library {
+	if lib.Nodejs != nil && lib.Nodejs.PackageName != "" {
+		return lib
+	}
+	if suffix, ok := strings.CutPrefix(lib.Name, "google-cloud-"); ok && suffix != "" {
```
JoeWang1127, line 32: nit: move this to a helper function, e.g., `fillPackageName`.

internal/librarian/nodejs/generate.go
```diff
@@ -94,7 +97,8 @@ func Generate(ctx context.Context, cfg *config.Config, library *config.Library,
 }
 
 var (
-	errToolNotInstalled = errors.New("tool not installed in librarian cache")
+	errToolNotInstalled    = errors.New("tool not installed in librarian cache")
+	errPackageNameRequired = errors.New("nodejs.package_name is required; non-cloud libraries must be configured in librarian.yaml or populated during librarian add")
```
JoeWang1127: nit: move var declaration to the top of the file.

internal/librarian/library_test.go
```diff
@@ -855,6 +855,24 @@ func TestApplyDefaults_Error(t *testing.T) {
...
+	lib := &config.Library{
+		Name: "google-cloud-secretmanager",
+		APIs: []*config.API{{Path: "google/cloud/secretmanager/v1"}},
+	}
+	got, err := applyDefaults(config.LanguageNodejs, lib, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if got.Nodejs == nil {
+		t.Error("got.Nodejs is nil")
```
JoeWang1127:
```suggestion
		t.Fatalf("got.Nodejs is nil")
```

Then you don't need the `return`.

JoeWang1127 (approved):

internal/librarian/nodejs/default.go
zhumin8: Right now Fill only populates the package name, so extracting it leaves Fill as just a 1-line pass-through wrapper around 8 lines of logic.
Kept it inline for simplicity for now, but happy to split it out if we have more fields in Fill soon.

PR #7586, opened 2026-09-14
feat(sidekick/swift): generate unknown field deserialization

chingor13:
This removes the `CustomSerialization` determination because we always are using custom serialization (unknown field handling)


internal/sidekick/swift/templates/common/message.mustache
```diff
@@ -79,11 +81,27 @@ public struct {{Codec.Name}}: Codable, Equatable, {{Codec.Model.WktPackage}}._An
...
+    static let _knownKeys: Set<Swift.String> = [
+      {{#Fields}}
+      "{{JSONName}}",
+      {{/Fields}}
+    ]
+    {{/HasFields}}
+    {{^HasFields}}
+
+    static let _knownKeys: Set<Swift.String> = []
+    {{/HasFields}}
```
coryan: optional: the `static let _knownKeys` could be outside the `{{#HasFields}}` block and would work fine, no? That may be easier to grok, but I am not sure.

internal/sidekick/swift/templates/common/message.mustache
```diff
@@ -114,6 +132,10 @@ public struct {{Codec.Name}}: Codable, Equatable, {{Codec.Model.WktPackage}}._An
     {{/Fields}}
     self.{{Codec.PropertyName}} = {{Codec.PropertyName}}
     {{/OneOfs}}
+    for key in container.allKeys where !CodingKeys._knownKeys.contains(key.stringValue) {
```
coryan, line 131: TIL: `for .. in ... where`, neat.

internal/sidekick/swift/generate_field_swift_test.go
```diff
@@ -77,8 +77,14 @@ func TestGenerateField_InitFromDecoder(t *testing.T) {
...
 	wantBlock := `  public init(from decoder: Decoder) throws {
     let container = try decoder.container(keyedBy: CodingKeys.self)
-    self.normalField = try container.decode(Swift.String.self, forKey: .normalField)
+    if let value = try container.decodeIfPresent(Swift.String.self, forKey: .normalField) {
+      self.normalField = value
+    }
     self.optionalField = try container.decodeIfPresent(Swift.String.self, forKey: .optionalField)
+    for key in container.allKeys where !CodingKeys._knownKeys.contains(key.stringValue) {
+      self._unknownFields.json[key.stringValue] = try container.decode(
+        GoogleCloudWKT.Value.self, forKey: key)
```
coryan, line 86: When we test this, we should include some tests for unknown float fields that have `NaN` as a value.  I think I made `GoogleCloudWKT` too strict in that it does not allow serialization of `NaN`.  Not the end of the world if unknown floats with non-number values are weird for GA, we just need to know about the issue.

coryan (approved):

internal/sidekick/swift/generate_field_swift_test.go
chingor13: Track in https://github.com/googleapis/google-cloud-swift/issues/897

PR #7602, opened 2026-09-16
feat(internal/librarian/golang): support private protobuf copies

### LINKED ISSUE #7585: golang: Support custom plugin generation and internal output routing
### Description / Context

To support advanced performance optimizations (such as memory allocation tuning or custom codecs) in generated client libraries, we need the ability to invoke custom protoc plugins during generation and route their output to private, internal directories.

Currently, generation assumes standard public API layouts. For experimental or advanced optimizations, we need to ensure that the generated optimization code does not leak into the public API surface or pollute the protobuf global registry for end users.

### Requirements
We propose augmenting the Librarian toolchain to support the following capabilities:

1. **Custom Plugin Execution:** The ability to configure and invoke additional protoc plugins (such as protoc-gen-go-vtproto or similar optimization generators) as part of the standard generation pipeline.
2. **Internal Output Routing:** The ability to route the output of specific plugins to a designated internal package directory (e.g., internal/) rather than the public generation surface. In Go, placing these outputs in an internal/ directory ensures they are not surfaced as end-user accessible code.

### Benefits
**Encapsulation:** Allows library maintainers to add performance-critical hotpaths or alternative codecs without exposing experimental or unstable APIs to users.
**Safety:** Keeps public API surfaces clean and isolated from internal optimization details.
**Extensibility:** Makes the generation pipeline more flexible for libraries with extreme performance constraints.

Internal References: go/cloud-sdk-go:spanner-performance-escalation-notes, go/goproto-spanner-client-10x

rahul2393:
## What

Adds `internal_copies` to the Go API configuration. Each entry names an `internal/` Go import path, a replacement proto package, and the extra protoc plugins (for example `protoc-gen-go-vtproto`) to run on a private copy of the API's messages:

```yaml
- path: google/spanner/v1
  go:
    internal_copies:
      - import_path: spanner/internal/vtpb
        proto_package: google.spanner.v1.vtinternal
        plugins:
          - name: go-vtproto
            options:
              - features=marshal+unmarshal+size+pool+unmarshal_unsafe
              - pool=cloud.google.com/go/spanner/internal/vtpb.PartialResultSet
```

The plugin binary is declared under `tools.go` and resolved from the Go tool bin directory, like `protoc-gen-go_gapic`.

## Why a private copy with a renamed proto package

The issue asked for custom plugin execution and routing of the output to an internal package. For a plugin such as vtprotobuf that generates *methods* on message types, output cannot simply be redirected: Go only allows methods in the receiver's own package, so vt methods on the public `spannerpb` types would be exported API. A separate copy of the messages under `internal/` avoids that, but a verbatim copy panics at init because protobuf-go registers descriptors by proto full name and file path with a default conflict policy of `panic`. The copy therefore carries a renamed proto package and renamed file paths, which the wire format never sees. Well-known types stay on the standard `google.golang.org/protobuf` packages; vtprotobuf's WKT support is a named-type conversion, so `PartialResultSet.Values` stays `[]*structpb.Value`.

## How

For each copy, generation runs protoc to a descriptor set, rewrites the API's own files (package, file name, `go_package`, every type reference by exact full name, services dropped), then runs protoc again with `--descriptor_set_in` for `protoc-gen-go` plus each configured plugin, and moves the output into the repo. Cleanup removes only `*.pb.go` files under a copy directory, honouring `keep`.

Validation runs before any cleanup, both statically in `librarian tidy` and at generate time: canonical import paths with an `internal` element that resolve (through existing symlinks) inside the library output and do not overlap other output (case-insensitively); valid, unique proto packages; no extensions of messages outside the copy; open API level only; no plugin options that change the output layout, in any comma-separated position.

## Testing

- Descriptor rewrite, path validation, API-level resolution, cleanup, and regeneration tests in `internal/librarian/golang/internalcopy_*_test.go`, including a protoc-backed twin-import test that compiles the public and private packages into one binary and round-trips a message.
- The Go unit-test workflow now installs the test tools so the protoc-backed tests run in CI rather than skip; coverage of the package is 91% with the plugins present and 81% without.
- Generated `google/spanner/v1` against google-cloud-go: `spanner/internal/vtpb` (ten base files plus ten vtproto files, package `vtpb`, no service code), public `spanner/apiv1/spannerpb` byte-identical, regeneration byte-stable, `go build` and `go vet` clean, both packages registered side by side without conflict.

Known limits, documented on the config fields: non-`.pb.go` plugin output and directories of retired or renamed copies need manual removal; blank imports left by dropped services are not pruned.

Fixes #7585


rahul2393:
@JoeWang1127 please check

rahul2393:
Example librarian.yaml changes for google-cloud-go

Screenshot 2026-09-17 at 3 16 37 AM

internal/librarian/golang/internalcopy.go
```diff
@@ -0,0 +1,629 @@
...
+	}
+	// An extension is registered on its extendee by field number, and renaming
+	// the extension does not change that key: a copy extending a message that
+	// is not itself copied would register a second extension with the same
+	// number and panic at init as soon as both packages are linked.
+	for _, fd := range apiFds {
+		if err := checkExtendees(renamedTypes, fd.GetName(), "."+oldPackage, fd.GetExtension(), fd.GetMessageType()); err != nil {
+			return nil, err
+		}
+	}
```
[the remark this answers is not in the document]
rahul2393: Good catch. Rather than special-casing the prefix, the rewrite now refuses an API file that declares no proto package, with an error naming the file (90c5fcc0). Every googleapis proto declares one, so the guard only prevents the silent mismatch.

internal/librarian/golang/clean.go
```diff
@@ -59,6 +59,11 @@ var (
 // Clean cleans up a Go library and its associated snippets.
 func Clean(library *config.Library) error {
 	libraryDir := library.Output
```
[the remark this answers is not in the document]
rahul2393: The new validation does not depend on the caller's path form:  
validateInternalCopyPaths resolves every directory through  resolveExistingPath, which calls filepath.Abs before resolving symlinks, so  the containment and overlap checks already run on absolute, resolved paths.

.github/workflows/go.yaml
```diff
@@ -45,8 +45,19 @@ jobs:
         if: matrix.task == 'smoke-test'
         working-directory: google-cloud-go
         run: librarian install
+      - name: Install Go generation tools
```
JoeWang1127, line 48: Do do you need to change unit test workflow?

internal/cache/cache_test.go
```diff
@@ -43,9 +43,7 @@ func TestDirectory(t *testing.T) {
 		},
 	} {
 		t.Run(test.name, func(t *testing.T) {
-			if test.env != "" {
-				t.Setenv(EnvLibrarianCache, test.env)
-			}
+			t.Setenv(EnvLibrarianCache, test.env)
```
JoeWang1127: Why do you need to change this file?

internal/config/language.go
```diff
@@ -76,6 +107,32 @@ type GoAPI struct {
...
+	// Before regeneration, only `*.pb.go` files in the copy directory,
+	// including `*_vtproto.pb.go`, are cleaned, honoring library-relative keep
+	// entries, which are literal paths rather than globs. Other files are not
+	// cleaned. Stale non-`*.pb.go` plugin output and directories of retired or
+	// renamed copies require manual removal.
+	//
+	// A copy directory must not be, contain, or lie inside another copy or
+	// client directory; that check ignores letter case, so paths that differ
+	// only by case are rejected as overlapping.
+	InternalCopies []*GoInternalCopy `yaml:"internal_copies,omitempty"`
```
JoeWang1127, line 118: Could you simplify this comment, I think most of the content are implementation details.

.github/workflows/go.yaml
rahul2393: Yes,  the unit-test job had no protoc plugins, so the package's generation tests, old and new, always skipped there, and this PR's code cannot reach the 80% gate with its tests skipped. 

The step installs the pinned test tools in that job and makes skips fail, so the generation tests actually run in CI.

internal/librarian/golang/clean.go
```diff
@@ -59,6 +59,11 @@ var (
...
+	// A copy directory is derived from configuration and cleaned recursively,
+	// so validate every copy before removing anything.
+	if err := validateInternalCopyPaths(library, libraryDir); err != nil {
```
JoeWang1127, line 69: Why do you need to validate the directory before cleaning?

internal/librarian/golang/clean.go
```diff
@@ -142,6 +147,17 @@ func cleanClientDirectory(library *config.Library, libraryDir string, keepSet ma
 		if err := cleanGeneratedClientFiles(clientPath, libraryDir, keepSet); err != nil {
 			return err
 		}
+		for _, cp := range goAPI.InternalCopies {
```
JoeWang1127: Could you put this for loop in a helper function?

internal/librarian/golang/clean.go
```diff
@@ -151,6 +167,14 @@ func cleanClientDirectory(library *config.Library, libraryDir string, keepSet ma
 }
 
 func cleanGeneratedClientFiles(clientPath, libraryDir string, keepSet map[string]bool) error {
+	return cleanGeneratedFiles(clientPath, libraryDir, keepSet, func(name string) bool {
```
JoeWang1127: I think the current clean logic is
1. find the right directory to clean
2. clean a file if the file path or name followed a predefined pattern.

Could you update the clean function to follow this rules?

internal/librarian/golang/generate.go
```diff
@@ -53,6 +53,11 @@ func Generate(ctx context.Context, cfg *config.Config, library *config.Library,
 	if err != nil {
 		return fmt.Errorf("failed to get absolute path of output directory: %w", err)
 	}
+	// Reject a bad internal copy before any file, public or private, is
+	// generated so that a failed run leaves the repository untouched.
+	if err := validateInternalCopyPaths(library, outDir); err != nil {
```
JoeWang1127, line 59: Could you explain why we need to validate a file path before generation?

Most files are cleaned at this point so the repo is changed even if the validation failed.

internal/librarian/golang/internalcopy.go
```diff
@@ -0,0 +1,635 @@
...
+	errInternalCopyAPILevel       = errors.New("internal copy requires the open protobuf API level")
+	errProtocPluginNil            = errors.New("internal copy plugins entry must not be null")
+	errProtocPluginName           = errors.New("protoc plugin name must consist of letters, digits, \"-\" and \"_\"")
+	errProtocPluginOption         = errors.New("protoc plugin option changes the output layout")
+	errProtocPluginNotFound       = errors.New("protoc plugin not found")
+
+	protocPluginNameRe = regexp.MustCompile(`^[A-Za-z0-9_-]+$`)
+	// protocPluginLayoutOptions are the protoc-gen-go style options that move
+	// plugin output away from the Go import path the copy is collected from.
+	// M<file> import mappings override go_package and are also rejected.
```
JoeWang1127: What does this line mean?

internal/librarian/golang/internalcopy.go
```diff
@@ -0,0 +1,635 @@
...
+			}
+		}
+	}
+	return nil
+}
+
+// clientImportPath returns the import path of the API's generated client,
+// deriving the default from the API path when the configuration has not been
+// filled yet.
+func clientImportPath(apiPath string, goAPI *config.GoAPI) string {
```
JoeWang1127: the GoAPI has been filled at this point so I think you don't need this function.

internal/cache/cache_test.go
rahul2393: I can remove it, it was needed for my local full test run

internal/librarian/golang/internalcopy.go
```diff
@@ -0,0 +1,635 @@
...
+	protoFiles, err := collectProtoFiles(googleapisDir, apiPath, goAPI.NestedProtos)
+	if err != nil {
+		return err
+	}
+	// Only the files directly in the API directory are copied. Nested protos
+	// are supplied to protoc as dependencies only, whatever proto package they
+	// declare, so their messages keep their public Go types and get no plugin
+	// output.
+	apiDir := filepath.Join(googleapisDir, apiPath)
+	var apiFiles []string
```
JoeWang1127: Could you create a helper function to scan the api directory and collect protos? I think it's straightforward than recursively collecting protos and remove.

internal/librarian/golang/internalcopy.go
```diff
@@ -0,0 +1,635 @@
...
+	plugins, err := resolveProtocPlugins(cp.Plugins)
+	if err != nil {
+		return err
+	}
+	copyDir, err := os.MkdirTemp(tempDir, "internal-copy-")
+	if err != nil {
+		return err
+	}
+	// Source info is kept so that the proto comments end up in the generated
+	// Go code, as they do for the public package.
```
JoeWang1127: Why do you need this comment at this line?

internal/librarian/golang/internalcopy.go
```diff
@@ -0,0 +1,635 @@
...
+	}
+	args = append(args, protoFiles...)
+	if err := runProtoc(ctx, pc, args...); err != nil {
+		return err
+	}
+	fds, err := readDescriptorSet(apiSet)
+	if err != nil {
+		return err
+	}
+	if err := checkInternalCopyAPILevel(fds, apiFiles, goAPI.ProtoAPILevel); err != nil {
```
JoeWang1127: Why do you need a verification step in generation? Validation should happen outside of the generation, ideally in a CI.

internal/librarian/golang/internalcopy.go
```diff
@@ -0,0 +1,635 @@
...
+	}
+	for _, plugin := range cp.Plugins {
+		args = append(args, "--plugin="+protocPluginPrefix+plugin.Name+"="+plugins[plugin.Name])
+		args = append(args, "--"+plugin.Name+"_out="+copyDir)
+		for _, opt := range plugin.Options {
+			args = append(args, "--"+plugin.Name+"_opt="+opt)
+		}
+	}
+	args = append(args, renamed...)
+	if err := runProtoc(ctx, pc, args...); err != nil {
```
JoeWang1127, line 300: Why do you need a 2nd generation?

internal/librarian/golang/validate.go
```diff
@@ -0,0 +1,55 @@
...
+
+	"github.com/googleapis/librarian/internal/config"
+)
+
+// Validate checks the Go-specific configuration of every library. It applies
+// lexical checks to internal copies without consulting the filesystem, so
+// that "librarian tidy" rejects a copy that is
+// null, resolves outside its library, overlaps another generated directory,
+// reuses a proto package or lists a malformed plugin.
+func Validate(cfg *config.Config) error {
```
JoeWang1127: I don't think we need a validate function for now.

JoeWang1127:
@rahul2393 thanks for creating this pull request.

I think this pull request is tool large to merge in one go, could you create multiple smaller ones for quick review?

I haven't checked the unit tests but I found multiple ones that do not comply with table driven test, see https://github.com/googleapis/librarian/blob/main/doc/howwewritego.md#table-driven-tests

internal/config/language.go
```diff
@@ -49,6 +49,37 @@ const (
...
+type GoInternalCopy struct {
+	// ImportPath is the Go import path of the copy, relative to
+	// cloud.google.com/go, in canonical form and inside the library's own
+	// module. It must contain an "internal" path element so that the copy
+	// cannot be imported by users of the library, and must not overlap the
+	// directory of another copy or of a generated client. Existing symlinks
+	// below the library output directory are rejected before generation.
+	ImportPath string `yaml:"import_path"`
+	// Plugins lists the additional protoc plugins to run for the copy.
+	Plugins []*GoProtocPlugin `yaml:"plugins,omitempty"`
```
JoeWang1127, line 63: Do you need more than one plugin?

rahul2393: No, I needed only one for immediate use cae

rahul2393:
I have split this up in 4 PRs
1) https://github.com/googleapis/librarian/pull/7605
2) https://github.com/googleapis/librarian/pull/7606

PR #7605, opened 2026-09-17
feat(internal/config): define Go internal copy settings

rahul2393:
Add API-level internal copy settings with one required extra protoc plugin and optional plugin options. Document the schema and preserve the settings when tidying configuration.

This first step establishes the configuration contract without changing cleanup or generation behavior, so it can be reviewed independently.

For https://github.com/googleapis/librarian/issues/7585


JoeWang1127 (approved):

rahul2393:
@JoeWang1127 I don't have permission to merge pleae help

PR #7607, opened 2026-09-17
feat(internal/librarian/golang): clean internal copy directories

rahul2393:
An internal copy directory holds generated code only, so clean removes the whole directory. Handwritten code lives outside the copy, and removing the directory leaves no stale files behind when the API's proto files change.

Fill validates the copies before clean or generate touch the repository: each copy's `import_path` must be a clean relative path with an `internal` element that resolves inside the library, so that cleanup cannot reach a public package or another library, and no `keep` entry may lie under a copy directory. Clean itself only derives the directory and removes it. Proto package, plugin, and overlap validation belong to generation and are not part of this change.

For https://github.com/googleapis/librarian/issues/7585


internal/librarian/golang/clean.go
```diff
@@ -54,11 +54,21 @@ var (
...
+	generatedCopyFileSuffixes = []string{".pb.go"}
 )
 
 // Clean cleans up a Go library and its associated snippets.
 func Clean(library *config.Library) error {
 	libraryDir := library.Output
+	// Internal copy directories come from configuration and are cleaned
+	// recursively, so make sure they all point inside this library before
+	// deleting anything.
+	if err := validateInternalCopyPaths(library, libraryDir); err != nil {
```
JoeWang1127: I think we need to validate this path, at least not in clean function.

internal/librarian/golang/internalcopy_path.go
JoeWang1127, line 1: Could you merge functions in this file to `internalcopy.go`?

internal/librarian/golang/clean.go
```diff
@@ -183,3 +189,43 @@ func cleanGeneratedClientFiles(clientPath, libraryDir string, keepSet map[string
...
+	return nil
+}
+
+// checkInternalCopies resolves every internal copy directory before anything
+// is deleted, so that a bad copy configuration is rejected up front, and
+// rejects keep entries inside a copy directory, which is removed entirely
+// rather than cleaned file by file.
+func checkInternalCopies(library *config.Library, libraryDir string) error {
+	for _, api := range library.APIs {
+		if api.Go == nil {
```
JoeWang1127: clean happens after fill step so this should not happen.

internal/librarian/golang/clean.go
```diff
@@ -183,3 +189,43 @@ func cleanGeneratedClientFiles(clientPath, libraryDir string, keepSet map[string
...
+			continue
+		}
+		for _, cp := range api.Go.InternalCopies {
+			copyPath, err := internalCopyDir(library, libraryDir, cp)
```
JoeWang1127: Looks like this is a validation step and it should happen outside of clean.

internal/librarian/golang/internalcopy_test.go
```diff
@@ -385,3 +387,172 @@ func testEnum(name string) *descriptorpb.EnumDescriptorProto {
...
+		t.Run(test.name, func(t *testing.T) {
+			_, gotErr := internalCopyDir(test.library, filepath.FromSlash(test.outDir), test.copy)
+			if !errors.Is(gotErr, test.wantErr) {
+				t.Errorf("internalCopyDir error = %v, wantErr %v", gotErr, test.wantErr)
+			}
+		})
+	}
+}
+
+func TestInternalCopyLibrarySymlink(t *testing.T) {
```
JoeWang1127:
```suggestion
func TestClean_InternalCopy_LibrarySymlink(t *testing.T) {
```

JoeWang1127 (approved):

PR #7612, opened 2026-09-18
feat(internal/librarian/python): copy opt_args_by_api when adding new APIs

### LINKED ISSUE #7600: librarian(python): support 'librarian add' for existing libraries with custom GAPIC options (opt_args_by_api)
When adding a new API version (e.g., `google/cloud/networkservices/v1beta1`) to an existing Python client library that defines custom GAPIC options in `librarian.yaml` (e.g., `google-cloud-network-services`), `librarian add` fails immediately with an error and does not update `librarian.yaml`.
**Command & Error Output**
```text
$ librarian add google/cloud/networkservices/v1beta1
librarian: new APIs cannot be automatically added to a library with custom GAPIC options
exit status 1
```

When onboarding google/cloud/networkservices/v1beta1:

- `FindExistingLibraryForNewAPI` correctly matches google-cloud-network-services because they share the same versionless path (google/cloud/networkservices/).
- `ValidateNewAPIs` rejects the addition because `len(lib.Python.OptArgsByAPI) != 0`.
- As a result, automated onboarding workflows and scripts fail to add the API path, blocking subsequent librarian generate and librarian tidy runs unless librarian.yaml is manually modified beforehand.

I think I lack the context on this:
What is the reason we choose to return this error? Does it make sense to copy the existing setting for the new API path? Do we have any concern on blindly copying these settings?

### Proposed Solutions / Suggestions

Option 1 (Automatic Inheritance - Recommended):

When adding a new API version to an existing library that already has opt_args_by_api configured for sister API versions in the same service (e.g., google/cloud/networkservices/v1), automatically propagate matching sister options (like python-gapic-name) to the newly added API path in opt_args_by_api.
Option 2 (Permissive Addition with Warning):

Allow librarian add to append the new API path to apis:, emit a warning indicating that custom GAPIC options exist for sister APIs, and let subsequent generation/tidy handle or flag any discrepancies.

JoeWang1127:
Replace `ValidateNewAPIs` with `UpdateExistingLibrary` to encapsulate validating and updating an existing Python library when adding a new API version. When `OptArgsByAPI` is configured on the library, copy the custom GAPIC options from the first API entry to the newly added API path and log a warning prompting manual review.

Fixes #7600


internal/librarian/python/add.go
```diff
@@ -160,6 +145,47 @@ func ReleasePleaseExtraFiles(lib *config.Library) []any {
...
+		return
+	}
+	firstAPIPath := library.APIs[0].Path
+	library.Python.OptArgsByAPI[apiPath] = library.Python.OptArgsByAPI[firstAPIPath]
+	log.Printf(
+		"WARNING: customized opt_args_by_api is copied from %q to %q, but this may not be correct for %q."+
+			"Please review and edit opt_args_by_api in librarian.yaml if needed.",
+		firstAPIPath, apiPath, apiPath,
+	)
+}
```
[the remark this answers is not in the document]
JoeWang1127: Done.

internal/librarian/python/add.go
```diff
@@ -127,6 +138,25 @@ func FindExistingLibraryForNewAPI(libraries []*config.Library, apiPath string) *
...
+		return
+	}
+	firstAPIPath := library.APIs[0].Path
+	optArgs, ok := library.Python.OptArgsByAPI[firstAPIPath]
+	if !ok {
+		return
+	}
+	library.Python.OptArgsByAPI[apiPath] = slices.Clone(optArgs)
+	log.Printf(
+		"WARNING: customized opt_args_by_api is copied from %q to %q, but this may not be correct for %q."+
```
zhumin8: super nit: space between lines
```suggestion
		"WARNING: customized opt_args_by_api is copied from %q to %q, but this may not be correct for %q. "+
```

zhumin8 (approved):

internal/librarian/python/add.go
JoeWang1127: Done.

